PR #17998, opened 2022-02-10
Add bioembeddings + lock + jax-unirep + fair-esm + bio-embeddings  bio-embeddings-cpcprot  bio-embeddings-deepblast  bio-embeddings-plus tape_proteins

hadim:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/bio-embeddings-deepblast/meta.yaml
+++ b/recipes/bio-embeddings-deepblast/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
-  noarch: true
+  noarch: python
 
 requirements:
   host:
--- a/recipes/bio-embeddings-deepblast/meta.yaml
+++ b/recipes/bio-embeddings-deepblast/meta.yaml
@@ -12,11 +12,12 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
+  noarch: true
 
 requirements:
   host:
     - pip
-    - python
+    - python >=3.6
     - poetry >=1.1.0
   run:
     - matplotlib-base
@@ -24,7 +25,7 @@ requirements:
     - numpy
     - pandas
     - pillow
-    - python
+    - python >=3.6
     - pytorch-lightning >=0.8.1
     - scikit-learn
     - scipy
--- a/recipes/bio-embeddings-deepblast/meta.yaml
+++ b/recipes/bio-embeddings-deepblast/meta.yaml
@@ -33,10 +33,6 @@ requirements:
 test:
   imports:
     - deepblast
-  commands:
-    - pip check
-  requires:
-    - pip
 
 about:
   home: https://pypi.org/project/bio-embeddings-deepblast/
--- a/recipes/bio-embeddings-deepblast/meta.yaml
+++ b/recipes/bio-embeddings-deepblast/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "bio-embeddings-deepblast" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/bio-embeddings-deepblast-{{ version }}.tar.gz
+  sha256: 26c1c18b405eeea57ad20d1d4c39c35dc40c9706705662a0af188523757bfae7
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python
+  run:
+    - matplotlib-base
+    - numba
+    - numpy
+    - pandas
+    - pillow
+    - python
+    - pytorch-lightning >=0.8.1
+    - scikit-learn
+    - scipy
+    - pytorch >=1.4
+
+test:
+  imports:
+    - deepblast
+  commands:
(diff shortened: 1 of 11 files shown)


recipes/bio-embeddings-deepblast/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/bio-embeddings-deepblast-{{ version }}.tar.gz
+  sha256: 26c1c18b405eeea57ad20d1d4c39c35dc40c9706705662a0af188523757bfae7
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
```
carterbox, line 14: Could this package be noarch?

recipes/bio-embeddings/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  imports:
+    - bio_embeddings
+  commands:
+    - pip check
+    - bio_embeddings --help
+  requires:
+    - pip
+
+about:
+  home: https://pypi.org/project/bio-embeddings/
```
carterbox, line 56: The dev url is https://github.com/sacdallago/bio_embeddings ?

carterbox (changes requested):
Hi @hadim, conda-forge[ currently doesn't build pytorch for windows,](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/master/recipe/meta.yaml) and that *is* the reason why your Windows builds are failing, but[ our policy](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python) is to make `noarch` packages even if they are missing dependencies.

- [ ] Some of these packages, you have marked to skip Windows because they have entry points? These packages can still be marked as noarch if the entrypoints are implemented correctly because conda will automatically convert them for Windows.

recipes/bio-embeddings/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/bio_embeddings-{{ version }}.tar.gz
+  sha256: a82d27d895d971bb30339071440f9c5e75d31cf9f4a8a600d17efe7bea0e8615
+
+build:
+  entry_points:
+    - bio_embeddings = bio_embeddings.utilities.cli:main
+  skip: true  # [win]
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
```
carterbox, line 17: Could be noarch?

carterbox (changes requested):

recipes/bio-embeddings-deepblast/meta.yaml
hadim: yes

recipes/bio-embeddings/meta.yaml
hadim: yes

recipes/bio-embeddings/meta.yaml
hadim: yes

hadim:
@carterbox LGTM

recipes/lock/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/lock-{{ version }}.tar.gz
+  sha256: cc5ac770930493eed7a8cfd0cf2568a125faf112eb8aa6b6149b3e581523d0c7
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
+  entry_points:
+    - lock = lock:lock
```
carterbox, line 17: What's the deal with this entry point? It's listed in the setup.py as a console script, but it didn't work for me when I installed this package using pip. Is that why you aren't testing it here?

recipes/tape_proteins/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set name = "tape_proteins" %}
+{% set version = "0.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  # url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/tape_proteins-{{ version }}.tar.gz
+  # sha256: 5e092ce3b0867342ebd98bd16824c904ff4efdd25476e3e1728f066bfdaaf41a
```
carterbox, line 10:
```suggestion
```

carterbox (changes requested):

recipes/lock/meta.yaml
hadim: The entry point is set here https://github.com/wdbm/lock/blob/master/setup.py#L22.

Without setting it in the feedstock, the build failed on Windows. But now I just realize that this lib can't be used on Windows anyway. So, I can just remove it.

What do you prefer? Keep it here explicitly or remove it?

I don't know much about this package TBH. I just added it because it's a dep of the other packages in that PR.

carterbox: I am trying to figure out whether you forgot to test this entry point or if it was intentionally untested because the entrypoint doesn't look like it is implemented correctly (upstream). I'm guessing you don't intend to use this entry point, so we can leave it as-is and users can open a issue about it later if it doesn't work.

hadim:
Thank you @carterbox for the review.

PR #18013, opened 2022-02-11
Add MNE-Realtime

hoechenberger:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mne-realtime/meta.yaml
+++ b/recipes/mne-realtime/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   noarch: python
   number: 0
-  script: "{{ PYTHON }} -m pip install . -vv"
+  script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
   host:
--- a/recipes/mne-realtime/meta.yaml
+++ b/recipes/mne-realtime/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "mne-realtime" %}
-{% set version = "0.1.2" %}
+{% set version = "0.1.3" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: 4e67afd386589884318a557e35156552ee46f1bd181865cd4389c927aba317d7
+  sha256: c9be35f26695bbfeb79656ead7ada2ad02175212b849479653184f29d0d594a2
 
 build:
   noarch: python
--- a/recipes/mne-realtime/meta.yaml
+++ b/recipes/mne-realtime/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "mne-realtime" %}
-{% set version = "0.1.1" %}
+{% set version = "0.1.2" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: 7eb763dc75ad45572b7d3a7e26b08503ece6bcc9bdea82d8dde95b8ca339ad23
+  sha256: 4e67afd386589884318a557e35156552ee46f1bd181865cd4389c927aba317d7
 
 build:
   noarch: python
--- a/recipes/mne-realtime/meta.yaml
+++ b/recipes/mne-realtime/meta.yaml
@@ -36,7 +36,9 @@ about:
   home: https://mne.tools/mne-realtime
   license: BSD-3-Clause
   license_family: BSD
-  license_file: LICENSE.txt
+  license_file:
+    - LICENSE.txt
+    - mne_realtime/externals/LICENSE_BSD
   summary: 'Realtime data analysis with MNE-Python'
   dev_url: https://github.com/mne-tools/mne-realtime
 
--- a/recipes/mne-realtime/meta.yaml
+++ b/recipes/mne-realtime/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "mne-realtime" %}
+{% set version = "0.1.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 7eb763dc75ad45572b7d3a7e26b08503ece6bcc9bdea82d8dde95b8ca339ad23
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - numpy
+    - scipy
+    - mne-base
+
+test:
+  imports:
+    - mne_realtime
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://mne.tools/mne-realtime
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.txt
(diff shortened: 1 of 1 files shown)


hoechenberger:
@larsoner could you please confirm that you're willingly listed as a co-maintainer here? Thanks!

recipes/mne-realtime/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.txt
+  summary: 'Realtime data analysis with MNE-Python'
+  dev_url: https://github.com/mne-tools/mne-realtime
+
+extra:
+  recipe-maintainers:
+    - hoechenberger
+    - larsoner
```
larsoner, line 46: I'm happy to be listed as a maintainer

recipes/mne-realtime/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  sha256: 7eb763dc75ad45572b7d3a7e26b08503ece6bcc9bdea82d8dde95b8ca339ad23
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python >=3.7
```
BastianZim, line 19: I cannot find anything about 3.7 upstream. Where is that mentioned?

recipes/mne-realtime/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 7eb763dc75ad45572b7d3a7e26b08503ece6bcc9bdea82d8dde95b8ca339ad23
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
```
BastianZim, line 15:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

Just a cosmetic change really.

recipes/mne-realtime/meta.yaml
larsoner: That's a bug -- @hoechenberger I fixed it and pushed 0.1.2 to PyPi, can you update?

hoechenberger: I don't think it's explicitly mentioned anywhere, but I have to put a constraint here since it's a `noarch` package. Pratically, MNE-Realtime only works with Python 3.7+, as it depends on MNE-Python, which in turn requires Python 3.7 and up

recipes/mne-realtime/meta.yaml
hoechenberger: I took this straight from the template on `main`, maybe this should get updated too :)

recipes/mne-realtime/meta.yaml
hoechenberger: @larsoner I've bumped it to 0.1.2

BastianZim: If a dependency has a python requirement, it doesn't need to be added here. Only the minimum python of this package should be listed. The solver will make sure that the minimum python version is pulled in that can be supported by all dependencies and the package itself.
The reason for this is that incase MNE-Python ever has a lower bound in the future, you don't need to change it here, changing it there is sufficient.

But I saw that upstream is now has it as well so let's just keep it. :)

recipes/mne-realtime/meta.yaml
BastianZim: Ahh ok thanks for the heads up. By the way, the easiest way to generate packages is with Grayskull: https://github.com/conda-incubator/grayskull 
It also formats everything correctly and in 95% of all cases creates a perfect package.

recipes/mne-realtime/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  imports:
+    - mne_realtime
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://mne.tools/mne-realtime
+  license: BSD-3-Clause
```
BastianZim, line 37: This file seems to be from a different repo under a different license: https://github.com/mne-tools/mne-realtime/blob/main/mne_realtime/externals/FieldTrip.py

Its license either needs to be added here or the vendored version removed and the package added as a dependency.

recipes/mne-realtime/meta.yaml
hoechenberger: Thanks!

recipes/mne-realtime/meta.yaml
hoechenberger: @larsoner Could you take a look at this? I wasn't aware of this vendored module … @BastianZim's got eagle eyes! 🦅 👍 😮

larsoner: A fun little rabbit hole... It's also BSD 3-clause. I'll push out a 0.1.3 with this info

https://github.com/fieldtrip/fieldtrip/blob/master/realtime/src/buffer/python/FieldTrip.py
https://github.com/fieldtrip/fieldtrip/blob/master/realtime/src/buffer/COPYING
https://github.com/fieldtrip/fieldtrip/blob/master/realtime/src/buffer/LICENSE_BSD

BastianZim: Not sure how much work it is but before you continue with this, from a conda-forge perspective having it as a separate dependency would be better instead of it being vendored. 
But if that's not possible we can work around that as well. :)

larsoner: > But if that's not possible we can work around that as well. :)

Agreed it would be better, but in this case it's not really our code to maintain, and I doubt the original authors are interested in maintaining a conda-forge package. So it would be a lot more (political and coding) work to separate it I think -- but if there are other packages that end up depending on it, agreed it will be worth separating then!

BastianZim: Yeah, I just saw that it's not a proper package and only a script so no problem there then. Then let's continue with the workaround.

recipes/mne-realtime/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  license_family: BSD
+  license_file: LICENSE.txt
```
BastianZim, line 39: This path would now need to point to the path of the new license as well, otherwise conda-build doesn't copy it over.

You can just add it as a list.

```yaml
license_file: 
  - first_license
  - Second_license
```

recipes/mne-realtime/meta.yaml
BastianZim: Upstream lists multiple licenses based on the situation you're using it in. These would all need to be included:

https://github.com/fieldtrip/fieldtrip/blob/7d4608900c0754decbbfe24b1072901ea7b040de/realtime/src/buffer/COPYING#L5

BastianZim: It might be possible to use just one if that's preferred but then I'll just quickly ping conda-forge/core for verification as they have the final decision for licenses.

larsoner: I tried to follow the instructions in COPYING, which were to 1) choose one license:

https://github.com/fieldtrip/fieldtrip/blob/7d4608900c0754decbbfe24b1072901ea7b040de/realtime/src/buffer/COPYING#L7

Then, 2) add it to your codebase, and 3) acknowledge in the manual, which for us comes from the docstring here:

https://github.com/mne-tools/mne-realtime/blob/569c56ab80ac320e92401685ca62d7ffb5aff949/mne_realtime/fieldtrip_client.py#L57-L59

So I think we should be in the clear on this one, but I am no license expert...

BastianZim: Yes you're right, should be sufficient.

recipes/mne-realtime/meta.yaml
hoechenberger: @BastianZim Done!

PR #18031, opened 2022-02-13
Add recipe for icontract

kklein:
I couldn't find [icontract](https://github.com/Parquery/icontract) on conda forge.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/icontract/meta.yaml
+++ b/recipes/icontract/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "icontract" %}
-{% set version = "2.6.0" %}
+{% set version = "2.6.1" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/icontract-{{ version }}.tar.gz
-  sha256: 1ef9332ab4ca10f4da0370cec3743b88e10f8aac3cfcb9bc528008e1b2f09fae
+  sha256: dacfc55009993058cd17576faa291ab76b6f39d80ec9d1e524778cd79da11857
 
 build:
   noarch: python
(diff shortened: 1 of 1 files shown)


kklein:

kklein:

recipes/icontract/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - asttokens >=2,<3
+    - python >=3.6
+    - typing_extensions
+
+test:
+  imports:
+    - icontract
+    - tests_3_6
+    - tests_3_7
+    - tests_3_8
```
BastianZim, line 31: If this is passing, that means it's installing the test as a test package/module in the global package namespace, which is not good and can become a problem since they will pollute the end-user environment. This is not a problem that you can necessarily solve yourself but can you ask upstream to fix this, please? Test folders should never be installed as a separate package.

BastianZim (commented):
Looks good, just one comment regarding the test modules.

recipes/icontract/meta.yaml
kklein: Great catch!
I raised an issue [0] upstream. How would you suggest to proceed?

[0] https://github.com/Parquery/icontract/issues/239

BastianZim: The best way would be to add a PR to upstream fixing this and then including the PR as a [patch](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#patches) here.

Alternatively, you can also add some sort of script to the build section that removes it but that's really a dirty/last resort solution.

BastianZim:
Looks good! Ready to be merged?

recipes/icontract/meta.yaml
kklein: The upstream problem is fixed and a patched version already released on pypi. I updated the recipe to this new version on pypi: [85e458c](https://github.com/conda-forge/staged-recipes/pull/18031/commits/85e458cb0ac1ad2dad06fc4696adb7b106c0ce88)

Do I understand correctly that we no longer need aforementioned approach?

BastianZim: Yes, it's not required anymore. From my side this can be merged.

PR #18038, opened 2022-02-14
Added recipe for hover 0.6.0 [ready for review]

phurwicz:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/hover/meta.yaml
+++ b/recipes/hover/meta.yaml
@@ -1,0 +1,53 @@
+{% set name = "hover" %}
+{% set version = "0.6.0" %}
+
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 26b9b4bcc4582a88910b14240002d88fe05b9e816db52a93202691ec72f26055
+
+build:
+  number: 0
+  noarch: python
+  script: python -m pip install . -vv
+
+requirements:
+  host:
+    - python >=3.7
+    - pip
+  run:
+    - python >=3.7
+    - bokeh >=2.3.3
+    - scikit-learn >=0.20.0
+    - pytorch >=1.4.0
+    - pandas >=1.3.0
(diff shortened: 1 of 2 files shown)


phurwicz:
This is phurwicz, willing to be listed as a maintainer.

haochuanwei:
This is Haochuan Wei (haochuanwei), willing to be listed as a maintainer.

phurwicz:
@conda-forge-admin please ping conda-forge/help-python

recipes/hover/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  imports:
+    - hover
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://phurwicz.github.io/hover
+  license: MIT
```
BastianZim, line 45: This file seems to be copied from a different package with a different license. If that's the case, the license and file need to be included here.

https://github.com/phurwicz/hover/blob/main/hover/utils/copied/snorkel.py

Generally, it's better to have the package as a dependency than to include it but I can see how its restrictions make that difficult.

BastianZim (changes requested):

recipes/hover/meta.yaml
phurwicz: Thanks. Could you point me to an example of how to include it?

I'm thinking about implementing this copied function on my own next time. It's been a ton of overhead just to use one function that isn't quite sophisticated.

BastianZim: > Thanks. Could you point me to an example of how to include it?

I don‘t have one right now but the steps are 

1. Find out which license upstream uses
2. Add the license to the one you have here using `AND`  So `MIT AND Apache-2.0` for example
3. Copy the license of the dependency into the same folder as the meta.yaml
4. Add the license to the list of files  

Example:

```yaml
license_file:
  - license_file_1
  - license_file2
```

> The context is that snorkel used to be listed as a dependency, but then other dependency versions start to fall apart because snorkel won't keep up with upgrades.

Makes sense. It could be that the conda-forge version has different dependencies but in the end, writing it yourself or adding it manually is probably the best approach.

PR #18053, opened 2022-02-15
Adding recipe for TREXIO library

q-posev:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/trexio/meta.yaml
+++ b/recipes/trexio/meta.yaml
@@ -27,11 +27,11 @@ requirements:
 
   host:
     - hdf5
-    ## The dependencies below are not required by TREXIO but they get exported from HDF5
-    ## and if they are not listed here than the `conda build --error-overlinking` fails.
-    #- zlib    
-    #- openssl 
-    #- libcurl 
+    # the dependencies below are not required by TREXIO but they get exported from HDF5
+    # and if they are not listed here than the `conda-build --error-overlinking` fails
+    - zlib    
+    - openssl 
+    - libcurl 
 
   run:
     - hdf5 
--- a/recipes/trexio/meta.yaml
+++ b/recipes/trexio/meta.yaml
@@ -27,24 +27,17 @@ requirements:
 
   host:
     - hdf5
-    # the dependencies below are not required by TREXIO but they get exported from HDF5
-    # and if they are not listed here than the `conda-build --error-overlinking` fails
-    - zlib    
-    - openssl 
-    - libcurl 
-
-  run:
-    - hdf5 
 
 test:
   commands:
-    - test -f $PREFIX/lib/libtrexio.a  # [unix]
+    - test -f ${PREFIX}/lib/libtrexio${SHLIB_EXT}  # [unix]
+    - test -f ${PREFIX}/include/trexio.h           # [unix]
 
 about:
   home: https://github.com/TREX-CoE/trexio
   license: BSD-3-Clause
   license_family: BSD 
-  license_file: LICENSE
+  license_file: COPYING
   summary: 'TREX library for efficient I/O'
   description: |
     The TREXIO library defines a standard format for storing wave functions, 
--- a/recipes/trexio/meta.yaml
+++ b/recipes/trexio/meta.yaml
@@ -44,7 +44,7 @@ about:
   home: https://github.com/TREX-CoE/trexio
   license: BSD-3-Clause
   license_family: BSD 
-  license_file: COPYING
+  license_file: LICENSE
   summary: 'TREX library for efficient I/O'
   description: |
     The TREXIO library defines a standard format for storing wave functions, 
--- a/recipes/trexio/meta.yaml
+++ b/recipes/trexio/meta.yaml
@@ -1,6 +1,35 @@
 
 {% set name = "trexio" %}
 {% set version = "2.1.0" %}
+{% set build = 0 %}
+
+# recipe-lint fails if mpi is undefined
+{% set mpi = mpi or 'nompi' %}
+{% if mpi == "nompi" %}
+# prioritize nompi via build number
+{% set build = build + 100 %}
+{% endif %}
+
+{% if mpi != "nompi" %}
+{% set mpi_prefix = "mpi_" + mpi %}
+{% else %}
+{% set mpi_prefix = "nompi" %}
+{% endif %}
+# add build string so packages can depend on
+# mpi or nompi variants
+# dependencies:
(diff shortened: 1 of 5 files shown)


q-posev:
@conda-forge/staged-recipes This PR is ready for review, thank you in advance!

recipes/trexio/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libtrexio.a  # [unix]
+
+about:
+  home: https://github.com/TREX-CoE/trexio
+  license: BSD-3-Clause
+  license_family: BSD 
+  license_file: COPYING
```
carterbox, line 47:
```suggestion
  license_file: LICENSE
```
Copying is a symbolic link to LICENSE. Add LICENSE directly?

recipes/trexio/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - zlib    
+    - openssl 
+    - libcurl 
+
+  run:
+    - hdf5 
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libtrexio.a  # [unix]
```
carterbox, line 41: Please check for all the files that you are shipping. You should be shipping a `shared` library and header files instead of a static library with no headers.

recipes/trexio/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - gnuconfig 
+    - pkg-config
+
+  host:
+    - hdf5
+    # the dependencies below are not required by TREXIO but they get exported from HDF5
+    # and if they are not listed here than the `conda-build --error-overlinking` fails
+    - zlib    
+    - openssl 
+    - libcurl
```
carterbox, line 34:
```
  ERROR (trexio,lib/libtrexio.so.0.0.0): Needed DSO lib/libcurl.so.4 found in ['libcurl']
  ERROR (trexio,lib/libtrexio.so.0.0.0): .. but ['libcurl'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```
This error message from conda means that it has inspected this binary and found that it links to libcurl (like when you inspect a file with ldd). Since, you state in this comment that libcurl is not a dependency of libtrexio, you should double check your build process to prevent overlinking.

You should *not* add the packages to the host section if trexio doesn't need them. If they are needed by hdf5 at runtime, hdf5 will have a run requirement for them.

recipes/trexio/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - hdf5
+    # the dependencies below are not required by TREXIO but they get exported from HDF5
+    # and if they are not listed here than the `conda-build --error-overlinking` fails
+    - zlib    
+    - openssl 
+    - libcurl 
+
+  run:
+    - hdf5
```
carterbox, line 37: This shouldn't be required because hdf5 has a run export for itself.

carterbox (changes requested):
The README on the dev repository says that CMake is for persons building from the tarball. Is that a strict requirement or do you just prefer autotools?

For autotools, the README includes calls to autogen.sh

q-posev:
> The README on the dev repository says that CMake is for persons building from the tarball. Is that a strict requirement or do you just prefer autotools?

We actually support both builds, it is up to the user to decide whether he would like to use CMake or autotools. Would you recommend to use CMake when building for `conda` ?

recipes/trexio/meta.yaml
q-posev: Right, good idea! Done.

xhochy:
CMake is preferable to autotools because we have a (slightly) better infrastructure set up for it. But also take into account that there sometimes differences between those builds: https://conda-forge.org/docs/maintainer/knowledge_base.html#moving-from-an-autotools-build-to-a-cmake-build

recipes/trexio/meta.yaml
q-posev: Done.

recipes/trexio/meta.yaml
q-posev: In fact, I just realized that `COPYING` is a symbolic link only when building from the GitHub repo clone. In the distribution tarball produced by Autotools (`make dist`) it appears as a text file, so it's fine. The `LICENSE` is not shipped with the distribution tarball. So we can keep it for now.

recipes/trexio/meta.yaml
q-posev: Indeed, hdf5 has a run requirement for these packages (see their [meta.yaml](https://github.com/conda-forge/hdf5-feedstock/blob/master/recipe/meta.yaml)). Our build system based on Autotools parses the libraries that are linked against HDF5 and adds them to the `LIBS` list, which results in overlinking. I switched to the CMake build and there is no overlinking issue. Probably the `FindHDF5.cmake` module was designed with these things in mind.

recipes/trexio/meta.yaml
q-posev: Done. 

Here I had one question: since `trexio` depends on `hdf5`, should this dependency be somehow exported as a runtime dependency? Or this is taken into account by `conda` when listing `hdf5` under `requirements/host`?

q-posev:
@carterbox I think this PR is ready for a review. Changes in a nutshell:

- Switched to CMake build to prevent overlinking issue
- Added MPI variant
- Removed  hdf5-related `host` and `run` requirements
- Added `run_exports` for trexio itself

recipes/trexio/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+    - pkg-config
+
+  host:
+    - hdf5  # [mpi == 'nompi']
+    - hdf5 * {{ mpi_prefix }}_*  # [mpi != 'nompi']
+
+test:
+  commands:
+    - test -f ${PREFIX}/lib/libtrexio${SHLIB_EXT}  # [unix]
+    - test -f ${PREFIX}/include/trexio.h  # [unix]
```
carterbox, line 66:
```suggestion
    - test -f ${PREFIX}/include/trexio.h  # [unix]
    - test -f ${PREFIX}/include/trexio_f.f90  # [unix]
```

carterbox (approved):

recipes/trexio/meta.yaml
carterbox: Because the recipe for hdf5 contains a run export, conda-build automatically adds hdf5 to the `requirements/run` section of your recipe when you add it to the `requirements/host` section.

PR #18080, opened 2022-02-17
add colander

ocefpaf:
See https://pypi.org/project/colander/

### DIFF
--- a/recipes/colander/meta.yaml
+++ b/recipes/colander/meta.yaml
@@ -36,7 +36,9 @@ about:
   home: https://github.com/Pylons/colander
   summary: A simple schema-based serialization and deserialization library
   license: LicenseRef-BSD-ModificationVariant
-  license_file: COPYRIGHT.txt, LICENSE.txt
+  license_file: 
+    - COPYRIGHT.txt
+    - LICENSE.txt
 
 extra:
   recipe-maintainers:
--- a/recipes/colander/meta.yaml
+++ b/recipes/colander/meta.yaml
@@ -35,7 +35,7 @@ test:
 about:
   home: https://github.com/Pylons/colander
   summary: A simple schema-based serialization and deserialization library
-  license: LicenseRef-BSD-ModificationVariant
+  license: BSD-3-Clause-Modification
   license_file: 
     - COPYRIGHT.txt
     - LICENSE.txt
--- a/recipes/colander/meta.yaml
+++ b/recipes/colander/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "colander" %}
+{% set version = "1.8.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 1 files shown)


ocefpaf:

recipes/colander/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - colander
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/Pylons/colander
+  summary: A simple schema-based serialization and deserialization library
+  license: Mup
```
BastianZim, line 39: It's similar but I don't think it's actually `Mup`: https://spdx.org/licenses/Mup.html

BastianZim (changes requested):

recipes/colander/meta.yaml
ocefpaf: They say BSD-derived: https://github.com/Pylons/colander/blob/master/setup.cfg#L8

Looks like it is this one: https://fedoraproject.org/wiki/Licensing:BSD#Modification_Variant

But I could not find a SPDX form for it.

recipes/colander/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/Pylons/colander
+  summary: A simple schema-based serialization and deserialization library
+  license: LicenseRef-BSD-ModificationVariant
+  license_file: COPYRIGHT.txt, LICENSE.txt
```
BastianZim, line 39:
```suggestion
  license_file: 
    - COPYRIGHT.txt
    - LICENSE.txt
```
The original  is valid yaml but converts it to a string and not a list.

recipes/colander/meta.yaml
ocefpaf: Yep. Looks like it. Thanks!

BastianZim:
Hmm that linter complaint is strange – it's a valid expression. If you're happy with the recipe I can merge it and then we can file an issue in the linter repo.

recipes/colander/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/Pylons/colander
+  summary: A simple schema-based serialization and deserialization library
+  license: Mup
+  license_file: COPYRIGHT.txt
```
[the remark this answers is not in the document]
marcelotrevisani: indeed, I think it is because that file was the first that grayskull found, but indeed we can improve that

ocefpaf:
> Hmm that linter complaint is strange – it's a valid expression. If you're happy with the recipe I can merge it and then we can file an issue in the linter repo.

We need to add that license to the list of valid ones. We can do that later. Merge when you are ready. Thanks!

BastianZim:
Yeah I asked in gitter - seems like an extra license package would be better but that might be a first step.

PR #18081, opened 2022-02-17
Add vegafusion-python-embed package

jonmmease:
This PR adds a recipe for the [vegafusion-python-embed](https://pypi.org/project/vegafusion-python-embed/) package. Project hompage is at https://vegafusion.io/.

This is a native Python package that wraps Rust logic using [PyO3](https://pyo3.rs/v0.15.1/), compiled with the [maturin](https://github.com/PyO3/maturin) build tool.  The conda recipe pulls the official source from PyPI and builds it using maturin.  This source distribution contains the project's own `LICENSE` file. A `Thirdparty.yml` file containing the licenses for third-party Rust dependencies (which are statically linked during the build process) is generated using [`cargo-bundle-licenses`](https://github.com/sstadick/cargo-bundle-licenses) during the build processes.

For context, this package will rarely be used directly and after it is published two additional pure-python packages that depend on this one will follow ([`vegafusion`](https://pypi.org/project/vegafusion/) and [`vegafusion-jupyter`](https://pypi.org/project/vegafusion-jupyter/)).

I'll ping once CI is passing. Thanks!
---

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/vegafusion-python-embed/meta.yaml
+++ b/recipes/vegafusion-python-embed/meta.yaml
@@ -8,7 +8,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name_under[0] }}/{{ name_under }}/{{ name_under }}-{{ version }}.tar.gz
-  sha256: 049c6577589d84b43c2674abd8da240daf6189fce695be50bedb31d44f363949
+  sha256: 155530fce0a812facc6606f51cf4b2505e29f31d2d3ec29a2e58ab9968fdbb3c
 
 build:
   number: 0
--- a/recipes/vegafusion-python-embed/meta.yaml
+++ b/recipes/vegafusion-python-embed/meta.yaml
@@ -14,7 +14,7 @@ build:
   number: 0
   script: |
     export RUSTFLAGS="--cfg unsound_local_offset"     # [not win]
-    set RUSTFLAGS=--cfg unsound_local_offset          # [win]
+    set "RUSTFLAGS=--cfg unsound_local_offset"        # [win]
     cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
     {{ PYTHON }} -m pip install . -vv
 
--- a/recipes/vegafusion-python-embed/meta.yaml
+++ b/recipes/vegafusion-python-embed/meta.yaml
@@ -1,0 +1,55 @@
+{% set name = "vegafusion-python-embed" %}
+{% set name_under = "vegafusion_python_embed" %}
+{% set version = "0.0.4_rc.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name_under[0] }}/{{ name_under }}/{{ name_under }}-{{ version }}.tar.gz
+  sha256: d5582e82ec482e9327ab53ac6047826effa0e4ac6bc1d0e81b9fe5818f09d1da
+
+build:
+  number: 0
+  script: |
+    cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+    {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  build:
+    - {{ compiler('c') }}        # [not win]
+    - {{ compiler('m2w64_c') }}  # [win]
+    - {{ compiler('rust') }}
+  host:
+    - python
+    - pip
+    - maturin
+    - cargo-bundle-licenses
+  run:
+    - python
+
+test:
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://vegafusion.io
+  license: AGPL-1.0-only
+  license_family: GPL
+  license_file:
+    - LICENSE
+    - THIRDPARTY.yml
+
+  summary: 'Server-side acceleration for the Vega visualization grammar'
+  description: |
+    VegaFusion provides serverside acceleration for the Vega visualization grammer. The vegafusion-python-embed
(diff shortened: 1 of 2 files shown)


dhirschfeld:
The key error seems to be:
```
error: invalid `--cfg` argument: `--target` (expected `key` or `key="value"`)
```
From:
```
...
+HOST=x86_64-conda-linux-gnu
+LDFLAGS=-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -L$PREFIX/lib
Error: `cargo metadata` exited with an error: error: failed to run `rustc` to learn about target-specific information

Caused by:
  process didn't exit successfully: `rustc - --crate-name ___ --print=file-names --cfg --target x86_64-unknown-linux-gnu --crate-type bin --crate-type rlib --crate-type dylib --crate-type cdylib --crate-type staticlib --crate-type proc-macro --print=sysroot --print=cfg` (exit status: 1)
  --- stderr
  error: invalid `--cfg` argument: `--target` (expected `key` or `key="value"`)

```

jonmmease:
Ok, build is green again

recipes/vegafusion-python-embed/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+requirements:
+  build:
+    - {{ compiler('c') }}        # [not win]
+    - {{ compiler('m2w64_c') }}  # [win]
+    - {{ compiler('rust') }}
+  host:
+    - python
+    - pip
+    - maturin
+    - cargo-bundle-licenses
```
bollwyvl, line 30: `cargo-bundle-licenses` appears to usually go in `build`

jonmmease: works for me!

recipes/vegafusion-python-embed/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+source:
+  url: https://pypi.io/packages/source/{{ name_under[0] }}/{{ name_under }}/{{ name_under }}-{{ version }}.tar.gz
+  sha256: d5582e82ec482e9327ab53ac6047826effa0e4ac6bc1d0e81b9fe5818f09d1da
+
+build:
+  number: 0
+  script: |
+    export RUSTFLAGS="--cfg unsound_local_offset"     # [not win]
+    set RUSTFLAGS=--cfg unsound_local_offset          # [win]
```
wolfv, line 17: I think the "safe way" is to put a quotes around the entire env var plus value, like `set "RUSTFLAGS=--cfg unsound_local_offset"`

recipes/vegafusion-python-embed/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "vegafusion-python-embed" %}
+{% set name_under = "vegafusion_python_embed" %}
+{% set version = "0.0.4_rc.1" %}
```
wolfv, line 3: is this an RC?

dhirschfeld: Thanks for the review @wolfv!  I agree this should build from an actual release if at all possible.

recipes/vegafusion-python-embed/meta.yaml
jonmmease: Thanks! Giving that a try now

recipes/vegafusion-python-embed/meta.yaml
jonmmease: Yeah, this is an RC because the current stable version doesn't have sdist bundles published.  If this approach looks good otherwise, I can publish a new version.

dhirschfeld: I think if you're ready to release `0.0.4` proper then this recipe is in good shape to be merged.

Ofc, it's always possible further :eyes: will pick up some more issues but it's easier for others to review when it's ready to merge (otherwise it would need a further review when it was ready)

jonmmease: Tests running on version 0.0.4

jonmmease:
Thanks for the reviews. Now built against version 0.0.4 final

jonmmease:
friendly ping @wolfv @bollwyvl  Is there anything else you'd like to see here?  Thanks again for taking a look!

bollwyvl:
Seems alright to me: i can't merge, though!

dhirschfeld:
> Seems alright to me: i can't merge, though!

Ping @conda-forge/staged-recipes!

recipes/vegafusion-python-embed/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - pip
+    - maturin
+  run:
+    - python
+
+test:
+  requires:
+    - pip
+  commands:
+    - pip check
```
tdegeus, line 38:
```suggestion
  commands:
    - pip check
    - python -c "import vegafusion_embed; vegafusion_embed.PyTaskGraphRuntime(4)"
```

tdegeus (commented):
Not a member of the core, but I saw your message on gitter.
Mostly thing looks good, though:
* I would consider a different name, that keeps together the name of the package. E.g. `vegafusion_embed-python`, `python-vegafusion_embed`, or `python-vegafusion-embed`
* I'm not sure that `run_test.py` is actually evaluated. But in any case I would avoid it all-together as indicated

PR #18090, opened 2022-02-18
Add coscine

niklassiemer:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/coscine/meta.yaml
+++ b/recipes/coscine/meta.yaml
@@ -20,6 +20,11 @@ requirements:
   host:
     - python >=3.6
     - pip
+    - requests
+    - requests-toolbelt
+    - tqdm
+    - colorama
+    - prettytable
   run:
     - python >=3.6
     - requests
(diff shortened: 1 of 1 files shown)


niklassiemer:
@conda-forge/staged-recipes, this is ready for review.

recipes/coscine/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+    - requests
+    - requests-toolbelt
+    - tqdm
+    - colorama
+    - prettytable
```
beenje, line 26: Those should only be `run` requirements. Pure python packages should only require `python` and `pip` in `host` requirements.
But I see that the package is importing itself in [setup.py](https://git.rwth-aachen.de/coscine/docs/public/coscine-python-sdk/-/blob/master/setup.py): `from src.coscine.version import __version__`.

This is not recommended:

> Although this technique is common, beware that it will fail if sample/__init__.py imports packages from install_requires dependencies, which will very likely not be installed yet when setup.py is run.

 Could you please open an issue in the upstream repo to change that in next version?
You can link to https://packaging.python.org/en/latest/guides/single-sourcing-package-version/


Recipe looks good otherwise. I'll merge.
Think about removing those run requirements when the upstream repo is updated.

niklassiemer: Thanks a lot for the comment and the merge, I will ask for a corresponding change for the next version.

PR #18093, opened 2022-02-18
Add pyscamp-gpu and pyscamp-cpu packages.

zpzim:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pyscamp-cpu/meta.yaml
+++ b/recipes/pyscamp-cpu/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyscamp" %}
-{% set version = "3.0.2" %}
+{% set version = "3.0.3" %}
 
 
 package:
@@ -7,11 +7,8 @@ package:
   version: {{ version }}
 
 source:
-  # Temporarily use test pypi package to test new code.
-  url: https://test-files.pythonhosted.org/packages/8e/cc/4216ddadb2b7c2d008802e7c7b1c32d4483f6ccb6fb320a956a368fffef9/pyscamp-3.0.3.dev4.tar.gz
-  sha256: 80081eab782d96111151a50d6924aafb617568cb94fe81a8da304095cd286fcf
-  #url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyscamp-{{ version }}.tar.gz
-  #sha256: 1a9dcae45c85d993910ecb75bbb28188e72cf8f0c086c876ef73d145793d8060
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyscamp-{{ version }}.tar.gz
+  sha256: 047e699f69a087b72af1ea681821d9da37fa3b805abb9ab2396baf6b530a1a11
 
 build:
   number: 0
--- a/recipes/pyscamp-gpu/meta.yaml
+++ b/recipes/pyscamp-gpu/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "pyscamp" %}
-{% set version = "3.0.2" %}
+{% set version = "3.0.3" %}
 
 
 package:
@@ -7,11 +7,8 @@ package:
   version: {{ version }}
 
 source:
-  # Temporarily use test pypi package to test new code.
-  url: https://test-files.pythonhosted.org/packages/8e/cc/4216ddadb2b7c2d008802e7c7b1c32d4483f6ccb6fb320a956a368fffef9/pyscamp-3.0.3.dev4.tar.gz
-  sha256: 80081eab782d96111151a50d6924aafb617568cb94fe81a8da304095cd286fcf
-  #url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyscamp-{{ version }}.tar.gz
-  #sha256: 1a9dcae45c85d993910ecb75bbb28188e72cf8f0c086c876ef73d145793d8060
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyscamp-{{ version }}.tar.gz
+  sha256: 047e699f69a087b72af1ea681821d9da37fa3b805abb9ab2396baf6b530a1a11
 
 build:
   number: 0


zpzim:

Hi there! I'm a first time contributor trying to submit my first recipes. I have a couple of questions I was hoping could be answered.

First let me mention a little about this project:

-  It is a cuda/c++ library which uses pybind11 to wrap a python module around it.
-  The python module is hosted as an sdist on pypi, with releases triggered by the github [repo](https://github.com/zpzim/SCAMP).
- The project must be built on the system it will run on (e.g. compilation makes extensive use of -march=native and other methods of detecting what the host architecture is capable of) to allow for maximum performance.
- The module is configured to build with CUDA if it is available.
    - If built with CUDA the module will make use of Nvidia GPUs if they are available, otherwise it will fall back to a CPU implementation. The cuda build requires the CUFFT library from the cuda toolkit.
    - If not built with CUDA the module will simply use the CPU implementation.

I am wondering if there is an optimal way to set this up on conda-forge. 

- Currently I have the recipe split into a cpu recipe which does not require cuda and a gpu recipe which does. Though this may not be optimal given that my project can build with CUDA if it is available and fallback if it is not.
- Through some investigation, I noticed there are multiple ways cuda can be specified in a conda recipe (e.g. `cudatoolkit`, `cudatoolkit-dev`, `{{ compiler('cuda') }}`, etc.) 
- I am not familiar with how conda manages the cuda installation, I noticed this repo has build targets in testing which test with various cuda toolkit versions (e.g. `linux_64_cuda_112`). Are these targets with cuda installed on the host? Does this mean that conda is able to use the host system toolkit if it is available? This is a bit confusing because we also have the `cudatoolkit` and `cudatoolkit-dev` packages which seem to install cuda inside the conda environment. Trying to understand this a bit better. 
- I'd also like to understand a bit more about the windows cuda support. I noticed there are no CI tests if the recipe builds with windows + CUDA and the `build_locally.py` script does not support building with win64 to check.
- Ideally, what I want to be able to do is for a user to be able to install my package with minimal steps using conda-forge. 
    - If they install `pyscamp-gpu` it will force a build with cuda if it is available, otherwise we won't install. Ideally we could automatically install the cuda toolkit and nvcc if they don't have them.
    - If they install `pyscamp-cpu` it will install without cuda support. 


Thanks in advance for any assistance!

zpzim:

zpzim:
Hi!

After some more investigation I have figured out some of my previous questions regarding cuda support. If I understand correctly:

1. conda uses the host cuda toolkit. It looks like maybe some symlinks are created behind the scenes.
2. For windows. CUDA extensions have to be installed along with the appropriate supported visual studio build tools for conda. e.g. VS 2017

Please correct me if I am wrong on either of those conclusions. 

I am still having some trouble getting this build to work properly on Windows. It looks like the CI build defaults to NMake Makefiles, however CMAKE_GENERATOR_PLATFORM and CMAKE_GENERATOR_TOOLSET are still getting set in the CI environment somehow. This leads to cmake build errors in my project as these environment variables make no sense for NMake Makefiles. Not sure if I am configuring something wrong or there is a bug somewhere...

```
CMake Error at CMakeLists.txt:44 (project):
  Generator

    NMake Makefiles

  does not support toolset specification, but toolset
-- Configuring incomplete, errors occurred!

See also "C:/bld/pyscamp-cpu_1646779121394/work/build/temp.win-amd64-3.7/Release/CMakeFiles/CMakeOutput.log".
    v141

  was specified.
```

If I could configure the build to use VS build tools on windows that would probably work best, since then users could get GPU support if they wanted.

Also it looks like some of the CI cuda builds have recently started failing with a gcc compiler error. This seems related to a gcc compiler bug which was fixed in gcc 10.4, see here: https://github.com/NVIDIA/nccl/issues/494#issuecomment-855018515

zpzim:
The failing CUDA builds seem like they could be related to this commit: https://github.com/conda-forge/staged-recipes/commit/c4589d0a75eea282e2e84fadfc14f696e50936f6

zpzim:
cc: @isuruf 

I disabled Windows builds for the CPU recipe until I can figure out how to get it to work.

It looks like the GCC bug (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100102) is still impacting the cuda build:

```
/home/conda/staged-recipes/build_artifacts/pyscamp-gpu_1646780099926/_build_env/x86_64-conda-linux-gnu/include/c++/10.3.0/chrono: In substitution of 'template<class _Rep, class _Period> template<class _Period2> using __is_harmonic = std::__bool_constant<(std::ratio<((_Period2::num / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::num, _Period::num)) * (_Period::den / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::den, _Period::den))), ((_Period2::den / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::den, _Period::den)) * (_Period::num / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::num, _Period::num)))>::den == 1)> [with _Period2 = _Period2; _Rep = _Rep; _Period = _Period]':
/home/conda/staged-recipes/build_artifacts/pyscamp-gpu_1646780099926/_build_env/x86_64-conda-linux-gnu/include/c++/10.3.0/chrono:473:154:   required from here
/home/conda/staged-recipes/build_artifacts/pyscamp-gpu_1646780099926/_build_env/x86_64-conda-linux-gnu/include/c++/10.3.0/chrono:428:27: internal compiler error: Segmentation fault
  428 |  _S_gcd(intmax_t __m, intmax_t __n) noexcept
      |                           ^~~~~~
Please submit a full bug report,
with preprocessed source if appropriate.
See <[https://gcc.gnu.org/bugs/>](https://gcc.gnu.org/bugs/%3E) for instructions.
```

This bug has been patched in https://gcc.gnu.org/git/gitweb.cgi?p=gcc.git;h=5357ab75dedef403b0eebf9277d61d1cbeb5898f and is available in `gcc >= 10.4`

zpzim:

Hi, I think I managed to get everything set up and working correctly. This PR is finally ready for review!

zpzim:
TL;DR for whoever reviews:

I'm a first time contributor trying to submit my first recipes. Here's a brief summary of my project:

- This is a cuda/c/c++ library which uses pybind11 to wrap a python module around it.
- The python module is hosted as an sdist on pypi, with releases triggered by the github [repo](https://github.com/zpzim/SCAMP).

- The project is cross platform and has been tested on x86_64 windows/linux/macOS.
- The project is configured to build with CUDA if it is available.
- If built with CUDA the module will make use of Nvidia GPUs if they are available, otherwise it will fall back to a CPU implementation. The CUDA build requires the CUFFT library from the CUDA toolkit.
- If not built with CUDA the module will simply use the CPU implementation.

I have the recipe split into a `pyscamp-cpu` recipe which does not require cuda and a `pyscamp-gpu` recipe which does. Though this may not be optimal given that my project can build with CUDA if it is available and fallback if it is not. Perhaps there is a way to create a singular `pyscamp` recipe?

`pyscamp` has 3 dependencies which are packaged with it:

- `eigen`
- `cpu_features`
- `pybind11`

Everything is compiled together statically into a single python module. The only runtime dependencies are C++ and libcufft.

beckermr (changes requested):
I understand that the dependencies package with this recipe are header-only (except maybe cpu_features?).  However, we generally do not allow packages to vendor these things unless there are special circumstances. Why are those dependencies vendored?

zpzim:
Ah I didn't realize that was an issue. When building out this project I followed suggestions from the dependencies:

`cpu_features` mentions in their [documentation](https://github.com/google/cpu_features/blob/main/cmake/README.md#recommended-usage--incorporating-cpu_features-into-a-cmake-project) that the recommended way to integrate their project is to vendor it.

I chose to vendor `eigen` and `pybind11` in SCAMP to minimize the need for users to install dependencies when building it and to retain maximum control of the version of the dependencies I was using. SCAMP depends on Eigen for performance and being able to pin an exact version or commit could be useful.

I see that eigen and pybind11 have up to date conda forge packages. I suppose it could be possible to configure SCAMP to depend on preinstalled headers for these libraries. But as you mentioned, these are header libraries, and I'd prefer to not have to diverge the build process for the base project and the recipe if possible.

beckermr:
I understand the annoyance but the vast majority of existing recipes use options in makefiles or cmake to allow for external headers like this. Your recipe should do the same.

zpzim:
Acknowledged. I will update SCAMP with the option to depend on the conda versions of eigen and pybind11. Is it okay if I vendor cpu_features? That seems to be what is recommended by the project owners and there appears to be potential ABI compatibility problems if the project is not statically linked.

beckermr:
The CPU features stuff is fine. Thank you!

recipes/pyscamp-gpu/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [(win64 or linux64) and cuda_compiler_version != "None"]
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
+    - pybind11 2.9.2
+    - eigen 3.4.0
```
beckermr, line 29: You want these in host, not build AFAIK.

recipes/pyscamp-cpu/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
+    - pybind11 2.9.2
+    - eigen 3.4.0
```
beckermr, line 27: These should go in host AFAIK.

recipes/pyscamp-gpu/meta.yaml
zpzim: I was thinking that too, but since they were header only I put them in build. I guess it is good to be safe here and make sure there's not a chance the user will be missing something they need.

recipes/pyscamp-cpu/meta.yaml
zpzim: Done.

beckermr (approved):

beckermr (approved):

beckermr:
Will merge once it passes.

PR #18102, opened 2022-02-19
Add recipe for mlrose python package

onelharrison:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mlrose/meta.yaml
+++ b/recipes/mlrose/meta.yaml
@@ -12,7 +12,6 @@ source:
 build:
   noarch: python
   number: 0
-  skip: True # [py<3]
   script: "{{ PYTHON }} -m pip install . -vv"
 
 requirements:
@@ -20,10 +19,10 @@ requirements:
     - numpy
     - scipy
     - scikit-learn
-    - python >= 3
+    - python >=3
     - pip
   run:
-    - python
+    - python >=3
 
 test:
   imports:
@@ -35,7 +34,7 @@ test:
 
 about:
   home: https://github.com/gkhayes/mlrose
-  license: BSD 3-Clause
+  license: BSD-3-Clause
   license_family: BSD
   license_file: LICENSE
   summary: 'Python package for implementing a number of Machine Learning, Randomized Optimization and SEarch algorithms.'
--- a/recipes/mlrose/meta.yaml
+++ b/recipes/mlrose/meta.yaml
@@ -21,7 +21,7 @@ requirements:
   run:
     - numpy
     - scipy
-    - scikit-learn
+    - scikit-learn <0.21.0
     - python >=3
 
 test:
--- a/recipes/mlrose/meta.yaml
+++ b/recipes/mlrose/meta.yaml
@@ -16,12 +16,12 @@ build:
 
 requirements:
   host:
-    - numpy
-    - scipy
-    - scikit-learn
     - python >=3
     - pip
   run:
+    - numpy
+    - scipy
+    - scikit-learn
     - python >=3
 
 test:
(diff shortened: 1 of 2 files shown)


onelharrison:
@conda-forge-admin These builds are failing because a [LICENSE file cannot be detected](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=462045&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1191). However, a LICENSE file is indeed present in the [source's root directory](https://github.com/gkhayes/mlrose). I would appreciate some assistance resolving this issue.

```
```

recipes/mlrose/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - numpy
+    - scipy
+    - scikit-learn
```
beenje, line 21: Those are `run` requirements, not `host`.
Under `host`, for pure python packages, you should only have `python` and `pip`.

recipes/mlrose/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://github.com/gkhayes/mlrose
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE
```
beenje, line 39: The `LICENSE` is in the GitHub repo but not in the sdist archive.
Please open an issue in the upstream repo so it's part of next version.

For now, you should copy that [file](https://github.com/gkhayes/mlrose/blob/master/LICENSE) here (same directory as the meta.yaml file).

beenje (changes requested):

recipes/mlrose/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - scikit-learn <0.21.0
+    - python >=3
+
+test:
+  imports:
+    - mlrose
+  requires:
+    - pip
+  commands:
+    - pip check
```
beenje, line 33: The python package still requires `sklearn` instead of `scikit-learn`. This is why the test is failing. Remove this check (and the requires pip).
The import will be enough.

recipes/mlrose/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python >=3
+    - pip
+  run:
+    - numpy
+    - scipy
+    - scikit-learn <0.21.0
```
beenje, line 24: Does the package really need such an old version of `scikit-learn`?

Fine if it's really the case, but you should open an issue upstream so it's updated to support newer versions.
It will be difficult to install with other packages otherwise.

beenje (changes requested):

recipes/mlrose/meta.yaml
beenje: I just found this [PR](https://github.com/gkhayes/mlrose/pull/50) and a comment that recommends to move to [mlrose-hiive](https://github.com/hiive/mlrose).

Seems like `mlrose` isn't maintained anymore.
Might be better to submit a recipe for `mlrose-hiive`?

onelharrison: @beenje I also found that to be the case. I'll be updating this to use mlrose-hiive.

PR #18120, opened 2022-02-21
Add `editables` recipe

ofek:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/editables/meta.yaml
+++ b/recipes/editables/meta.yaml
@@ -1,0 +1,41 @@
+{% set name = "editables" %}
+{% set version = "0.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/editables-{{ version }}.tar.gz
+  sha256: 6918f16225258f24ef9800c2327e14eded42ddac344e77982380749464024f35
+
+build:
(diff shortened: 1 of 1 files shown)


ofek:
@conda-forge-admin ping conda-forge/help-python please 🙏

recipes/editables/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.5
+  run:
+    - python >=3.5
+
+test:
```
CAM-Gerlach, line 24: You might want to add the actual test runner deps/command(s) here, assuming it has one.

ofek: The tests are not shipped

CAM-Gerlach: They should be in the `git` tarball used as the basis for the build, are they not?

dopplershift: This recipe is using the PyPI tarball, which doesn't include them. Which is fine. The import check and running `pip check` should be enough to catch any packaging issues.

dopplershift (approved):

recipes/editables/meta.yaml
ofek: The sdist is the source and tests aren't there

PR #18133, opened 2022-02-22
Add opensim recipe to conda forge

aymanhab:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/opensim/build.sh
+++ b/recipes/opensim/build.sh
@@ -8,7 +8,8 @@ cmake ../opensim-core-source/dependencies -LAH \
     -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} \
     -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
     -DCMAKE_BUILD_TYPE="Release" \
-    -DSUPERBUILD_ezc3d=ON \
+    -DSUPERBUILD_ezc3d=OFF \
+    -DSUPERBUILD_simbody=OFF \
     -DCMAKE_INSTALL_PREFIX=../opensim_dependencies_install \
     -DCMAKE_INSTALL_LIBDIR="lib"
 make --jobs ${CPU_COUNT}
--- a/recipes/opensim/build.sh
+++ b/recipes/opensim/build.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+cd opensim-core-source
+cd ..
+mkdir build_dep
+cd build_dep
+cmake ../opensim-core-source/dependencies -LAH \
+    -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DSUPERBUILD_simbody=OFF \
+    -DSUPERBUILD_BTK=OFF \
+    -DSUPERBUILD_docopt=OFF \
+    -DCMAKE_INSTALL_PREFIX=../opensim_dependencies_install \
+    -DCMAKE_INSTALL_LIBDIR="lib"
+make --jobs ${CPU_COUNT}
+
+cd ..
+mkdir build
+cd build
+cmake ../opensim-core-source -LAH \
+    -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DOPENSIM_DEPENDENCIES_DIR=../opensim_dependencies_install \
+    -DCMAKE_PREFIX_PATH="${PREFIX}" \
+    -DSIMBODY_HOME="${PREFIX}" \
+    -DBUILD_PYTHON_WRAPPING=ON \
+    -DBUILD_TESTING=OFF \
+    -DCMAKE_INSTALL_PREFIX="${PREFIX}" \
+    -DCMAKE_INSTALL_LIBDIR="lib" 
+make --jobs ${CPU_COUNT}
+make doxygen
+make install
--- a/recipes/opensim/build.sh
+++ b/recipes/opensim/build.sh
@@ -8,9 +8,7 @@ cmake ../opensim-core-source/dependencies -LAH \
     -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} \
     -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
     -DCMAKE_BUILD_TYPE="Release" \
-    -DSUPERBUILD_simbody=OFF \
-    -DSUPERBUILD_BTK=OFF \
-    -DSUPERBUILD_docopt=OFF \
+    -DSUPERBUILD_ezc3d=ON \
     -DCMAKE_INSTALL_PREFIX=../opensim_dependencies_install \
     -DCMAKE_INSTALL_LIBDIR="lib"
 make --jobs ${CPU_COUNT}
--- a/recipes/opensim/meta.yaml
+++ b/recipes/opensim/meta.yaml
@@ -6,7 +6,7 @@ package:
 
 source:
     url: https://github.com/opensim-org/opensim-core/archive/refs/tags/{{ version }}.tar.gz
-    sha256: 121DA8FC9ABE1CC5A37A62FA592A2EE45B6FCA1D42CADF22CD713BB5BD987974
+    sha256: 121da8fc9abe1cc5a37a62fa592a2ee45b6fca1d42cadf22cd713bb5bd987974
     folder: opensim-core-source
     patches:
         # This patch is included in opensim-core after the 4.1 release.
--- a/recipes/opensim/meta.yaml
+++ b/recipes/opensim/meta.yaml
@@ -6,7 +6,7 @@ package:
 
 source:
     url: https://github.com/opensim-org/opensim-core/archive/refs/tags/{{ version }}.tar.gz
-    # sha256: 4cd301f6e2cf3a50af2491dd002f1701c33d89e4085fe7bb7dcfacc5c7c92691
+    sha256: 121DA8FC9ABE1CC5A37A62FA592A2EE45B6FCA1D42CADF22CD713BB5BD987974
     folder: opensim-core-source
     patches:
         # This patch is included in opensim-core after the 4.1 release.
--- a/recipes/opensim/meta.yaml
+++ b/recipes/opensim/meta.yaml
@@ -1,0 +1,60 @@
+{% set version = "4.1" %}
+
+package:
+    name: opensim
+    version: {{ version }}
+
+source:
+    url: https://github.com/opensim-org/opensim-core/archive/{{ version }}.tar.gz
+    sha256: 4cd301f6e2cf3a50af2491dd002f1701c33d89e4085fe7bb7dcfacc5c7c92691
+    folder: opensim-core-source
+    patches:
+        # This patch is included in opensim-core after the 4.1 release.
+        - fix-osx-python-linker-flags.patch
+
+build:
+    number: 0
+    run_exports:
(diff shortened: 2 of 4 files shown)


aymanhab:
@moorepants Here's a revival of the old opensim recipe updated to the latest opensim build settings. I'm working on it on the side since the cycle is too long. Not sure if there's a better way to reproduce locally with some confidence. If you see something totally off please let me know. Thank you

moorepants:
You can build the recipe locally with `conda build .` in the recipe directory.

recipes/opensim/build.sh
```diff
@@ -0,0 +1,32 @@
...
+
+cd opensim-core-source
+cd ..
+mkdir build_dep
+cd build_dep
+cmake ../opensim-core-source/dependencies -LAH \
+    -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT} \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=${MACOSX_DEPLOYMENT_TARGET} \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DSUPERBUILD_ezc3d=ON \
```
moorepants, line 11: I assume superbuild is vendored this package. THere is a feedstock we should use for this, if possible: https://github.com/conda-forge/ezc3d-feedstock. If not, there can't be any clashes if someone also has the conda forge ezc3d installed.

aymanhab:
@moorepants I have been using conda build locally for a while before trying the conda-forge pathway though using VS generator in CMake rather than Ninja. using Ninja the build fails to find libblas which seems a common issue. Any clues/suggestions? Thank you

recipes/opensim/build.sh
moorepants: If it only works with one version of ezc3d, then you can vendor it, but is it install only in the opensim source tree so there would be no clobbering if exc3d is also installed sepearately?

You can pin the version in the requires section, but that will also create incompatibilities.

Ideally CMake is setup so that you can specify the path of all the dependencies, so that we use the conda forge binaries.

moorepants:
> @moorepants I have been using conda build locally for a while before trying the conda-forge pathway though using VS generator in CMake rather than Ninja. using Ninja the build fails to find libblas which seems a common issue. Any clues/suggestions? Thank you

You'll have to post the error messages here for us to help.

aymanhab:
2022-02-23T02:50:27.1032180Z FAILED: simbody/stamp/simbody-build C:/bld/opensim_1645584378296/work/build_dep/simbody/stamp/simbody-build 
2022-02-23T02:50:27.1033042Z cmd.exe /C "cd /D %SRC_DIR%\build_dep\simbody\build && %BUILD_PREFIX%\Library\bin\cmake.exe --build . && %BUILD_PREFIX%\Library\bin\cmake.exe -E touch C:/bld/opensim_1645584378296/work/build_dep/simbody/stamp/simbody-build"
2022-02-23T02:50:27.1033923Z ninja: error: 'libblas.lib', needed by 'Platform/CMakeFiles/PlatformFiles_subdir', missing and no known rule to

recipes/opensim/build.sh
moorepants: From the first comment box: " Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged)." That's the ideal situation. So having `ezc3d==1.4.6` in the requirements is the first thing to try. But yes, opensim's build script would need to support pointing to an arbitrary path for the package.

moorepants:
Have you read this: https://conda-forge.org/docs/maintainer/knowledge_base.html#blas

recipes/opensim/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - cmake >=3.15
+        - ninja  # [win]
+        - doxygen
+        - swig 4.0.2
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - python
+        - numpy
+    host:
+        - libblas
```
moorepants, line 30: This looks right.

recipes/opensim/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+        - simbody >=3.7
+        - docopt.cpp >=0.6.3
+        - python
+        - numpy
+    run:
+        - simbody >=3.7
+        - docopt.cpp >=0.6.3
+        - python
+        - {{ pin_compatible('numpy') }}
+        - libopenblas  # [unix]
```
moorepants, line 40: This shouldn't be here, I don't think.

recipes/opensim/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    build:
+        # CMake 3.15 provides python3 components to locate numpy
+        - cmake >=3.15
+        - ninja  # [win]
+        - doxygen
+        - swig 4.0.2
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - python
+        - numpy
```
moorepants, line 28: I don't think numpy should be in the build section.

moorepants: Python probably doesn't belong here either.

aymanhab:
> Have you read this: https://conda-forge.org/docs/maintainer/knowledge_base.html#blas

Yes, I did yesterday and tried a bunch of combinations/suggestions but went back to the question of why Ninja fails while VS generator succeeds 😕

moorepants:
I personally don't know anything about ninja.

I'm getting this error with conda build locally on linux:
```
CMake Error at CMakeLists.txt:834 (find_package):
  By not providing "Findcasadi.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "casadi", but
  CMake did not find one.

  Could not find a package configuration file provided by "casadi" (requested
  version 3.4.4) with any of the following names:

    casadiConfig.cmake
    casadi-config.cmake

  Add the installation prefix of "casadi" to CMAKE_PREFIX_PATH or set
  "casadi_DIR" to a directory containing one of the above files.  If "casadi"
  provides a separate development package or SDK, be sure it has been
  installed.


-- Configuring incomplete, errors occurred!
See also "$SRC_DIR/build/CMakeFiles/CMakeOutput.log".
See also "$SRC_DIR/build/CMakeFiles/CMakeError.log".
```

moorepants:
Do you have a cmake flag to point directly to libblas?

I don't have windows, so you'd can check in your anaconda or miniconda directory what the name of the libblas is on windows.

recipes/opensim/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    folder: opensim-core-source
+
+build:
+    number: 0
+    run_exports:
+        - {{ pin_subpackage('opensim', max_pin='x.x') }}
+    detect_binary_files_with_prefix: true
+
+requirements:
+    build:
```
moorepants, line 19: I think you need `make` here for the unix builds.

aymanhab:
Great feedback @moorepants Will take another pass with your recommendations. Thanks for all your help 👍

moorepants:
One general question though. I'm pretty sure you all vendor all your dependencies in some way. I don't really know precisely what "superbuild" is or means, but it would be good to know if opensim's cmake setup is designed to allow not vendoring all these package. conda-forge is much like a linux distribution, in the sense that dependencies are not vendored. Also, since you vendor everything you probably want to pin to specific versions of the dependencies, which can also be problematic for conda forge. It'd be good to know this up front so we can have a sense of if it is even possible to get this into conda forge. I don't really know.

moorepants:
On the mac build on the CI here is it failed when building simbody.

```
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/share/doc/simbody/SimbodyAndMolmodelUserGuide.pdf
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/share/doc/simbody/SimbodyTheoryManual.pdf
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/lib/libSimTKsimbody.3.8.dylib
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/lib/libSimTKsimbody.dylib
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/libexec/simbody/simbody-visualizer.app
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/libexec/simbody/simbody-visualizer.app/Contents
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/libexec/simbody/simbody-visualizer.app/Contents/MacOS
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/libexec/simbody/simbody-visualizer.app/Contents/MacOS/simbody-visualizer
-- Installing: $SRC_DIR/opensim_dependencies_install/simbody/libexec/simbody/simbody-visualizer.app/Contents/Info.plist
[ 72%] Completed 'simbody'
[ 72%] Built target simbody
make: *** [all] Error 2
```

It shouldn't be building simbody, as simbody is already packaged in conda forge. I'm pretty sure there is a cmake flag in opensim to build against an existing simbody lib. I think I remember doing that many years ago.

moorepants:
Windows build on the CI is failing with:
```
// The location of the Simbody installation to use; you can change this. Set as empty to let CMake search for Simbody automatically.
SIMBODY_HOME:PATH=C:/bld/opensim_1645584378296/_h_env/Library

// The directory containing a CMake configuration file for spdlog.
spdlog_DIR:PATH=spdlog_DIR-NOTFOUND

ninja: error: loading 'build.ninja': The system cannot find the file specified.


ninja: error: loading 'build.ninja': The system cannot find the file specified.


ninja: error: loading 'build.ninja': The system cannot find the file specified.```

aymanhab:
ok, I didn't know simbody was packaged already, thanks for the pointer. 
Superbuild allows a build process to download source of a dependency and configure/compile/install for use in a parent project. Superbuild(s) work with a source  tree/tag rather than an installed/prebuilt library.

moorepants:
So, is there a way to build opensim without superbuild and dynamically link to existing binaries for all the dependencies?

moorepants:
> ok, I didn't know simbody was packaged already, thanks for the pointer.

Yes, here is the recipe: https://github.com/conda-forge/simbody-feedstock and you can see how it builds against existing dependencies. I think simbody vendors some things (like an old ipopt and some other stuff) but those don't conflict if you also install, for example, conda forge's ipopt.

aymanhab:
I'm trying this out locally on windows now (Using conda-forge packages for simbody and ezc3d). I wouldn't need to deal with libblas then as it's not used directly by opensim.

moorepants:
I was trying to see where all the dependencies come from and see https://github.com/opensim-org/opensim-moco#building-moco. Probably best not to build with moco here. We can call this opensim core. If moco can be built by dynamically linking to a prebuilt opensim, then we should make a separate conda package for moco.

moorepants:
> So, is there a way to build opensim without superbuild and dynamically link to existing binaries for all the dependencies?

Before we waste time trying to get this build here, is there an answer to this question? We may be on a wild goose chase if opensim and it's dependencies are not reasonably separable.

moorepants:
@aymanhab I chatted with Adam Kewley here at Delft and I no longer think this PR is worth pursuing. Unless there is a very clear way to decouple all dependencies and link dynamically to conda forge binary versions of the dependencies for all OSes, this probably won't work. Building your own isolated conda package for opensim like others have done, may be the only solution for now.

aymanhab:
@moorepants It is possible to build OpenSim without moco and that saves us a ton of depndencies so a reasonable place to start at conda-forge packaging, however longer term our users will need it. I'm leaning to go back to the old plan to have a conda package in our own channel that replicates OpenSim+Moco just with the added flexibility of having different python/numpy variants and slowly "Vendor" out depndencies here.

aymanhab:
Glad we reached the same conclusion @moorepants 😃

moorepants:
After browsing this https://github.com/opensim-org/opensim-core/issues/2869 it also seems to messy. If you want to have a widely compatible opensim you need to make the dependency management more flexible.

aymanhab:
Thanks for looking into this. In practice, we had an old packaging for opensim-rl on conda that people in reinforcement learning have been using for years and it included simbody and all dependencies within it (no vendoring), the only practical issue using it had to do with python/numpy combinations that conflicted with tensorflow and other recent ML packages and it didn't update with OpenSim, so while not perfect the PR here https://github.com/opensim-org/conda-opensim/pull/8 should be good enough so we may pick up the discussion there.

moorepants:
> the only practical issue using it had to do with python/numpy combinations that conflicted with tensorflow

Yes, that is exactly why trying to build into this conda forge system is a good idea.

PR #9162, opened 2019-08-15
Add UCX

jakirkham:
This builds a package for [UCX]( https://github.com/openucx/ucx ), which provides an optimized communication layer that can be leveraged by other libraries like OpenMPI and [Dask]( https://blog.dask.org/2019/06/09/ucx-dgx ). As such this can be incredibly useful for enabling high speed networking in HPC environments.

This relies on a few CDTs that have already been integrated to provide `libibverbs`, `libnuma`, and `librdma`.

Currently this is only building the CPU part of the package. Would like to sort out the GPU bits in the feedstock. This will rely on solving issue ( https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/237 ). That said, UCX provides a good use case for that work.

Checklist

- [x] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml#L57-L66) for an example)
- [x] Source is from official source
- [x] Package does not vend other packages
- [x] Build number is 0
- [x] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your
      recipe (see
      [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos)
      for more details)
- [x] GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there

PR #18166, opened 2022-02-25
add dtc-package generating dtc, libfdt and pylibfdt

timsnyder:
mostly copied from LiteX-Hub's recipe but merged in conda-forgeness
* split into 3 packages dtc, libfdt, pylibfdt.
   * licenses are different between the three
   * minmize impact of python linkage & dep
* added optional yaml dependency to dtc

Checklist
- [X] Build number is 0.


recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+       Device Tree binary format.
+      license: GPL-2.0-or-later OR BSD-2-Clause
+      license_family: BSD
+      license_file: 
+        - GPL
+        - BSD-2-Clause
+        - README.license
+  - name: libfdt
+    script: install.sh  # [not win]
+    script: install.bat # [win]
```
timsnyder, line 87:
```suggestion
    script: install.sh   # [not win]
    script: install.bat  # [win]
```

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+        - {{ pin_subpackage("libfdt", max_pin="x.x") }}
+      missing_dso_whitelist:
+        - /lib64/libc.so.6
+    requirements:
+      build:
+        - make
+    test:
+      commands:
+        - test -f $PREFIX/include/libfdt.h                    # [not win]
+        - test -f $PREFIX/lib/libfdt-{{version}}${SHLIB_EXT}  # [not win]
```
timsnyder, line 99:
```suggestion
        - test -f $PREFIX/lib/libfdt-{{ version }}${SHLIB_EXT}  # [not win]
```

timsnyder:
Seems like the linter is complaining about there not being a top-level about when all of the split packages have their own about sections.  This is already captured in conda-smithy#1124 by @jakirkham and I'll just do what he ended up doing in #9162 and add the toplevel about section since I know it won't be used.

timsnyder:
Since dtc is a very linux-specific thing, I was thinking it might only make sense on linux but the fact that flex and bison aren't available on windows makes it definitely not buildable on Windows.  Will set it up to only build on unixen.

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+      license_file: 
+        - GPL
+        - BSD-2-Clause
+        - README.license
+
+
+about:
+  home: https://git.kernel.org/pub/scm/utils/dtc/dtc.git
+  summary: 'Toplevel for subpackages dtc, libfdt and pylibfdt'
+  license: 'Depends on the subpackage'
```
timsnyder, line 114:
```suggestion
  license: GPL-2.0-or-later
```

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+      license_file: 
+        - GPL
+        - BSD-2-Clause
+        - README.license
+
+
+about:
+  home: https://git.kernel.org/pub/scm/utils/dtc/dtc.git
+  summary: 'Toplevel for subpackages dtc, libfdt and pylibfdt'
+  license: GPL-2.0-or-later
```
timsnyder, line 114:
```suggestion
  license: GPL-2.0-or-later
  license_family: GPL
  license_file:
    - GPL
    - README.license
```

recipes/dtc/conda-forge.yml
```diff
@@ -0,0 +1,2 @@
+win:
+  enabled: False
```
timsnyder, line 2: I couldn't remember whether I need to do this manually here or if it happens because I have `build.skip: True  # [win]` in my meta.yaml once we convert this into a feedstock.  I'm beginning to think I should delete this file before we merge this.

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,122 @@
...
+    - yaml
+  run:
+    - python {{ python }}
+
+outputs:
+  - name: dtc
+    script: install.sh   # [not win]
+    build:
+      missing_dso_whitelist:
+        - /lib64/libc.so.6
```
timsnyder, line 37: I also don't remember having to specify `/lib64/libc.so.6` in the missing_dso_whitelist before.  I searched the other feedstocks and found several that do this.  Does this mean that I'm directly linking to the system libc instead of the one in the compiler sysroot?  I'm rereading https://github.com/conda/conda-build/blob/0a6efdcfa02484954e7a3fdf5c2ec4bb3cb6682e/conda_build/post.py.  It's been awhile.

timsnyder: Looking through the other recipes for CONDA_BUILD_SYSROOT, I see a bunch of packages adding `-isysroot $CONDA_BUILD_SYSROOT` into CFLAGS CXXFLAGS and LDFLAGS.  I tried doing that and removing the missing_dso_whitelist and I get the errors again.  They look like this:
```
Warning: rpath /home/conda/staged-recipes/build_artifacts/dtc-package_1645834285171/_build_env/lib is outside prefix /home/conda/staged-recipes/build_artifacts/dtc-package_1645834285171/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold (removing it) for /home/conda/staged-recipes/build_artifacts/dtc-package_1645834285171/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/libfdt-1.6.1.so
WARNING :: Failed to get_static_lib_exports(/home/conda/staged-recipes/build_artifacts/dtc-package_1645834285171/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/libfdt.a)
Unknown format
  ERROR (libfdt,lib/libfdt-1.6.1.so): /lib64/libc.so.6 not found in packages, sysroot(s) nor the missing_dso_whitelist.
```

@conda-forge/help-c-cpp what's the deal with libc and the missing_dso_whitelist?  At one point, I thought that we expected stuff to link against the sysroot libc at build time and then at runtime link the system libc.  Should I just leave the missing_dso_whitelist in my recipe or do I need to modify the build so that it links differently?

timsnyder: I think I got this figured out after looking more at https://github.com/conda/conda-build/blob/77463386355f3a726437311d3fa1f39a59afc2e9/conda_build/post.py#L1062-L1072, I realized that the split packages need to list the compiler in the `requirements.build` section of their output so that the sysroot will be found in build_prefix and libc will be found to be okay by the overlinking logic.  Testing this locally and will push an update to remove the missing_dso_whitelist and also remove the `-isysroot` arguments because I'm pretty sure that those are baked into the compilers correctly for linux.  Perhaps they are needed for osx though?

timsnyder:
@conda-forge/staged-recipes @conda-forge/help-c-cpp this one is ready to merge if I can get a review.  Thanks.

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+    script: install.sh   # [not win]
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - make
+        - swig
+      host:
+        - python {{ python }}
+      run:
+        - python {{ python }}
```
carterbox, line 67: pylibfdt statically links to libfdt? Is there a compile option to link dynamically?

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+        - {{ compiler('c') }}
+        - make
+      host:
+        - {{ pin_subpackage("libfdt", exact=True) }}
+        - yaml
+      run:
+        - {{ pin_subpackage("libfdt", max_pin="x.x") }}
+    test:
+      commands:
+        - dtc --help
```
carterbox, line 46: What about fdtget, fdtdump, etc.? Can you not test those binaries too?

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+      run_exports:
+        - {{ pin_subpackage("libfdt", max_pin="x.x") }}
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - make
+    test:
+      commands:
+        - test -f $PREFIX/include/libfdt.h                    # [not win]
+        - test -f $PREFIX/lib/libfdt-{{ version }}${SHLIB_EXT}  # [not win]
```
carterbox, line 94: There are two additional headers being shipped.

Test for .so .so.1 .a too!

carterbox (changes requested):
Hi @timsnyder,

You can check the outputs of these builds by downloading the build artifacts from azure. https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=465595&view=artifacts&pathAsName=false&type=publishedArtifacts

I noticed that there were a few more things being shipped compared to what you tested for, so double check that you're only shipping what you expect to be included!

recipes/dtc/meta.yaml
timsnyder: The extension links against `libfdt.a` when creating (for example) `lib/python3.8/site-packages/_libfdt.cpython-38-x86_64-linux-gnu.so`.  So, the pylibfdt package, does not depend at runtime on the libfdt package.  I don't see a way to change the linkage but I'm not an expert when it comes to building python extensions.

recipes/dtc/meta.yaml
timsnyder: Sure.  I'll add more tests.

recipes/dtc/meta.yaml
timsnyder: Done

timsnyder:
@carterbox Thanks for the review.  I added more tests and assuming that I didn't break the osx build, I think this should be ready to merge if you can give it another pass.

recipes/dtc/meta.yaml
```diff
@@ -0,0 +1,166 @@
...
+        - cp util.c util.h version_gen.h tests
+        # use Makefile.tests directly so that binaries won't be built but used from installed package
+        # NO_PYTHON because we test that in pylibfdt
+        # VECHO=: to enable Makefile.tests to stand on it's own
+        # LDLIBS=-lfdt to correctly find installed libfdt instead of having in-tree library an explicit dependency listed in the link command
+        # LIBDL= needs to also include -lfdt because of how it is used in rules to override the  builtin and not use LDLIBS
+        # TESTS_PREFIX=./ to work around 'cd $TEST_PREFIX' in check recipe
+        # TEST_BINDIR="$PREFIX/bin" because testutils.sh hardcodes the paths to binaries, defaulting to ../
+        - 'NO_PYTHON=1 TEST_BINDIR="$PREFIX/bin" make -C tests -f Makefile.tests VECHO=: LDLIBS=-lfdt tests_clean check TESTS_PREFIX=./ LIBDL="-lfdt -ldl"'  # [linux]
+        - 'NO_PYTHON=1 TEST_BINDIR="$PREFIX/bin" make -C tests -f Makefile.tests VECHO=: LDLIBS=-lfdt tests_clean check TESTS_PREFIX=./ LIBDL=-lfdt'  # [osx]
```
carterbox, line 79: Could you explain what this test is for? It looks like you already tested all of the binaries in previous tests (either by running help or checking for their existence).

recipes/dtc/meta.yaml
carterbox: Lemme look into this whilst you fix the broken OSX build. Otherwise, we'll just ship it statically linked.

carterbox (changes requested):

timsnyder:
'make check' is running the actual test suite in the package to make sure
that more functionality works in binaries and that the libraries can be
linked against once installed.

We could probably remove the existence tests now but I left them in since
they aren't hurting anything.

timsnyder:
I'll need to look into the osx failure later this evening.

timsnyder:
In dtc, `tests/trees.S` uses GCC-specific assembly directives  to build up test collateral binary files into a utility called dumptrees and then spit them out into various .dtb files that the tests manipulate.  I looked briefly into getting that to build using clang and I don't think it will be trivial.  I'm going to do the additional testing only on Linux where it is known to work and only do existence and import testing on osx.

timsnyder:
@carterbox any luck with figuring out how to have the python package use the dynamic library from libfdt?  I'm poking around trying a few things this morning.

timsnyder:
I don't see a way to make the python module use the dynamic libfdt.  @carterbox can we just ship the statically linked one?

timsnyder:
@carterbox , I saw an email with a big comment that I guess you deleted about how pylibfdt is being built.  It looks like the upstream folks that work on dtc are also noticing and working on some fixes to get pylibfdt to depend on libfdt in https://www.spinics.net/lists/devicetree-compiler/msg03891.html from about a month ago.  I'll keep an eye on that and update the recipe when the next version is cut that includes these improvements (hopefully).

carterbox:
@timsnyder, not deleted, just [buried in a thread](https://github.com/conda-forge/staged-recipes/pull/18166#discussion_r821911058). 😄

> I'll keep an eye on that and update the recipe when the next version 

Sounds good.

PR #18181, opened 2022-02-28
Add `rich-click`

ewels:
Checklist
- [x] Build number is 0.

<details>

Thank you very much for putting in this recipe PR!

This repository is very active, so if you need help with
a PR or once it's ready for review, please let the right people know.
There are language-specific teams for reviewing recipes.

Currently available teams are:
- python `@conda-forge/help-python`
- python/c hybrid `@conda-forge/help-python-c`
- r `@conda-forge/help-r`
- java `@conda-forge/help-java`
- nodejs `@conda-forge/help-nodejs`
- c/c++ `@conda-forge/help-c-cpp`
- perl `@conda-forge/help-perl`
- Julia `@conda-forge/help-julia`
- ruby `@conda-forge/help-ruby`

If your PR doesn't fall into those categories please contact
the full review team `@conda-forge/staged-recipes`.

Due to GitHub limitations first time contributors to conda-forge are unable
to ping these teams. You can [ping the team][1] using a special command in
a comment on the PR to get the attention of the `staged-recipes` team. You can
also consider asking on our [Gitter channel][2] if your recipe isn't reviewed promptly.

[1]: https://conda-forge.org/docs/maintainer/infrastructure.html#conda-forge-admin-please-ping-team
[2]: https://gitter.im/conda-forge/conda-forge.github.io

</details>

### DIFF
--- a/recipes/rich-click/meta.yaml
+++ b/recipes/rich-click/meta.yaml
@@ -11,6 +11,7 @@ source:
 
 build:
   number: 0
+  noarch: python
   script: "{{ PYTHON }} -m pip install . -vv"
 
 requirements:
--- a/recipes/rich-click/meta.yaml
+++ b/recipes/rich-click/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "rich-click" %}
+{% set version = "1.1.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ewels/rich-click/archive/v{{ version }}.tar.gz
+  sha256: 6204e3ae465abca7cf630649d9c9b7853d4ec404c85eeab03b8af77db91616f3
+
+build:
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - rich >=10
+    - click >=7.0
+
+test:
+  imports:
+    - rich_click
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://github.com/ewels/rich-click
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+  summary: 'Format click help output nicely with rich'
+  description: |
+    Provides attractive help output from click,
+    formatted with rich, with minimal customisation required.
+  doc_url: https://github.com/ewels/rich-click
+  dev_url: https://github.com/ewels/rich-click
+
+extra:
+  recipe-maintainers:
+    - ewels
--- a/recipes/rich-click/meta.yaml
+++ b/recipes/rich-click/meta.yaml
@@ -19,8 +19,8 @@ requirements:
     - pip
   run:
     - python
-    - rich >=10
-    - click >=7.0
+    - rich >=10,<12
+    - click >=7,<9
 
 test:
   imports:
--- a/recipes/rich-click/meta.yaml
+++ b/recipes/rich-click/meta.yaml
@@ -16,10 +16,10 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.6
     - pip
   run:
-    - python
+    - python >=3.6
     - rich >=10,<12
     - click >=7,<9
 


ewels:
Ready for review @conda-forge/help-python 🙏🏻

recipes/rich-click/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - rich >=10
```
saraedum, line 22: Maybe these should say `>=10,<11` and `>=6?

recipes/rich-click/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - click >=7.0
```
saraedum, line 23: I don't think you need the `.0` here. So, `>=7,<8` maybe?

recipes/rich-click/meta.yaml
ewels: Why less than 11?

recipes/rich-click/meta.yaml
ewels: Ah I went back and forth on the `.0`! I can certainly remove that 👍🏻 

What's the reason for the `<8` though? It works fine with v8, just needs at least v7 or later..

recipes/rich-click/meta.yaml
saraedum: I mean, `>=10,<unreleased version`. Since you don't actually know whether it is going to break with later versions.

recipes/rich-click/meta.yaml
saraedum: Sorry, `<9`, i.e., the next unreleased version.

recipes/rich-click/meta.yaml
saraedum: So, then `<12` actually. (I had not checked what's the current version.)

ewels: Ah I see - is that common practice? I don't think that I've ever done that with other packages 🤔 I'd be a bit worried that I'd forget and cause issues for people down the line even if there are no compatibility issues..

saraedum: Since 12 is supposedly going to contain some breaking changes, you can't really know whether rich-click 1.1.1 is going to work with rich 12. Without this pin, people might be able to install a broken version of rich-click in the future and then we have to mark these builds as broken or patch the metadata.

saraedum:
Should this package be `noarch: python`? See https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#architecture-independent-packages

recipes/rich-click/meta.yaml
ewels: Fair enough 👍🏻

In this particular case, @willmcgugan pumps out the major semver versions on a pretty regular basis, which is why I'm a bit nervous (it's very likely that several major versions might pass between any releases of rich-click). But I'll try to remember - I guess I can bump the conda-forge build number and change the requirements if that happens..

```suggestion
    - rich >=10,<12
```

recipes/rich-click/meta.yaml
ewels:
```suggestion
    - click >=7.0,<9
```

saraedum:
```suggestion
    - click >=7,<9
```

recipes/rich-click/meta.yaml
saraedum: > I guess I can bump the conda-forge build number and change the requirements if that happens..

Yes, that's the safe approach :)

ewels:
For the `noarch` thing, I left it because there [is a PR](https://github.com/ewels/rich-click/pull/13) for `rich-click` likely to be merged imminently which will add a `console_script`.

I was reading the [conda-forge docs](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python) about it which made me nervous so I thought maybe I should play it safe and leave it out. Specifically:

> Scripts argument in setup.py is not used

Re-reading this I see the following line though:

> If `console_scripts` `entry_points` are defined in `setup.py` or `setup.cfg`, they are also listed in the build section of `meta.yaml`

Which I guess covers this use case. I'm not entirely sure what it means by listing in the build section, but I guess I can look into that again when I have a release with the console_script stuff included.

recipes/rich-click/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ewels/rich-click/archive/v{{ version }}.tar.gz
+  sha256: 6204e3ae465abca7cf630649d9c9b7853d4ec404c85eeab03b8af77db91616f3
+
+build:
+  number: 0
```
ewels, line 13:
```suggestion
  number: 0
  noarch: python
```

recipes/rich-click/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
```
ewels, line 22:
```suggestion
    - python >=3.6
```

saraedum:
Yes, I would imagine that your use case is covered. You'd just need to add a line like this to the build section: https://github.com/conda-forge/svgdigitizer-feedstock/blob/master/recipe/meta.yaml#L13

ewels:
Ah, many thanks for the reference @saraedum - super helpful ⭐

saraedum:
Thanks for your contribution @ewels. I am looking forward to trying out rich-click in some of my Python packages.

ewels:
Thanks for the quick and helpful review @saraedum!

PR #18187, opened 2022-03-01
Add dash-mantine-components

snehilvj:
Checklist
- [x] Build number is 0.


snehilvj:
@conda-forge/staged-recipes ready for review.

recipes/dash_mantine_components/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  run:
+    - python >=3.6
+    - setuptools
+    - dash >=2.0.0
+    - dash-core-components ==2.0.0
+    - dash-html-components ==2.0.0
+    - dash-table ==5.0.0
+    - flask >=1.0.4
+    - flask-compress
+    - importlib-metadata ==4.8.3  # [py<37]
```
chrisburr, line 31: Selectors can't be used with `noarch: python` recipes. As conda-forge has dropped support for Python 3.6 and it's already end-of-life I'd be inclined to suggest changing the minimum dependency to `python >=3.7`

snehilvj: Thank for the feedback. I have made the corrections.

snehilvj:
This is ready for review @conda-forge-admin , please ping team

recipes/dash_mantine_components/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/dash_mantine_components-{{ version }}.tar.gz
+  sha256: 8a9a6f038cfa130f70043dbcd88d1baab1d4f90c9a8c6725804565eed489996f
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
BastianZim, line 14:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

recipes/dash_mantine_components/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - python >=3.7
+    - setuptools
+    - dash >=2.0.0
+    - dash-core-components ==2.0.0
+    - dash-html-components ==2.0.0
+    - dash-table ==5.0.0
+    - flask >=1.0.4
+    - flask-compress
+    - importlib-metadata ==4.8.3
+    - plotly >=5.0.0
```
BastianZim, line 32: I cannot find any of these in the reqs, are they really needed? 

Ref https://github.com/snehilvj/dash-mantine-components/blob/33476cd1a3c8eb72e72ff22eb29bf33f26faf3ec/setup.py#L21

BastianZim (changes requested):

recipes/dash_mantine_components/meta.yaml
snehilvj: This library is not a stand alone library but an extension pack for the library `dash`. There is an internal check for `dash` when `dash-mantine-components` is imported. Hence I had to mention them here.

BastianZim: Yes but in that case, only `dash` should be required, no? Also, conda-forge tries to be in sync with upstream so it would be good to have the dependencies listed there as well then.

carterbox: @snehilvj, the point is that the dash package [already depends on these other packages](https://github.com/conda-forge/dash-feedstock/blob/main/recipe/meta.yaml) such as `flask`, `plotly` so if your package doesn't import `flask`, `plotly`, (or other packages directly), then they shouldn't be listed here. Otherwise, if a future `dask` swaps `flask` for some other package, then users of your package will still be installing `flask` when it is not needed. I don't see any evidence that your extension uses `flask` directly.

carterbox (changes requested):

recipes/dash_mantine_components/meta.yaml
snehilvj: Sorry for the late resolution, but made the changes now.

snehilvj: I had to add
```
- dash-core-components ==2.0.0
- dash-html-components ==2.0.0
- dash-table ==5.0.0
```
otherwise the tests would still fail.

BastianZim: Yes, that's ok. The dash feedstock only has the minimum tools required, anything else you need to add yourself.

snehilvj: So is the PR good to merge now?

BastianZim: Yes just giving it a final check.

BastianZim (approved):

PR #18246, opened 2022-03-06
PM4Py Conda Forge recipe

fit-alessandro-berti:
PM4Py - Process Mining for Python

Developed and maintained by Fraunhofer FIT

https://pm4py.fit.fraunhofer.de/

### DIFF
--- a/recipes/pm4py/meta.yaml
+++ b/recipes/pm4py/meta.yaml
@@ -1,0 +1,67 @@
+{% set name = "pm4py" %}
+{% set version = "2.2.19.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/ab/19/aa8e3014e9d606a9cf81dc435c4fd39aa89a3f57c6d64c55c730c36685b3/pm4py-2.2.19.1.tar.gz
+  sha256: 2104e253e16720706896df2cb359a85fef09d219da923b0a093ab307d4dc4d17
+
+build:
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - pip
+    - numpy
+    - pandas
+    - deprecation
+    - lxml
+    - scipy
+    - graphviz
+    - intervaltree
+    - matplotlib-base
+    - networkx
+    - tqdm
+    - cvxopt
+  run:
+    - python
+    - numpy
+    - pandas
+    - deprecation
+    - lxml
+    - scipy
+    - graphviz
+    - intervaltree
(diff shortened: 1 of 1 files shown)


recipes/pm4py/meta.yaml
```diff
@@ -0,0 +1,67 @@
+{% set name = "pm4py" %}
+{% set version = "2.2.19.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/ab/19/aa8e3014e9d606a9cf81dc435c4fd39aa89a3f57c6d64c55c730c36685b3/pm4py-2.2.19.1.tar.gz
```
ngam, line 9: Use a more standard approach, either the tar.gz from pypi or the source code from github so that the bots can automatically update it

recipes/pm4py/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  sha256: 2104e253e16720706896df2cb359a85fef09d219da923b0a093ab307d4dc4d17
+
+build:
+  number: 0
```
ngam, line 13: Let's make it a noarch pls
```suggestion
  number: 0
  noarch: python
```

recipes/pm4py/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - graphviz
+    - intervaltree
+    - matplotlib-base
+    - networkx
+    - tqdm
+    - cvxopt
+
+test:
+  imports:
+    - pm4py
```
ngam, line 49: Please add a pip check here.

recipes/pm4py/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+about:
+  home: https://github.com/pm4py/pm4py-core
```
ngam, line 52: Is there another derivative of this? Why is it called pm4py-core in the source but not here?

ngam (changes requested):

recipes/pm4py/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    PM4Py is a python library that supports (state-of-the-art) process mining algorithms in python.
+    It is completely open source and intended to be used in both academia and industry projects.
+    PM4Py is a product of the Fraunhofer Institute for Applied Information Technology.
+  doc_url: https://pm4py.fit.fraunhofer.de/
+  dev_url: https://pm4py.fit.fraunhofer.de/static/assets/api/2.2.18/index.html
+
+extra:
+  recipe-maintainers:
+    - fit-alessandro-berti
+    - fit-sebastiaan-van-zelst
```
ngam, line 67: Both maintainers need to explicitly consent here to be listed. Also ping appropriate team when ready.

PR #18278, opened 2022-03-09
Adding 'baobzi' package

blackwer:
Checklist
- [x] Build number is 0.


blackwer:
I've left the vendoring info unchecked because this package uses `cmake`'s `FetchContent` module to download `msgpack-c`, `Catch2`, and `Eigen`. All of which are header based or statically linked, but I wasn't sure what to do about the license info regardless.

blackwer:
@conda-forge-admin, please ping team/help-c-cpp

This seems ready, aside from my license questions. I doubt they're a problem, but
> I've left the vendoring info unchecked because this package uses cmake's FetchContent module to download msgpack-c, Catch2, and Eigen. All of which are header based or statically linked, but I wasn't sure what to do about the license info regardless.

blackwer:
@conda-forge-admin, please ping help-c-cpp

carterbox:

carterbox (changes requested):
@blackwer, to answer your questions about licenses. Eigen and Catch2 are header only, and Catch2 is a testing framework library?, so if you aren't shipping the tests, then you don't need to worry about the licenses.

msgpack-c is not header-only. You need to patch the build process to dynamically link against our shared libraries on conda-forge.

blackwer:
@carterbox Thanks for the info!

> @blackwer, to answer your questions about licenses. Eigen and Catch2 are header only, and Catch2 is a testing framework library?, so if you aren't shipping the tests, then you don't need to worry about the licenses.
Not shipping the tests, so this is all clear!

> msgpack-c is not header-only. You need to patch the build process to dynamically link against our shared libraries on conda-forge.
I'm using the header-only version of msgpack-c(++) from tag 3.3.0. So this should also be fine.

carterbox (changes requested):
This package is shipping 3 separate language APIs simultaneously: c++, python, and julia. The python API adds a python version specific runtime requirement which creates 3 build variants, one for each python version. It also causes anyone who wants to link against the c++/julia API to require Python to be installed. I'm not familiar with julia, but I assume wrapping a C library with Julia doesn't imply a Julia version requirement.

Is the compilation of the Python API separable from the C++ API? i.e. Can you build the C++ library alone, then build each of the Python APIs by linking against the prebuilt C++ library? If so, add a new folder for `libbaobzi` and create a recipe that only builds the C++ library. Then the `baobzi` recipe depends on that package and just builds the python wrapper. This is the preferable approach.

If not, then we need to add an [outputs section](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#outputs-section) which separates the python/c-julia components/requirements into separate packages from the same recipe. An example of this is [libopencv](https://github.com/conda-forge/opencv-feedstock/blob/main/recipe/meta.yaml), which, as noted in that recipe, doesn't actually keep the files separate like the approach above, but keeps the dependencies separate.

blackwer:
> I'm not familiar with julia, but I assume wrapping a C library with Julia doesn't imply a Julia version requirement.

There is no Julia version requirement, afaik.

> Is the compilation of the Python API separable from the C++ API? i.e. Can you build the C++ library alone, then build each of the Python APIs by linking against the prebuilt C++ library? If so, add a new folder for `libbaobzi` and create a recipe that only builds the C++ library. Then the `baobzi` recipe depends on that package and just builds the python wrapper. This is the preferable approach.

The python API isn't actually compiled. It's just a thin wrapper around a C API using `cdll`. I don't use `cython` or any other direct language interop. The C API is a thin wrapper around some C++ templates, and the C++ API is a completely independent wrapper around those templates. I just go ahead and compile this all into the same shared object.

> If not, then we need to add an [outputs section](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#outputs-section) which separates the python/c-julia components/requirements into separate packages from the same recipe. An example of this is [libopencv](https://github.com/conda-forge/opencv-feedstock/blob/main/recipe/meta.yaml), which, as noted in that recipe, doesn't actually keep the files separate like the approach above, but keeps the dependencies separate.

Right now I'm happy actually just providing python support. But if it's significantly better to resolve all of this now rather than later, I can take a look at the `libopencv` recipe when I have some free cycles next week

carterbox:
> The python API isn't actually compiled. It's just a thin wrapper around a C API using cdll.

CDLL from the [built-in ctypes](https://docs.python.org/3/library/ctypes.html) module? That actually makes things even simpler because then your python wrapper probably qualifies for [noarch](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python), so it can be built once and used with `.dylib` or `.so`.

PR #18303, opened 2022-03-11
Adding DBSTEP

luchini18:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/dbstep/meta.yaml
+++ b/recipes/dbstep/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   noarch: python
   number: 0
-  script: "{{ PYTHON }} -m pip install . -vv"
+  script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
   host:
--- a/recipes/dbstep/meta.yaml
+++ b/recipes/dbstep/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "dbstep" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: bdd560fdd0bdf689a795f167d2b2511fec5ec6dc7140fb604088e5dcb2b2b12d
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - python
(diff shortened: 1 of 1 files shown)


luchini18:

recipes/dbstep/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+requirements:
+  host:
+    - pip
+    - python >=3.6
+  run:
+    - cclib
+    - numba 0.52
+    - numpy 1.19
+    - python >=3.6
+    - scipy
```
BastianZim, line 26: Upstream does not list the versions. Are you sure that they are required?

Ref: https://github.com/patonlab/DBSTEP/blob/dffbe0ccb5e2935ab424683b50b794fe2bf10641/setup.py#L23

Additionally, such strict bounds will make it very difficult for users to install it in an environment with other packages. Can they be relaxed?

recipes/dbstep/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: bdd560fdd0bdf689a795f167d2b2511fec5ec6dc7140fb604088e5dcb2b2b12d
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
```
BastianZim, line 15:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

BastianZim (changes requested):

recipes/dbstep/meta.yaml
luchini18: I think the numba version requirement can be removed but it requires specific versioning for numpy.

> error: numpy 1.22.3 is installed but numpy<1.22,>=1.18 is required by {'numba'}

Is this something I should worry about?

BastianZim: Do you have a link to the logs? Normally, the conda solver should take care of that. 

If these versions are correct, they should be reflected upstream as well.

luchini18: Okay, if the conda solver is able to resolve this I'll take them out.

BastianZim: Any version listed here should only be listed because you need a special functionality. 

Interoperability will be handled by the conda solver.

luchini18: Okay, I see, thank you!

PR #18313, opened 2022-03-12
Adding Polyglotdb with dependencies conch_sounds and pyraat 

michaelhaaf:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/conch_sounds/meta.yaml
+++ b/recipes/conch_sounds/meta.yaml
@@ -20,13 +20,13 @@ requirements:
     - pip
     - pyraat
     - pytest
-    - python
+    - python >=3.6
   run:
     - librosa >=0.8.0
     - numpy
     - praatio >=5.0
     - pyraat
-    - python
+    - python >=3.6
     - scipy
 
 test:
--- a/recipes/conch_sounds/meta.yaml
+++ b/recipes/conch_sounds/meta.yaml
@@ -16,15 +16,15 @@ build:
 
 requirements:
   host:
-    - librosa >=0.8.0
     - pip
-    - pyraat
     - pytest
     - python >=3.6
+    - librosa >=0.8.0,=0.8.*
+    - pyraat
   run:
-    - librosa >=0.8.0
+    - librosa >=0.8.0,=0.8.*
     - numpy
-    - praatio >=5.0
+    - praatio >=5.0,=5.*
     - pyraat
     - python >=3.6
     - scipy
--- a/recipes/conch_sounds/meta.yaml
+++ b/recipes/conch_sounds/meta.yaml
@@ -1,0 +1,50 @@
+{% set name = "conch_sounds" %}
+{% set version = "0.3.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/conch-sounds-{{ version }}.tar.gz
+  sha256: 8dffbbf154baec8f265eaa0205b26f6edfcf0a39f94bb588cc540c8aa244fe30
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - librosa >=0.8.0
+    - pip
+    - pyraat
+    - pytest
+    - python
+  run:
+    - librosa >=0.8.0
+    - numpy
+    - praatio >=5.0
+    - pyraat
+    - python
+    - scipy
+
+test:
+  imports:
+    - conch
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/mmcauliffe/Conch
+  summary: Analyze acoustic similarity in Python
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - michaelhaaf
+    - mmcauliffe
--- a/recipes/polyglotdb/meta.yaml
+++ b/recipes/polyglotdb/meta.yaml
@@ -2,8 +2,8 @@
 {% set version = "1.2.1" %}
 
 package:
-  name: "{{ name|lower }}"
-  version: "{{ version }}"
+  name: {{ name|lower }}
+  version: {{ version }}
 
 source:
   url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
--- a/recipes/polyglotdb/meta.yaml
+++ b/recipes/polyglotdb/meta.yaml
@@ -1,0 +1,78 @@
+{% set name = "polyglotdb" %}
+{% set version = "1.2.1" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
+  sha256: 89d2b4a19595a7bc14d1fba124afc63454baa9678275cd093d091d0430fc36c2
+
+build:
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - conch_sounds
+    - influxdb
+    - librosa
+    - neo4j
+    - pip
+    - praatio
+    - python
+    - requests
+    - textgrid
+    - tqdm
+  run:
+    - conch_sounds
+    - influxdb
+    - librosa
+    - neo4j-driver
+    - praatio
+    - python
+    - requests
+    - textgrid
+    - tqdm
+
+test:
+  imports:
+    - polyglotdb
+    - polyglotdb.acoustics
+    - polyglotdb.acoustics.formants
+    - polyglotdb.acoustics.pitch
+    - polyglotdb.acoustics.vot
+    - polyglotdb.client
+    - polyglotdb.corpus
+    - polyglotdb.databases
+    - polyglotdb.io
+    - polyglotdb.io.enrichment
+    - polyglotdb.io.exporters
+    - polyglotdb.io.importer
+    - polyglotdb.io.inspect
+    - polyglotdb.io.parsers
+    - polyglotdb.io.types
+    - polyglotdb.query
+    - polyglotdb.query.annotations
+    - polyglotdb.query.annotations.attributes
+    - polyglotdb.query.base
+    - polyglotdb.query.discourse
+    - polyglotdb.query.lexicon
+    - polyglotdb.query.metadata
+    - polyglotdb.query.speaker
+    - polyglotdb.syllabification
+
+about:
+  home: "https://github.com/MontrealCorpusTools/PolyglotDB"
+  license: MIT
+  license_family: MIT
+  license_file: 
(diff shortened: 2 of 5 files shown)


mmcauliffe:
I am willing to be listed as a maintainer, thanks!

michaelhaaf:
@conda-forge/staged-recipes recipe looks ready to go!

recipes/conch_sounds/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - librosa >=0.8.0
+    - pip
+    - pyraat
```
BastianZim, line 21: Are you sure that all these extra dependencies are required for building the package?

recipes/conch_sounds/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - pytest
+    - python >=3.6
+  run:
+    - librosa >=0.8.0
+    - numpy
+    - praatio >=5.0
```
BastianZim, line 27: Upstream uses `~=` do you know which one is correct? (Note: conda-build doesn't support `~=` so that would need to be translated into the respective `>=,=` version).

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,78 @@
+{% set name = "polyglotdb" %}
+{% set version = "1.2.1" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
```
BastianZim, line 6:
```suggestion
  name: {{ name|lower }}
  version: {{ version }}
```

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
```
BastianZim, line 9:
```suggestion
  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+  sha256: 89d2b4a19595a7bc14d1fba124afc63454baa9678275cd093d091d0430fc36c2
+
+build:
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
```
BastianZim, line 14:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

Would it be possible to convert the `scripts` tag upstream to `entry points`?
Then this could be noarch.

Ref: https://github.com/MontrealCorpusTools/PolyglotDB/blob/f021a403566388a9e13c76c2d727979baf4264e1/setup.py#L102

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - conch_sounds
+    - influxdb
+    - librosa
+    - neo4j-python-driver
+    - pip
+    - praatio
+    - python
+    - requests
+    - textgrid
+    - tqdm
```
BastianZim, line 27: Are you sure all of this is needed?

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+  run:
+    - conch_sounds
+    - influxdb
+    - librosa
+    - neo4j-python-driver
+    - praatio
+    - python
+    - requests
+    - textgrid
+    - tqdm
```
BastianZim, line 37: Please add the version requirements here from upstream.

recipes/pyraat/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - pyraat
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/mmcauliffe/Pyraat
+  summary: Interface for running Praat scripts through Python
+  license: MIT
```
BastianZim, line 36: Upstream has two licenses. Since some here are also maintainer there. Can you please clarify which one is correct?

BastianZim (changes requested):

recipes/pyraat/meta.yaml
michaelhaaf: @mmcauliffe is the maintainer for both. My understanding is that MIT is the correct license.

recipes/conch_sounds/meta.yaml
michaelhaaf: Librosa and pyraat are direct dependencies:
- https://github.com/mmcauliffe/Conch-sounds/search?q=librosa
- https://github.com/mmcauliffe/Conch-sounds/search?q=pyraat
I believe pip is required for building the package? I'm new to conda-forge so I could be mistaken.

recipes/conch_sounds/meta.yaml
michaelhaaf: ">=,=" should be correct. I'll test.

recipes/polyglotdb/meta.yaml
michaelhaaf: These are direct dependencies, I don't think the package would build/run without them, unless I'm missing something.

recipes/conch_sounds/meta.yaml
BastianZim: Yes but from what I understand they are only required when running the package, not when building it?

Yes, pip and python are almost always required.

recipes/polyglotdb/meta.yaml
michaelhaaf: - Version requirements as lower bounds acceptable? (version >=, = x.x ?)
- Should I put these requirements for all dependencies in this PR?

Thanks for the feedback by the way!

recipes/polyglotdb/meta.yaml
BastianZim: `Host` are packages that are needed when building the package: `pip`, `cmake` etc. `Run` is what is used when the package is installed.

Ref: https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-host-and-run

recipes/polyglotdb/meta.yaml
BastianZim: Yes that's fine. You should put all required dependencies and their version constraints here. I would recommend using https://github.com/conda-incubator/grayskull it's normally pretty good at detecting everything automatically.

michaelhaaf: Sounds good, my template came from grayskull so I'll manually fill in the rest. Thanks again

recipes/polyglotdb/meta.yaml
michaelhaaf: https://github.com/MontrealCorpusTools/PolyglotDB/pull/167

michaelhaaf:
@conda-forge/staged-recipes I believe the changes requested have been addressed. thanks for the feedback!

recipes/conch_sounds/meta.yaml
michaelhaaf: It turns out that removing librosa and pyraat caused build errors, I've kept them in the subsequent commits. I was able to remove similar dependencies from the polyglotdb meta.yaml

recipes/pyraat/meta.yaml
BastianZim: @mmcauliffe Can you please confirm this?

In that case the license should be `license: MIT OR other_licencse`

recipes/conch_sounds/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+requirements:
+  host:
+    - pip
+    - pytest
+    - python >=3.6
+    - librosa >=0.8.0,=0.8.*
+    - pyraat
+  run:
+    - librosa >=0.8.0,=0.8.*
```
BastianZim, line 25: I cannot find the version upstream: https://github.com/mmcauliffe/Conch-sounds/blob/7668612d7a610d0f5ae3332f990e71b26c5e8b34/setup.py#L63 

Is this listed somewhere else?

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 89d2b4a19595a7bc14d1fba124afc63454baa9678275cd093d091d0430fc36c2
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  noarch: python
```
BastianZim, line 15: This needs to be removed until https://github.com/MontrealCorpusTools/PolyglotDB/pull/167 is available.
```suggestion
```

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+requirements:
+  host:
+    - pip
+    - pytest
+    - python >=3.6
+  run:
+    - conch_sounds
+    - influxdb
+    - librosa >=0.8.0,=0.8.*
+    - neo4j-python-driver >=4.3,=4.*
```
BastianZim, line 26:
```suggestion
    - neo4j-python-driver >=4.3,==4.*
```
I think `~=` gets translated into that. You might need to double-check the others as well.

recipes/polyglotdb/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+requirements:
+  host:
+    - pip
+    - pytest
+    - python >=3.6
+  run:
+    - conch_sounds
+    - influxdb
+    - librosa >=0.8.0,=0.8.*
```
BastianZim, line 25: I cannot find a bound here: https://github.com/MontrealCorpusTools/PolyglotDB/blob/f021a403566388a9e13c76c2d727979baf4264e1/setup.py#L97

recipes/pyraat/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  license_file: LICENSE
```
BastianZim, line 37:
```suggestion
  license_file: 
    - LICENSE
    - LICENSE.txt
```

This includes both license files.

BastianZim (changes requested):
Just a couple of comments then this is ready.

recipes/conch_sounds/meta.yaml
michaelhaaf: I pinned the librosa dependency to >= 0.8.0 due to build errors when testing locally. There was an update to a librosa subdependency which now breaks builds from earlier versions: https://github.com/numba/numba/issues/5865 https://github.com/librosa/librosa/pull/1107

The dependency should be pinned upstream as well (though it isn't yet).

recipes/polyglotdb/meta.yaml
michaelhaaf: I tested both, it seems to be single "="

recipes/polyglotdb/meta.yaml
michaelhaaf: see https://github.com/conda-forge/staged-recipes/pull/18313#discussion_r827264804

PR #18334, opened 2022-03-14
Skip http errors in actions workflow and remove review team check

BastianZim:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/.github/workflows/automate-review-labels.yml
+++ b/.github/workflows/automate-review-labels.yml
@@ -36,11 +36,11 @@
                 '@conda-forge/help-r',
                 '@conda-forge/help-ruby'
               ];
+              let found_label = false;
               for (const team of teams) {
                   let text = context.payload.comment.body;
                   const regex = new RegExp(team + '[^\-]|' + team + '$');
                   let result = regex.test(text);
-                  let found_label = false;
                   if (result) {
                       label = team.replace("@conda-forge/help-", "");
                       label_no_staged = label.replace("@conda-forge/", "");
--- a/.github/workflows/automate-review-labels.yml
+++ b/.github/workflows/automate-review-labels.yml
@@ -18,6 +18,7 @@
       runs-on: ubuntu-latest
       steps:
         - name: check-teams
+          id: check_teams
           uses: actions/github-script@v6
           with:
             result-encoding: json
@@ -44,7 +45,7 @@
                   if (result) {
                       label = team.replace("@conda-forge/help-", "");
                       label_no_staged = label.replace("@conda-forge/", "");
-                      let found_label = true;
+                      found_label = true;
                       github.rest.issues.addLabels({
                           issue_number: context.issue.number,
                           owner: context.repo.owner,
@@ -55,9 +56,9 @@
                   }
               };
               return found_label
-        - name: add-labels
+        - name: remove-labels
           if: >
-            steps.check-teams.outputs.result
+            (steps.check_teams.outputs.result == 'true')
             && contains(github.event.issue.labels.*.name, 'Awaiting author contribution')
           uses: actions/github-script@v6
           with:
@@ -86,6 +87,7 @@
       runs-on: ubuntu-latest
       steps:
         - name: check-membership
+          id: check_membership
           uses: actions/github-script@v6
           with:
             github-token: ${{ secrets.GH_TOKEN }}
@@ -102,13 +104,15 @@
               membership_response.then(function(result) {
                  console.log(result);
                  if (result.data.state == 'active'){
-                   return;
+                   return true;
                  } else {
-                   throw new Error('Reviewer not active member of conda-forge/staged-recipes');
+                   return false;
                  }
               });
-        - name: swap-labels
-          if: ${{ success() }}
+        - name: remove-labels
+          if: >
+            (steps.check_membership.outputs.result == 'true')
+            && contains(github.event.issue.labels.*.name, 'review-requested')
           uses: actions/github-script@v6
           with:
             script: |
@@ -118,6 +122,12 @@
                   repo: context.repo.repo,
                   name: ['review-requested']
               });
+        - name: add-labels
+          if: >
+            (steps.check_membership.outputs.result == 'true')
+          uses: actions/github-script@v6
+          with:
+            script: |
               github.rest.issues.addLabels({
                   issue_number: context.issue.number,
                   owner: context.repo.owner,
--- a/.github/workflows/automate-review-labels.yml
+++ b/.github/workflows/automate-review-labels.yml
@@ -57,7 +57,7 @@
               return found_label
         - name: add-labels
           if: >
-            steps.set-result.outputs.result
+            steps.check-teams.outputs.result
             && contains(github.event.issue.labels.*.name, 'Awaiting author contribution')
           uses: actions/github-script@v6
           with:
--- a/.github/workflows/automate-review-labels.yml
+++ b/.github/workflows/automate-review-labels.yml
@@ -17,9 +17,10 @@
         && !contains(github.event.issue.labels.*.name, 'review-requested')
       runs-on: ubuntu-latest
       steps:
-        - name: add-labels
+        - name: check-teams
           uses: actions/github-script@v6
           with:
+            result-encoding: json
             script: |
               const teams = [
                 '@conda-forge/staged-recipes',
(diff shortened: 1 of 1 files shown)


BastianZim:
This does two things:

1. If the label doesn't exist the action normally falls. Since that can be the case though, we are skipping that. One example is here: https://github.com/conda-forge/staged-recipes/runs/5541701357?check_suite_focus=true Ref: https://github.com/BastianZim/test-labeler/issues/14
2. We are checking if someone is a member of staged-recipes before removing the label. This API fails if someone is not, instead of returning false. Therefore, we are catching this error: https://github.com/conda-forge/staged-recipes/runs/5542553830?check_suite_focus=true Ref: https://github.com/BastianZim/test-labeler/issues/15

.github/workflows/automate-review-labels.yml
```diff
@@ -82,19 +90,19 @@
...
+                    team_slug: 'staged-recipes',
+                    username: '${{ github.event.sender.login }}'
+                });
+              } catch (e) {
+                if (e.name == 'HttpError') {
+                 // Reviewer is not a member
+                } else {
+                  throw e; // let other errors bubble up
+                }
+              }
```
carterbox, line 105: In this section, I think we should avoid try-catch. It is possible that the response from the server is "this person is not a member" without raising an error.

Furthermore, the response is asynchronous, so I think we need to use continue to use `.then()` in order to wait for the response from the server.

If we want to avoid masking uncaught errors with the custom `Error('Reviewer not an active...')`, then we should raise the Error on the IF instead of the ELSE like this:


```javascript
console.log(result);
if (result.status != 200 or result.data.state != 'active'){
   throw new Error('Reviewer not a member');
}
```

.github/workflows/automate-review-labels.yml
```diff
@@ -48,12 +48,20 @@
...
+                            repo: context.repo.repo,
+                            name: ['Awaiting author contribution']
+                        });
+                      } catch (e) {
+                        if (e.name == 'HttpError') {
+                         // Label doesn't exist.
+                        } else {
+                          throw e; // let other errors bubble up
+                        }
+                      };
```
carterbox, line 64: I'm not sure what `github.rest.issues.removeLabel` returns. We need to figure that out. If it's an asynchronous result like below, then we should use `.then()` and figure out what STATUS the server responds with instead of catching all `HttpError`.

carterbox (changes requested):

.github/workflows/automate-review-labels.yml
carterbox: It looks like we want to ignore `status == 404` with message "label not found", but the server could also respond with `status = 410` which we we would not want to ignore, but it might also be raised as an `HttpError` in this API.

carterbox: - [ ] Also apply this patch to the other removeLabel call at the end of this file.

.github/workflows/automate-review-labels.yml
BastianZim: From what I can see the response will always be `404` so a full fault if the commenter is not part of the team: https://docs.github.com/en/rest/reference/teams#get-team-membership-for-a-user 
Not sure why GitHub chose not to return a normal response but anyways...

We can try to catch other cases as well but from what I can see there aren't any.

The problem with throwing the error is that this pollutes the notifications of anyone watching the repo as failed runs are added there, unlike what Azure is doing. So we need to finish the job successfully, even if someone is not a member. 
Maybe we can return a variable that will either be true or false and then instead of checking for success check for that variable? Something like in here: https://github.com/actions/github-script#reading-step-results

If you agree, I'll update it accordingly.

.github/workflows/automate-review-labels.yml
BastianZim: Makes sense. One other approach could also be to split this up into two steps.

The first step adds the label and if the label can be added, sets a variable to true. That variable is then returned (Like above) and can be used for an `if` for the second step. The `if` will then also check `contains(github.event.issue.labels.*.name, 'Awaiting author contribution')`

.github/workflows/automate-review-labels.yml
carterbox: > pollutes the notifications of anyone watching the repo [...] we need to finish the job successfully, even if someone is not a member.

That's a great point. I didn't realize, it did that.

> Maybe we can return a variable that will either be true or false

Yes. Try this approach. The only reason I didn't try this originally was that this logic cannot go in `if:` I has to go in `script:`? But not having notification spam is better than executing one less step in a job.

.github/workflows/automate-review-labels.yml
carterbox: Yes. I'm a fan of the second step with conditional execution. Seems easier to read than error handling, but the conditional should have two conditions:
```yaml
if: >
  previous_step_found_team_mention
  and contains(github.event.issue.labels.*.name, 'Awaiting author contribution')
script: |
   // remove label
```

.github/workflows/automate-review-labels.yml
BastianZim: Ok, will update accordingly. My JS is pretty poor so does your snippet above work already? Then I'm just going to copy/paste and add the Actions logic.

By the way, feel free to push to my branch as well if you want to change something.

.github/workflows/automate-review-labels.yml
BastianZim: Yes, that sounds good. We might need the `||` again for handling issue comments vs PRs but let me check if that's really the case.

.github/workflows/automate-review-labels.yml
carterbox: > My JS is pretty poor so does your snippet above work already?

Almost? I think javascript uses `||`, `&&` instead of `or`, `and`.

.github/workflows/automate-review-labels.yml
BastianZim: Have added that in https://github.com/conda-forge/staged-recipes/pull/18334/commits/937032331378243f7ce9631cdfdbd6b3a89a7878

Can you have a quick look if this makes sense and then I will copy it to the repo to test it.

.github/workflows/automate-review-labels.yml
```diff
@@ -39,24 +40,35 @@
...
-                          name: ['Awaiting author contribution']
                       });
                       // TODO: Maybe use GitHub API to request review / assign team
                   }
-              }
+              };
+              return found_label
+        - name: add-labels
+          if: >
+            steps.set-result.outputs.result
```
BastianZim, line 60: I'm not sure if the `True` is actually evaluated like that. I will test that in the test repo.

.github/workflows/automate-review-labels.yml
BastianZim: From what I can see the error is not directly returned so I will try to trigger that in the test repo to find out what the structure of the error message is.

.github/workflows/automate-review-labels.yml
carterbox:
```suggestion
            steps.check-teams.outputs.result
```
You need to use the name?/id? field of the step you want output from.

.github/workflows/automate-review-labels.yml
```diff
@@ -39,24 +40,35 @@
                   let text = context.payload.comment.body;
                   const regex = new RegExp(team + '[^\-]|' + team + '$');
                   let result = regex.test(text);
+                  let found_label = false;
```
carterbox, line 43: This variable needs to be set outside the loop. Otherwise, it is reset to false in each iteration.

carterbox (changes requested):

.github/workflows/automate-review-labels.yml
BastianZim: Whops forgot that, thanks.

.github/workflows/automate-review-labels.yml
BastianZim: Is it not? I thought the loop starts later? Or am I missing something?

carterbox:
I pushed a commit that I think resolves https://github.com/BastianZim/test-labeler/issues/14. We still need to figure out how to get access to team membership on PRs from forks or change the `check-membership` step so it doesn't need special access.

BastianZim:
That commit looks great, thank you!

I have opened an issue in the GitHub community but haven't heard anything back about the membership issue. If we don't get a reply over the weekend, I'll see if I can use my company's paid Github support for this. In any case, I'll keep this PR updated.

.github/workflows/automate-review-labels.yml
```diff
@@ -21,7 +21,7 @@
           id: check_teams
           uses: actions/github-script@v6
           with:
-            result-encoding: json
```
BastianZim: I think we need to keep `result-encoding: json`

.github/workflows/automate-review-labels.yml
```diff
@@ -35,80 +37,112 @@
...
-          && (github.event.sender.login != github.event.issue.user.login)
-          && contains(github.event.issue.labels.*.name, 'review-requested')
-        ) || (
-          github.event.pull_request
-          && (github.event.sender.login != github.event.pull_request.user.login)
-          && contains(github.event.pull_request.labels.*.name, 'review-requested')
-        )
+        github.event.pull_request
+        && (github.event.sender.login != github.event.pull_request.user.login)
+        && contains(github.event.pull_request.labels.*.name, 'review-requested')
```
carterbox, line 86: We only activate this task when it was triggered by a review because we need the review request to be dismissed.

carterbox (approved):

PR #18382, opened 2022-03-17
Add uproot-browser

chrisburr:
Requires https://github.com/conda-forge/textual-feedstock/pull/7

cc @henryiii

### DIFF
--- a/recipes/uproot-browser/meta.yaml
+++ b/recipes/uproot-browser/meta.yaml
@@ -21,7 +21,7 @@ requirements:
   host:
     - pip
     - python >=3.7
-    - flit-core
+    - flit-core >=3.3
   run:
     - awkward >=1
     - click >=8
--- a/recipes/uproot-browser/meta.yaml
+++ b/recipes/uproot-browser/meta.yaml
@@ -1,0 +1,55 @@
+{% set name = "uproot-browser" %}
+{% set version = "0.4.0" %}
+
+
+package:
+  name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


henryiii:
0.4.0 is about out, and that will fix Python 3.7 & 3.8, which will be nice.

chrisburr:
@conda-forge/staged-recipes, ready for review. Windows is failing due to what appears to be a bug in the package however it's `noarch: python` so it doesn't really matter.

(cc @henryiii in case you want to look into the Windows failure for the next release)

recipes/uproot-browser/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  entry_points:
+    - uproot-browser = uproot_browser.__main__:main
+
+requirements:
+  host:
+    - pip
+    - python >=3.7
+    - flit-core
```
BastianZim, line 24:
```suggestion
    - flit-core >=3.3
```
https://github.com/scikit-hep/uproot-browser/blob/5c28fd2290fcbb02ba39ac18a7d6677402591084/pyproject.toml#L2

BastianZim (commented):
Not super important but might be nice to have.

chrisburr:
Thanks @BastianZim! 🍰

henryiii:
This looks interesting for the windows issue: https://stackoverflow.com/a/52372390/2402816

PR #18426, opened 2022-03-23
Add recipe for pycraf-gui

bwinkel:
Checklist
- [x] Build number is 0.


bwinkel:

recipes/pycraf-gui/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+{% set sha256 = "1403ac72528b2f4600e89d4a1746888162cadc734490d819f2c08c6a4d1a4910" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
```
BastianZim, line 12: I would recommend regenerating the recipe with Grayskull as the `fn:` part is not really needed anymore. Some other things like `--no-deps` can also be removed and will not be included in the Grayskull generated recipe.

I can also review it manually but that seems faster.

Ref: https://github.com/conda-incubator/grayskull

BastianZim (changes requested):

recipes/pycraf-gui/meta.yaml
bwinkel: Ok, thanks. I'll give that a try.

bwinkel: Now, that was quite convenient, I must say. Thanks for the hint. Could you have another look?

recipes/pycraf-gui/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pycraf-gui-{{ version }}.tar.gz
+  sha256: 1403ac72528b2f4600e89d4a1746888162cadc734490d819f2c08c6a4d1a4910
+
+build:
+  entry_points:
+    - pycraf-gui = pycraf_gui.__main__:main
```
BastianZim, line 14: Only `console_scripts` need to be provided here: https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python

BastianZim (changes requested):
Just a couple of comments, then this is ready.

recipes/pycraf-gui/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  run:
+    - python >=3.8
+    - astropy
+    - numpy
+    - matplotlib >=3.3
+    - pycraf
+    - pyqt
+    - pytest
+    - pytest-astropy
+    - pytest-qt
```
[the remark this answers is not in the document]
bwinkel: pycraf and pycraf-gui make use of astropy's [TestRunner](https://docs.astropy.org/en/latest/testhelpers.html#astropy-test-runner), which allows to run the test via a function defined within the module. It may seem a bit overkill, but most of our users don't know much about programming etc., and will never run pytest standalone.

recipes/pycraf-gui/meta.yaml
bwinkel: In my first attempt, I left that out, which had the consequence that on Windows an error was thrown (it created a binary, pycraf-gui.exe, which was considered as not suitable for noarch). The above snippet with the entry_points was actually suggested by the grayskull [webapp](https://www.marcelotrevisani.com/grayskull).

recipes/pycraf-gui/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+about:
+  home: https://github.com/bwinkel/pycraf-gui
+  summary: A simple GUI for pycraf
+  description: |
+    The pycraf package provides functions and procedures for various tasks in spectrum-management compatibility studies. This package, pycraf-gui
+    provides a simple user interface for the path attenuation functionality of pycraf.
+  dev_url: https://github.com/bwinkel/pycraf-gui
+  license: GPL-3.0
+  license_file: COPYING
```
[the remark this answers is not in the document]
bwinkel: Good catch. I added another "license_file".

recipes/pycraf-gui/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - pip
+    - setuptools
+    - setuptools_scm
+    - astropy
+
+  run:
+    - python >=3.8
+    - astropy
+    - numpy
+    - matplotlib >=3.3
```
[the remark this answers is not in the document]
bwinkel: Oh, you hit a weak spot. Actually, I'm not even sure, which (older) matplotlib versions are still compatible. They change their API so often, that I sometimes struggle to keep up. I'll make some checks tomorrow. In the meantime, I change the pin to >=3, which I'm 99% sure should also work. Whether v2 is still OK, I need to try...

recipes/pycraf-gui/meta.yaml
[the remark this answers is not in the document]
BastianZim: Ahh ok makes sense then.

recipes/pycraf-gui/meta.yaml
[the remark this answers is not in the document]
BastianZim: Ok then let‘s keep the pin for now.

recipes/pycraf-gui/meta.yaml
BastianZim: Ahh you‘re right - ok then that‘s fine

BastianZim (approved):

bwinkel:
Thanks a lot for your help! I appreciate this very much.

PR #18477, opened 2022-03-26
Add apache-airflow-providers-influxdb

xylar:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/apache-airflow-providers-influxdb/meta.yaml
+++ b/recipes/apache-airflow-providers-influxdb/meta.yaml
@@ -41,7 +41,7 @@ about:
     - LICENSE
     - NOTICE
   license_family: Apache
-  doc_url: https://airflow.apache.org/docs/{{ name }}/stable/index.html
+  doc_url: https://airflow.apache.org/docs/apache-airflow-providers-influxdb/stable/index.html
   dev_url: https://github.com/apache/airflow/
 
 extra:
--- a/recipes/apache-airflow-providers-influxdb/meta.yaml
+++ b/recipes/apache-airflow-providers-influxdb/meta.yaml
@@ -37,7 +37,9 @@ about:
   home: https://airflow.apache.org/
   summary: Provider for Apache Airflow. Implements apache-airflow-providers-influxdb package
   license: Apache-2.0
-  license_file: LICENSE
+  license_file: 
+    - LICENSE
+    - NOTICE
   license_family: Apache
   doc_url: https://airflow.apache.org/docs/{{ name }}/stable/index.html
   dev_url: https://github.com/apache/airflow/
--- a/recipes/apache-airflow-providers-influxdb/meta.yaml
+++ b/recipes/apache-airflow-providers-influxdb/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "apache-airflow-providers-influxdb" %}
+{% set version = "1.1.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/apache-airflow-providers-influxdb-{{ version }}.tar.gz
+  sha256: de9db24d26c6cd7645abd317806c0ed916dd90656d27d8fc29a0c0a248114ffc
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7
+    - pip
+    - setuptools
+  run:
+    - python >=3.7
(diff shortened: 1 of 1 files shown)


xylar:
@conda-forge/help-python, this is ready to review. Thank you in advance!

recipes/apache-airflow-providers-influxdb/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://airflow.apache.org/
+  summary: Provider for Apache Airflow. Implements apache-airflow-providers-influxdb package
+  license: Apache-2.0
+  license_file: LICENSE
```
BastianZim, line 40:
```suggestion
  license_file: 
    - LICENSE
    - NOTICE
```

BastianZim (changes requested):
Just one small thing

recipes/apache-airflow-providers-influxdb/meta.yaml
BastianZim: This needs to be included in the other apache-airflow-providers packages as well please. I missed that originally.

xylar: @BastianZim, if you are referring to this file:
https://github.com/apache/airflow/blob/main/NOTICE
are you sure it is appropriate for the providers?  I could see it being necessary to add to the `apache-airflow` package itself.

xylar: It is not packaged with the providers so it will be a real pain to add it.  There are something like 50 providers.

BastianZim: Yes, that's the one. It's distributed with the source distribution on PyPI so I think it's important.

BastianZim: > It is not packaged with the providers so it will be a real pain to add it. There are something like 50 providers.

There is one included with the providers.

BastianZim: This is what's included on PyPI:

```sh
.
├── LICENSE
├── MANIFEST.in
├── NOTICE
├── PKG-INFO
├── README.rst
├── airflow
│   └── providers
│       └── influxdb
│           ├── __init__.py
│           ├── example_dags
│           │   ├── __init__.py
│           │   ├── example_influxdb.py
│           │   └── example_influxdb_query.py
│           ├── get_provider_info.py
│           ├── hooks
│           │   ├── __init__.py
│           │   └── influxdb.py
│           └── operators
│               ├── __init__.py
│               └── influxdb.py
├── apache_airflow_providers_influxdb.egg-info
│   ├── PKG-INFO
│   ├── SOURCES.txt
│   ├── dependency_links.txt
│   ├── entry_points.txt
│   ├── not-zip-safe
│   ├── requires.txt
│   └── top_level.txt
├── pyproject.toml
├── setup.cfg
└── setup.py

7 directories, 24 files
```

xylar: I will double check but I downloaded the `microsoft-azure` provider and it wasn't there.

BastianZim: Hmm strange then maybe it's not included all the time.

xylar: Okay, I see it for this one.  I'll try to find out if there's an easy way to detect if it's there or now in my automated update process for all the providers.

xylar:
Thanks for noticing the `NOTICE` file.  I'll do my best to clean things ups.

recipes/apache-airflow-providers-influxdb/meta.yaml
BastianZim: > Okay, I see it for this one. I'll try to find out if there's an easy way to detect if it's there or now in my automated update process for all the providers.

If you run the build locally in your process, I would just add it and then see if the CI fails because it can find it or not.

BastianZim: Might be the easiest and fastest way.

xylar: I think `NOTICE` is likely to appear in all future provider updates because they're all generated the same way as far as I can tell.  But if I run into trouble, I'll add code to download the PyPI package and check if `NOTICE` is here.

Running build locally is way too time consuming when I need to update 72 providers.

xylar:
@BastianZim, thank you again.  I believe the remaining 4 providers are ready for you to re-review when you get a chance.

PR #18504, opened 2022-03-29
Add Sapiens 1.0.0 (ready for review)

prihoda:
Add Sapiens 1.0.0 from PyPI: https://pypi.org/project/sapiens/

### DIFF
--- a/recipes/sapiens/LICENSE
+++ b/recipes/sapiens/LICENSE
@@ -1,0 +1,21 @@
+The MIT License
+
+Copyright © 2022 Merck Sharp & Dohme Corp. a subsidiary of Merck & Co., Inc., Kenilworth, NJ, USA."
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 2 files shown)


prihoda:
This is ready to review. Windows is not supported due to missing pytorch dependency.

recipes/sapiens/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 41a8242da8ced4c4390ab6167950c6abb53ce84e3a7519282ccd680254d3119a
+
+build:
```
synapticarbors, line 12: It looks like this project has an `entry_point` for a console script. It should be added. See:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-entry-points

and then a simple test of the command line tool should be added to the test section:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#test-commands

recipes/sapiens/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License
```
synapticarbors, line 1: I've opened a PR upstream to add the license file to the source distribution. If/when that's merged and there is a new upstream release of `sapiens`, this manually added copy should be removed.

synapticarbors (changes requested):

recipes/sapiens/meta.yaml
prihoda: Oh right, I created the entrypoint in setup.py, but it doesn't actually exist. So I deleted it from setup.py for now.

recipes/sapiens/LICENSE
prihoda: Merged and released 1.0.1, shall we change it here already? 

The new sha256 is `49eedaec6213410423ec3b46e2655fbf8731caee7de69fa2873a59497274bae7`

synapticarbors: If you want you can update the version and sha256 here and delete the manually packaged license file

synapticarbors:
conda-forge dropped using the `defaults` channel back in late-September 2021. There is a note in https://conda-forge.org/docs/user/announcements.html#announcements that says, "You can get the previous behaviour by using the `channel_sources` setting in `conda-forge.yml`." I don't see that field currently, so maybe it's been replaced by:
https://conda-forge.org/docs/maintainer/conda_forge_yml.html#channels

That said, I think it's ok to ignore the windows failure since this is a `noarch` package. I think it will be up to the end user to specify that they want to use the `defaults` channel when installing `sapiens`.

prihoda:
> I think it will be up to the end user to specify that they want to use the defaults channel when installing sapiens.

Makes sense.

prihoda:
@synapticarbors Let's merge this version, so that we include release 1.0.0. I can remove the LICENSE later.

synapticarbors (approved):

PR #18511, opened 2022-03-30
Adding geospatial recipe

giswqs:
Checklist
- [x] Build number is 0.


giswqs:
@conda-forge/help-python The recipe has passed all the checks and is ready for your review. Thanks.

giswqs:
@conda-forge/help-python The recipe has passed all the checks and is ready for your review. Thanks.

recipes/geospatial/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.7
```
BastianZim, line 20: You list 3.5 upstream. Are you sure that this is correct?
```suggestion
    - python >=3.5
```

Ref: https://github.com/giswqs/geospatial/blob/73ca0ab34e20d5790b92ec3364389323057255e2/setup.py#L30

recipes/geospatial/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  run:
+    - geemap
+    - leafmap
+    - python >=3.7
```
BastianZim, line 24:
```suggestion
    - python >=3.5
```

recipes/geospatial/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.7
+  run:
```
BastianZim, line 21: The list of reqs is much longer upstream. Are you sure that only the ones below are needed?

Ref: https://github.com/giswqs/geospatial/blob/v0.1.0/requirements.txt

recipes/geospatial/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - geemap
+    - leafmap
+    - python >=3.7
+    - xarray
+
+test:
```
BastianZim, line 27: It would be good to add a pip check here as well. Something like this one: https://github.com/conda-forge/azure-quantum-feedstock/blob/1864c7c7528cbad2af029f83de1c3ddda5267726/recipe/meta.yaml#L44-L47

BastianZim (changes requested):

recipes/geospatial/meta.yaml
giswqs: I have updated the upstream to require Python >=3.7

recipes/geospatial/meta.yaml
giswqs: I have updated the upstream to require Python >=3.7

recipes/geospatial/meta.yaml
giswqs: This list of reqs has changed since this PR was initially submitted. I have updated the list in the recipe.

recipes/geospatial/meta.yaml
giswqs: I think I added the `pip check` before, but the checks failed and I am not sure why. I just re-added `pip check` to see if it can pass the checks this time.

giswqs: I have fixed all `pip check` errors except `pointpats 2.2.0 requires opencv-contrib-python, which is not installed.` However, `pointpats` is ready listed in the recipe and it has been installed. See Line 1276 docker build. I am not sure which package depends on `pointpats`. I also added `libopencv`, `opencv`, and `py-opencv`, but it still does not solve the issue. 
Should this issue be resolved by the `pointpats` package on the conda-forge channel? How could have`pointpats` passed checks if it is missing some required packages?

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=484880&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1276

giswqs: The issue has been resolved after removing the `pysal` package from the reqs. All checks have passes and it is ready for your review. Thanks.

recipes/geospatial/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - rioxarray
+    - sqlalchemy
+    - streamlit-folium
+    - xarray-spatial
+    - quilt3
+    - httplib2shim
+    - python-box <=5.4
+    - ipyevents <=0.9.0
+    - setuptools >=60
+    - google-api-python-client <=2
```
BastianZim, line 49: None of these are listed upstream. Are you sure that they are needed? Especially `setuptools` seems strange as that is normally not needed when running the package.

BastianZim (changes requested):

recipes/geospatial/meta.yaml
giswqs: The issue was caused by `pip check`. In the past, the recommended recipe template does not include `pip check`. Some of the submitted recipes were accepted because `pip check` was not used at the time even though they would fail should `pip check` be used . Now they are causing issues for the downstream packages because of them due to the use of `pip check`. Therefore, I had to add those missing packages to the end of the recipe in order to pass `pip check`. Anyway, I have removed the packages in question, and it should now resolve the issue.

giswqs:
All checks have passed and it is ready for your review. Thanks.

BastianZim (approved):

PR #18527, opened 2022-03-30
Add elyra-code-viewer-extension

akchinSTC:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/elyra-code-viewer-extension/meta.yaml
+++ b/recipes/elyra-code-viewer-extension/meta.yaml
@@ -25,8 +25,8 @@ requirements:
 
 test:
   commands:
-    - test -f ${PREFIX}/share/jupyter/labextensions/@elyra/code-viewer-extension/package.json
-    - if not exist %PREFIX%\\share\\jupyter\\labextensions\\@elyra\\code-viewer-extension\\package.json (exit 1)
+    - test -f ${PREFIX}/share/jupyter/labextensions/@elyra/code-viewer-extension/package.json                            # [unix]
+    - if not exist %PREFIX%\\share\\jupyter\\labextensions\\@elyra\\code-viewer-extension\\package.json (exit 1)         # [win]
 
 about:
   home: https://github.com/elyra-ai/elyra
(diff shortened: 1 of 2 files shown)


recipes/elyra-code-viewer-extension/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - jupyter-packaging
+    - pip
+    - python >=3.7
+  run:
+    - elyra-server ==3.7.0
```
dopplershift, line 23: Would it be correct (and better) to say:
```suggestion
    - elyra-server =={{ version }}
```
?

recipes/elyra-code-viewer-extension/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - python >=3.7
+
+test:
+  commands:
+    - exit 0
```
dopplershift, line 28: We need some kind of test to validate that the package works as needed. Can you do like this [other extension](https://github.com/conda-forge/jupyterlab-unfold-feedstock/blob/30f98b549dbe016584ebc9eb5ba5afcde2d95861/recipe/meta.yaml#L28-L31) does?

dopplershift: Is there any reason you didn't also copy the Windows test from the example I linked to?

akchinSTC: Re-added the test in the interest of getting the past CI. It was removed because the project itself doesnt explicitly support windows, just osx and unix environments (although we're working on it) =)
Is there an way to exclude windows based testing from CI or is it a requirement as part of the conda community?

dopplershift (approved):

recipes/elyra-code-viewer-extension/meta.yaml
dopplershift: Well, it's fine if it doesn't support. Right now, though, as a noarch package it gets built on Linux and is installable on all the platforms. (I'll note that `elyra` itself is the same way.) Usually the way to keep it off Windows is to have a dependency that doesn't support it. :wink:

If you want to make sure this can't be installed on Windows, then I *think* you can depend on the `__unix` virtual package (though that's only a thing for newish versions of conda.

akchinSTC:
@marcelotrevisani - can i get a merge for this ? =)

akchinSTC:
Closing PR due to package being made redundant (merged upstream into jupyterlab)

PR #18551, opened 2022-04-02
Adding xarray-einstats feedstock.

ericmjl:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/xarray-einstats/meta.yaml
+++ b/recipes/xarray-einstats/meta.yaml
@@ -21,7 +21,7 @@ requirements:
     - python >=3.8
   run:
     - numpy >=1.21
-    - python
+    - python >=3.7
     - scipy
     - xarray >=0.20
 
--- a/recipes/xarray-einstats/meta.yaml
+++ b/recipes/xarray-einstats/meta.yaml
@@ -36,7 +36,8 @@ test:
 about:
   home: https://pypi.org/project/xarray-einstats/
   summary: Stats, linear algebra and einops for xarray.
-  license_file: PLEASE_ADD_LICENSE_FILE  # help!
+  license: Apache-2.0
+  license_file: LICENSE
 
 extra:
   recipe-maintainers:
--- a/recipes/xarray-einstats/meta.yaml
+++ b/recipes/xarray-einstats/meta.yaml
@@ -1,0 +1,44 @@
+# Generated using grayskull; edited by @ericmjl.
+{% set name = "xarray-einstats" %}
+{% set version = "0.2.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/xarray-einstats-{{ version }}.tar.gz
+  sha256: 5c4e00d09172daf665cbd5e533a897896b84f37feb5cab0fd75829477fbcda7d
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python
+  run:
+    - numpy >=1.21
+    - python
+    - scipy
+    - xarray >=0.20
+
+test:
+  imports:
+    - xarray_einstats
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://pypi.org/project/xarray-einstats/
+  summary: Stats, linear algebra and einops for xarray.
+  license_file: PLEASE_ADD_LICENSE_FILE  # help!
+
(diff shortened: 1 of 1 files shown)


OriolAbril:
I am willing and happy to be a maintainer of the feedstock

recipes/xarray-einstats/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  sha256: 5c4e00d09172daf665cbd5e533a897896b84f37feb5cab0fd75829477fbcda7d
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
```
OriolAbril, line 20: as the library has pyproject.toml only and no setup.py nor setup.cfg, it needs pip>=21.3, it also looks like flit should be installed for everything to work as it is a build system requirement (which I think means it should go here in the host)

ericmjl: Great catch, thank you, @OriolAbril!

recipes/xarray-einstats/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip >=21.3
+    - python
```
OriolAbril, line 21:
```suggestion
    - python >=3.7
    - flit
```

recipes/xarray-einstats/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  run:
+    - numpy >=1.21
+    - python
```
OriolAbril, line 24:
```suggestion
    - python >=3.7
```

recipes/xarray-einstats/meta.yaml
OriolAbril: the azure jobs are complaining about flit, it is a build system dependency in pyproject: https://github.com/arviz-devs/xarray-einstats/blob/main/pyproject.toml#L2. I think it needs to be here like it is in the poliastro recipe for example: https://github.com/conda-forge/poliastro-feedstock/blob/main/recipe/meta.yaml

python could be bigger than 3.6 as the linter suggests, but the run is 3.7 so it is probably easier and less prone to confusion to use 3.7 in both

ericmjl: Thanks for the catch, @OriolAbril! Okie just made the changes.

ericmjl:
@OriolAbril can I check, is the license file distributed as part of the `pip`-installable package? I believe that may need to be added into the distributed source code in order for us to be able to do a conda-forge package.

OriolAbril:
> @OriolAbril can I check, is the license file distributed as part of the pip-installable package? I believe that may need to be added into the distributed source code in order for us to be able to do a conda-forge package.

Not sure, let me see if I can check this somehow.

recipes/xarray-einstats/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+test:
+  imports:
+    - xarray_einstats
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://pypi.org/project/xarray-einstats/
```
OriolAbril, line 38:
```suggestion
  home: https://github.com/arviz-devs/xarray-einstats
```

I think the arviz feedstock takes the license from github directly using the link provided here: https://github.com/conda-forge/arviz-feedstock/blob/main/recipe/meta.yaml#L41

OriolAbril: I can try to add the license to pip package too but I'd need an example of how to do it (I tried to copy arviz and saw arviz doesn't include it in the pip package)

ericmjl: I think I found an example to copy, @OriolAbril! 

`pyjanitor` uses a MANIFEST.in file to declare which other files need to be packaged together and distributed. 

When I looked up other references, I stumbled upon the rules by which Python package files are included: https://packaging.python.org/en/latest/guides/using-manifest-in/. Perhaps that could help as well.

OriolAbril:
Preparing a patch release with a manifest file in https://github.com/arviz-devs/xarray-einstats/pull/12

OriolAbril:
Should be fixed now with 0.2.2 release, thanks for the pointers.

Can you add https://xarray-einstats.readthedocs.io as [`doc_url`](https://github.com/conda-forge/arviz-feedstock/blob/main/recipe/meta.yaml#L52)?

xhochy (approved):

PR #18555, opened 2022-04-02
DEP turn off drone

### LINKED ISSUE #18545: turn off drone
We should turn off drone since it no longer works.

beckermr:
This PR turns off drone since it is dead to us.

closes #18545


beckermr:
@conda-forge/core Are we ok with this?

PR #18556, opened 2022-04-03
Add fish2eod

aaronshifman:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/fish2eod/meta.yaml
+++ b/recipes/fish2eod/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
-  skip: true  # [win]
+  noarch: python
 
 requirements:
   host:
--- a/recipes/fish2eod/meta.yaml
+++ b/recipes/fish2eod/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/aaronshifman/{{name}}/archive/refs/tags/v{{version}}.tar.gz
+  url: https://github.com/aaronshifman/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
   sha256: 5ff1a9d5e5e8f10871ced326d4b86f9ecd0616e760e6beb983dfbebcae26f6c2
 
 build:
--- a/recipes/fish2eod/meta.yaml
+++ b/recipes/fish2eod/meta.yaml
@@ -15,11 +15,6 @@ build:
   skip: true  # [win]
 
 requirements:
-  build:
-    - python
-    - setuptools
-    - {{ pin_compatible('numpy') }}
-    - pip
   host:
     - python
     - pip
@@ -32,7 +27,7 @@ requirements:
     - h5py
     - matplotlib-base
     - meshio <5
-    - {{ pin_compatible('numpy') }}
+    - numpy
     - scipy
     - pandas
     - tqdm
@@ -40,7 +35,6 @@ requirements:
     - Shapely
     - lxml
     - dataclasses
-    - pip
 
 test:
   imports:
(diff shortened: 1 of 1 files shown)


aaronshifman:
@conda-forge-admin, please ping conda-forge/python

aaronshifman:
Ready for review

aaronshifman:
@conda-forge-admin, please ping conda-forge/python

aaronshifman:

recipes/fish2eod/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/aaronshifman/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 5ff1a9d5e5e8f10871ced326d4b86f9ecd0616e760e6beb983dfbebcae26f6c2
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
```
dopplershift, line 15: I see this package is pure Python (though it only advertises support for LInux/MacOS). Any particular need for the skip on Windows? If it's due to missing dependencies, we can make this a noarch python package instead by:
```suggestion
  noarch: python
```

recipes/fish2eod/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - {{ pin_compatible('numpy') }}
+    - pip
```
dopplershift, line 22: Not needed since you're not compiling any code.
```suggestion
```

recipes/fish2eod/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  run:
+    - python
+    - fenics-dolfin ==2019.1.0
+    - superlu_dist ==7.2.0
+    - gmsh ==3.0.6
+    - pytables
+    - h5py
+    - matplotlib-base
+    - meshio <5
+    - {{ pin_compatible('numpy') }}
```
dopplershift, line 35: Since you're not linking against the numpy C-API, just do:
```suggestion
    - numpy
```

recipes/fish2eod/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - scipy
+    - pandas
+    - tqdm
+    - PyYAML
+    - Shapely
+    - lxml
+    - dataclasses
+    - pip
```
dopplershift, line 43: Is pip actually needed at *run* time?
```suggestion
```

dopplershift (changes requested):
Overall in good shape, just some changes to some dependencies and a question about making it noarch.

recipes/fish2eod/meta.yaml
aaronshifman: Thank you @dopplershift for the comments. I've taken your suggestions for the other comments. The reason I didn't go with the noarch build was because there are datafiles (specified in MANIFEST.IN) that are needed at runtime. There was an issue with the naming in the packaging process and the required files weren't being found. Having separate builds seemed to resolve the issue

dopplershift: Would you be willing to switch this over to noarch at least temporarily in this build so I can see the problem (there are no windows failures in the CI logs I can see on this PR)? Having a pure Python package as noarch is preferable because it gets us down to a single build rather than the 20-30 builds that you get with all the python version and architecture permutations we support.

aaronshifman: @dopplershift I must be mis-remembering from a previous recipe experiment because the macos / linux builds ran fine, the windows build failed as expected because the dependencies don't exist for windows. So I'm fine with `noarch`

dopplershift (approved):
Looks like this is failing due to a missing dependency on Windows, so I'm good to merge this as is if you're happy with it.

aaronshifman:
Yep, I'm fine with it, wasn't planning on supporting windows because the dep's are missing.

PR #18626, opened 2022-04-08
Adding NEST Desktop from PyPI

babsey:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/nest-desktop/meta.yaml
+++ b/recipes/nest-desktop/meta.yaml
@@ -26,9 +26,7 @@ test:
     - nest_desktop
   commands:
     - pip check
-    - nest-desktop  # [unix]
-    - nest-desktop start -d  # [unix]
-    - nest-desktop stop  # [unix]
+    - nest-desktop status  # [unix]
   requires:
     - pip
 


babsey:
Hi @conda-forge/staged-recipes, PR is ready.

Two remarks:
- Initially, I set the version to 3.0.0 and would like to build all NEST Desktop to 3.1.2. Is it possible?
- Builds in Linux and Mac jobs are fine but in win job it seems that it built wrong package "tango-gateway".
   -> Maybe you have to resubmit the win job.

carterbox (changes requested):
> Initially, I set the version to 3.0.0 and would like to build all NEST Desktop to 3.1.2. Is it possible?

If you want to maintain multiple versions of your package. e.g. Release 3.0.x and 3.1.x versions at the same time, then create a separate branch for each release once your feedstock is created. All (merge?) commits in the feedstock will be published; this is why you must use a personal fork to make changes the feedstock.

If you just want to build each version from 3.0.0 to 3.1.2 once, then you need to make a PR for each version to your feedstock and merge them into the feedstock.

Does that make sense?

https://conda-forge.org/docs/maintainer/updating_pkgs.html#forking-and-pull-requests

> Builds in Linux and Mac jobs are fine but it seems that win job built wrong package "tango-gateway".

Reopening and closing the PR restarts the CI. Adding new commits restarts the CI. I don't know why the other package was built instead of this one.

babsey:
Thanks for the review.

We see your point for multiple versions and we decide that we do use the latest release version (3.1.2) instead of multiple.

I need to understand what triggers entry_points. 
NEST Desktop has actually bash script `nest-desktop` to manage server instance.
I prefer using this because I can change the environment variable for server host / port.

babsey:
@carterbox Friendly ping.

babsey:
Dear Conda-Forge Team,

the PR is ready for more review.  

Thanks!

recipes/nest-desktop/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - pip
+    - python >=3.6
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - nest_desktop
+  commands:
+    - pip check
```
carterbox, line 30:
```suggestion
  commands:
    - pip check
    - nest_desktop --help
```
Normally, we should also test that the CLI is installed correctly, but if we start your app, it doesn't return and must be force quit?

recipes/nest-desktop/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/nest-desktop-{{ version }}.tar.gz
+  sha256: 49051aeefc8be189ae7dfb64ad983b7a0a5636765a79febd3347194b57ba61ba
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  noarch: python
```
carterbox, line 15:
```suggestion
  skip: True  # [not unix]
```
On second review, it looks like your recipe does not qualify for `noarch: python` because you are shipping a bash script (which doesn't work on Windows).

recipes/nest-desktop/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  entry_points:
+    - nest_desktop = nest_desktop.app:run
```
carterbox, line 17:
```suggestion
```
Actually, your app also doesn't have this entry point because it is not registered in your setup.py.

carterbox (changes requested):

recipes/nest-desktop/meta.yaml
babsey: First, to clarify our concept: we have a python package `nest_desktop` and a bash script `nest-desktop`.
The user can start the server instance of NEST Desktop in both ways.

NEST Desktop uses Python module `http.server` module which runs on `from nest_desktop import app; app.run()`.

The bash script to start NEST Desktop is only a wrapper to Python module, we could ignore it for `noarch`.
```python3 -m nest_desktop.app ${HOST} ${PORT}```

File: recipes/nest-desktop/meta.yaml (hunk 3 above)
Comment (babsey) [reply recipes/nest-desktop/meta.yaml:17]: Ok, we do not have any entry points. Thus, we will remove it.

File: recipes/nest-desktop/meta.yaml (hunk 1 above)
Comment (babsey) [reply recipes/nest-desktop/meta.yaml:30]: The bash script `nest-desktop` shows help when no sub-command is used:

```
~$ nest-desktop
NEST Desktop 3.1.2
------------------
Usage: nest-desktop status|start|stop|restart [-h <HOST>] [-p <PORT>]

Commands:
  status      display the status of NEST Desktop
  start       start a new server instance for NEST Desktop
  stop        stop a server instance running on <HOST>:<PORT>
  restart     restart (i.e. stop and start) a server on <HOST>:<PORT>

Options:
  -h <HOST>   use hostname/IP address <HOST> for the server [default: 127.0.0.1]
  -p <PORT>   use port <PORT> for opening the socket [default: 8000]
~$
```

To answer your question: Yes, it has to be force quit since it is served forever. (Press CRTL + C to stop.)

We could add an argument to test it without running forever:
```
def run(host=NEST_DESKTOP_HOST, port=NEST_DESKTOP_PORT, test=False):

    web_dir = os.path.join(os.path.dirname(__file__), "app")
    os.chdir(web_dir)

    Handler = http.server.SimpleHTTPRequestHandler
    
    if test:
        socketserver.TCPServer((host, port), Handler)
        print('Test successful.')
    else:
        with socketserver.TCPServer((host, port), Handler) as httpd:
            httpd.serve_forever()
```

What do you think?

**Review by babsey — COMMENTED** [review]
Thank you for fast review. We placed comments after your points.

File: recipes/nest-desktop/meta.yaml (hunk 3 above)
Comment (carterbox) [reply recipes/nest-desktop/meta.yaml:17]: Yes, please remove these lines because your app currently does not have a console_script entry_point.

File: recipes/nest-desktop/meta.yaml (hunk 1 above)
Comment (carterbox) [reply recipes/nest-desktop/meta.yaml:30]: If the bash script returns after printing the help, then use the suggestion above. If not, then I don't know how to kill your app after it has been started. Perhaps:

```suggestion
  commands:
    - pip check
    - nest-desktop start  # [unix]
    - nest-desktop stop   # [unix]
```

These tests are for testing that the packaging was done correctly, so we don't need to test your app, just test whether the app is installed into the conda environment correctly. i.e. that the bash script is found on the path.

I think it would be helpful for Windows users (who cannot use the bash script) to add the help when launching the app from python directly, but this is not required.

recipes/nest-desktop/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - python >=3.6
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - nest_desktop
+  commands:
+    - pip check
+    - nest-desktop start  # [unix]
```
babsey, line 29: when it should run as daemon:

```suggestion
    - nest-desktop start -d  # [unix]
```

babsey:
I apologize, I do not understand why the test breaks.
`nest-desktop` shows only command doc like `--help`.

babsey:
Ah, the deamon argument is not existing in this version. 
I am working on this.

Issue #18023, opened 2022-02-12
Add automated labelling to staged-recipes


BastianZim:
This is a proposal to add automated labelling to the staged-recipes repo using GitHub actions.
I'm opening this as an issue instead of a PR as I wanted to get some input first what everybody else thinks. 
As conda-forge is generally very hands-on I didn't want to add too much automation and thought that the following would be a good start:

1. Label PRs based on the requested team. 
  This is pretty straightforward and would only add a label with the pinged team. It would make browsing PRs easier and quicker as you can see if it's in your domain or not.
  Additionally, even PRs by new contributors that cannot yet ping a team are labelled and can be marked as ready for review.
  
    The GitHub actions yaml would be something like this:
    ```yaml
    name: Label PRs

    on:
      issue_comment:
        types: [created]

    jobs:
      label-pr:
        name: Label PR
        if: ${{ github.event.issue.pull_request }}
        runs-on: ubuntu-latest
        strategy:
          matrix:
           include:
             - label: 'staged-recipes'
               comment: '@conda-forge/staged-recipes'

             - label: 'python'
               comment: '@conda-forge/help-python'

             - label: 'python-c'
               comment: '@conda-forge/help-python-c'

             - label: 'r'
               comment: '@conda-forge/help-r'

             - label: 'java'
               comment: '@conda-forge/help-java'

             - label: 'nodejs'
               comment: '@conda-forge/help-nodejs'

             - label: 'c-cpp'
               comment: '@conda-forge/help-c-cpp'

             - label: 'perl'
               comment: '@conda-forge/help-perl'

             - label: 'julia'
               comment: '@conda-forge/help-julia'

             - label: 'ruby'
               comment: '@conda-forge/help-ruby'

        steps:
          - name: team-label
            if: contains(github.event.comment.body, matrix.comment)
            env:
              label: ${{ matrix.label }}
            uses: actions/github-script@v6
            with:
              script: |
                github.rest.issues.addLabels({
                  issue_number: context.issue.number,
                  owner: context.repo.owner,
                  repo: context.repo.repo,
                  labels: [process.env.label, 'review-requested']
                })
    ```

    I tested the code locally and noticed a couple of small issues:
    1. conda-forge/help-python-c also includes conda-forge/help-python as it checks if the string is included.
    2. Currently runs for every label as the `if` is only checked during the run. Maybe there is a better way.

2. Add review-requested lables
[...]

PR #18631, opened 2022-04-09
Add recipe for Sparrow Python bindings

awvwgk:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/scine-sparrow-python/meta.yaml
+++ b/recipes/scine-sparrow-python/meta.yaml
@@ -1,0 +1,80 @@
+{% set name = "scine-sparrow-python" %}
+{% set version = "3.0.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/qcscine/sparrow/archive/refs/tags/{{ version }}.tar.gz
+    sha256: 70636871694c9363ae3fb2df5050bddb22667b71d875d5a7e9afd872f6a2b65d
+    fn: {{ name }}.tar.gz
+    patches:
+      - skip_library_build.patch
+  - url: https://github.com/qcscine/development-utils/archive/refs/tags/4.0.0.tar.gz
+    sha256: 54002c2082b6bb75672ec66bf9cf3935bbdf6b085ed9b4d7174cbdedb7c2275d
+    fn: dev.tar.gz
+    folder: dev
+
+build:
+  number: 0
+  skip: true  # [win]
+  script:
+    - set -ex
+    - >
+       cmake
+       ${CMAKE_ARGS}
+       -B _build -G Ninja
+       -DPYTHON_EXECUTABLE={{ PYTHON }}
+       -DSCINE_MARCH=""
+       -DSCINE_SKIP_LIBRARY=ON
+       -DSCINE_BUILD_TESTS=OFF
+       -DSCINE_BUILD_PYTHON_BINDINGS=ON
+    - cmake --build _build
+    - cmake --install _build
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - pkg-config
+    - llvm-openmp  # [osx]
+    - libgomp  # [linux]
+  host:
+    - boost-cpp
+    - eigen
+    - libblas
+    - pip
+    - python
+    - scine-sparrow {{ version }}
+    - yaml-cpp
+  run:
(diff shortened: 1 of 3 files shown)


awvwgk:
@conda-forge/help-python-c This Python binding package for the C++ library sparrow is ready for review.

- see https://github.com/conda-forge/scine-sparrow-feedstock for C++ library

awvwgk:
@conda-forge/staged-recipes This recipe is ready for review

recipes/scine-sparrow-python/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/qcscine/sparrow/archive/refs/tags/{{ version }}.tar.gz
+    sha256: 70636871694c9363ae3fb2df5050bddb22667b71d875d5a7e9afd872f6a2b65d
+    fn: {{ name }}.tar.gz
+    patches:
+      - skip_library_build.patch
+  - url: https://github.com/qcscine/development-utils/archive/refs/tags/4.0.0.tar.gz
```
wolfv, line 14: would it make sense to distribute this as a seperate package if it is used across multiple packages?

awvwgk: Those are some kind of common CMake files rather than an actual installable/distributable package, which are actually used as git submodule upstream but easier to include with multiple sources.

awvwgk:
Anything left open to address to get this recipe merged?

carterbox:
@awvwgk In future, please combine related recipes into one PR with separate folders for each feedstock. i.e. all of the scine/sparrow related recipes could have been reviewed at one time which is helpful for reviewers to understand dependency relationships and consistent naming.

awvwgk:
Sure, the first batch of scine related packages went in together, retriggering the feedstock builds afterwards due to CDN delays of ~2hrs per package upload took a while get everything build. Therefore, I split the second batch into several recipes to incrementally add the remaining packages, which is also suboptimal but ensures that the feedstocks will at least build cleanly once created.

carterbox:
>  ensures that the feedstocks will at least build cleanly once created.

Only if their PRs are merged in the correct order! 😆 Unfortunately, our review process is somewhat random.

awvwgk:
I can gather the three scine related PRs in one, if this helps. However, multiple dependent recipes are hard for this particular software stack, because this is template heavy code and compile times are quite long. For the first three recipes it were already ~40 min (see https://github.com/conda-forge/staged-recipes/pull/18465).

carterbox (changes requested):
This package is just the python API wrapper around scine-sparrow. It contains just the python module with one __init__.py file which loads a shared library from scine-sparrow.

The way this is implemented is python-only, so this package could be noarch python.

It should only depend on python and scine-sparrow. Run dependencies on blas and gcc and yaml, etc are not correct. 

```
    - _openmp_mutex >=4.5
    - libblas >=3.8.0,<4.0a0
    - libgcc-ng >=10.3.0
    - libstdcxx-ng >=10.3.0
    - python >=3.7,<3.8.0a0
    - python_abi 3.7.* *_cp37m
    - scine-sparrow 3.0.0.*
    - scine-sparrow >=3.0.0,<4.0a0
    - scine-utilities-python
    - yaml-cpp >=0.6.3,<0.7.0a0
```

recipes/scine-sparrow-python/meta.yaml
carterbox: This is fine.

awvwgk:
Good point, I changed the build script to just do the few steps implemented in CMake to build the setup.py and than install directly with pip.

awvwgk:
@carterbox Is there something I have to keep in mind with the new labels?

carterbox:
They are triggered by anyone mentioning a help-team to ask for review. They exist to reviewers can quickly see whether a review was requested or not because we get notified for all activity even if a PR isn't ready for review.

awvwgk:
Alright, so I can set the label again by mentioning the team?

@conda-forge/help-python Review comments have been addressed.

carterbox:
Nice use of sed!

awvwgk:
Thanks a lot. Also, I like the new label system.

Is there an issue for providing further feedback?

carterbox:
The original discussion of automated labels is https://github.com/conda-forge/staged-recipes/issues/18023. If you want to suggest improvements probably open a new issue with reference to the old.

PR #18664, opened 2022-04-13
Add aerobulk-python

jbusecke:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/aerobulk-python/meta.yaml
+++ b/recipes/aerobulk-python/meta.yaml
@@ -13,7 +13,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://github.com/xgcm/aerobulk-python/archive/refs/tags/{{version}}.tar.gz
+  url: https://github.com/xgcm/aerobulk-python/archive/refs/tags/{{ version }}.tar.gz
   sha256: 36e8aecab8a9a3b13a1e28cfcec2f077359cbe3ab38745b9951a1517652b3ec6
 
 
@@ -51,7 +51,7 @@ about:
   home: https://github.com/xgcm/aerobulk-python
   summary: 'A python wrapper for the aerobulk air-sea flux fortran code'
   license: GPL-3.0-only
-  license_family: GLP
+  license_family: GPL
   license_file: LICENSE.txt
 
 extra:
--- a/recipes/aerobulk-python/meta.yaml
+++ b/recipes/aerobulk-python/meta.yaml
@@ -40,7 +40,7 @@ test:
   # Some packages might need a `test/commands` key to check CLI.
   # List all the packages/modules that `run_test.py` imports.
   imports:
-    - aerobulk.test
+    - aerobulk.mod_aerobulk_wrap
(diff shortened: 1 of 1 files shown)


jbusecke:
@rabernat could you confirm that we got everything covered in terms of the license (two unchecked bullets in the top comment). Once those are checked I think this is ready to go.

jbusecke:
Oh also @rabernat I added you as a maintainer, hope that was ok.

rabernat:
of course! :+1:

rabernat:
> * If static libraries are linked in, the license of the static library is packaged.

We are using the same license as the static library (GPL-3):
https://github.com/brodeau/aerobulk/blob/master/LICENSE

I hope this is correct.

jbusecke:

Please see the above discussion. With this package we provide a python wrapper for the [aerobulk](https://github.com/brodeau/aerobulk/issues/8). We currently use aerobulk as a submodule and are unsure what that means in terms of the license. Is there some guidance for packages that are wrappers for other packages by any chance? Many thanks

recipes/aerobulk-python/meta.yaml
```diff
@@ -0,0 +1,62 @@
+# Note: there are many handy hints in comments in this example -- remove them when you've finalized your recipe
```
ocefpaf, line 1: Ping! BTW, you can probably create recipes faster and almost merge-ready with grayskull. See the comment below.

jbusecke: Ah yes, thanks. I meant to do that, actually. Let me remove the hints now.

jbusecke:
@ocefpaf do you have an opinion on the issue with the license discussed above?

ocefpaf:
> @ocefpaf do you have an opinion on the issue with the license discussed above?

That is fine for python wrappers. There are no source packages for Fortran code and the license is compatible.

The last issue for us to merge is the version number. 0.0.0 is not really valid.

jbusecke:
Ok Ill update the pypi release to 0.1.0 and then correct it here? Thanks for the help.

jbusecke:
Hmm I am running into issues here again. The v0.0.0 that the original version of this recipe checked out was uploaded manually, and included a binary wheel (https://pypi.org/project/aerobulk-python/0.0.0/#files). I then tried to automate the pypi release via a github action [here](https://github.com/xgcm/aerobulk-python/blob/main/.github/workflows/publish.yml), but I had to restrict it to a source distribution, which does not include the fortran code from the aerobulk repo, which we keep as a git submodule.
I believe that the issue for the current CI failures are simply missing files since the source tarball does not include the git submodule content. 

Is there a recommended way to deal with this in a conda-forge recipe? Or do I need to deal with this on the pypi upload side? 

Do I need to change the source url in the recipe to a github release instead of a pypi tarball? I might also be missing something else entirely.

jbusecke:
After looking at it with fresh eyes, that makes total sense. I think I have it figured out now. Thank you very much for the pointer @ocefpaf.

PR #18675, opened 2022-04-14
Adding meta-package for Oracle JDK

phreed:
This is a meta-package i.e. it does not package the Oracle JDK but rather it adds a locally installed JDK to the environment.

@conda-forge/help-java

This package is not yet ready for review.


recipes/example/recipe.yaml
```diff
@@ -0,0 +1,90 @@
+context:
```
beckermr, line 1: Delete this file.

recipes/oracle-jdk/recipe.yaml
```diff
@@ -0,0 +1,46 @@
+
```
beckermr, line 1: Remove this file too.

phreed:
We do not like `boa`?

recipes/oracle-jdk/recipe.yaml
phreed: We do not like `boa`?

recipes/oracle-jdk/post-link.sh
```diff
@@ -0,0 +1,48 @@
...
+
+# Discovery
+SetLocal EnableExtensions EnableDelayedExpansion
+if [ ! -d "${ORACLE_JDK_DIR}" ]; then
+  {
+    echo "The target JDK version has not been installed. ${ORACLE_JDK_DIR}";
+    echo "see https://www.oracle.com/java/technologies/downloads/#java8-linux";
+    echo " jdk-8u321-linux-x64.rpm "
+  } >> "${CONDA_PREFIX}/.messages.txt"
+  exit 0
```
phreed, line 26: This should be `exit 1` as it indicates failure.
I set it to `exit 0` so the pipeline test would pass.

recipes/oracle-jdk/post-link.bat
```diff
@@ -0,0 +1,46 @@
...
+echo "   PKG_NAME:     %PKG_NAME%"     >> "%CONDA_PREFIX%\.messages.txt"
+echo "   PKG_VERSION:  %PKG_VERSION%"  >> "%CONDA_PREFIX%\.messages.txt"
+echo "   PKG_BUILDNUM: %PKG_BUILDNUM%" >> "%CONDA_PREFIX%\.messages.txt"
+:: set | FindStr "PKG" >> "%CONDA_PREFIX%\.messages.txt"
+
+set "PKG_BIN=%CONDA_PREFIX%\bin"
+set "PKG_UUID=%PKG_NAME%-%PKG_VERSION%_%PKG_BUILDNUM%"
+
+:: should I be using `reg query ` to find the jdk directory path?
+set "ORACLE_JDK_DIR=%ProgramFiles%\Java\jdk1.8.0_321"
```
phreed, line 15: I will be needing to change discovery to work with any (latest) release of a particular version as it changes fairly regularly.
Rather than be specific about jdk1.8.0_321 the discovery should pick the latest revision; only match on the prefix jdk1.8.0

recipes/oracle-jdk/recipe.yaml
beckermr: We like boa. We don't use it in feedstocks and so do not need the files.

recipes/oracle-jdk/post-link.bat
```diff
@@ -0,0 +1,46 @@
...
+
+:: Discovery
```
phreed, line 17: Given how often the revision for v8 changes. It would be better to find the latest revision. 
i.e. match on jdk1.8.0* and not the specific jdk1.8.0_321

Ideally, I would like to discover the ORACLE_JDK_DIR in the `post-link` and then make it available to `pre-unlink`, `activate`, and `deactivate`. Is there a way to do that?

beckermr: IDK if there is an official way to do it. I am sure putting a file with the path in the prefix would work fine.

recipes/oracle-jdk/post-link.sh
beckermr: We likely cannot fully test this recipe. You'll need to find a way to make the recipe pass on CI and still fail later. Conda build has a special env var to indicate it is in its testing phase. Maybe you can use that?

recipes/oracle-jdk/post-link.bat
phreed: I used a similar approach. `post-link` writes a file to the "conda-meso" folder.
```
set "CONDA_MESO=%CONDA_PREFIX%\conda-meso\%PKG_NAME%-%PKG_VERSION%_%PKG_BUILDNUM%"
set "DISCOVERY_SCRIPT=%CONDA_MESO%\discovery.bat"
```

beckermr: One thing here is that the `conda-*` namespace in the PREFIX should not be used by packages. We should pick something very obvious that reflects the package that is dealing with the files, e.g. `oracle-jdk-meta` or whatever. This makes it painfully obvious where files might be if we are debugging user environments later.

phreed: I am trying to parallel the pattern set by `conda-meta` i.e. `%CONDA_PREFIX%\conda-meta\%PKG_NAME%-%PKG_VERSION%_%PKG_BUILDNUM%.json"
I have several packages that I need to do a similar treatment on and I do not think cluttering the `%CONDA_PREFIX%` would be wise.

beckermr: Hmmm. The vast majority of environments won't have these packages installed and so this didn't concern me. 

Maybe @conda-forge/core has thoughts?

isuruf:
What's the purpose of this package?

phreed:
@isuruf 
I have written a bunch about this so rather than repeating myself.
https://github.com/conda-forge/conda-forge.github.io/issues/1706
[There is also a README.md in this project](https://github.com/conda-forge/staged-recipes/pull/18675/commits/c291ee799d14e9513f142b44f3c3906ede0e66c7#diff-760ee96dffb36277321bf258904370e07ea7e95eaec913cdabd22101972804dd)

phreed:
@beckermr I just saw your message about converting this to a draft.
In any case, it works for my purposes.

phreed:
I am switching this back to being a draft.
I am studying the following to see how they handle `external`  

- [ ]  https://conda-forge.org/docs/user/tipsandtricks.html#using-external-message-passing-interface-mpi-libraries
- [ ] https://github.com/conda-forge/openmpi-feedstock/blob/main/recipe/meta.yaml
- [ ] https://github.com/conda-forge/mpich-feedstock/blob/main/recipe/meta.yaml

recipes/oracle-jdk/post-link.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash -euo
```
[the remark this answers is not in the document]
phreed: Doing the discovery during activate would work as well.
There are actually a bunch of executables in addition to `java`.
It just seemed like doing discovery during install was the right way to go.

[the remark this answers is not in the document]
hmaarrfk: The issue is that your environment might change from install time to run time.

I find it best to delay "auto" as long as possible.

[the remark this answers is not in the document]
hmaarrfk: Anyway, my question was stemming from curiosity.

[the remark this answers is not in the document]
phreed: I guess I would like to try it the way it is.

phreed:
I am going to add this to an Anaconda organization in its current form.
I would like it to find its way into conda-forge at some point but I understand there may be reasons to not accept it in its current form.

hmaarrfk:
I think it would be interesting to have this in conda-forge. I think Oracle JDK is widely used and would provide an alternative, at least for us to test, compared to OpenJDK. I do however object to the discovery being run in in the `post-link` phase. I don't think tit is necessary, and seems more robust to run in the activate phase.

However, it is tricky to get these packages merged into conda-forge. For example cuda 11.6 still isn't in, however, it exists in nvidia's channels for a while. Mostly due to the difficult in testing downstream effects which are "hidden" from conda-forge.

Hosting the packages in your own conda channel, will allow people to test and review.

phreed:
Once I have put this in https://anaconda.org/mesomorph/dashboard I will try out moving the discovery logic into activate.

recipes/oracle-jdk/post-link.sh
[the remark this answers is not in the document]
phreed: I will try out moving the discovery logic into activate.

PR #18676, opened 2022-04-15
DOC: Reword PR template to encourage autolabel action use [ci skip]

carterbox:
The new autolabeling GitHub action is better than @mentions because it
provides more information that is filterable/sortable from the PR tab.
This saves reviewers time because they can see whether another reviewer
has already responded or (for old PRs) if the PR is waiting for author action.

However, these benefits require that users mention a team directly instead
of asking the conda-forge-admin. WHY? The action ignores comments from other
bots. AND most users use "ping team" instead of "ping conda-forge/language-team"

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,32 +1,31 @@
 <!--
 Thank you very much for putting in this recipe PR!
 
-This repository is very active, so if you need help with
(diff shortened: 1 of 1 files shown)


BastianZim (approved):
If you’re happy with everything, feel free to merge. The changes look good to me.

PR #18731, opened 2022-04-22
Add `ibis-heavyai`

tupui:
This recipe will replace https://github.com/conda-forge/ibis-omniscidb-feedstock which will be archived.

It's part of the OmniSci rebranding to HeavyAI.

PyPI version of `ibis-heavyai` not yet released.

- https://github.com/heavyai/ibis-heavyai/pull/64

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/ibis-heavyai/meta.yaml
+++ b/recipes/ibis-heavyai/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "ibis-heavyai" %}
+{% set version = "1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ibis-heavyai-{{ version }}.tar.gz
+  sha256: ...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.7
+  run:
+    - python >=3.7
+    - ibis-framework >=2.0
+    - heavyai
+    - rbc-project >=0.8.0
+    - pyarrow >=3.0.0
+    - regex
+    - pandas
+    - shapely
+    - importlib_metadata
+
+test:
+  imports:
+    - ibis_heavyai
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/heavyai/ibis-heavyai
+  summary: Ibis HeavyDB backend
+  dev_url: https://github.com/heavyai/ibis-heavyai
+  doc_url: https://github.com/heavyai/ibis-heavyai/docs/heavyai.rst
+  license: Apache-2.0
+  license_file: LICENSE.md
+  description: |
+    Ibis is a Python framework to write data analytics code, in a similar way
+    to pandas. Ibis can execute the queries not only in memory
+    (as pandas does), but in different backends, including SQL databases,
+    and analytics databases, like HeavyDB. HeavyDB is SQL-based, relational,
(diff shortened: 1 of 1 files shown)


tupui:
@pearu @andrewseidl please acknowledge that you would like to maintain this recipe.

andrewseidl:
Confirmed

pearu:
Confirmed

recipes/ibis-heavyai/meta.yaml
```diff
@@ -0,0 +1,61 @@
+{% set name = "ibis-heavyai" %}
+{% set version = "1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ibis-heavyai-{{ version }}.tar.gz
+  sha256: ...
```
tupui, line 10:
```suggestion
  sha256: ccebecc331eeb8c4f50df7984e7445bfebd25285bd674d2ec2a6326002b7ece0
```

recipes/ibis-heavyai/meta.yaml
```diff
@@ -0,0 +1,59 @@
+{% set name = "ibis-heavyai" %}
+{% set version = "1.0" %}
```
carterbox, line 2: Why are you releasing version 1.0 when version 2.0.5 was released in January? You can also get the source tar directly from GitHub instead of via PYPI; just swap the url and SHA.

recipes/ibis-heavyai/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - pyarrow >=3.0.0
+    - regex
+    - pandas
+    - shapely
+    - geopandas
+    - importlib_metadata
+
+test:
+  imports:
+    - ibis_heavyai
```
carterbox, line 36: Please remember to add pip check back if updating to the latest version solves the pinning problem.

carterbox (changes requested):

recipes/ibis-heavyai/meta.yaml
tupui: Because it's a new package in the end so it does not really matters. In the end, end users have to adjust things and a hard break is legitimate here.

As for where to get the package from, I followed the recipe template maker Grayskull. Why should we prefer GitHub over PyPI? Because to me PyPI seems to be the better way.

recipes/ibis-heavyai/meta.yaml
tupui: We will check this for the next release sure.

tupui:

recipes/ibis-heavyai/meta.yaml
carterbox: If you keep your pypi releases up to date, there are no missing files from the pypi release, and there is no conflicting release information on GitHub, then there is no difference between getting the source from pypi or github.

However, the development URL for this package points here: https://github.com/heavyai/ibis-heavyai

This repository already has releases in version 2.0.x. I thought that maybe your PYPI releases are out of date which is why I suggested GitHub releases, but now I realize it's probably that you kept all of the git history from before the name change, so all the old tags and releases are still there.

If you start with version 1.0 under the new name you will eventually run into tag/release clashes in your GitHub repo which already has version 2.0.x and has already been forked by others and so every fork may contain these tags too. So I am suggesting that you continue versioning from 2.0.5 or start with 3.0.0 or clear out all those old tags, so there is no confusion.

tupui: Thank you for pointing this out. I will keep this in mind.

PR #18735, opened 2022-04-22
Add `dials`

ndevenish:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/dials/meta.yaml
+++ b/recipes/dials/meta.yaml
@@ -49,22 +49,6 @@ requirements:
     - tqdm
     - wxpython
 
-test:
-  source_files:
-    - dials/tests
-    - dials/conftest.py
-  requires:
-    - dials-data
-    - pytest
-    - pytest-mock
-    - pytest-forked
-    - pytest-xdist
-    - pip
-  imports:
-    - dials
-  commands:
-    - pytest $SRC_DIR/dials/tests --regression -n auto
-    - pip check
 
 about:
   home: https://github.com/dials/dials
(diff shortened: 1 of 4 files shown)


Anthchirp:
Happy to be a maintainer

rjgildea:
Happy to be added as a maintainer 👍

recipes/dials/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - {{ pin_compatible('boost') }}
+    - boost-cpp
+    - {{ pin_compatible('cctbx-base') }}
+    - cctbx
+    - {{ pin_compatible('dxtbx') }}
+    - gemmi
+    - h5py
+    - jinja2
+    - matplotlib-base
+    - {{ pin_compatible('numpy') }}
```
Anthchirp, line 40: Is this compiling against numpy or just using numpy?
If it is compiling against, then should add numpy to host dependencies for completeness sake.

Also alphabetical sorting of dependencies wouldn't go amiss 😉

Anthchirp (approved):

recipes/dials/meta.yaml
Anthchirp: (And if only importing then just list it as `numpy`)

ndevenish: That's a a good point, I don't think dials uses it at compile time, only dxtbx and cctbx. I think I left the pin_compatible _for_ these dependencies, but imagine that their restrictions would work.

Also, I'd tried to make it package-alphabetical but it's probably cleaner to do it in a way that can be simply sorted.

Anthchirp:
Looks good to me

PR #18742, opened 2022-04-23
Add histolab

ernestoarbitrio:
Checklist
- [x] Build number is 0.


ernestoarbitrio:

recipes/histolab/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/histolab-{{ version }}.tar.gz
+  sha256: 7bbcfbc2c61a44ff38e0f4e618bb8e172516e793326beba23f793845750b01fb
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: True  # [win]
```
ocefpaf, line 15: Is the reason to skip Windows some missing dependency there? If so, can this be noarch instead? We prefere noarch b/c it reduces the strain on our CIs.

ernestoarbitrio: The only reason is that we support windows but only for python 3.7 due to some incompatibility with `openslide`. Is it possible to skip specific os and specific py versions?

ocefpaf: In that case a noarch package would be broken. If you don't care about Windows you can leave as-is or you can do:

```yaml
  skip: True  # [win and (not py37)]
```

(untested but something like that :-)

ernestoarbitrio: It seems that `openslide` is not on the windows build, so I'm gonna skit the windows at all

Issue #17059, opened 2021-11-24
Package scalene


h-vetinari:
Just saw this in an awesome talk by Emery Berger as a really state-of-the art profiler, and incidentally that it only listed pip install instructions.

See https://github.com/plasma-umass/scalene

PR #18747, opened 2022-04-24
Adding Scalene.

emeryberger:
Checklist
- [X] Build number is 0.

(I have self listed myself in the maintainer section)

Fixes https://github.com/conda-forge/staged-recipes/issues/17059
Closes https://github.com/conda-forge/staged-recipes/pull/17325


recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/{{ org }}/{{ package_name }}/archive/refs/tags/v{{ version }}.tar.gz
+  # See https://pypi.org/project/scalene/#files
+  sha256: 2d77a1d5f2a1b9314e64e9ec97d068e6e2e0cb74a28f219924477c8b251b806a
+
+build:
+  skip: True  # [py<38]
+  number: 0
```
jakirkham, line 18: Would suggest just using `script` and dropping `build.sh` & `bld.bat`. This line is [taken from the example]( https://github.com/conda-forge/staged-recipes/blob/66111e0d8cb287af412285f00a81a1cf9fdd67f5/recipes/example/meta.yaml#L38 )

```suggestion
  number: 0
  script: {{ PYTHON }} -m pip install . -vv
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set package_name = "scalene" %}
+{% set initial = "s" %}
```
jakirkham, line 2: This appears unused. Should we drop it?

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+{% set version = "1.5.7" %}
+{% set org = "plasma-umass" %}
+{% set author = "emeryberger" %}
```
jakirkham, line 5: These are used once. Would suggest we inline these

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - make
+    - pip
+  host:
+    - python
```
jakirkham, line 27: `pip` belongs in `host` along with the `python` used

```suggestion
  host:
    - python
    - pip
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - rich
+    - cloudpickle
+    - numpy
+    - pynvml
+
+test:
+  imports:
+    - {{ package_name }}
+  source_files:
+    - test/testme.py
```
jakirkham, line 41: This file appears to be missing. Should it be included in the recipe as well?

recipes/scalene/meta.yaml
emeryberger: Removing, thanks.

recipes/scalene/meta.yaml
emeryberger: Will do.

recipes/scalene/meta.yaml
emeryberger: It's in the repository... Is that not enough?

emeryberger: I just verified that `conda build .` works on my platform and executes that test, as intended.

emeryberger:
@jakirkham Thanks for your help. Finally a successful run on all supported platforms, and I believe I have addressed all concerns.

nicovank:

jakirkham:

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,56 @@
+{% set package_name = "scalene" %}
+{% set version = "conda_2022_08_07" %}
```
ocefpaf, line 2: The previous tag was [v1.5.9.1](https://github.com/plasma-umass/scalene/releases/tag/v1.5.9.1). What was the reason for this change? It is quite unusual version  string. You can either use [SemVer](https://semver.org/) or [CalVer](https://calver.org/) in conda-forge. If you need to use that you should at least drop the `conda_` from the version string and make that a full CalVer replacing the `_` and `.`.

jakirkham: Right am curious was this just done for testing purposes (to grab currently unreleased changes for testing)? If so (assuming we are happy with other things here), does it make sense to do a proper release that could then be used in the recipe?

emeryberger: I do normally use SemVer. The version in the meta tag was indeed for testing. I am happy to issue a new release of Scalene (which would be 1.5.10). Just let me know if that's what needed, and I will do it.

One question I have is what the process is for issuing new releases on Conda; is it a new PR every time (like this one), or is there a more streamlined possibility going forwards?

ocefpaf: Once the PR is merged a feedstock will be created, see https://conda-forge.org/docs/maintainer/adding_pkgs.html#feedstock-repository-structure

If you follow SemVer, the bots will find the new tags and send PRs to the feedstock to update it. If there are no big changes you can usually just merge that one. Sometimes  you may need to adjust the dependencies.

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - cloudpickle
+    - numpy
+    - pynvml
+
+test:
+  imports:
+    - {{ package_name }}
+  source_files:
+    - test/testme.py
+  commands:
```
jakirkham, line 43: Would suggest adding this to validate dependencies. This can help with validating autoupdate PRs that Filipe mentioned above

```suggestion
  requires:
    - pip
  commands:
    - pip check
```

recipes/scalene/testme.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
+#import math
+
+from numpy import linalg as LA
+
+arr = [i for i in range(1,1000)]
+
+def doit1(x):
+#    x = [i*i for i in range(1,1000)][0]
```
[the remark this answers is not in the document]
emeryberger: Removing them would be fine.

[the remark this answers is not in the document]
emeryberger: (Doing this now.)

recipes/scalene/meta.yaml
emeryberger: Sounds great. Please let me know if I need to issue a new release to get this merged.

recipes/scalene/testme.py
```diff
@@ -0,0 +1,51 @@
+import numpy as np
+#import math
```
jakirkham, line 2:
```suggestion
```

recipes/scalene/meta.yaml
jakirkham: Yeah cutting a new release would be great. Seems like things are working otherwise

emeryberger: On it.

emeryberger:

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
+{% set package_name = "scalene" %}
+{% set version = "v1.5.10" %}
```
jakirkham, line 2:
```suggestion
{% set version = "1.5.10" %}
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+package:
+  name: {{ package_name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/0b/9d/54e252b4d573fd65e5301ae75fc56d397d64027f7953e13394ba3692db26/scalene-1.5.10.tar.gz
```
jakirkham, line 9: Would suggest following [the `url` from the `example` recipe here]( https://github.com/conda-forge/staged-recipes/blob/85f1d3abc6fcec742b0ab569b103cf30b4f93f8e/recipes/example/meta.yaml#L16 ). This makes it easier for the autoupdate bot to send version update PRs

```suggestion
  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```

jakirkham (commented):
Thanks Emery! 🙏

Added a couple small suggestions below. Otherwise LGTM

jakirkham:

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  skip: true  # [win]
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - make
+    - pip
```
carterbox, line 24:
```suggestion
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - python
+    - pip
+    - setuptools_scm
+    - wheel
+  run:
+    - python
+    - rich
+    - cloudpickle
+    - numpy
+    - pynvml
```
carterbox, line 34:
```suggestion
    - rich >=9.2.0
    - cloudpickle >=1.5.0
    - numpy
    - pynvml >=11.0.0
```
numpy is not listed in setup.py as a requirement and is only imported in the "old" module. Is it really required?

carterbox (changes requested):

recipes/scalene/meta.yaml
emeryberger: It was required for the test, which I thought was necessary, but I guess not?

emeryberger: Scratch that, it should in theory be coming from the tarball, but that test _does_ require numpy. I could use a different test that does not.

emeryberger: Let me see how that goes.

carterbox: If numpy is only requires for the tests, then it should be listed in the `test/requires` not `requirements/run`.

emeryberger: Thanks for the clarification. If this test works, it'll be fine (it's really just to make sure that it produces a profile without a hitch; the numpy test is not needed - it's just what we use in our CI).

jakirkham:

jakirkham (approved):
LGTM. Thoughts from anyone else?

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  imports:
+    - {{ package_name }}
+  source_files:
+    - test/original/bm_mdp.py
+  requires:
+    - pyperf
+    - pip
+  commands:
+    - pip check
+    - python3 -m scalene --cli test/original/bm_mdp.py  # [not win]
```
ocefpaf, line 46:
```suggestion
    - {{ PYTHON }} -m scalene --cli test/original/bm_mdp.py  # [not win]
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - {{ PYTHON }} -m scalene --cli test/original/bm_mdp.py  # [win]
+    
+about:
+  home: https://github.com/plasma-umass/{{ package_name }}
```
ocefpaf, line 50: Using jinja like this obfuscate the recipe, render the links useless when reading/inspecting the recipe only, and doesn't really save us anything. I prefer the actual pkg name here instead.

emeryberger: This was a suggested change from a reviewer that I incorporated. It also meets [the DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Anyway, in the interest of getting out of this churn cycle, can we just leave it?

emeryberger: Suggested change was here: https://github.com/conda-forge/staged-recipes/pull/18747/commits/2d28eb0cb3e3f7cffde34296e03ebfb78ca1ae46#r886185791

recipes/scalene/meta.yaml
ocefpaf: The thing is, it is working but with which Python? If that is failing with our Python it is not working then. Changes are that calling `python3` you are not calling the conda-forge Python on CIs.

emeryberger: OK, let's see what happens!

recipes/scalene/meta.yaml
emeryberger: OK, will change.

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
+{% set package_name = "scalene" %}
```
ocefpaf, line 1: You'll need to change it here too.

```suggestion
{% set name = "scalene" %}
```

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - {{ name }}
+  source_files:
+    - test/original/bm_mdp.py
+  requires:
+    - pyperf
+    - pip
+  commands:
+    - pip check
+    - {{ PYTHON }} -m scalene --cli test/original/bm_mdp.py  # [not win]
+    - {{ PYTHON }} -m scalene --cli test/original/bm_mdp.py  # [win]
```
ocefpaf, line 47:
```suggestion
    - {{ PYTHON }} -m scalene --cli test\\original\\bm_mdp.py  # [win]
```

emeryberger:
Failing because it can't find "python", apparently?

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=556369&view=logs&j=e35d4f76-8ff2-5536-d795-df91e63eb9f7&t=fa7b4b17-b6ff-5c9c-8cfc-15f888c92310&l=2213

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - {{ name }}
+  source_files:
+    - test/original/bm_mdp.py
+  requires:
+    - pyperf
+    - pip
+  commands:
+    - pip check
+    - {{ PYTHON }} -m scalene --cli test/original/bm_mdp.py  # [not win]
+    - {{ PYTHON }} -m scalene --cli test\\original\\bm_mdp.py  # [win]
```
jakirkham, line 47: Since `{{ PYTHON }}` seems to be expand to `python` in the `host` environment, using `{{ PYTHON }}` here doesn't seem to work. Would just use `python`

```suggestion
    - python -m scalene --cli test/original/bm_mdp.py  # [not win]
    - python -m scalene --cli test\\original\\bm_mdp.py  # [win]
```

emeryberger: Fingers crossed!

emeryberger: If it just expands to "python", won't this have the same effect? I will try it but next thing to try seems to be "python3". (Unless there is a Jinja variable`PYTHON3`...)

jakirkham: `{{ PYTHON }}` is Jiinja syntax. This gets filled in by an initial parsing of the recipe. Ideally `{{ PYTHON }}` would be set to something context dependent, but that doesn't not appear to be the case (hence why the path to the `host` `python` is used in the `test` phase).

It shouldn't present an issue as things like `python` are not expanded. The `{{ ... }}` syntax is needed to look up a Jinja variable and it expand it (as happens with `{{ PYTHON }}`)

carterbox (changes requested):
It's still unclear to me whether `scalene.old` is intended to be packaged. This module does not have an `__init__.py` If this module is not part of the package, then it should be excluded and `numpy` is not a run dependency.

emeryberger:
@carterbox `scalene.old` is no longer in the PR (unless I am missing something).

jakirkham:
Yeah I don't see this either. Could you please clarify what you are referring to Daniel?

carterbox:
The `scalene.old` module is [included in the source](https://github.com/plasma-umass/scalene/tree/master/scalene/old) and is being shipped. It is mentioned in the logs multiple times.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=556415&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1459

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=556415&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1259

emeryberger:
I see, you mean the `old/` in the `scalene` source directory. I'm not sure why minutiae about the contents of the source repo matter for the purposes of this package... in any event, that code is not a module, but a currently-disabled piece of functionality from Scalene's leak analysis algorithm.

jakirkham:
Thanks for the context Daniel! 🙏

Given this is not unique to the Conda package (both the `sdist` and `whl`s include it), would suggest we leave it alone (instead of doing something special/different for the Conda package).

If we want this addressed, think it would be best to [raise a new issue upstream]( https://github.com/plasma-umass/scalene/issues/new/choose ) and suggest [excluding these files via the MANIFEST.in]( https://packaging.python.org/en/latest/guides/using-manifest-in/#manifest-in-commands ). That way they would be removed from all redistributables (including Conda packages).

emeryberger:
Thanks for the suggestion re: the `MANIFEST.in` file, @jakirkham - I've made that modification (excluding `old/`) and it will be reflected in the next release.

jakirkham:
Thanks Emery! 🙏

For reference, that change is here ( https://github.com/plasma-umass/scalene/commit/482284cdbfe8cdaadb3e639681d7f1d4c72c0f74 )

carterbox:
>  I'm not sure why minutiae about the contents of the source repo matter for the purposes of this package

It's related to packaging because it's the only place that `numpy` is imported. [Since `numpy` is not listed in `setup.py`](https://github.com/plasma-umass/scalene/blob/482284cdbfe8cdaadb3e639681d7f1d4c72c0f74/setup.py#L167), you are adding an extra dependency for conda users.

emeryberger:
OK - that said, that file is never included anywhere, so I don't understand why it's an extra dependency (it's currently orphaned). In any event, I'd hate to cut a new release to just exclude this one file, but let me know if I need to. Thanks.

recipes/scalene/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - python
+    - pip
+    - setuptools_scm
+    - wheel
+  run:
+    - python
+    - rich >=9.2.0
+    - cloudpickle >=1.5.0
+    - numpy
```
carterbox, line 33:
```suggestion
```
I think you're not understanding my point. `numpy` is not listed in the package `install_requires`. Your conda recipe should not list `numpy` as a dependency if `numpy` is not imported (which it isn't; except in `scalene.old`).

carterbox:
`scalene.old` will no longer be shipped with some unknown future release of setuptools or you can manually exclude it from the next release.

h-vetinari:
Woohoo! 🥳

recipes/scalene/meta.yaml
jakirkham: Gotcha. Outside of this file, it looks like [there are a few uses]( https://github.com/plasma-umass/scalene/search?p=2&q=numpy ), but mainly tests. Not sure how [this profile]( https://github.com/plasma-umass/scalene/blob/56719723b694030905f25223dd207c6a97909e0f/scalene/scalene-gui/profile.json#L80 ) figures in.

jakirkham:
Thanks all! 🙏

emeryberger:
Thanks, everyone, for all the help in this process! I expect our next package to integrate more smoothly :).

Quick question: what do I need to do get a "osx-arm64" build?

BastianZim:
Just add the name here and a bot will make a PR: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/migrations/osx_arm64.txt

jakirkham:
xref: https://github.com/conda-forge/scalene-feedstock/issues/10

PR #18753, opened 2022-04-25
Add pydevtool

tupui:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pydevtool/meta.yaml
+++ b/recipes/pydevtool/meta.yaml
@@ -1,12 +1,11 @@
-{% set name = "pydevtool" %}
 {% set version = "0.2.0" %}
 
 package:
-  name: {{ name|lower }}
+  name: pydevtool
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pydevtool-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/p/pydevtool/pydevtool-{{ version }}.tar.gz
   sha256: e821ce0b637b1bbf7638b1de23522b76c23b8e9900678ff46e6c846dcd9464e2
 
 build:
--- a/recipes/pydevtool/meta.yaml
+++ b/recipes/pydevtool/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "pydevtool" %}
+{% set version = "0.2.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pydevtool-{{ version }}.tar.gz
+  sha256: e821ce0b637b1bbf7638b1de23522b76c23b8e9900678ff46e6c846dcd9464e2
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+  run:
+    - doit
+    - python >=3.8
+
+test:
+  imports:
+    - pydevtool
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/pydoit/pydevtool
+  summary: CLI dev tools powered by pydoit
+  dev_url: https://github.com/pydoit/pydevtool
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - tupui
+    - schettino72
--- a/recipes/pydevtool/meta.yaml
+++ b/recipes/pydevtool/meta.yaml
@@ -40,4 +40,5 @@ about:
 extra:
   recipe-maintainers:
     - tupui
+    - rgommers
     - schettino72


tupui:

schettino72:
> @schettino72 can you confirm here that you would like to help with maintaining this recipe?

Yes, I can help.

rgommers:
Please add me too @tupui

recipes/pydevtool/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  home: https://github.com/pydoit/pydevtool
+  summary: CLI dev tools powered by pydoit
+  dev_url: https://github.com/pydoit/pydevtool
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - tupui
+    - schettino72
```
tupui, line 43:
```suggestion
  recipe-maintainers:
    - tupui
    - rgommers
    - schettino72
```

tupui:
Sure done 😃

recipes/pydevtool/meta.yaml
```diff
@@ -0,0 +1,44 @@
+{% set name = "pydevtool" %}
+{% set version = "0.2.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pydevtool-{{ version }}.tar.gz
```
h-vetinari, line 9: Having a `name` variable (kinda) makes sense in the example to help people populate their pypi URLs quickly and correctly, but it's a "fake" kind of variable, because it's not like it can be changed within this recipe. I'd suggest to remove it. 

```suggestion
{% set version = "0.2.0" %}

package:
  name: pydevtool
  version: {{ version }}

source:
  url: https://pypi.io/packages/source/p/pydevtool/pydevtool-{{ version }}.tar.gz
```

recipes/pydevtool/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - python >=3.8
+  run:
+    - doit
+    - python >=3.8
+
+test:
+  imports:
+    - pydevtool
+  commands:
+    - pip check
```
h-vetinari, line 29: I don't know pydevtool, but from looking at the source, it mainly (or even predominantly) is made up of a CLI.

I'd suggest that some common commands be added to he `commands:` section.
```suggestion
  commands:
    - pip check
    - pydevtool something something
```

h-vetinari (commented):
Aside from a nit and increasing test coverage, this looks pretty good! :)

Thanks @tupui!

recipes/pydevtool/meta.yaml
schettino72: It is not a CLI itself (for now), but provide some helpers to build CLIs.

tupui: I disagree with this point. It's not the responsibility of a package manager to test the package but the packaging. Plus it's a simple Python set of functions. Hence if we can import the namespace, from a packaging point of view, the job is done. If there are other issues there are to be fixed/found on the package repo.

recipes/pydevtool/meta.yaml
tupui: Why not. Note that this was generated using the recommended template generator. If this is not wanted anymore, it should be fixed there.

recipes/pydevtool/meta.yaml
h-vetinari: Well... If only the packaging were independent of the functionality of the package.

As it is, it is best practice to run the upstream test suite for the feedstock, in order to make sure no functionality has been inadvertently broken by the packaging (and you can believe me from many many many painful experiences, that a package import is not even close to an adequate proxy for correct functionality - more like the barest minimum).

That said, there are cases where there's indeed not much to test (and this may be one of them), but I disagree in turn with your characterization of the responsibilities of the package manager (which isn't conda-specific; distros and other package managers generally also use the project's test suites to likewise ensure functionality of their packaging).

tupui: > Well... If only the packaging were independent of the functionality of the package.

We are talking about a pure Python package which consist in a single file. I stand by my statement.

tupui: > ... other package managers generally also use the project's test suites to likewise ensure functionality of their packaging).

Exactly my point. It's not the responsibility of individual packages to do that. Package managers can use these package if they want in their own test suite.

tupui: Anyway, feel free to add a suggestion. I don't plan to do more now.

h-vetinari: As you will find out after enough time in packaging, that is not a sustainable approach long-term. Packages evolve, the packaging often doesn't - until something breaks somewhere. The work to add some proper testing is a good investment compared to fixing the fallout of a breakage later (even when factoring in ~low likelihood), but well; some things need to be experienced I guess.

recipes/pydevtool/meta.yaml
h-vetinari: As I wrote, the template makes sense (+/-) so that newcomers don't have to figure out the PyPI links themselves; for a final recipe, it's useless though.

tupui: As I said, I used Grayskull to generate this file. There is nothing that would prevent this tool to resolve the link and just leave the version as a variable.

recipes/pydevtool/meta.yaml
h-vetinari: [some of your comments arrived in my browser after my last comment]

> > ... other package managers generally also use the project's test suites to likewise ensure functionality of their packaging).
> 
> Exactly my point. It's not the responsibility of individual packages to do that. Package managers can use these package if they want in their own test suite.

conda/mamba + conda-forge is a package manager, or maybe we have a crossed wire somewhere? People developing using conda-forge packages (like scipy does) expect correct functionality - where should that be ensured IYO? With our current infrastructure, there's nothing _but_ the feedstocks (for individual packages) that can reasonably ensure correctness of each consumable output.

> Anyway, feel free to add a suggestion. I don't plan to do more now.

Fine by me

tupui:
https://github.com/orgs/conda-forge/teams/help-python, ready for review!

h-vetinari:
Better to ping @conda-forge/staged-recipes for that 🙃 

(also, review bandwidth on this repo is quite a bottleneck, so it can take some time...)

h-vetinari (approved):
To accelerate the process to the extent of my abilities: LGTM

recipes/pydevtool/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+  run:
+    - doit
```
BastianZim, line 21:
```suggestion
    - doit >=0.36.0
```
Ref: https://github.com/pydoit/pydevtool/blob/5556488ca8eb40d1368ef7fc1e46c8a843ccff6d/setup.cfg#L28

BastianZim (changes requested):
Just one thing then this is good to go!

BastianZim:
> Better to ping https://github.com/orgs/conda-forge/teams/staged-recipes for that 🙃

Because of the new labelling, it's actually better now to ping the relevant team again as we then can see it in the overview. :)

h-vetinari:
> Because of the new labelling, it's actually better now to ping the relevant team again as we then can see it in the overview. :)

Ah, that's a recent change where I seem to have missed some details, apologies!

BastianZim:
Yes, no announcement has been made yet, so you couldn't have known. 😄

recipes/pydevtool/meta.yaml
BastianZim: Ahh just saw that that was introduced after the release. All good.

tupui:
Thank you everyone 🎉

PR #18774, opened 2022-04-27
Add kenlm recipe

cgr71ii:
Checklist
- [x] Build number is 0.


cgr71ii:
I want to be in the maintainer section

cgr71ii:
I know I shouldn't be using `git_url`, but there are not tarballs available (no [releases](https://github.com/kpu/kenlm/releases) nor [tags](https://github.com/kpu/kenlm/tags)). I don't know if this recipe will be accepted or not, but if there are other solutions in order to avoid using `git_url` and yet use the main source of the tool, which is the GitHub repo, I'd like to know and give it a try.

cgr71ii:

recipes/kenlm/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "kenlm" %}
+{% set commit = "0760f4c4df76f3286656e7232dc3ad6495248bc2" %}
+{% set version = commit[:7] %}
```
ocefpaf, line 3: Do we have a stable release? This is usually frowned upon. If not you can try the https://github.com/conda-forge/cfep/blob/main/cfep-05.md route.

cgr71ii: No, there is no stable release (https://github.com/conda-forge/staged-recipes/pull/18774#issuecomment-1110933919), and I agree it is not the best solution. The problem is that this tool doesn't seem to be releasing specific versions, but using master branch instead. In fact, there is no even version scheme defined in their `setup.py` (that's why I use the short commit as version).

The proposed method about making a pre-release package does not seem to be the best option here, since this recipe wouldn't be a pre-release itself, but I agree it might be the middle road of a solution. Do you think that pre-releasing is what I should do instead of letting the recipe as it is right now? I guess I would use `dev` label if you think this is the best option and the one I should follow.

I haven't been able to find any other recipe with the same use case I'm having here, what is strange to me since there are repositories (not a few if I'm not wrong) which follow the same methodology about not releasing and just using the code in the master branch.

cgr71ii: Perfect! Just one question, how can I apply dev label? In https://github.com/conda-forge/cfep/blob/main/cfep-05.md specifies

> To create a dev or rc package a PR can be issued into the dev or rc branch of the feedstock. This branch must change the recipe/conda_build_config.yaml to point to the <package_name>_dev or <package_name>_rc label.

but this is not a feedstock yet.

ocefpaf: It is the same procedure, just add the `onda_build_config.yaml ` with the dev channel info in the recipe directory. Something like this:

```yaml
channel_targets:
  - conda-forge kenml_dev
```

cgr71ii: Thank you! It is done! Is it ready the recipe to be accepted or it is necessary to do something else?

cgr71ii:
The recipe has `build: skip: True  # [not linux64]`, so I don't know why `win_64 win` CI tests failed

PR #18788, opened 2022-04-28
New py2dm recipe from PyPI w/ version specific edits

bedaro:
This is my first time doing a conda recipe; review and feedback welcome. Generated from py2dm PyPI w/ greyskull

There are some dependencies that only apply to older python versions, so noarch was removed and preprocessing selectors added. I think that's the right way to do it? @conda-forge-admin, please ping conda-forge/help-python

Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -11,24 +11,20 @@ source:
 
 build:
   script: {{ PYTHON }} -m pip install . -vv
-  number: 0
+  number: 1
 
 requirements:
   host:
-    - python >=3.6
+    - python
     - pip
   run:
-    - python >=3.6
-    - typing_extensions # [py<3.8]
-    - cached-property # [py<3.8]
+    - python
+    - typing_extensions   # [py<3.8]
+    - cached-property     # [py<3.8]
 
 test:
   imports:
     - py2dm
-  commands:
-    - pip check
-  requires:
-    - pip
 
 about:
   home: https://github.com/leonhard-s/Py2DM
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -33,6 +33,8 @@ requirements:
     - cached-property     # [py<38]
 
 test:
+  requires:
+    - pip
   commands:
     - pip check
   imports:
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -16,10 +16,6 @@ build:
 requirements:
   build:
     - {{ compiler('cxx') }}
-    # Without these the compilation of the C++ extension fails on Python 3.7
-    - typing-extensions   # [py<38]
-    - typing_extensions   # [py<38]
-    - cached-property     # [py<38]
   host:
     - python
     # It looks like host directives are being used during the run of
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -16,6 +16,9 @@ build:
 requirements:
   build:
     - {{ compiler('cxx') }}
+    # Without these the compilation of the C++ extension fails on Python 3.7
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
   host:
     - python
     - pip
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -33,6 +33,8 @@ requirements:
     - cached-property     # [py<38]
 
 test:
+  commands:
+    - pip check
   imports:
     - py2dm
 
--- a/recipes/py2dm/meta.yaml
+++ b/recipes/py2dm/meta.yaml
@@ -1,0 +1,41 @@
+{% set name = "py2dm" %}
+{% set version = "0.2.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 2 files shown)


bedaro:
OSX build error is:
`Needed DSO lib/libc++.1.dylib found in ['libcxx'] but ['libcxx'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)`

I don't know what's going on here and could use some help. Google didn't turn up anything useful.

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+requirements:
+  build:
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+    - setuptools
+  host:
+    - python
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
```
chrisburr, line 24: These are likely not needed in host:
```suggestion
```

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 1
+
+requirements:
+  build:
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+    - setuptools
```
chrisburr, line 20:
```suggestion
    - {{ compiler('c++') }}
```

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  host:
+    - python
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+    - pip
+  run:
+    - python
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+    - libcxx              # [osx]
```
chrisburr, line 30:
```suggestion
```

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+test:
+  imports:
+    - py2dm
```
chrisburr, line 34: It looks like this provides both a Python and compiled implementation. I presume you want the compiled one to work in conda-forge?

If so, is there a way of testing to make sure it was compiled rather than it silently falling back (like the Linux build currently is).

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/py2dm-{{ version }}.tar.gz
+  sha256: 447e3af1e5adfc7a898acd5e5744ab58f86d995cf615147629fec7b4bf605be7
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 1
```
chrisburr, line 14: This should only be incremented for "deployed" builds (i.e. PRs which have been merged into the main branch of a feedstock):
```suggestion
  number: 0
```

recipes/py2dm/meta.yaml
bedaro: Great idea, I'm writing a more thorough test case and will commit your other suggestions.

recipes/py2dm/meta.yaml
bedaro: Turns out they are needed, otherwise there's a build error while running setup.py egg_info. I don't know enough about python module packaging to understand why

recipes/py2dm/meta.yaml
bedaro: Needed to be "cxx" rather than "c++" but this is done

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    # Without these the compilation of the C++ extension fails on Python 3.7
+    - typing-extensions   # [py<38]
+    - typing_extensions   # [py<38]
+    - cached-property     # [py<38]
```
chrisburr, line 22: These are definitely not needed in build. If removing them causes an error then it's a sign of a different problem in the recipe.

recipes/py2dm/meta.yaml
chrisburr: Indeed, [upstream is importing itself during the build](https://github.com/leonhard-s/Py2DM/blob/b67fac13a3465323a218f52f62885ab38bb727c1/setup.py#L6) so this is fine as-is.

For reference: Ideally the `version = attr: py2dm.__version__` syntax would be used instead ([see here](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html)) as that doesn't require the package to be importable.

recipes/py2dm/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    # setup.py egg_info, and that requires all the dependencies of py2dm
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+    - pip
+  run:
+    - python
+    - typing-extensions   # [py<38]
+    - cached-property     # [py<38]
+
+test:
```
chrisburr, line 35:
```suggestion
test:
  commands:
    - pip check
```

recipes/py2dm/meta.yaml
bedaro: Okay, I will investigate further.

PR #18803, opened 2022-04-29
Adding MicroED tools

hattne:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/microed-tools/run_test.sh
+++ b/recipes/microed-tools/run_test.sh
@@ -1,22 +1,22 @@
 #! /bin/sh
 
-idoc2smv                                      \
-    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT \
-    "${SRC_DIR}/test/movie23.idoc"
+idoc2smv                                        \
+    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT   \
+    "${PREFIX}/share/microed-data/movie23.idoc"
 cat << EOF | md5sum -c -
 7018e62b3e52122ae3be875642f33bad  t_001.img
 EOF
 
-tiff2smv                                      \
-    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT \
-    "${SRC_DIR}/test/movie23_000.tif"
+tiff2smv                                           \
+    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT      \
+    "${PREFIX}/share/microed-data/movie23_000.tif"
 cat << EOF | md5sum -c -
 9f01e41a1e01ad59668450160c873214  t_001.img
 EOF
 
-tvips2smv                                     \
-    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT \
-    "${SRC_DIR}/test/movie23_000.tvips"
+tvips2smv                                            \
+    -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT        \
+    "${PREFIX}/share/microed-data/movie23_000.tvips"
 cat << EOF | md5sum -c -
 03232857ed7ce025016d94343719ac98  t_001.img
 a23a49265bbca8d6996dc0e06aeb808b  t_002.img
--- a/recipes/microed-tools/run_test.sh
+++ b/recipes/microed-tools/run_test.sh
@@ -1,20 +1,20 @@
 #! /bin/sh
 
-"${PREFIX}/bin/idoc2smv"                      \
+idoc2smv                                      \
     -f -g 5 -o "t_###.img" -r 0.09 -z EST5EDT \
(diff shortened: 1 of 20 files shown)


recipes/microed-tools/run_test.sh
```diff
@@ -0,0 +1,23 @@
+#! /bin/sh
+
+"${PREFIX}/bin/idoc2smv"                      \
```
carterbox, line 3:
```suggestion
"idoc2smv"                      \
```
Your tests shouldn't require you to provide the absolute path of the binaries to work correctly. Also, this is the wrong prefix for testing. The test environment prefix is `CONDA_PREFIX`; the test environment is separate from the build/install environment so we can test that your package is relocatable.

carterbox (changes requested):

recipes/microed-tools/run_test.sh
hattne: Fixed in f812a275efef3df391b7c2ccf74ba0f1067b7afe.

hattne:
> OPTIONAL: You can place pre-release builds into a special channel to keep them separate from non-pre-release-builds
> 
> https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=prerelease#pre-release-builds

OK, thanks!  I set `channel_targets` to `conda-forge microed-tools_dev` in `conda_build_config.yaml`. I suppose the rendered recipe should be merged into a `dev` branch once the pull request is resolved.  Can you do that?

hattne:
> I'm unsure about iconv linking: on Windows, we always depend on an external libiconv package, and that's fine. On macOS (and Linux, at least with glibc) it's built in. I would prefer to link against the built-in functions and eliminate the dependency on libiconv where possible, but I see the potential for underlinking if one of the dependencies explicitly links against libiconv. Any pointers on how to deal with this?
> 
> Otherwise, @conda-forge/help-c-cpp: Ready for review.

@carterbox Any opinions on the iconv issue?

recipes/microed-tools/run_test.sh
hattne: Also added a separate `noarch` package for the test data in 372d3010b0c5519f53703441ce8554058bcd0385 to avoid duplicating 9 MiB of identical data across all architectures.

hattne:
> > I suppose the rendered recipe should be merged into a dev branch once the pull request is resolved. Can you do that?
> 
> No. You do that. A feedstock is just a github repo, so you can create whatever branches you want. Any branch in the feedstock is published, so if you want to have parallel release types, then each should get its own branch. All experimentation/debugging of your recipes should occur in a personal fork via pull request so that they are not published.

OK, but I can't do that until there is an actual feedstock (i.e. until the PR is merged)?

> > @carterbox Any opinions on the iconv issue?
> 
> You can have different dependencies for each platform by using selectors. ~Is that what you're asking about?~ You already know this because you are using selectors, so I don't understand your question.

I suppose it's really more of a CMake problem: for instance, assume that I depend on iconv and libxml2, and libxml2 also depends on iconv. If my package is properly linked against a built-in iconv I shouldn't list  iconv as a dependency, no matter what libxml2 does.  But if libxml2 depends on, and is linked against, an external iconv, it will be installed in the build environment and CMake _may_ pick up the external iconv library rather than the built-in one. Then, I'd be underlinking.

Maybe I sort that out once/if it actually happens.

carterbox (changes requested):

hattne:
> > > > I suppose the rendered recipe should be merged into a dev branch once the pull request is resolved. Can you do that?
> > > 
> > > 
> > > No. You do that. A feedstock is just a github repo, so you can create whatever branches you want. Any branch in the feedstock is published, so if you want to have parallel release types, then each should get its own branch. All experimentation/debugging of your recipes should occur in a personal fork via pull request so that they are not published.
> > 
> > 
> > OK, but I can't do that until there is an actual feedstock (i.e. until the PR is merged)?
> 
> Yes. Exactly.
> 
> > > > @carterbox Any opinions on the iconv issue?
> > > 
> > > 
> > > You can have different dependencies for each platform by using selectors. ~Is that what you're asking about?~ You already know this because you are using selectors, so I don't understand your question.
> > 
> > 
> > I suppose it's really more of a CMake problem: for instance, assume that I depend on iconv and libxml2, and libxml2 also depends on iconv. If my package is properly linked against a built-in iconv I shouldn't list iconv as a dependency, no matter what libxml2 does. But if libxml2 depends on, and is linked against, an external iconv, it will be installed in the build environment and CMake _may_ pick up the external iconv library rather than the built-in one. Then, I'd be underlinking.
> > Maybe I sort that out once/if it actually happens.
> 
> You need to control how your program links with the build script. If you don't want your binary to link to something, have a build parameter that prevents it from happening.

OK, thanks!  All clear here, then.

carterbox (approved):

PR #18853, opened 2022-05-04
added recipe for semi-ate-spyder

nerohmot:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -12,7 +12,6 @@ source:
 
 build:
   number: 0
-  noarch: python
   script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - semi-ate-project-database
     - semi-ate-sammy
     - spyder >=5.3
-    - Jinja2<3.1 >=2.10
+    - Jinja2 <3.1,>=2.10
     - pyobjc
     
 test:
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -21,7 +21,7 @@ requirements:
   run:
     - numpy
     - pydantic
-    - python 
+    - python >=3.8
     - qtawesome
     - semi-ate-common
     - semi-ate-plugins
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -12,7 +12,7 @@ source:
 
 build:
   number: 0
-  skip: True  # [osx]
+#  skip: True  # [osx]
   noarch: python
 #  script: {{ PYTHON }} -m pip install . -vv
 
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -29,7 +29,6 @@ requirements:
     - semi-ate-project-database
     - semi-ate-sammy
     - spyder >=5.3
-    - numpydoc ==1.3.1
     
 test:
   imports:
@@ -39,6 +38,8 @@ test:
     - pip check  # [linux or win]
   requires:
     - pip
+    # This build of numpydoc has broken requirements
+    - numpydoc !=1.2.1=*_0
 
 about:
   home: https://pypi.org/project/semi-ate-spyder/
--- a/recipes/semi-ate-spyder/meta.yaml
+++ b/recipes/semi-ate-spyder/meta.yaml
@@ -31,6 +31,8 @@ requirements:
     - spyder >=5.3
     
 test:
+  requires:
+    - applaunchservices
   imports:
     - ate_spyder
     - ate_spyder.widgets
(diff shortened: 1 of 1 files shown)


nerohmot:
@chrisburr could you merge ?

chrisburr:
As I mentioned in https://github.com/conda-forge/staged-recipes/pull/18852#issuecomment-1118426485:

> Please [request reviews from the staged recipes team](https://github.com/conda-forge/staged-recipes#12-my-pull-request-passes-all-checks-but-hasnt-received-any-attention-how-do-i-call-attention-to-my-pr--what-is-the-customary-amount-of-time-to-wait) rather than pinging users directly.

recipes/semi-ate-spyder/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - pydantic
+    - python >=3.8
+    - qtawesome
+    - semi-ate-common
+    - semi-ate-plugins
+    - semi-ate-project-database
+    - semi-ate-sammy
+    - spyder >=5.3
+    - numpydoc ==1.3.1
+#    - applaunchservices >=0.1.7
```
goanpeca, line 33: Could you remove this comment?

recipes/semi-ate-spyder/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    
+    
+test:
+  imports:
+    - ate_spyder
+    - ate_spyder.widgets
+#  commands:
+#    - pip check
+#  requires:
+#    - pip
```
goanpeca, line 44: Could you remove these comments?

nerohmot:
@goanpeca
Hi Gonzalo,
if I add the `pip check` again then OSX starts failing 
image
with as a trace :
image
`applauncherservices` is introduced by `spyder`, and as `spyder` is ofcourse a dependency of this spyder-plugin ...
image
as a compromise, is it acceptable that I do the same as the `spyder` recipe ie : `pip check  # [linux or win]` ?
image

recipes/semi-ate-spyder/meta.yaml
nerohmot: ok

recipes/semi-ate-spyder/meta.yaml
nerohmot: :-)

nerohmot:
I also asked @ccordoba12 for a 'cleaner' solution from spyder side in spyder-ide/spyder#17835, but I would appreciate it if we can proceed as proposed, and not have to wait for the cleaner solution :innocent:

BastianZim:
Since it’s noarch it will only be built on Linux in the feedstock and we can ignore the OSX failure here.

recipes/semi-ate-spyder/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - numpy
+    - pydantic
+    - python >=3.8
+    - qtawesome
+    - semi-ate-common
+    - semi-ate-plugins
+    - semi-ate-project-database
+    - semi-ate-sammy
+    - spyder >=5.3
+    - numpydoc ==1.3.1
```
carterbox, line 32:
```suggestion
```
This is not listed in the requirements upstream.

carterbox (changes requested):

nerohmot:
@carterbox, you would think so, however when applying your suggestion I get :
image
With as win_64 trace: 
image
and as trace for linux_64 :
image
So before going down the `jinja2` rabbit hole, I first tried to force `numpydoc` to the latest version (hence the `=1.3.1`) and luckily for me conda found a solution for that!!! :sweat_smile:

recipes/semi-ate-spyder/meta.yaml
goanpeca: I meant delete the lines, just to make the recipe cleaner ;-) 

pip check does not work all the time, due to package name differences or other issues.

recipes/semi-ate-spyder/meta.yaml
goanpeca: I also meant to delete it if it was not going to be used

nerohmot:
@conda-forge/staged-recipes Can we merge now?

carterbox:
> numpydoc 1.2.1 has requirement Jinja2<3.1,>=2.10, but you have jinja2 3.1.2.

This error does not say that your package requires numpydoc. It says that numpydoc was installed in the test environment and the numpydoc=1.2.1 package is broken.

recipes/semi-ate-spyder/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    
+test:
+  imports:
+    - ate_spyder
+    - ate_spyder.widgets
+  commands:
+    - pip check  # [linux or win]
+  requires:
+    - pip
```
carterbox, line 41:
```suggestion
    
test:
  imports:
    - ate_spyder
    - ate_spyder.widgets
  commands:
    - pip check  # [linux or win]
  requires:
    - pip
    # This build of numpydoc has broken requirements
    - numpydoc !=1.2.1=*_0
```
Something like this, so the test environment doesn't choose the broken numpydoc build.

carterbox:
Also, you don't need to take screenshots. If you want to quote relevant text, just copy the log text directly.

carterbox (changes requested):

carterbox:
There's probably a similar issue with applaunchservices. That package incorrectly specified their dependencies / didn't do pip check in their feedstock, so the pip check error shows up here. Figure out which build of applaunchservices isn't broken and add that as a constraint to your test environment.

recipes/semi-ate-spyder/meta.yaml
nerohmot: I see

PR #18857, opened 2022-05-05
Add pds4_tools

LevN0:
Checklist
- [x] Build number is 0.


LevN0:

Windows build fails due to non-ASCII characters in test-data filenames, previously reported in https://github.com/conda/conda-build/issues/4437. However unlike in that issue, I am able to install a noarch build (made on Linux) just fine on Windows, possibly because it's only the test-data that has the issue and test-data for this package isn't included on install. Is this acceptable?

Separately, for future releases I need to fix tests upstream to use absolute imports as right now `source_files` requires the entire package rather than just the tests directory.

recipes/pds4_tools/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=2.6,>=3.3
```
carterbox, line 20:
```suggestion
    - python >=3.3
```
These requirements are redundant. Please fix in both places.

recipes/pds4_tools/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - pds4_tools
+  requires:
+    - pytest
+    - pip
+    - xorg-libx11
+  source_files:
+    - pds4_tools
+  commands:
+    - pip check
+    - pytest pds4_tools/tests/test_core.py
```
carterbox, line 37:
```suggestion
```
This is a noarch python package. Full correctness tests are the responsibility of upstream and not needed here. Tests in the conda recipe should show that the packaging/dependencies/compilation was successful. Import tests and pip check are sufficient because there is no compilation.

carterbox (changes requested):
Agreed. The windows failure can be ignored because it is probably path related.

recipes/pds4_tools/meta.yaml
LevN0: What is the suggestion `python >=3.3` as opposed to `python >=2.6`?

LevN0:

recipes/pds4_tools/meta.yaml
carterbox: That's a good question. My reasons are that we don't publish new packages for python=2.7 on conda-forge anymore, so we have no way of testing if this package is actually compatible with python=2.7? I'm fine with accepting the minimum version as 2.6.

recipes/pds4_tools/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+requirements:
+  host:
+    - pip
+    - python >=2.6
+  run:
+    - numpy
+    - python >=2.6
+    - tk
+    - matplotlib-base
```
carterbox, line 25: Is matplotlib optional or not? If there's a good use case for not generating GUI elements then you may not want to require all users to install them.

The [setup.py](https://github.com/Small-Bodies-Node/pds4_tools/blob/50df4401897131fcccd37f5758d31a361ff34b37/setup.py#L90) lists matplotlib and tk/tkinter as optional.

carterbox (changes requested):

recipes/pds4_tools/meta.yaml
LevN0: For users not interested in the GUI aspect it's optional. conda doesn't support marking optional dependencies directly as far as I can tell, and some people recommend then just marking those as required given the broad ecosystem support. Personally I am unsure which abroad is best.

LevN0: Thinking on this more, since other tools have used `pds4_tools` as a library for its read functionality, it's probably best to keep the optional dependencies for visualization as optional. 

However tk is not optional despite upstream claims. That's an oversight I need to fix upstream, it should be optional. Never caught it since tk generally ships with python, only incidentally caught it here due to the need for libx11. Thus I've removed matplotlib-base but am keeping tk for now.

PR #18884, opened 2022-05-09
pyparam initial build commit

jayramr:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/pyparam/meta.yaml
+++ b/recipes/pyparam/meta.yaml
@@ -1,0 +1,34 @@
+{% set name = "pyparam" %}
+{% set version = "0.4.17" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 17be473262414f86dcf8fa519f0ea2a52eb18b0c9d73d7df0ea647b983dee3ff 
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - pip
+  run:
+    - python
+
+test:
+  commands:
+    - python -m pyparam
+  requires:
+    - pip
+
+about:
+  home: https://github.com/pwwang/pyparam
+  summary: Powerful parameter processing, command line argument parser
+  license: MIT


recipes/pyparam/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - pip
```
jayramr, line 21: - poetry

recipes/pyparam/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - poetry
+  run:
+    - python >=3.7
+
+test:
+  commands:
+    - python -m pyparam
+  requires:
+    - pip
```
chrisburr, line 29:
```suggestion
    - pip check
  requires:
    - pip
  imports:
    - pyparam
```

recipes/pyparam/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+about:
+  home: https://github.com/pwwang/pyparam
+  summary: Powerful parameter processing, command line argument parser
+  license: MIT
+  license_file: LICENSE
+  
+extra:
+  recipe-maintainers:
+    - pwwang
```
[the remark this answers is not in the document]
jayramr: sorry I'm new to this package buidiling. 
I though in the maintainer section i need to put the actual owner who develop the package from pypi.org. 
Let me correct this with mine

jayramr:
This is included in v0.5.1 , Updated the meta.yaml

ndevenish:
The builds are still failing - the SHA256 also needs to be updated, because the tarball you are downloading has changed.

ndevenish:
Also, I'd suggest accepting the testing code change suggestion above, and unless I'm mistaken the explicit import step in that suggestion makes the `python -m pyparam` command redundant.

jayramr:
done updated the sha256 checksum.  And i submitted the imports test suggestion. 
Thanks for your support

ndevenish:
So, the problem now appears to be that you've put the dependencies into the `host:`, not the `run:` section. The difference is described in https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#requirements-section but to be honest it can sometimes be hard to understand the differences. 

Thankfully, since this is python-only it's a little easier; Essentially, things in `host` are required when _building_ the package (so, when running `pip install`), and things in `run` are required when _using_ the package. The tests don't run in the "build" stage, the tests run  in a new, separate environment, so that you are testing the package in exactly the kind of environment a user would test them (this helps avoid errors in differences between build and deployed environments). If you scroll up on the failed output a bit you can see mamba installing an environment, without diot.

So, move diot to the `run:` requirements.

jayramr:
nothing provides requested simpleconf -> not available in conda which results the build process failed.

recipes/pyparam/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - python >=3.7
+    - pip
+    - poetry
+    - simpleconf
+  run:
+    - python >=3.7
+    - diot
+    - rich
+    - poetry
+    - simpleconf
```
ndevenish, line 28:
```suggestion
    - python-simpleconf
```
Now your upstream [feedstock](https://github.com/conda-forge/python-simpleconf-feedstock) is created, you should point to that.

And, I don't think poetry is a runtime dependency?

PR #18887, opened 2022-05-10
Add MNE-ICALabel

hoechenberger:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mne-icalabel/meta.yaml
+++ b/recipes/mne-icalabel/meta.yaml
@@ -1,0 +1,51 @@
+{% set name = "mne-icalabel" %}
+{% set version = "0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 6d8b9aced23ed936791d47322b5c998a14e5340fa7f7709b050fa776b1c0673e
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+    - setuptools >=46.4.0
+    - wheel
+  run:
+    - python
+    - numpy
+    - scipy
+    - mne-base >=1.0
+    - pytorch
+
+test:
+  imports:
+    - mne_icalabel
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://mne.tools/mne-icalabel/stable/
+  summary: mne-icalabel is a Python package for labeling independent components that stem from an Independent Component Analysis (ICA).
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE
+  dev_url: https://github.com/mne-tools/mne-icalabel
+
+extra:
+  recipe-maintainers:
+    - hoechenberger
+    - larsoner
+    - adam2392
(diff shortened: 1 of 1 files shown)


hoechenberger:

recipes/mne-icalabel/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  imports:
+    - mne_icalabel
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://mne.tools/mne-icalabel/stable/
+  summary: mne-icalabel is a Python package for labeling independent components that stem from an Independent Component Analysis (ICA).
```
hoechenberger, line 40:
```suggestion
  summary: MNE-ICALabel is a Python package for labeling independent components that stem from an Independent Component Analysis (ICA).
```

hoechenberger:
@larsoner @mscheltienne @adam2392 Could you please confirm that you'd like to co-maintain this package? Also feel free to add additional co-maintainers. But we'll need their written confirmation here, too :)

larsoner:
Yes I'm happy to be a maintainer

hoechenberger:

hoechenberger:

hoechenberger:
Unfortunately we cannot depend on `pytorch` on Windows due to https://github.com/conda-forge/pytorch-cpu-feedstock/issues/32
So for Windows, we'll have to instruct users to install `pytorch` manually from the `pytorch` channel

adam2392:
I confirm I can be a maintainer

hoechenberger:
@conda-forge/staged-recipes, this one is ready for review. Thank you!

recipes/mne-icalabel/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - pip
+    - setuptools >=46.4.0
+    - wheel
+  run:
+    - python
+    - importlib-resources  # [py<39]
+    - mne-base >=1.0
+    - numpy >=1.16.0
+    - scipy >=1.2.0
+    - pytorch  # [not win]  https://github.com/conda-forge/pytorch-cpu-feedstock/issues/32
```
ocefpaf, line 29: Does this package works without pytorch on Windows? If not we are producing a broken package on Windows here and it would be better to make it noarch and let Windows fail the installation. (When pytorch is available this package would be ready on Windows already as noarch.)

hoechenberger: Wait, I cannot make it `noarch` because we have:

```
importlib-resources  # [py<39]
```

hoechenberger: I've now made `importlib-resources` a dependency on all platforms even though we actually only need it for `py<39`, but this allows us to make the package `noarch`

[e5b8000](https://github.com/conda-forge/staged-recipes/pull/18887/commits/e5b80008fe4c8f6392fe4b9244c03301a6d9812b)

hoechenberger:
@ocefpaf This is now passing on all platforms except for Windows, as expected.

I think this is good to merge, then?

The built `noarch` package will not be installable on Windows if only using the `conda-forge` channel, which is what we want.

recipes/mne-icalabel/meta.yaml
ocefpaf: It is fine to have `importlib-resources` in all platforms. In theory, of things work as they are supposed to, it will default to the standard lib in places were that is available.

hoechenberger:
Thanks, @ocefpaf!

PR #18892, opened 2022-05-10
added libid3tag

notestaff:
Checklist
- [x] Build number is 0.

### DIFF
--- a/.travis_scripts/create_feedstocks
+++ b/.travis_scripts/create_feedstocks
@@ -22,7 +22,7 @@ curl -L https://github.com/conda-forge/miniforge/releases/latest/download/Mambaf
 bash ~/Mambaforge.sh -b -p ~/Mambaforge
 touch ~/Mambaforge/conda-meta/pinned
 (
-    source ~/Mambaforge/bin/activate root
+    source ~/Mambaforge/bin/activate base
 
     # Configure conda.
     echo ""
@@ -32,7 +32,7 @@ touch ~/Mambaforge/conda-meta/pinned
     conda config --set add_pip_as_python_dependency false
 
     unset conda
-    mamba update -n root --yes --quiet conda mamba
+    mamba update -n base --yes --quiet "conda<4.13" mamba
 )
 source ~/Mambaforge/bin/activate root
 
--- a/.travis_scripts/create_feedstocks.py
+++ b/.travis_scripts/create_feedstocks.py
@@ -31,7 +31,7 @@ from ruamel.yaml import YAML
 # Enable DEBUG to run the diagnostics, without actually creating new feedstocks.
 DEBUG = False
 
-REPO_SKIP_LIST = ["core", "bot", "staged-recipes", "arm-arch", "systems"]
+REPO_SKIP_LIST = ["core", "bot", "staged-recipes", "arm-arch", "systems", "ctx"]
 
 recipe_directory_name = 'recipes'
 
@@ -225,6 +225,9 @@ if __name__ == '__main__':
         for recipe_dir, name in list_recipes():
             if name.lower() in REPO_SKIP_LIST:
                 continue
+            if name.lower() == "ctx":
+                sys.exit(1)
+
             feedstock_dir = os.path.join(feedstocks_dir, name + '-feedstock')
             print('Making feedstock for {}'.format(name))
             try:


notestaff:

recipes/libid3tag/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/tenacityteam/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  skip: True  # [win]
+  skip: True  # [osx]
```
carterbox, line 16:
```suggestion
  skip: True  # [osx]
  run_exports:
    # TODO: Relax pinning when API/ABI reaches 1.0 
    - {{ pin_subpackage('libid3tag', exact=True) }}
  ignore_run_exports_from:
    - {{ compiler('cxx') }}
```
[Link inspector reports](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=504385&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1265) overdepending on the cxx standard library. Need to either ignore exports from the compiler or remove the compiler if it is not needed.

recipes/libid3tag/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make
+    - pkg-config
+  host:
+    - gperf
+    - zlib
+  run:
+    - zlib
```
carterbox, line 29:
```suggestion
```
zlib has a run_export on libzlib. If your tests need the zlib headers, then add zlib to the test requirements.

recipes/libid3tag/run_test.sh
```diff
@@ -0,0 +1,21 @@
...
+   struct id3_frame *frame = id3_tag_findframe(tag, ID3_FRAME_TITLE, 0);
+   id3_file_close(fp);
+   return 0;
+}
+EOF
+
+
+pkg_config_cflags=$(pkg-config --cflags --libs id3tag)
+${CC} "test_libid3tag.c" ${pkg_config_cflags} "-o" "test_libid3tag"
+"./test_libid3tag"
```
carterbox, line 21: Please also add [some existence tests ](https://conda-forge.org/docs/maintainer/adding_pkgs.html#simple-existence-tests)for one of each of the file types you are shipping. i.e.

- [ ] header
- [ ] shared library
- [ ] cmake config
- [ ] pkgconfig file

carterbox (changes requested):

recipes/libid3tag/meta.yaml
notestaff: Tried that, things break, see above: https://github.com/conda-forge/staged-recipes/pull/18892/commits/7ac01e61e513a686da82fb0b34b46e23e831320f
Test does not use zlib headers.   I too was surprised why the run_export isn't working, but there is nothing zlib-specific in the tests, so wouldn't moving the requirement to the tests mean things won't work for users?

carterbox:
You need to use @ mention not the url of the team page! 😆 Also, I don't have time this week to do reviews. I will come back eventually if someone else from the team doesn't before me.

@conda-forge/help-c-cpp, ready for another review.

recipes/libid3tag/meta.yaml
carterbox: Correct, that's why I didn't suggest adding zlib to the test environment. I can't see the build logs for your old build now, but libzlib only ships the so versioned libraries. i.e. libz.so.1 If your linker is looking for the unversioned libz, that will cause a runtime error.

I don't know specifically why we aren't shipping the unversioned libz, so you may need to patch the build.

carterbox: The link inspector from the latest build reports:

```
   INFO (libid3tag,lib/libid3tag.so.0.16.2): Needed DSO lib/libz.so.1 found in conda-forge::libzlib-1.2.11-h166bdaf_1014
```

It looks like you are linking to a versioned libz, and this libz will be included in the run requirements because libzlib is exported from libz. My guess is that your test *does* require headers or pkgconfig from the libz package.

carterbox: @notestaff, are you a part of tenacityteam?

I think the run dependency on the zlib headers is caused by this [PUBLIC link declaration](https://github.com/tenacityteam/libid3tag/blob/3df1882b76c2f44a274af1af33497125c95679b9/CMakeLists.txt#L69). PUBLIC (vs PRIVATE) implies that linkers to libid3tag need to know about the zlib API (that the zlib API is used in the libid3tag headers). If this is the case, then zlib is required for building against libid3tag. If not, then this is bug that needs to be patched. I am not sure which is the correct solution, so discuss with upstream.

recipes/libid3tag/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make
+    - pkg-config
+  host:
+    - gperf
+    - zlib
+
```
carterbox, line 33:
```suggestion
  host:
    - gperf
    - zlib
  run:
    # 31 May 2022
    # Upstream CMakeLists declares that zlib headers are required to build against libid3tag.
    # https://github.com/tenacityteam/libid3tag/blob/3df1882b76c2f44a274af1af33497125c95679b9/CMakeLists.txt#L69
    - zlib
```

carterbox (changes requested):

recipes/libid3tag/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+test:
+  files:
+    - test.mp3
+  requires:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - pkg-config
+    - zlib
```
carterbox, line 41:
```suggestion
```

carterbox (changes requested):

PR #18896, opened 2022-05-10
Adding elapid

earth-chris:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/elapid/meta.yaml
+++ b/recipes/elapid/meta.yaml
@@ -21,7 +21,7 @@ requirements:
     - python
     - descartes >=1.1.0
     - geopandas >=0.7.0
-    - glmnet  # [py<39 and not win]
+    - glmnet  # [not win] windows build unavailable; elapid will use sklearn features instead
     - numpy >=1.18
     - pandas >=1.0.3
     - rasterio >=1.2.1
--- a/recipes/elapid/meta.yaml
+++ b/recipes/elapid/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
+  noarch: python
 
 requirements:
   host:
@@ -19,16 +20,15 @@ requirements:
     - pip
   run:
     - python
-    - descartes >=1.1.0
-    - geopandas >=0.7.0
-    - glmnet >=2.0,<2.3  # [not win] windows build unavailable; elapid will use sklearn features instead
+    - descartes >=1.1.0,<2
+    - geopandas >=0.7.0,<2
     - numpy >=1.18
-    - pandas >=1.0.3
-    - rasterio >=1.2.1
-    - rtree >=0.9
-    - scikit-learn >=0.22.2
-    - shapely >=1.7.0
-    - tqdm >=4.60.0
+    - pandas >=1.0.3,<2
+    - rasterio >=1.2.1,<2
+    - rtree >=0.9,<2
+    - scikit-learn >=0.22.2,<2
+    - shapely >=1.7.0,<2
+    - tqdm >=4.60.0,<5
 
 test:
   imports:
--- a/recipes/elapid/meta.yaml
+++ b/recipes/elapid/meta.yaml
@@ -10,7 +10,6 @@ source:
   sha256: a713f02c5fcaec6cfe584c03cc1faed57dcb837b96d9bd8b21c479cd591164dd
 
 build:
-  skip: true  # [py<37]
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
@@ -41,6 +40,7 @@ test:
 
 about:
   home: https://elapid.org
+  dev_url: https://github.com/earth-chris/elapid
   summary: Species distribution modeling support tools
   license: MIT
   license_file: LICENSE
--- a/recipes/elapid/meta.yaml
+++ b/recipes/elapid/meta.yaml
@@ -16,10 +16,10 @@ build:
 
 requirements:
   host:
-    - python>=3.7
+    - python >=3.7
     - pip
   run:
-    - python>=3.7
+    - python >=3.7
     - descartes >=1.1.0,<2
     - geopandas >=0.7.0,<2
     - numpy >=1.18
--- a/recipes/elapid/meta.yaml
+++ b/recipes/elapid/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "elapid" %}
-{% set version = "0.3.6" %}
+{% set version = "0.3.12" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


earth-chris:

recipes/elapid/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - descartes >=1.1.0
+    - geopandas >=0.7.0
+    - glmnet  # [py<39 and not win]
```
saraedum, line 25: This seems to be the only thing here that makes this not a `noarch: python` recipe. Could you explain what this is about exactly? I.e., how come there is a dependency that is only needed when Python is old? (Might also be nice to add a comment to the recipe so contributors to the recipe in the future do not stumble upon this.

recipes/elapid/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  imports:
+    - elapid
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://elapid.org
+  summary: Species distribution modeling support tools
```
saraedum, line 44:
```suggestion
  dev_url: https://github.com/earth-chris/elapid
  summary: Species distribution modeling support tools
```

recipes/elapid/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+build:
+  skip: true  # [py<37]
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
```
saraedum, line 21: All these pins have no upper bound. I see that you are doing the same thing in the upstream package. You could consider being a bit more conservative here and bound things if you know/find out how their versioning works. Say, likely if there's ever a descartes 2.0.0 (I have no idea what that package does) then likely elapid 0.3.13 won't work with that version of descartes. So if people somehow try to install an old version of elapid, they'll get a new version of descartes (and everything else) and end up with a non-functional environment. On the other hand, if descartes 2.0.0 comes out and elapid works fine with it, then you can just relax the bound in a `build.number: 1`.
Anyway, just a suggestion.

saraedum (changes requested):
Looks alright to me. Please have a look at the comments above.

recipes/elapid/meta.yaml
earth-chris: it's mostly because the `glmnet` package doesn't include a windows build. I was under the impression from [this issue](https://github.com/civisanalytics/python-glmnet/issues/76) that python 3.9 and above was also unavailable. After checking the [conda-forge page](https://anaconda.org/conda-forge/glmnet) that doesn't appear to be the case, so it looks like this could just be `[not win]`.

this is a statistical modeling package that has some nice-to-have features that are unavailable in `sklearn`. `elapid` works around this by using `glmnet` if it's installed, `sklearn` otherwise. I could technically drop the requirement entirely, but imho it's worth having those nice-to-haves.

recipes/elapid/meta.yaml
earth-chris: this is my first time preparing a distribution like this so I'm not familiar with best practices here. I was under the impression that I don't need to solve problems I don't yet have. That is, I don't need to anticipate breaking changes when nothing is yet broken. Is this naive?

I'm not sure I understand what setting the build number to 1 does. Does this mean that it's safer to restrict the major versions of most packages until they're released and tested to functionally work, at which point i'd change the requirement in the recipe to lift that restriction?

earth-chris:
hi @saraedum, thanks so much for your review! I've accepted the edits you recommended and relaxed the python version constraints on `glmnet` (but kept the windows constraint; see [package page](https://anaconda.org/conda-forge/glmnet) for reference if needed). I just have one outstanding question clarifying your comments on major versions.

recipes/elapid/meta.yaml
earth-chris: reviewing the dependencies, it looks like there's only 1 or 2 packages that I might expect to introduce breaking changes, which are `geopandas` and `sklearn`.

`geopandas` seems unlikely to change with a 1.0 release, as they develop separate software systems that make fundamental changes to the api/dependencies, like [pyogrio](https://github.com/geopandas/pyogrio). I suspect the 1.0 release will still be backwards compatible but add features and increase stability.

`sklearn` has changed their API and deprecated features somewhat regularly, but have done so less frequently since the 1.0 release. I'm not sure how to best mitigate this - maybe bound it to <2.0?

The remaining dependencies are fairly stable (`numpy`, `rasterio`, `rtree`), are no longer in active development (`glmnet`), or breaking dependencies are handled upstream by `geopandas` (`descartes`, `shapely`).

So it doesn't seem super critical to pin upper bounds to me until we run into an issue, but I'm also willing to heed advice.

recipes/elapid/meta.yaml
amirhessam88: @earth-chris FYI I did follow up with @civisanalytics to be able to maintain `glmnet` project; No one responded tho

earth-chris: yea, i saw that, thanks @amirhessam88. seems unlikely that they're gonna resume maintenance on the package, unfortunately, so future updates would probably have to come from someone who forks it and pushes a new package.

recipes/elapid/meta.yaml
earth-chris: I've at least pinned the upper version dependency of `glmnet` in case they resume development on the project, at which point I might expect some updates to their API.

Also, re-pinging @saraedum to kindly ask that they update their review.

earth-chris:

recipes/elapid/meta.yaml
saraedum: > this is my first time preparing a distribution like this so I'm not familiar with best practices here. I was under the impression that I don't need to solve problems I don't yet have. That is, I don't need to anticipate breaking changes when nothing is yet broken. Is this naive?

As I tried to outline above, having no upper bounds means that you might have to patch existing builds later. Say `descartes 2` breaks elapid, then somebody doing a `conda create -n something elapid=0.3.14` would get the latest version of `descartes` which won't work. So you need to patch the existing builds (which is somewhat cumbersome) to restrict `descartes <2` everywhere after the fact.

> I'm not sure I understand what setting the build number to 1 does. Does this mean that it's safer to restrict the major versions of most packages until they're released and tested to functionally work, at which point i'd change the requirement in the recipe to lift that restriction?

Yes.

saraedum: I made a more conservative suggestion below.

Feel free to ignore it if you don't like it. But it's easier to restrict things now than to react to breakage later imho.

recipes/elapid/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - python
+    - descartes >=1.1.0
+    - geopandas >=0.7.0
+    - glmnet >=2.0,<2.3  # [not win] windows build unavailable; elapid will use sklearn features instead
```
saraedum, line 24: As a suggestion for the next release of elapid, you could consider just issuing a warning if glmnet is not available, remove this dependency and set this recipe to `noarch: python`. The upside is that it will work without any further changes on apple silicon, aarch64, ….

saraedum:
Sorry for the delayed reaction here. This looks alright to me. Let me know if you want to make any further changes.

recipes/elapid/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - numpy >=1.18
+    - pandas >=1.0.3
+    - rasterio >=1.2.1
+    - rtree >=0.9
+    - scikit-learn >=0.22.2
+    - shapely >=1.7.0
+    - tqdm >=4.60.0
```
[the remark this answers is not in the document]
earth-chris: ok, I've added upper bounds to some packages. I didn't accept these changes as-is because they are unnecessarily strict (elapid has been tested and works with `sklearn` 1.1.2, with `geopandas` 0.11.1, etc.).

recipes/elapid/meta.yaml
earth-chris: I've removed the `glmnet` requirement in the recipe and set it to `noarch: python`. It already has a try/except for importing glmnet so it should be fine as-is.

earth-chris:
alright @saraedum I think I've addressed your requested changes.

PR #18901, opened 2022-05-11
add atmosphere-virtual-lab

srepmub:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/atmosphere-virtual-lab/conda_build_config.yaml
+++ b/recipes/atmosphere-virtual-lab/conda_build_config.yaml
@@ -1,8 +1,0 @@
-cudnn:                                    # [linux64]
-  - undefined                             # [linux64]
-cuda_compiler_version:                    # [linux64]
-  - None                                  # [linux64]
-docker_image:                             # [linux64]
-  - condaforge/linux-anvil-cos7-x86_64    # [linux64]
-cdt_name:                                 # [linux64]
-  - cos7                                  # [linux64]
--- a/recipes/atmosphere-virtual-lab/conda_build_config.yaml
+++ b/recipes/atmosphere-virtual-lab/conda_build_config.yaml
@@ -1,0 +1,8 @@
+cudnn:                                    # [linux64]
+  - undefined                             # [linux64]
+cuda_compiler_version:                    # [linux64]
+  - None                                  # [linux64]
+docker_image:                             # [linux64]
+  - condaforge/linux-anvil-cos7-x86_64    # [linux64]
+cdt_name:                                 # [linux64]
+  - cos7                                  # [linux64]
--- a/recipes/atmosphere-virtual-lab/conda-forge.yml
+++ b/recipes/atmosphere-virtual-lab/conda-forge.yml
@@ -1,0 +1,2 @@
+os_version:
+  linux_64: cos7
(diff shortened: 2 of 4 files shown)


srepmub:
fails because the test cannot import https://github.com/conda-forge/staged-recipes/pull/18900. should work once that is merged, and a dependency on it added here.

svniemeijer:
I confirm to be a maintainer

srepmub:
@conda-forge/staged-recipes please review

recipes/atmosphere-virtual-lab/conda_build_config.yaml
```diff
@@ -0,0 +1,8 @@
+cudnn:                                    # [linux64]
+  - undefined                             # [linux64]
+cuda_compiler_version:                    # [linux64]
+  - None                                  # [linux64]
+docker_image:                             # [linux64]
+  - condaforge/linux-anvil-cos7-x86_64    # [linux64]
+cdt_name:                                 # [linux64]
+  - cos7                                  # [linux64]
```
ocefpaf, line 8: I'm not sure you need this file. Can you elaborate a bit why you added it? The library seems to be pure Python (in fact it is already noarch).

srepmub: thank you. the vtk dependency causes GLIBC errors, and this is the only way I could find so that doesn't happen. for another recipe I worked on recently, we added a sysroot dependency in 'build', but since this is python there's no building..

ocefpaf: You can add a build section, on top of host, with sysroot. I believe that is preferred. Can you try it?

srepmub: @ocefpaf I'm pretty sure that was the first thing I tried, and it didn't work.

@carterbox, thanks! that doesn't look very different to me, but I will see if that does the trick.

srepmub: seems to have worked ^^

srepmub:
https://github.com/orgs/conda-forge/teams/staged-recipes please review changes

recipes/atmosphere-virtual-lab/conda_build_config.yaml
ocefpaf: What is puzzling me here is that this module is pure python and we should not need a compiler or to specify cos7 to make it work. Maybe we have a problem with the vtk package. Let me test a few things here before we merge this one.

svniemeijer: Would it be possible to at least go ahead with the creation of the feedstock and then solve this specific cos7 issue as a separate issue? We want to use the packages for a demonstration at a conference next week, so we are keen to get this published as soon as possible.

ocefpaf: Give me a few hours. I'm investigating this now. You'll make the deadline conf :-)

ocefpaf: setup.py did not give me any clue but looking at the logs this does have a compiled extension, so this is correct. Let's merge!

PR #18919, opened 2022-05-12
Add versioningit

jwodder:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/versioningit/meta.yaml
+++ b/recipes/versioningit/meta.yaml
@@ -1,0 +1,68 @@
+{% set name = "versioningit" %}
+{% set version = "1.1.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 3b7565a2db92f1206b5b833a72cc6936c8a74d65e94224f0a9eb54799d1561d9
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  entry_points:
+    - versioningit = versioningit.__main__:main
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+    - setuptools
+  run:
+    - python >=3.6
+    - dataclasses
+    - importlib-metadata >=3.6
+    - packaging
+    - setuptools
+    - tomli >=1.2,<3.0
+
+test:
+  requires:
+    - build >=0.7.0
+    - git >=2.32
+    - mercurial >=5.2
+    - setuptools >=42
+    - pip
+    - pydantic >=1.8,<2.0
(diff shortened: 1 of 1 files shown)


jwodder:
@conda-forge/help-python Ready for review.

recipes/versioningit/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - pytest-cov >=3.0
+    - pytest-mock >=3.0
+    - wheel
+  source_files:
+    - test
+    - tox.ini
+  imports:
+    - versioningit
+  commands:
+    - python -m pytest -vv test
```
carterbox, line 50:
```suggestion
    - pip check
    - versioningit --help
    - python -m pytest -vv test
```
You *can* run your entire test suite, but we're more interested in testing whether the packaging was correct. For a pure-python package this means checking that entry-points and dependencies were installed correctly.

recipes/versioningit/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - build >=0.7.0
+    - git >=2.32
+    - mercurial >=5.2
+    - setuptools >=42
+    - pip
+    - pydantic >=1.8,<2.0
+    - pytest >=7.0
+    - pytest-cov >=3.0
+    - pytest-mock >=3.0
+    - wheel
```
carterbox, line 43: The requirements are different than upstream.

https://github.com/jwodder/versioningit/blob/3390f75620874afc3260f7c96f5f69434e791a80/tox.ini#L9

carterbox (changes requested):

recipes/versioningit/meta.yaml
carterbox: Maybe I'm just not familiar with tox syntax? What does "~=" mean? Here ">=" maps to all of "~=" and ">=" and ">=,<" in the tox.ini.

jwodder: `~= X.Y` is [PEP 440](https://peps.python.org/pep-0440/) syntax for `>= X.Y, < (X+1).0`.  Does the lack of an upper bound really matter here?

carterbox: Cool! Thanks for teaching me something about PEP440 syntax! 👍🏼 

I believe the equivalent conda recipe syntax would be:
```yaml
{{ pin_compatible('build', lower_bound='0.7.0' ) }}
```

Since you're the developer of the package, you tell me if the upper bound is important or not.

carterbox (approved):

recipes/versioningit/meta.yaml
jwodder: The upper bound is only important while developing.  When testing a full release (which is the only time when the Conda recipe would be run), having it in the Conda recipe would just cause more harm than good.

jwodder:
@conda-forge/help-python What do I still need to do to get this merged?

PR #18925, opened 2022-05-12
Initial gekko package for conda-forge

APMonitor:
Thank you for reviewing this initial contribution @conda-forge/help-python

### DIFF
--- a/recipes/gekko/LICENSE
+++ b/recipes/gekko/LICENSE
@@ -1,0 +1,25 @@
+Copyright 2022 Advanced Process Solutions, LLC. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification, are
+permitted provided that the following conditions are met:
+
+   1. Redistributions of source code must retain the above copyright notice, this list of
+      conditions and the following disclaimer.
+
+   2. Redistributions in binary form must reproduce the above copyright notice, this list
+      of conditions and the following disclaimer in the documentation and/or other materials
+      provided with the distribution.
+
+THIS SOFTWARE IS PROVIDED BY APMONITOR "AS IS" AND ANY EXPRESS OR IMPLIED
+WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL APMONITOR OR
+CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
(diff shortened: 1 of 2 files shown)


calvinwy:
@conda-forge/staged-recipes @conda-forge/core

Could someone from the conda-forge team have a look into the problem @APMonitor is experiencing above?  I am waiting for the Gekko package to be included in conda-forge, so that I can also include my package CDST on conda-forge which depends on the Gekko package.

carterbox (changes requested):
Please do not ignore the checklist included in our pull request template. I have re-added the checklist for you.

Unfortunately, this package (in its current state) cannot be accepted because it contains a precompiled binary (apm). We consider this "vending another package". Additionally, all binaries need to be compiled using the conda-forge  toolchain.

To continue with this recipe you will need to do one of the following:

- Find this apm package on conda-forge already and add it as a dependency
- Find the source code of apm and add a separate recipe for it to the PR and add it as a dependency for gekko
- Make a strong case to the core team for why apm cannot be built using the conda-forge toolchain

APMonitor:
> Unfortunately, this package (in its current state) cannot be accepted because it contains a precompiled binary (apm). We consider this "vending another package". Additionally, all binaries need to be compiled using the conda-forge toolchain.
> 
> To continue with this recipe you will need to do one of the following:
> 
>     * Find this apm package on conda-forge already and add it as a dependency

Unfortunately, the `apm` package isn't on `conda-forge`.

>     * Make a strong case to the core team for why apm cannot be built using the conda-forge toolchain

The `apm` binary is from about 70,000 lines of F90 and C++ code that form that core of the APMonitor Optimization Suite. There are optional 4 binaries that allow local solves (Windows: apm.exe, Linux: apm, MacOS: apm_mac, ARM Linux: apm_arm). Unfortunately, the mixed-code does not compile with `gcc` compilers so forming a tool chain to compile on each platform seems unlikely. One option is to just remove the local binaries in the `bin` folder for the `conda-forge` distribution. By default, `gekko` uses cloud based computing to solve the optimization and simulation problems and returns the solution to the python program. Is that a potential route forward?

It appears that someone (not me) already made `gekko` available for conda install with `conda install -c bjrn gekko` as seen [here](https://anaconda.org/bjrn/gekko). However, this is 4 minor releases behind the [gekko version on pypi](https://pypi.org/project/gekko/). Maybe I should just focus on making the package available through `conda` instead?

carterbox:
> By default, gekko uses cloud based computing to solve the optimization and simulation problems and returns the solution to the python program. Is that a potential route forward?

Yes, you could patch this distribution so the binaries are not included. Making requests to a server is allowed.

> It appears that someone (not me) already made gekko available for conda install with conda install -c bjrn gekko as seen [here](https://anaconda.org/bjrn/gekko). [...] Maybe I should just focus on making the package available through conda instead?

I'm not sure what you're asking. conda-forge is a conda channel. bjrn is a conda channel. The bjrn and conda-forge channels are not affiliated in any way. The conda-forge policies regarding repackaging and ABI compatibility are independent of other channels.

PR #18942, opened 2022-05-14
Adding fABBA_package

chenxinye:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/fABBA/meta.yaml
+++ b/recipes/fABBA/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "fABBA" %}
+{% set version = "0.9.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 6401f15d91523318566426debed4553f73d355e77ced108eec90d623ab18a6e8
+
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+    
+requirements:
+  build:
+    - python >=3.6
+    - pip
+    - numpy 
+    - cython
+
+  run:
+    - python >=3.6
+    - pandas
+    - requests
+    - numpy
+    - scikit-learn
+    - scipy ==1.2.1
+    - matplotlib-base
+    
+
+test:
+  imports:
+    - fABBA
+
+
+about:
+  home: https://github.com/nla-group/fABBA
+  summary: 'An efficient aggregation method for the symbolic representation of temporal data.'
+  license: BSD 3-Clause
+  license_file: LICENSE
+  dev_url: https://github.com/nla-group/fABBA
+
+extra:
+  recipe-maintainers:
+    - chenxinye
+    - guettel
--- a/recipes/fABBA/meta.yaml
+++ b/recipes/fABBA/meta.yaml
@@ -35,6 +35,7 @@ test:
     - fABBA
   requires:
    - pip
+   - cython
   commands:
    - pip check
    
--- a/recipes/fABBA/meta.yaml
+++ b/recipes/fABBA/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "fABBA" %}
-{% set version = "0.9.6" %}
+{% set version = "0.9.7" %}
 
 package:
   name: {{ name|lower }}
@@ -9,7 +9,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
-  sha256: 04e69601c3ef2e8604ea58023a8e8f5b097bad8edd726011b65a4efb7e9f96c9
+  sha256: bf0672c2d0be449627380f51a7e695ddd43bf548a292de9dcfe8562e256ff95a
 
 
 build:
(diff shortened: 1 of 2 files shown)


chenxinye:
@conda-forge/help-python This is ready for review, thanks!

recipes/fABBA/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  home: https://github.com/nla-group/fABBA
+  summary: 'An efficient aggregation method for the symbolic representation of temporal data.'
+  license: BSD-3-Clause
+  license_file: LICENSE
+  dev_url: https://github.com/nla-group/fABBA
+
+extra:
+  recipe-maintainers:
+    - chenxinye
+    - guettel
```
ndevenish, line 53: I believe conda-forge needs @guettel to confirm that they are willing to be listed as a maintainer here

recipes/fABBA/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: bf0672c2d0be449627380f51a7e695ddd43bf548a292de9dcfe8562e256ff95a
+
+
+build:
+  noarch: python
```
carterbox, line 16:
```suggestion
```
This is not a noarch package because it contains a compiled module using cython.

recipes/fABBA/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+  run:
+    - python >=3.6
+    - pandas
+    - requests
+    - numpy
+    - tqdm
+    - scikit-learn
+    - scipy ==1.7.2
+    - matplotlib-base
```
carterbox, line 35:
```suggestion
requirements:
  build:
    - {{ compiler('c') }}
  host:
    - python
    - pip
    - numpy 
    - cython
  run:
    - python
    - pandas
    - requests
    - {{ pin_compatible('numpy') }}
    - scikit-learn
    - scipy >=1.2.1
```
Upstream states these run requirements. No matplotlib or tqdm.

recipes/fABBA/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    
+
+test:
+  imports:
+    - fABBA
```
carterbox, line 40:
```suggestion
test:
  imports:
    - fABBA
  requires:
   - pip
  commands:
   - pip check
```
Because you are compiling some code, it is also good to run some tests to see that the compiled functions execute properly.

recipes/fABBA/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "fABBA" %}
+{% set version = "0.9.7" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+
```
carterbox, line 8:
```suggestion
```

carterbox (changes requested):

recipes/fABBA/meta.yaml
carterbox: I'm also pretty sure that your setup.py is wrong regarding cython. cython is listed in the `install_requires` section, but it should probably be listed in the `setup_requires` section.

recipes/fABBA/meta.yaml
chenxinye: Fix, thanks for pointing out this, I copy the recipe from our another pure python package and forget to change it.

recipes/fABBA/meta.yaml
chenxinye: Thanks so much for pointing out this!

recipes/fABBA/meta.yaml
chenxinye: Thanks for mentioning that, we actually have codecov test in our repo, which works fine, so we did not arrange in conda recipe.

ndevenish:
So, i'm guessing that the change at https://github.com/nla-group/fABBA/commit/d7aa28fd4181911b42c1713db59fa7dc8a304c61 was made _after_ uploading to pypi, so `pip check` is thinking that it's actually a requirement, where it was actually mis-specified. (and this is where the error in the test was coming from).

Adding the requirement to the `test:` part only probably won't help, because although the tests will pass here in the CI - it leaves the pip dependency tree in an invalid state, and I believe any conda-forge sub-package that might attempt to use fABBA will also run into the issue when it tries to do `pip check`.

I think your options are probably:
- Make an updated release on pypi (possibly overkill)
- Make a post-release version on github
- [Patch](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#patches) the sources in the recipe to remote the unnecessary requirement. Depending on the actual commit that corresponds to the pypi package, https://github.com/nla-group/fABBA/commit/d7aa28fd4181911b42c1713db59fa7dc8a304c61.patch might be the only patch file to include.

The latter is what my instinct would be to do, until the next release picks up the correct setup.py, because then the current public pypi package is still the root source. But there are pros-and-cons to weigh.

(FWIW, although it doesn't affect this conda-forge recipe, I'd also encourage making a tag on the git repository - then you have a record of _exactly_ which commit corresponds to the published pypi package).

chenxinye:
Hi  Reviews,

This is indeed nonarch build, cause fABBA allows user's machine without C/C++ compiler. 
When the Cython compiling raises errors, then the installation will trigger trivial python installation, i.e., Cython is not necessary·

carterbox:
> This is indeed noarch build, cause fABBA allows user's machine without C/C++ compiler.

Yes, I see now that cython is optional in your setup.py, but because we added the c compiler to the build requirements, this conda package is no longer noarch. Conda packages are pre-built (meaning setup.py is not run on the end user's computer; it is run on our servers). Your users will download fABBA with the compiled extensions.

carterbox (changes requested):
The only thing missing is a comment from @guettel saying that they are willing to be a maintainer.

PR #18963, opened 2022-05-17
Add parquet-python

timkpaine:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/parquet-python/meta.yaml
+++ b/recipes/parquet-python/meta.yaml
@@ -11,7 +11,6 @@ source:
 
 build:
   number: 0
-  noarch: python
   script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
--- a/recipes/parquet-python/meta.yaml
+++ b/recipes/parquet-python/meta.yaml
@@ -27,15 +27,32 @@ requirements:
 test:
   imports:
     - parquet
+    host:
+      - python
+    run:
+      - python
 
 outputs:
   - name: parquet-python
-  - name: parquet_python
-    build:
-        noarch: generic
     requirements:
+      host:
+        - python
+        - pip
+      run:
+        - python
+        - thriftpy2
+
+    test:
+      imports:
+        - parquet
+
+  - name: parquet_python
+    requirements:
+      host:
+        - python
       run:
         - {{ pin_subpackage("parquet-python", exact=True) }}
+
     test:
       imports:
         - parquet
--- a/recipes/parquet-python/meta.yaml
+++ b/recipes/parquet-python/meta.yaml
@@ -27,6 +27,7 @@ outputs:
     requirements:
       host:
         - python
+        - pip
       run:
         - python
         - thriftpy2
--- a/recipes/parquet-python/meta.yaml
+++ b/recipes/parquet-python/meta.yaml
@@ -16,17 +16,17 @@ build:
   entry_points:
     - parquet = parquet.__main__:main',
 
-  requirements:
-    host:
-      - python
-      - pip
-    run:
-      - python
-      - thriftpy2
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - thriftpy2
 
-  test:
-    imports:
-      - parquet
+test:
+  imports:
+    - parquet
 
 outputs:
   - name: parquet-python
--- a/recipes/parquet-python/meta.yaml
+++ b/recipes/parquet-python/meta.yaml
@@ -11,26 +11,20 @@ source:
 
 build:
   number: 0
+  noarch: python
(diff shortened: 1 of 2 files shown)


timkpaine:
@conda-forge/help-python all set! had to make an arch package since it creates a binary `parquet.exe`

recipes/parquet-python/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/p/parquet/parquet-{{ version }}.tar.gz
+  sha256: fb1c90768c1b9159d4d6a9b3112ea8107b0b46d7491c5ac452ba7350f333bf0a
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
ndevenish, line 14:
```suggestion
build:
  noarch: python
  number: 0
  script: {{ PYTHON }} -m pip install . --no-deps -vv
  entry_points:
    - parquet = parquet.__main__:main
```
The `.exe` was complained about created because pip creates them for console_scripts entry points on windows; if you provide conda-build with a specific list of entry_points, it knows that it needs to handle that issue for noarch packages on windows, so you can remain noarch.

carterbox: >it knows that it needs to handle that issue for noarch packages on windows, so you can remain noarch.

Yes. Note that this is mentioned in the docs for noarch python

> I also believe --no-deps isn't required on conda-forge.

Agreed. I think there is some environment variable that sets this for all builds.

recipes/parquet-python/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+
+outputs:
+  - name: parquet-python
+    requirements:
+      host:
+        - python >=3.6
```
carterbox, line 27: Where does it say that the minimum version is 3.6? Might be good all the way back to 3.0.

carterbox (changes requested):

recipes/parquet-python/meta.yaml
timkpaine: conda-forge only builds 3.7+ so ive been turning off <=3.6 for everything as a default. But since this is supposed to be `noarch: python` i'll remove

timkpaine:

timkpaine:
as worked through on other PRs, the "extraneous" `python` deps are actually required to enable the various jinja functions which would not otherwise work

carterbox:
Multiple ways to do it, actually. And sorry for messing up the name of your feedstock last night.

https://github.com/conda-forge/jupyterlab_nbconvert_nocode-meta-feedstock/blob/main/recipe/meta.yaml
https://github.com/conda-forge/importlib_metadata-feedstock/blob/main/recipe/meta.yaml
https://github.com/conda-forge/importlib_resources-feedstock/blob/main/recipe/meta.yaml 

Let's see if I can get it right this time.

recipes/parquet-python/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+source:
+  url: https://pypi.io/packages/source/p/parquet/parquet-{{ version }}.tar.gz
+  sha256: fb1c90768c1b9159d4d6a9b3112ea8107b0b46d7491c5ac452ba7350f333bf0a
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  entry_points:
+    - parquet = parquet.__main__:main',
```
carterbox, line 17:
```suggestion
    - parquet = parquet.__main__:main
```
Function names can't contain punctuation.

timkpaine:
@carterbox that didnt work

timkpaine:
https://github.com/conda-forge/jupyterlab_nbconvert_nocode-meta-feedstock this is insanely complicated.

carterbox:
> insanely complicated

I think that's a bit of an overstatement. The only thing it does it move pip call to a build script because the PYTHON environment variable is defined when the Jinja variable PYTHON isn't.

timkpaine:
This one is better than the other repo. The other one has the build number variable, noarch generic and i need to remember that "-meta" is randomly in the name

carterbox:
@timkpaine, thanks again for your patience while we iterated over various ways to add an alias for your packages. I think this one is the best solution? I encourage you to update your other feedstocks with this approach when you make new releases.

timkpaine:
@carterbox thank you!! I will do so, I only updated 1-2 of my feedstocks with this (old) approach, will update them then proceed to apply it to all my packages. No more "pip install X-Y works but for conda I have to install X_Y"

PR #19000, opened 2022-05-18
DOC: Add an example recipe for compiled packages

carterbox:
The purpose of this PR is to add a meta.yaml for compiled packages. The existing example is for python projects only and does not include some of the key features of a compiled library recipe. i.e. appropriate run_exports, existence testing for headers / package configuration, and checking the logs for linker warnings.

### DIFF
--- a/recipes/example/meta-compiled.yaml
+++ b/recipes/example/meta-compiled.yaml
@@ -1,0 +1,103 @@
+# This recipe example is for packages which are compiled libraries and binaries.
+
+# Jinja variables help maintain the recipe as you'll update the version only
+# here. Using the name variable with the URL in line 7 is convenient when
+# copying and pasting from another recipe, but not really needed.
+{% set name = "simplejson" %}
+{% set version = "3.8.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  # The URL for GitHub releases
+  url: https://github.com/simplejson/simplejson/releases/download/{{ version }}/simplejson-{{ version }}.tar.gz
+  # otherwise fall back to archive:
+  # url: https://github.com/simplejson/simplejson/archive/v{{ version }}.tar.gz
+  sha256: 2b3a0c466fb4a1014ea131c2b8ea7c519f9278eba73d6fcb361b7bdb4fd494e9
+  # sha256 is the preferred checksum -- you can get it for a file with:
+  #  `openssl sha256 <file name>`.
+  # You may need the openssl package, available on conda-forge:
+  #  `conda install openssl -c conda-forge``
+
+build:
+  # Use separate bld.bat and build.sh files to define the build steps. By
+  # default, the package will be built for all major OSs. Add the line "skip:
+  # True  # [not win]" to limit to Windows. More info about selectors can be
+  # found in the conda-build docs:
+  # https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors
+  number: 0
+  # run_exports add a dependency to packages which dynamically link against
+  # your package and need it at run time. You should adjust this pinning to
+  # ensure ABI compatability.
+  run_exports:
+    - {{ pin_subpackage( name|lower ) }}
+
+requirements:
+  build:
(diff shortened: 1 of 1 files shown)


recipes/example/meta-compiled.yaml
```diff
@@ -0,0 +1,103 @@
...
+  # run_exports add a dependency to packages which dynamically link against
+  # your package and need it at run time. You should adjust this pinning to
+  # ensure ABI compatability.
+  run_exports:
+    - {{ pin_subpackage( name|lower ) }}
+
+requirements:
+  build:
+    # The build section is for packages that are used on the build platform.
+    # Compilers are named 'c', 'cxx' and 'fortran'.
```
ngam, line 40: + go, cuda, etc.? (link to list?)

recipes/example/meta-compiled.yaml
```diff
@@ -0,0 +1,103 @@
...
+  # otherwise fall back to archive:
+  # url: https://github.com/simplejson/simplejson/archive/v{{ version }}.tar.gz
+  sha256: 2b3a0c466fb4a1014ea131c2b8ea7c519f9278eba73d6fcb361b7bdb4fd494e9
+  # sha256 is the preferred checksum -- you can get it for a file with:
+  #  `openssl sha256 <file name>`.
+  # You may need the openssl package, available on conda-forge:
+  #  `conda install openssl -c conda-forge``
+
+build:
+  # Use separate bld.bat and build.sh files to define the build steps. By
```
ngam, line 25: I'd also clarify this a little further to say it is possible to list the build steps here instead of separate file (especially if it is just `cmake ..`)

recipes/example/meta-compiled.yaml
```diff
@@ -0,0 +1,103 @@
+# This recipe example is for packages which are compiled libraries and binaries.
```
ngam, line 1: Would a more realistic example be a python package that uses compiled binaries/libraries? That example may prove more useful to explain the linking and prefix location, etc.

recipes/example/meta-compiled.yaml
ngam: Either way, I would also include a supplementary build.sh file anyway

ocefpaf:
Maintaining the examples is kind of an extra burden and we usually get folks changing them or deleting them when submitting packages. We should probably do this in the docs and eliminate the example folder altogether.

BastianZim:
How about adding this as a readme? Then people won‘t change it (I hope).

ocefpaf:
> How about adding this as a readme? Then people won‘t change it (I hope).

My only concern is having the README out of sync with the docs but that is a minor problem and I believe having is in the README is way better than in the examples folder.

BastianZim:
Hmm true. Maybe just a link then, with also a link to Grayskull? 
Anyways, +1 from my side for having this in the docs because the example recipe gets changed often by first-time submitters.

ocefpaf:
> Maybe a staged-recipes-examples repo is a good alternative? Again though, maintenance...

I think that I prefer a curated set of examples, like a list of existing feedstocks URLs in the README, than another repo.

ngam:
> > Maybe a staged-recipes-examples repo is a good alternative? Again though, maintenance...
> 
> I think that I prefer a curated set of examples, like a list of existing feedstocks URLs in the README, than another repo.

Sure, though the only problem then is the mapping between a feedstock recipe and a staged-recipes recipe, but that could be explained (e.g. that certain yaml files could be placed in the folder, etc.)

carterbox:
Sorry! This conversation got lost in my notifications. What I'm reading is that we should have a curated list of live examples feedstocks (not a separate examples repo). I think this would work, but these special example repos should have comment annotations explaining them like the current example. We could then replace the example folder in staged-recipes with a README that links to the feedstocks and explains to make a new folder in staged-recipes?

### PRO
- Examples are actively being used, so they will be maintained by the feedstock maintainers not by conda-forge documentation maintainers or staged-recipes
- A variety of examples could be provided for various languages and recipe complexities

### CON
- Example feedstocks will be cluttered with explanation text

PR #19003, opened 2022-05-19
Add emsarray-core recipe

mx-moth:
Checklist
- [x] Build number is 0.


mx-moth:

mx-moth:

mx-moth:

BastianZim (changes requested):
Hi @mx-moth, thanks for your contribution. Just a couple of points:
1. Please redo the recipe with Grayskull just for the formatting and some extra tests.
2. As this is an alpha version, you'll need to target the pre-release channel: https://conda-forge.org/docs/maintainer/knowledge_base.html#pre-release-builds
3. You added `-core` to your package but we prefer to keep the same name as PyPI. Is there a specific reason for it?

mx-moth:
1. I'll redo with Greyskull tomorrow as requested
2. It was alpha only because I was hoping to test out the release process before actually making a release. I intend to release the current code base as 0.1.0. If it makes the process easier, I'll publish the 0.1.0 PyPI package and update this recipe to match.
3. I am intending on adding a `emsarray` package with all of the optional extras (matplotlib for plotting, tutorial dataset support, etc) once the `emsarray-core` package is published. This is a similar pattern to what I found in [`dask`](https://anaconda.org/conda-forge/dask) / [`dask-core`](https://anaconda.org/conda-forge/dask-core) ([Installation instructions for dask / dask-core](https://docs.dask.org/en/latest/install.html#conda))

BastianZim:
1. Thanks
2. Ahh ok. Then I would suggest going with a full release as that makes it easier here.
3. For that, I would recommend going with a multi-output recipe and submitting everything in this PR. One example is this feedstock: https://github.com/conda-forge/sagemaker-inference-toolkit-feedstock/blob/main/recipe/meta.yaml

mx-moth:
Redone with grayskull! The output recipe used `{{ PYTHON }}` in the build step, which was undefined in Jinja and caused the build to fail. I've replaced that with `python` as suggested in various github issues or stackoverflow questions.

I think I've got all the build set ups working. It creates two packages, and one depends on the other as desired, with the appropriate version pins. I'm unsure about host vs. run requirements and where `python` needs to be in those, and how to specify that this is for Python 3.8 and higher. It seems to work so far...

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
+{% set name = "emsarray" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
```
BastianZim, line 5:
```suggestion
  name: {{ name|lower }}-meta
```
This is a meta-package which will not actually get built.

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/emsarray-{{ version }}.tar.gz
+  sha256: 63c92c95483ec263aecf88266acf102513e2ce74099c035172affe4eff959729
+
+build:
+  number: 0
```
BastianZim, line 13:
```suggestion
  number: {{ build_num }}
```

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+outputs:
+  - name: emsarray-core
+    build:
+      noarch: python
+      script: python -m pip install . -vv --no-deps
```
BastianZim, line 19:
```suggestion
      script: python -m pip install . -vv --no-deps
```
This should be moved into a separate file, then it should work.

Example: https://github.com/conda-forge/sagemaker-inference-toolkit-feedstock/tree/main/recipe

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+        - emsarray
+      commands:
+        - pip check
+        - emsarray --help
+      requires:
+        - pip
+
+  - name: emsarray
+    build:
+      noarch: python
```
BastianZim, line 46:
```suggestion
      noarch: generic
      string: {{ build_num }}_fullpackage
```
Generic since you're not actually building anything and string so that it is clearer what this is.

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+    requirements:
+      host:
+        - python >=3.8
+        - pip
```
BastianZim, line 50: Since you're not building anything, this can be removed.
```suggestion
```

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  summary: Utilities for working with non-CF grid datasets
+  license: LicenseRef-CSIRO-Open-Source-Software-Licence-Agreement
+  license_family: BSD
+  license_file: LICENCE
+  doc_url: https://emsarray.readthedocs.io/
+  dev_url: https://github.com/csiro-coasts/emsarray/
+
+extra:
+  recipe-maintainers:
+    - mx-moth
```
BastianZim, line 76:
```suggestion
    - mx-moth
  feedstock-name: {{ name|lower }}
```
This makes sure that the right name is chosen for the feedstock.

BastianZim (changes requested):

BastianZim:
> Redone with grayskull! The output recipe used `{{ PYTHON }}` in the build step, which was undefined in Jinja and caused the build to fail. I've replaced that with `python` as suggested in various github issues or stackoverflow questions.

Yes, that's due to the way it is made available. Generally, it is preferred to keep the variable as that includes some stuff from conda so please go with the suggested separate file here.

> I think I've got all the build set ups working. It creates two packages, and one depends on the other as desired, with the appropriate version pins. I'm unsure about host vs. run requirements and where `python` needs to be in those, and how to specify that this is for Python 3.8 and higher. It seems to work so far...

Yes, everything is correct.
The host section is for building which is why this can be removed for the second package since you don't build anything there.

mx-moth:
Changes made. I had to unindent the `script: build_base.sh` line one level to make it work, which matches the example repository you linked. It builds locally now, lets see what CI thinks of it

(Apologies for the small flurry of changes just then)

recipes/emsarray/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+build:
+  number: {{ build_num }}
+
+outputs:
+  - name: emsarray-core
+    build:
+      noarch: python
+      entry_points:
+        - emsarray = emsarray.cli:main
+    script: ./build_base.sh
```
BastianZim, line 22:
```suggestion
    script: build_base.sh
```
Just to make it simpler.

mx-moth:
Thanks!

PR #19033, opened 2022-05-20
Add recipe for meson-python

awvwgk:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/meson-python/meta.yaml
+++ b/recipes/meson-python/meta.yaml
@@ -1,0 +1,66 @@
+{% set name = "meson-python" %}
+{% set version = "0.5.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/meson_python-{{ version }}.tar.gz
+  sha256: 84d2610829b1773c098055b83a04258cd92079f3cf8e96d3bcc2b31513e5cf83
+  patches:
+    - ninja-overdep.patch
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - colorama
+    - meson >=0.60.0
+    - ninja
+    - pep621 >=0.3.0
+    - pip
+    - python >=3.7
+    - tomli >=1.0.0
+    - typing-extensions >=3.7.4
+  run:
+    - colorama
+    - meson >=0.60.0
+    - ninja
+    - pep621 >=0.3.0
+    - python >=3.7
+    - tomli >=1.0.0
+    - typing-extensions >=3.7.4
+    - wheel >=0.36.0
+
+test:
+  imports:
+    - mesonpy
+  commands:
+    - pip check
+    - pytest tests -vv -k "not test_contents_unstaged"
+  source_files:
+    - tests/
+  requires:
+    - {{ compiler('c') }}
+    - git
+    - gitpython
+    - patchelf  # [linux]
+    - pip
+    - pytest
+    - pytest-mock
+
+about:
+  home: https://github.com/FFY00/meson-python
+  doc_url: https://meson-python.readthedocs.io/en/latest/
+  summary: Meson Python build backend (PEP 517)
+  license: EUPL-1.2
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - awvwgk
+    - rgommers
--- a/recipes/meson-python/ninja-overdep.patch
+++ b/recipes/meson-python/ninja-overdep.patch
@@ -1,0 +1,32 @@
+diff --git a/PKG-INFO b/PKG-INFO
+index 40355c0..e0f6a39 100644
+--- a/PKG-INFO
++++ b/PKG-INFO
+@@ -14,7 +14,6 @@ Project-URL: Changelog, https://mesonpy.readthedocs.io/en/latest/changelog.html
+ Requires-Python: >=3.7
+ Requires-Dist: colorama; os_name == "nt"
+ Requires-Dist: meson>=0.60.0
+-Requires-Dist: ninja
+ Requires-Dist: pep621>=0.3.0
+ Requires-Dist: tomli>=1.0.0
+ Requires-Dist: typing-extensions>=3.7.4; python_version < "3.8"
+diff --git a/pyproject.toml b/pyproject.toml
+index d61a210..b66565a 100644
+--- a/pyproject.toml
++++ b/pyproject.toml
+@@ -3,7 +3,6 @@ build-backend = 'mesonpy'
+ backend-path = ['.']
+ requires = [
+   'meson>=0.62.0',
+-  'ninja',
+   'pep621>=0.3.0',
+   'tomli>=1.0.0',
+   'typing-extensions>=3.7.4; python_version<"3.8"',
+@@ -28,7 +27,6 @@ classifiers = [
+ dependencies = [
+   'colorama; os_name == "nt"',
+   'meson>=0.60.0',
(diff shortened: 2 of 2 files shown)


awvwgk:
@conda-forge/help-python This recipe is ready for review. While I would prefer `noarch: python`, it seems more important to support all Python versions to not limit other Python packages planning to use this one as build system (like SciPy).

rgommers (changes requested):
Thanks for working on this @awvwgk!

The patch here looks wrong to me. If there's something to fix, let's do it upstream. But the reason for changing `ninja` requirements here is not correct either way: when one uses build isolation in a Python venv, a `ninja` installed by conda-forge cannot be used.

This package should be `noarch`. The `pip check` failure may be due to incorrect metadata in `pyproject.toml`; if needed a patch can be carried for that until it's fixed in the next `meson-python` release.

recipes/meson-python/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+requirements:
+  host:
+    - colorama
+    - meson >=0.60.0
+    - ninja
+    - pep621 >=0.3.0
+    - pip
+    - python >=3.7
+    - tomli >=1.0.0
+    - typing-extensions >=3.7.4
```
rgommers, line 28: `patchelf` is also needed, see `mesonpy/_elf.py`.

rgommers: On Linux only

awvwgk: Since we are building for noarch now tests will only run on Linux.

rgommers: There's actually a good argument for leaving it out, but I'm not sure if it's currently optional. It will depend on whether there's a standalone non-Python library being vendored.

Actually, on second though - maybe fine to leave it out until someone needs it. WDYT?

awvwgk: We cannot depend on platform specific libraries with a noarch package, if platform specific dependencies are required, this package cannot be noarch.

rgommers: Ah, then let's leave it out - SciPy doesn't need it, and I don't think most other packages need it either.

carterbox: > We cannot depend on platform specific libraries with a noarch package, if platform specific dependencies are required, this package cannot be noarch.

That's [technically not true](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python). If you have a dependency that is available only available for some platforms, we still recommend noarch it's just that the package will be uninstallable for platforms which do not have the dependency.

There are two ways to remedy this:

1. If the dependency is required, accept that the package is uninstallable on some platforms because it doesn't work on those platforms but the package itself is pure python.
2. If the dependency is optional, you can remove the package from the run to run_constrained and users need to manually install if they want. OR ask upstream to add an empty metapackage for the platforms that are missing the package.

awvwgk: Well, that's not what I meant, but point taken ;). I'll switch to noarch and drop patchelf from the run requirements.

awvwgk:
Seems like this is ready for review again, now as `noarch; python`.

recipes/meson-python/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+about:
+  home: https://github.com/FFY00/meson-python
+  doc_url: https://meson-python.readthedocs.io/en/latest/
+  summary: Meson Python build backend (PEP 517)
+  license: EUPL-1.2
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - awvwgk
```
rgommers, line 64: Can you add me too @awvwgk? SciPy builds will depend on `meson-python`, so I'm interested in helping maintain this package.

recipes/meson-python/ninja-overdep.patch
```diff
@@ -0,0 +1,32 @@
+diff --git a/PKG-INFO b/PKG-INFO
+index 8999cfd..6a86c0d 100644
+--- a/PKG-INFO
++++ b/PKG-INFO
+@@ -14,7 +14,6 @@ Project-URL: Changelog, https://mesonpy.readthedocs.io/en/latest/changelog.html
+ Requires-Python: >=3.7
+ Requires-Dist: colorama; os_name == "nt"
+ Requires-Dist: meson>=0.60.0
+-Requires-Dist: ninja
```
rgommers, line 9: This patch is kinda unusual, so let me summarize my understanding:
- `PKG-INFO` and `pyproject.toml` are not used at all by `conda` tooling, so no harm done there
- This is added to make `pip check` in CI happy

Maybe worth adding a comment if we're keeping this patch permanently?

rgommers (approved):
LGTM overall now, thanks @awvwgk

recipes/meson-python/meta.yaml
awvwgk: Sure

recipes/meson-python/ninja-overdep.patch
awvwgk: I'm not an expert in Python package metadata, but at least the `project.dependencies` list in `pyproject.toml` creates incorrect information on the requirements of this package, which will be present in the installed package and are therefore flagged by `pip`.

The `Requires-Dist:` tag and `build-system.requires` entries are probably less critical here, but are patched for consistency to make the intent.

Patching dependency metadata in projects from PyPI is not unusual IMO, at least for the projects I packaged so far I found it to be incorrect or incomplete in many cases.

rgommers: I can't say I understand this. `ninja` _is_ a required dependency. Beyond avoiding `pip check`, I'm not sure what this patch does or incorrect information means. Maybe one of the other reviewers knows?

awvwgk: I guess the question boils down to whether `pip` is supposed to manage the dependency on a C++ program or not.

rgommers: Ah, I guess we just came to the same conclusion of that being the question:)  https://github.com/FFY00/meson-python/issues/60#issuecomment-1133957875

BastianZim: Just as an fyi mostly: Pip check is not necessary. It’s nice to have but fails in many situations where a dependency doesn‘t properly declare itself (like here). 
I‘d be happy to merge the PR without it if that's easier for you.

awvwgk: I would prefer to keep this patch to have tools like `build`, `pip` or `wheel` which work with the package metadata not flag the unrecognized dependency from the conda package. Of course it is possible to manually silence those warnings and checks, but this moves the burden to the user.

recipes/meson-python/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - colorama
+    - meson >=0.60.0
+    - ninja
```
rgommers, line 23: I think `wheel` should be here; must have overlooked that before. It gets pulled in anyway via `python` I guess, but it's a direct dependency so good to capture.

recipes/meson-python/ninja-overdep.patch
rgommers: 0.5.0 should be released later today with some changes to dependencies, may be worth waiting for that and updating this tomorrow.

awvwgk: Updated to 0.5.0

rgommers (approved):
LGTM, thanks!

rgommers:
@conda-forge/staged-recipes can this be merged?

PR #19035, opened 2022-05-20
Adding torchhd

dheyay:
Checklist
- [x] Build number is 0.


dheyay:
@conda-forge/help-python, PR is good to be reviewed

recipes/torchhd/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+test:
+  imports:
+    - torchhd
+    - torchhd.datasets
+
+about:
+  home: "https://github.com/hyperdimensional-computing/torchhd"
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
```
ndevenish, line 37: The `LICENCE.txt` in this PR actually isn't doing anything, as this line is (correctly) picking up the file `LICENSE` from the package sources

ndevenish: Although I just noticed the LICENSE packaged with the pypi [and repository](https://github.com/hyperdimensional-computing/torchhd/blob/f414c9efb17683eccdfb36f3d3de63330aaa717e/LICENSE) is different, as-in, different names under copyright. Do you know that this LICENCE.txt is actually more accurate and this is being corrected upstream? (in which case, this should be changed to point to the internal `LICENSE.txt` so that it isn't ignored)

dheyay: I have updated both the meta.yaml file to point to the correct LICENSE.txt and updated the license text file to be in line with the copyright on the repository as well.

recipes/torchhd/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
+  sha256: 4ffb7c627441aa71a3e286cab035bbe9c52d7ccfaf2605c613a22cdfea1eeccb
+
+build:
+  number: 0
+  skip: true  # [win]
```
BastianZim, line 14: If the only reason is a missing dependency, this can continue to stay noarch.

recipes/torchhd/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - numpy
+    - pandas
+    - python
+    - requests
+    - pytorch
+
+test:
+  imports:
+    - torchhd
+    - torchhd.datasets
```
BastianZim, line 31:
```suggestion
test:
  imports:
    - torchhd
    - torchhd.datasets
  commands:
    - pip check
  requires:
    - pip
```
This just double checks the dependencies.

recipes/torchhd/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
BastianZim, line 1: This license is already included in the tarball that you can download from PyPI so you can remove it.

recipes/torchhd/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+about:
+  home: "https://github.com/hyperdimensional-computing/torchhd"
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE.txt
```
BastianZim, line 37:
```suggestion
  license_file: LICENSE
```

BastianZim (changes requested):
Just a couple of things, then this is ready to go.

recipes/torchhd/meta.yaml
BastianZim: Yes, you can just specify PyTorch and ignore the failure on Windows. That way, if PyTorch ever becomes available on Windows, both packages can be installed.

dheyay: Right, I changed it to stay noarch in python. Also, How would I explicitly ignore the failure on Windows?

BastianZim: Ahh, ignoring means literally that. We’ll just ignore it here. :smile:
In the feedstock, noarch packages are only built on Linux so you won‘t see the error anymore. 
Ready to get merged then?

dheyay: Alright, I understand now haha. Thank you!
I just made the final change and made it noarch.

BastianZim:
Just one thing, the folder should be the same name as the recipe ("torch-hd")

PR #19081, opened 2022-05-26
 conda-forge recipe for the OpenMM ATMMetaForce plugin

egallicc:
Checklist
- [x] Build number is 0.


egallicc:
@conda-forge/staged-recipes please review. Thanks!

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Gallicchio-Lab/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: cf7975639c285ed0fab441c2ac36e3cbf6e499256a708240d4acfeb562939bae
+
+build:
+  number: 0
+  skip: true  # [win or osx or ( linux64 and cuda_compiler_version in (undefined, 'None') ) ]
```
carterbox, line 14:
```suggestion
  skip: true  # [not linux64]
  skip: true  # [cuda_compiler_version in (undefined, 'None')]
```
This is easier to change later if you want to add more platforms.

carterbox (commented):
I'm reviewing [another openmm plugin](https://github.com/conda-forge/staged-recipes/pull/18968), but I'm still not familiar with how these are used.

Are these plugins intended to always be used with/via python or do people ever use the C API part by itself? If people use the C API alone, then we want to split python and C libs so that C users aren't required to install python.

Additionally, the build matrix for python + CUDA packages is quite large (4 CUDA versions x 3 python versions) by separating python and CUDA we can have 7 builds instead of 12.

recipes/openmm-atmmetaforce-plugin/meta.yaml
egallicc: Thank you @carterbox. This OpenMM plugin is currently used exclusively through its python interface because the C++ API does not implement utility functions indispensable for the calculations.

I have committed the suggested change.

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,84 @@
...
+  run:
+    - python
+    - openmm
+    - fftw
+    - numpy
+    # OpenCL ICD
+    - ocl-icd  # [linux]
+    - ocl-icd-system  # [linux]
+    - khronos-opencl-icd-loader  # [win or (osx and opencl_impl == 'khronos')]
+    - ocl_icd_wrapper_apple      # [osx and opencl_impl == 'khronos']
```
carterbox, line 53: Please double check these requirements. I searched your plugins source code and didn't see any mention of numpy, fftw, or cython. Also, the build section is for build tools, host is for libraries or tools that must match the target platform when cross-compiling, and the run section is for python dependencies.

Do you need doxygen? I doesn't look like your are shipping docs pages.

carterbox (changes requested):
> This OpenMM plugin is currently used exclusively through its python interface because the C++ API does not implement utility functions indispensable for the calculations.

Thanks for clearing that up. Going forward, I will assume that this is a python package with C extension and not a C library with a python wrapper.

Please see notes below about reducing your declared dependencies.

egallicc (commented):
Thank you. Confirmed.

recipes/openmm-atmmetaforce-plugin/meta.yaml
egallicc: Thank you. I have revised the build and host requirements as suggested and following the examples.

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [linux]
+    - cmake
+
+  host:
+    - python
+    - openmm
+    - swig
```
carterbox, line 28:
```suggestion
  build:
    - {{ compiler('c') }}
    - {{ compiler('cxx') }}
    - {{ compiler('cuda') }}  # [linux]
    - cmake
    - swig

  host:
    - python
    - openmm
```
SWIG is a build tool, so it belongs in build.

recipes/openmm-atmmetaforce-plugin/build.sh
```diff
@@ -0,0 +1,29 @@
...
+
+mkdir build
+cd build
+cmake ${CMAKE_ARGS} ${SRC_DIR}
+make -j $CPU_COUNT
+make install PythonInstall
+
+for lib in ${PREFIX}/lib/plugins/*${SHLIB_EXT}; do
+    ln -s $lib ${PREFIX}/lib/$(basename $lib) || true
+done
```
carterbox, line 29: Since this is a python library with c extension, you shouldn't need to ship any c-headers. Please add a step to the build script which removes the files added to /include.

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - lxml
+    # OpenCL ICD
+    - ocl-icd  # [linux]
+    - khronos-opencl-icd-loader  # [win or (osx and opencl_impl == 'khronos')]
+    
+  run:
+    - python
+    - openmm
+    # OpenCL ICD
+    - ocl-icd  # [linux]
```
carterbox, line 39:
```suggestion
    - ocl-icd  # [linux]
```
The link inspector [reports that this package is unused](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=517904&view=logs&j=eb6ecf2b-0363-5c1a-baa7-2d771889d197&t=d8576026-8998-52ff-d027-1d631b0fb837&l=1853). Why is it added as a requirement?

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - khronos-opencl-icd-loader  # [win or (osx and opencl_impl == 'khronos')]
+    - ocl_icd_wrapper_apple      # [osx and opencl_impl == 'khronos']
+
+test:
+  requires:
+    - pocl  # [unix]
+    - pip
+  imports:
+    - simtk
+    - simtk.openmm
```
carterbox, line 50:
```suggestion
    - openmm
```
There is a [deprecation warning](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=517904&view=logs&j=eb6ecf2b-0363-5c1a-baa7-2d771889d197&t=d8576026-8998-52ff-d027-1d631b0fb837&l=3148) for this test.

carterbox (changes requested):

recipes/openmm-atmmetaforce-plugin/build.sh
egallicc: Is this step absolutely necessary? The OpenMM package, of which this is a plugin, ships C++ header files. All of the OpenMM plugins on conda-forge with a python interface that I have seen ship C++ headers.

The C++ header of this plugin is very small. Removing it will not save a significant amount in terms of file size.

An additional reason to preserve C++ headers is that it would allow users to use the C++ API in their derivative codes, even though it would not have the full functionality of the python interface.

If this step is absolutely necessary, could you please provide an example of the commands to delete C++ headers from the default include directory? Or please provide an example of a package with a build script that does this. Thank you!

recipes/openmm-atmmetaforce-plugin/meta.yaml
egallicc: Done

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - pocl  # [unix]
+    - pip
+  imports:
+    - simtk
+    - openmm
+  commands:
+    - pip check
+    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginCUDA$SHLIB_EXT  # [unix]
+    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginOpenCL$SHLIB_EXT  # [unix]
+    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginReference$SHLIB_EXT  # [unix]
```
carterbox, line 50:
```suggestion
  commands:
    - pip check
    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginCUDA$SHLIB_EXT  # [unix]
    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginOpenCL$SHLIB_EXT  # [unix]
    - test -f $PREFIX/lib/plugins/libATMMetaForcePluginReference$SHLIB_EXT  # [unix]
    - test -f $PREFIX/include/ATMMetaForce.h  # [unix]
    - test -f $PREFIX/include/internal/ATMMetaForceImpl.h  # [unix]
```
Test the existence of a header since you intend to ship them. Helps catch moving or renaming of headers at a new release.

recipes/openmm-atmmetaforce-plugin/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - openmm
+    - pthread-stubs
+    - lxml
+    # OpenCL ICD
+    - ocl-icd  # [linux]
+    - khronos-opencl-icd-loader  # [win or (osx and opencl_impl == 'khronos')]
+    
+  run:
+    - python
+    - openmm
```
carterbox, line 37:
```suggestion
    # openmm package is missing a run_export as of 9 June 2022
    - {{ pin_compatible('openmm') }}
```
Ensures that future API changes to openmm's C interface don't break your package.

carterbox (changes requested):

recipes/openmm-atmmetaforce-plugin/build.sh
egallicc: Thank you for the help and explanations. I have committed the changes indicated.

carterbox (changes requested):
Please see comment above about overdepending on ocl-icd.

recipes/openmm-atmmetaforce-plugin/meta.yaml
egallicc: Apologies. It should be clean now.

.scripts/build_steps.sh
```diff
@@ -55,3 +55,4 @@ find ~/staged-recipes-copy/recipes -mindepth 2 -maxdepth 2 -type f -name "yum_re
 python ${CI_SUPPORT}/build_all.py
 
 touch "/home/conda/staged-recipes/build_artifacts/conda-forge-build-done"
+sleep infinity
```
carterbox, line 58: Why are you trying to timeout the CI?

egallicc: Ooops ... I did not mean to push that. My mistake. Sorry again. It has been removed.

PR #19103, opened 2022-05-29
[WIP] Add pytorch dependencies

hmaarrfk:
Results:
* libonnx https://github.com/conda-forge/onnx-feedstock/pull/105

Checklist
- [ ] Build number is 0.


ngam:
xref: https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108

h-vetinari:
Awesome work!

Let me know if something specific needs reviewing. :)

hmaarrfk:
One thing I would want to do is make the patches for the CMake changes upstreamable. I've learned that some dependencies are header only.

Those are stated as `STATIC` libraries, while we should only be looking for the `INTERFACE`. This seems like something that upstream might accept. 

Finally, I think using `FindPackage` is likely good for shared system libraries, but I haven't had a chance to test this hypothesis.

Ultimately, this is mostly an academic exercise at this stage, since I don't know how to maintain this.
We would likely have to build these libraries statically, but they don't really move the needle in terms of compilation time for pytorch.

h-vetinari:
@hmaarrfk 
Should we still pursue this...? Seems it was pretty far (passing everywhere except osx)?

hmaarrfk:
I think I did some good damage. But ultimately, one has to show that it gets pytorch GPU under the 6 hour limit.

Upstream doesn't really feel compelled to create "real versions" so maintaining will also be hard.

hmaarrfk:
ns

recipes/torch_tb_profiler/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+source:
+  url: https://github.com/pytorch/kineto/archive/{{ gitrev }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  skip: true  # already built
+  noarch: python
+  script:
+    - cd tb_plugin/
+    - python setup.py install --single-version-externally-managed --record=record.txt
```
hmaarrfk, line 20:
```suggestion
  script: "{{ PYTHON }} -m pip install . -vv"
```

recipes/torch_tb_profiler/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - pandas
+    - tensorboard
+
+test:
+  imports:
+    - torch_tb_profiler
+
+about:
+  home: https://github.com/pytorch/kineto/tree/main/tb_plugin
+  license: BSD
```
hmaarrfk, line 37:
```suggestion
  license: BSD-3-Clause
```

PR #19110, opened 2022-05-30
Add cpppo and greenery recipes

beenje:
Checklist
- [X] Build number is 0.

greenery >=2.0,<3.0 is required by cpppo, this is why I submitted 2.1 and not the latest version.
The LICENSE isn't included in that older package.

cpppo includes the LICENSE file but the copy of the GPL (COPYING) isn't part of the package.

### DIFF
--- a/recipes/cpppo/meta.yaml
+++ b/recipes/cpppo/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "cpppo" %}
+{% set version = "4.4.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cpppo-{{ version }}.tar.gz
+  sha256: 8c5a2947d6e9fd732d2d887fa02216beebb4068ab99e2590590f491425992bc6
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  entry_points:
+    - enip_server = cpppo.server.enip.main:main
+    - enip_client = cpppo.server.enip.client:main
+    - enip_get_attribute = cpppo.server.enip.get_attribute:main
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+    - greenery >=2.0,<3.0
+
+test:
+  imports:
+    - cpppo
+    - cpppo.server
+    - cpppo.server.enip
+  commands:
+    - pip check
+    - enip_server --help
+    - enip_client --help
+    - enip_get_attribute --help
+  requires:
+    - pip
+
+about:
+  home: https://github.com/pjkundert/cpppo
+  summary: Cpppo is a Communication Protocol Python Parser and Originator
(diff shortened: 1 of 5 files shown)


beenje:
@conda-forge/help-python, this PR is ready for review

beenje:
Added the deprecated `enip_getattr` entry point to avoid error on windows: `noarch_python] Noarch package contains binary script: enip_getattr.exe`

Note that it's not tested as the import doesn't work (was removed in master).

ndevenish:
If it outright doesn't work, and it's removed in master (but not a release), it might be better to temporarily carry a patch here (e.g. [4c217b.diff](https://github.com/pjkundert/cpppo/commit/4c217b6c06b88bede3888cc5ea2731f271a95086.diff)) rather than package a broken entry point

ndevenish:
Like https://github.com/ndevenish/staged-recipes/commit/b0b616e4cc16217bded6240852e2889f7dbcac39

beenje:
Very good point! Thanks a lot for the patch!

recipes/cpppo/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - enip_server = cpppo.server.enip.main:main
+    - enip_client = cpppo.server.enip.client:main
+    - enip_get_attribute = cpppo.server.enip.get_attribute:main
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
```
carterbox, line 28:
```suggestion
    - python >=3.0
    - pip
  run:
    - python >=3.0
```
Technically, it says it's compatible with python>=2.7, but that requires a backport.

carterbox (changes requested):

recipes/cpppo/meta.yaml
beenje: I don't mind using 3.0 but I thought conda-forge encouraged to pin to the "supported" version of Python. Isn't that the case?

carterbox: I'm not familiar with the reasons behind that policy.

Since this package is noarch, I don't see why providing the version spec stated by upstream is a problem. If this wasn't noarch, then I would understand that reasoning.

carterbox: AFAIK version constraints on noarch: python packages are completely optional. The docs imply that need not constrain python if the package supports all versions.

https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=noarch%20python#noarch-python

beenje: Thanks for the merge!
The link above above with "no python specific requirement" is a condition for a package to be noarch.

I think the minimum python version for noarch package is to help conda to reduce the search space when resolving environment. But I don't see anything about that in the documentation so it might be something that was removed or that I misunderstood :-)

BastianZim: Just my two cents but the 3.6 restriction was indeed for helping the search. Since we now have mamba, I don‘t think this is a problem anymore since it’s generally more performant. 
It never was in the docs though, was just passed around knowledge.😄

PR #19126, opened 2022-06-01
Add stimuli package

mscheltienne:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/stimuli/meta.yaml
+++ b/recipes/stimuli/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "stimuli" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/stimuli-{{ version }}.tar.gz
+  sha256: 1d54937cacbf678bed3431f173d94c3cdc613bc7259bc3a494be83821c37bbde
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >= 3.8
+    - setuptools >= 61.0.0
+    - wheel
+  run:
+    - matplotlib-base
+    - numpy
+    - opencv
+    - python >= 3.8
+    - scipy
+    - screeninfo
+    - python-sounddevice
+
+test:
+  imports:
+    - stimuli
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://github.com/mscheltienne/simple-stimuli
+  summary: Simple auditory and visual stimuli that do not require PsychoPy.
+  license: MIT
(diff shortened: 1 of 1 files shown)


mscheltienne:
I confirm I would like to maintain this recipe.

mscheltienne:
@conda-forge/help-python It should be good to go. This is my first time writing a recipe for conda-forge, so please let me know what can be changed or improved!
Should the tests run by `pytest` be executed when building the conda-forge feedstock?

mscheltienne:
Couple of questions:
- noarch seems like the way to go most of the time, making everything simpler. Any-chance to get it noarch with this dependency on `screeninfo`?
- Does `skip: true  # [py < 38)]` correctly limits the distribution of this package on the conda-forge channel to python version before 3.8? If so, why is this different than putting a constraint on the python-requirement?

mscheltienne:
I don't know how to fix the `cython` dependency for `screeninfo`. Any help would be appreciated 🙏

BastianZim:
1. > Should the tests run by pytest be executed when building the conda-forge feedstock?
  
  Depends. Generally, we like to run tests when something is compiled to make sure that everything worked correctly. When it's just python, the import test is sufficient. When testing only requires pytest and it's already included in the source distribution it's an easy "win" and you can add it, even if it's pure python but it's not required. 

2. > noarch seems like the way to go most of the time, making everything simpler. Any-chance to get it noarch with this dependency on screeninfo?

Yes, this can be noarch.

3. > Does skip: true # [py < 38)] correctly limits the distribution of this package on the conda-forge channel to python version before 3.8? If so, why is this different than putting a constraint on the python-requirement?

That will be after. Basically, any specified versions will be skipped.
This is about compiled vs noarch. For noarch only one version is built and the python limits come from the version constraints. For arch packages, each version is built and the skip: True part removes builds that would build for a python version that is not supported. So, if you have noarch, limiting using version constraints is sufficient.

3. I don't know how to fix the python dependency for screeninfo. Any help would be appreciated 🙏

That's nothing you can fix. The screeninfo package didn't include a dependency correctly.
Their list: https://github.com/conda-forge/screeninfo-feedstock/blob/7b79fa209fe6dfcf46558ee633e39e3ab26523cc/recipe/meta.yaml#L21-L24
Upstream: https://github.com/rr-/screeninfo/blob/9269bc36fd690c26caf9ce9d2b463d23ce873f0a/pyproject.toml#L31

What I would suggest is making a PR to that feedstock and fixing the dependency. Once that is updated, your package should work. If they're not responsive or if you want to get this out immediately, you can also add it here as a temporary fix.

mscheltienne:
Done: https://github.com/conda-forge/screeninfo-feedstock/pull/7 Hopefully, the PR is correct..
I had a feeling the issue was on the `screeninfo` feedstock..

Thanks for all the information and for taking the time to explain this!

mscheltienne:
@BastianZim Looks good now, what are the next steps?

BastianZim:
Nothing. That’s it. :)

PR #19156, opened 2022-06-06
Add xcb-util-image and xcb-util-renderutil for linux

erykoff:
With the update of `qt-main` to 5.15, additional xcb modules are necessary (see https://github.com/conda-forge/qt-main-feedstock/issues/32).  Although these could be CDT, they are also easy to package with conda-forge and some already are. Specifically, the additional packages used by `qt-main` are `xcb-util-wm`, `xcb-util-keysyms`, `xcb-util-renderutil`, and `xcb-util-image`.  The first two of these are already available on conda-forge.  This PR adds the last two, following the template of the first two.  These are very straightforward c compilations.

I have confirmed locally that the outputs from these feedstocks provide the necessary libraries and allow `qt-main=5.15` to work without external provision of these packages (and therefore not requiring any additional CDT packages when upgrading from `qt=5.12` to `qt-main=5.15`).

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/xcb-util-image/COPYING
+++ b/recipes/xcb-util-image/COPYING
@@ -1,0 +1,27 @@
+Copyright © 2007-2008 Bart Massey <bart@cs.pdx.edu>
+Copyright © 2008 Julien Danjou <julien@danjou.info>
+Copyright © 2008 Keith Packard <keithp@keithp.com>
+
+Permission is hereby granted, free of charge, to any person
+obtaining a copy of this software and associated documentation
+files (the "Software"), to deal in the Software without
+restriction, including without limitation the rights to use, copy,
+modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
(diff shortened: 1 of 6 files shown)


erykoff:
@conda-forge/staged-recipes This is ready.

ryanvolz:
I have no review powers, but I very much would like to see this merged to give users options in dealing with https://github.com/conda-forge/qt-main-feedstock/issues/32.

As for the actual recipes, two quick points that shouldn't be blockers:

1. I don't think you need to add the licenses to the recipe. I only looked at one of the source tarballs, but that one seemed to include the license and it would be enough to just specify the filename as you have already done.
2. Both packages include a static library. This matches the existing `xcb-util-*` packages, but [CFEP 18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md) does express a preference that they not be included. If you do want to follow the CFEP, it's easy enough to delete the static libraries from the build script after `make install`.

pkgw:
Thanks! I agree with @ryanvolz's two points but I also agree that they're minor, so I've gone ahead and merged.

PR #19160, opened 2022-06-06
Add recipe for pyston_lite

kmod:
Checklist
  - We use a tool from the LuaJIT package as a build step, but don't include any LuaJIT code in the build artifacts
- [x] Build number is 0.


kmod:

kmod:
@conda-forge-admin please ping conda-forge/help-python-c

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+about:
+  home: https://github.com/pyston/pyston
+  summary: 'A JIT for Python'
+  description: |
+      pyston-lite is the JIT component of Pyston, a faster implementation of Python.
+  license: PSF-2.0
+  license_file: LICENSE.txt
+
+extra:
```
xhochy, line 85:
```suggestion
extra:
  feedstock-name: pyston_lite
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+      commands:
+        - python -c "import sys; assert 'pyston_lite' in sys.modules"
+
+about:
+  home: https://github.com/pyston/pyston
+  summary: 'A JIT for Python'
+  description: |
+      pyston-lite is the JIT component of Pyston, a faster implementation of Python.
+  license: PSF-2.0
+  license_file: LICENSE.txt
```
xhochy, line 83: LICENSE is part of the source
```suggestion
  license_file: LICENSE
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - python
+      run:
+        - {{ pin_subpackage('pyston_lite', exact=True) }}
+
+    test:
+      commands:
+        - python -c "import sys; assert 'pyston_lite' in sys.modules"
+
+about:
+  home: https://github.com/pyston/pyston
```
xhochy, line 78: This doesn't match to where the sources come from

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+{% set pyston_version = "2.3.4" %}
+{% set python_version = "3.8.* *_cpython" %}
+{% set build_num = 0 %}
+
+package:
+  name: {{ name|lower }}-split
+  version: {{ pyston_version }}
+
+source:
+  - url: https://github.com/kmod/pyston/archive/bcbb8e939c28162e39d536622625961522424117.tar.gz
```
xhochy, line 11: Why is this not referring to any version?

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    folder: pyston/LuaJIT/
+
+build:
+  number: {{ build_num }}
+
+  skip: True  # [not linux]
+  skip: True  # [not x86_64 and not aarch64 and not arm64]
+  # Despite the pin on a Python version, the build script will still build a
+  # "python 3.9" version of this package which depends on python 3.8. So skip it as well:
+  skip: True  # [py!=38]
```
xhochy, line 27: The following should be sufficient
```suggestion
  skip: True  # [not linux]
  skip: True  # [py!=38]
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    script: build_pyston_lite.sh
+
+    build:
+      number: {{ build_num }}
+
+    requirements:
+      build:
+        # This dependency is required even though we specified it at the top-level scope
+        - {{ compiler('c') }}
+        # NB: listing a "python" dependence in this section will cause the package to misbuild
```
xhochy, line 48: I would also remove this is the case with all non-cross-compiled packages
```suggestion
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - name: pyston_lite
+    # This 'script' line must go under 'outputs', not 'build'
+    script: build_pyston_lite.sh
+
+    build:
+      number: {{ build_num }}
+
+    requirements:
+      build:
+        # This dependency is required even though we specified it at the top-level scope
```
xhochy, line 46: These are separate environments, so if you need something, you always need to redeclare it.
```suggestion
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    sha256: 753f3687d49173e860c737758b81d78419b74f3fd05abe6f05eb06c072cbf42c
+  # pyston-lite uses a preprocessing script from inside the LuaJIT repository (that is not
+  # packaged as part of the luajit package), but only as a build tool (ie is not included
+  # in the built artifacts).
+  - url: https://github.com/LuaJIT/LuaJIT/archive/e2c312e0deb874aa5fa8ce502c08d87deb38e82f.tar.gz
+    sha256: bff4764c19466def90667d2f7fcbd0781b8e62476b12105868f8e44babd03f8e
+    folder: pyston/LuaJIT/
+
+build:
+  number: {{ build_num }}
```
xhochy, line 21: Please put a plain number here, it will be propagated to the outputs
```suggestion
  number: 0
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
+{% set name = "pyston_lite" %}
+{% set pyston_version = "2.3.4" %}
+{% set python_version = "3.8.* *_cpython" %}
+{% set build_num = 0 %}
```
xhochy, line 4:
```suggestion
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+  build:
+    - {{ compiler('c') }}
+
+outputs:
+  - name: pyston_lite
+    # This 'script' line must go under 'outputs', not 'build'
+    script: build_pyston_lite.sh
+
+    build:
+      number: {{ build_num }}
```
xhochy, line 42:
```suggestion
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+        - python {{ python_version }}
+
+    test:
+      imports:
+        - pyston_lite
+
+  - name: pyston_lite_autoload
+
+    build:
+      number: {{ build_num }}
```
xhochy, line 62:
```suggestion
```

kmod:
Sorry for testing via pushing -- it was working locally but failing in CI. Is there a better way to debug those situations?

carterbox:
If you can run docker/podman on your local computer, you can use the `build-locall.py` script to build this recipe in the same environment as our CI (for any linux build).

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,86 @@
...
+  skip: True  # [py!=38]
+
+# In theory the requirements can/should be specified per-output, but in practice
+# conda-build will produce a corrupt package if the compiler dependency is not
+# additionally specified at this top scope. It is required to declare the dependency
+# both here and in the outputs: section as well; removing either will cause the c
+# compiler to not be installed.
+requirements:
+  build:
+    - {{ compiler('c') }}
```
carterbox, line 31: The top-level package can be one of the outputs, so you can do something like this to avoid redundant requirements sections:

```yaml

package:
  name: pyston_lite

# define pyston_lite metadata here because it is the main recipe

outputs:
  - name: pyston_lite  # just name as output, but it's already defined above
  - name: pyston_lite_autoload
    build:
      noarch: python
    # complete secondary output information

# all outputs get the same about section
about:

```

carterbox (changes requested):
This package is currently not building anything on the CI.

Also, the todo list was marked as complete, but not all users listed in the maintainers section have commented on this PR. Please don't check boxes that have not been completed.

kmod:
> If you can run docker/podman on your local computer, you can use the `build-locally.py` script to build this recipe in the same environment as our CI (for any linux build).

Sorry if I could have been more clear, but this is what I meant by "it was working locally". There seems to be some sort of difference between `build-locally.py` and what the CI is doing.

kmod:
> This package is currently not building anything on the CI.
> 
> Also, the todo list was marked as complete, but not all users listed in the maintainers section have commented on this PR. Please don't check boxes that have not been completed.

Apologies, I checked with them out-of-band but I guess I overlooked the part about having them comment here. I'll remove them for now because they are travelling without internet access.

I'll take a look at why it's not building anything, thanks for noticing that!

kmod:
Ok, I believe this is ready for review again.

I discovered that [conda-build is nondeterministic](https://github.com/conda/conda-build/issues/4512), which I suspect is why I was getting different results locally and on CI.

recipes/pyston_lite/meta.yaml
xhochy: No, please have the top-level package and outputs separate. This has bitten me multiple times. It works but only in very specific circumstances.

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+      # This build section cannot be empty or else it will crash the linter
+      script: true
+
+    requirements:
+      build:
+        # (See comment for the top-level requirements block)
+        - {{ compiler('c') }}
+        # NB: listing a "python" dependence in this section will cause the package to misbuild
+      host:
+        - python 3.8.* *_cpython
```
xhochy, line 56: This will be automatically set by the above skips. Better omit this implementation detail of naming the python implementation.
```suggestion
        - python
```

recipes/pyston_lite/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+        - make
+      run:
+        - python 3.8.* *_cpython
```
xhochy, line 59:
```suggestion
        - python
```

xhochy (commented):
Minor nits but otherwise looks OK

recipes/pyston_lite/meta.yaml
kmod: I just tried this, and `build-locally.py` failed because it tried to build a 3.10 version of this package. I might be able to remove this pin by duplicating the skip line into this output, if you prefer that approach.

Also pyston-lite is by nature highly coupled with the CPython implementation, using many internal implementation details that aren't part of the C API. I added this pin because it doesn't work with Pyston, which is otherwise a highly-compatible implementation. I added a comment, hopefully this is ok.

recipes/pyston_lite/meta.yaml
kmod: Ok I removed this one

recipes/pyston_lite/meta.yaml
carterbox: Can confirm that skip statement does not prevent conda-build from trying to use python=3.10.

kmod:
I don't have any particular horse in the package-naming race, though it does seem unfortunate that the feedstock name will have "split" in it. It feels more important to me to have the actual packages have good names, and I'm not sure if we can have it all?

Anyway let me know if there's anything else I can do to on the pr!

kmod:
Awesome thanks!

PR #19166, opened 2022-06-07
Add multivisor recipe

beenje:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/multivisor/meta.yaml
+++ b/recipes/multivisor/meta.yaml
@@ -8,6 +8,8 @@ package:
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/multivisor-{{ version }}.tar.gz
   sha256: f834f7b6968d478e1e2e24638d4fc29a9a57947fe5642cb266f109745112835b
+  patches:
+    - setup-exclude-tests.patch
 
 build:
   number: 0
--- a/recipes/multivisor/meta.yaml
+++ b/recipes/multivisor/meta.yaml
@@ -105,5 +105,6 @@ about:
   license_file: LICENSE
 
 extra:
+  feedstock-name: multivisor
   recipe-maintainers:
     - beenje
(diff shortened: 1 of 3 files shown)


beenje:
Build is failing on windows because `supervisor` isn't available, but package is noarch.

recipes/multivisor/meta.yaml
```diff
@@ -0,0 +1,107 @@
...
+  - name: multivisor-web
+    build:
+      noarch: python
+    requirements:
+      host:
+        - pip
+        - python >=3.6
+      run:
+        - python >=3.6
+        - {{ pin_subpackage('multivisor', exact=True) }}
```
jaimergp, line 49: I think there were some issues with `exact=True`, multi output recipes and noarch, which might require you to use `max_pin='x.x.x.x'`, instead, but I can't recall what was the problem :/

recipes/multivisor/meta.yaml
```diff
@@ -0,0 +1,107 @@
...
+      commands:
+        - multivisor-cli --help
+
+about:
+  home: https://github.com/tiagocoutinho/multivisor
+  summary: A centralized supervisor UI (web & CLI)
+  license: GPL-3.0-or-later
+  license_file: LICENSE
+
+extra:
```
jaimergp, line 105: You might want to add `feedstock-name` here to make sure you get a `multivisor-feedstock` repo and not a `multivisor-suite-feedstock` repo, if that's preferred.

recipes/multivisor/meta.yaml
beenje: I think I had to put the same host requirement everywhere to get the same build string.
I don't see any issue using max_pin. I changed to that.

beenje:
Nice catch about the `tests`! I added a patch and created a PR in the upstream repo: https://github.com/tiagocoutinho/multivisor/pull/90

Thanks!

recipes/multivisor/meta.yaml
jaimergp:
```suggestion
extra:
  feedstock-name: multivisor
```

Then I think we need this ^

PR #19220, opened 2022-06-10
Add recipe for skprogs

awvwgk:
Checklist
- [x] Build number is 0.

cc @aradi

### DIFF
--- a/recipes/skprogs/meta.yaml
+++ b/recipes/skprogs/meta.yaml
@@ -22,11 +22,7 @@ requirements:
     - cmake
     - ninja
     - pkg-config
-    - libgomp  # [linux]
-    - llvm-openmp  # [osx]
   host:
-    - libblas
-    - liblapack
     - libxc
   run:
     - sktools ={{ version }}
--- a/recipes/skprogs/meta.yaml
+++ b/recipes/skprogs/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     - liblapack
     - libxc
   run:
-    - sktools
+    - sktools ={{ version }}
 
 test:
   commands:
--- a/recipes/skprogs/meta.yaml
+++ b/recipes/skprogs/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "skprogs" %}
+{% set version = "0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/dftbplus/skprogs/archive/refs/tags/{{ version }}.tar.gz
+  sha256: e63e6a38f920bbb85f1f8fdff2da5ec195a3f2ded354f6f09cf68cf69785b59d
+  patches:
+    - no-python.patch
+
+build:
+  number: 0
+  skip: true  # [win or osx]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('fortran') }}
+    - cmake
+    - ninja
+    - pkg-config
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+  host:
+    - libblas
+    - liblapack
+    - libxc
+  run:
+    - sktools
+
+test:
+  commands:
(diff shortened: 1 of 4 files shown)


awvwgk:
@aradi Let me know if you want to be a co-maintainer of this recipe.

aradi:
@awvwgk Sure, I would be happy to join as package maintainer :smile:

recipes/sktools/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/dftbplus/skprogs/archive/refs/tags/{{ version }}.tar.gz
+  sha256: e63e6a38f920bbb85f1f8fdff2da5ec195a3f2ded354f6f09cf68cf69785b59d
+
+build:
+  number: 0
+  noarch: python
```
carterbox, line 14: This is not a noarch package because it uses the scripts argument in setup.py.

After switching to non-noarch you may need to patch the install so the scripts are installed to the correct place for windows environments. I'm not sure.

https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=noarch#noarch-python

recipes/sktools/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  run:
+    - numpy
+    - python >=3.6
+
+test:
+  imports:
+    - {{ name }}
+    - {{ name }}.hsd
+    - {{ name }}.calculators
+    - {{ name }}.skgen
```
carterbox, line 30: Are you planning to change the package name? Don't need templates for constant values.

recipes/skprogs/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - ninja
+    - pkg-config
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+  host:
+    - libblas
+    - liblapack
+    - libxc
+  run:
+    - sktools
```
carterbox, line 32:
```suggestion
    - sktools ={{ version }}
```

recipes/skprogs/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - {{ compiler('fortran') }}
+    - cmake
+    - ninja
+    - pkg-config
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+  host:
+    - libblas
+    - liblapack
+    - libxc
```
carterbox, line 30: Please check the [build logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=519794&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=2363). It may be that blas is not used or that it is being statically linked instead of dynamically linked.

carterbox (changes requested):

recipes/sktools/meta.yaml
awvwgk: The consuming package is currently not Windows compatible, I'll try to get upstream there, but no promises.

recipes/sktools/meta.yaml
awvwgk: Package name is likely to stay, if preferred I can add the explicit name here.

recipes/skprogs/meta.yaml
awvwgk: Looks like upstream is overdepending on LAPACK/BLAS. Maybe @aradi can comment on the necessity of having a linear algebra library around.

recipes/sktools/meta.yaml
awvwgk: Raised in https://github.com/dftbplus/skprogs/issues/35

awvwgk:
@conda-forge/help-c-cpp Review comments have been addressed.

PR #19267, opened 2022-06-11
Adding Anuga

stoiver:
@conda-forge/help-python-c this package needs to use the mingw compiler on windows. I tried to setup the build section of the meta.yaml to use m2w64-toolchain and libpython on windows. Hope it works. 

Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -21,6 +21,8 @@ requirements:
     - {{ compiler('cxx') }}      # [not win]
     - m2w64-gcc                  # [win]
     - libpython                  # [win]
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
 
   host:
     - python 
@@ -31,7 +33,6 @@ requirements:
 
   run:
     - python
-    - _openmp_mutex   # [linux]
     - {{ pin_compatible('numpy') }}
     - dill
     - future
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -53,12 +53,15 @@ requirements:
     - python
     - m2w64-gcc-libs  # [win]
     - libpython       # [win]
+    - libstdcxx-ng    # [linux]
+    - libgcc-ng       # [linux]
+    - _openmp_mutex   # [linux]
     - cython
     - dill
     - future
     - gdal
     - gitpython
-    - matplotlib
+    - matplotlib-base
     - meshpy
     - netcdf4
     - numpy
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -22,6 +22,7 @@ requirements:
     - m2w64-gcc                  # [win]
     - libpython                  # [win]
     - cython
+    - pip
 
   host:
     - python 
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -21,12 +21,12 @@ requirements:
     - {{ compiler('cxx') }}      # [not win]
     - m2w64-gcc                  # [win]
     - libpython                  # [win]
-    - cython
-    - pip
 
   host:
     - python 
     - libpython  # [win]
+    - pip
+    - cython
 
   run:
     - python
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "anuga" %}
-{% set version = "3.1.7" %}
+{% set version = "3.1.8" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/anuga-{{ version }}.tar.gz
-  sha256: 45b213438efc26dfc296347b98087fc8fbcde92b8d2d7ff11f1ec14c241c9a80
+  sha256: 1f15690caf02ed9c45d088536bee22d10a41d1954077e884b3cdcc361d422904
 
 build:
   noarch: python
--- a/recipes/anuga/meta.yaml
+++ b/recipes/anuga/meta.yaml
@@ -37,7 +37,6 @@ requirements:
     - dill
     - future
     - gdal >=3.0.4
-    - gitpython
     - matplotlib-base
     - meshpy
     - netcdf4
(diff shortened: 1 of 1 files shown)


stoiver:
@conda-forge/help-python-c This recipe should be ready for review.

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/anuga-{{ version }}.tar.gz
+  sha256: {{ sha }}
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: True  # [py<3.6]
+
+requirements:
```
carterbox, line 18: It looks like there are too many dependencies in each requirements sub-section.

The build section is for build tools like compilers, cmake, etc. These packages must match the host platform when cross compiling.

The host section is for compiled libraries and python build requirements like python, setuptools, libpython, sometimes numpy (if compiled against their c API).

The run section is for packages only needed at runtime like scipy, matplotlib. Cython is rarely needed here.

Packages only needed to run the tests, go in a separate test/requires section. i.e. pytest

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - setuptools
+    - utm
+    - wheel
+    - pip
+  run:
+    - python
+    - m2w64-gcc-libs  # [win]
+    - libpython       # [win]
+    - libstdcxx-ng    # [linux]
+    - libgcc-ng       # [linux]
```
carterbox, line 57:
```suggestion
```
These are added automatically as run_exports from the compilers, and many times the cxx standard library isn't even used.

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - Pmw
+    - pymetis
+    - pytest
+    - pytz
+    - scipy
+    - setuptools
+    - utm
+    - wheel
+    - pip
+  run:
```
carterbox, line 52: These run dependencies should have version specifications that match the metadata in the setup.py.

carterbox (changes requested):
This package is using an uncommon layout and build system. Thanks for you patience in advance.

recipes/anuga/meta.yaml
stoiver: @carterbox  thanks. I have tried to minimize the requirements as suggested. Just pushed changes so hopefully I have left all the requirements.

recipes/anuga/meta.yaml
stoiver: @carterbox  removed as suggested.

recipes/anuga/meta.yaml
stoiver: @carterbox I ended up removing these from host requirements.

stoiver:
@conda-forge/help-python-c This recipe should be ready for review.

recipes/anuga/meta.yaml
carterbox: The packages in requirements/run should have version specifications which match upstream.

https://github.com/anuga-community/anuga_core/blob/c657e1e41900bcf47f568b2b9bf75892bb46430a/setup.py#L51-L71

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - python
+    - _openmp_mutex   # [linux]
+    - dill
+    - future
+    - gdal
+    - gitpython
+    - matplotlib-base
+    - meshpy
+    - netcdf4
+    - numpy
```
carterbox, line 42:
```suggestion
    - {{ pin_compatible('numpy') }}
```
The numpy version at compile time should match the numpy version at runtime.

carterbox (changes requested):

stoiver:
@conda-forge/help-python-c This recipe should be ready for review.

@carterbox, I changed the numpy requirement to {{ pin_compatible('numpy') }} as suggested. 

I actually went back to the setup.py include requirements and removed most of the specific version constraints (leaving just a constraint on gdal >=3.0.4).

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  host:
+    - python 
+    - libpython  # [win]
+    - pip
+    - cython
+    - numpy
+
+  run:
+    - python
+    - _openmp_mutex   # [linux]
```
carterbox, line 34:
```suggestion
  build:
    - {{ compiler('c') }}        # [not win]
    - {{ compiler('cxx') }}      # [not win]
    - m2w64-gcc                  # [win]
    - libpython                  # [win]
    - llvm-openmp  # [osx]
    - libgomp      # [linux]

  host:
    - python 
    - libpython  # [win]
    - pip
    - cython
    - numpy

  run:
    - python
```

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - dill
+    - future
+    - gdal >=3.0.4
+    - gitpython
+    - matplotlib-base
+    - meshpy
+    - netcdf4
+    - Pmw
+    - pymetis
+    - pytest
```
carterbox, line 45:
```suggestion
```
Why does your package require pytest at runtime? pytest is ususally for development only.

recipes/anuga/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - numpy
+
+  run:
+    - python
+    - _openmp_mutex   # [linux]
+    - {{ pin_compatible('numpy') }}
+    - dill
+    - future
+    - gdal >=3.0.4
+    - gitpython
```
carterbox, line 39:
```suggestion
```
gitpython also seems like development tool that may not be needed for the main functionality of modeling floods.

recipes/anuga/meta.yaml
carterbox: Just want to double check: Does your compiled code link against the numpy C API?

carterbox (changes requested):

recipes/anuga/meta.yaml
stoiver: @carterbox, we are using the test  structure provided by numpy which based on pytest. And in addition we use pytest to skip some of those tests. With pytest we can run the tests via  `import anuga; anuga.test()`. So should keep `pytest` in the run requirements.

stoiver:
@conda-forge/help-python-c This recipe should be ready for review.

@carterbox, regarding "Does your compiled code link against the numpy C API?" We use cython to link between our low level C (and in one case C++) code and python. I don't think we now have any explicit use of the numpy C API, other than through the C code created by cython.

PR #19268, opened 2022-06-11
ENH check for core too

beckermr:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/.github/workflows/automate-review-labels.yml
+++ b/.github/workflows/automate-review-labels.yml
@@ -94,12 +94,24 @@
                 team_slug: 'staged-recipes',
                 username: context.payload.sender.login,
               });
+              const core_membership = github.rest.teams.getMembershipForUserInOrg({
+                org: 'conda-forge',
+                team_slug: 'core',
+                username: context.payload.sender.login,
+              });
               return membership.then((result) => {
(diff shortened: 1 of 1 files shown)


carterbox (commented):
This PR adds core team members to the list of accounts allowed to dismiss review-requested tags. Implementation makes sense. Thanks for your time!

carterbox:
So actually, this doesn't work when the person is not in the core team.

https://github.com/conda-forge/staged-recipes/actions/runs/2484529299

carterbox:
I suspect it is because the async handling of one response is inside the other. Instead the result checking should be flat. Or they should be checked together instead of nested. If the result is not checked but returns an error state, then the error probably bubbles up and that is the cause of the error above.

i.e. Because I am a member of staged recipes, but not core the 404 error from core membership is not handled before the function tries to return.

beckermr:
Weird. I added a handler to both. I'll back out the changes in a bit.  Sorry for the trouble.

carterbox:
I tried some things too, but they didn't work. Maybe we should debug this issue over on the test repo. We just need to figure out how to catch those 404s while having multiple team checks.

beckermr:
Sure. I don't mind debugging live in staged recipes personally.

PR #19295, opened 2022-06-14
Adding mfront, code_aster

ldallolio:
Checklist
- [X] Build number is 0.


recipes/code-aster/conda_build_config.yaml
```diff
@@ -0,0 +1,42 @@
+boost:
```
carterbox, line 1: The conda-forge channel has a [repository of package pinnings ](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml) that is loaded into our build environment automatically. You should not be defining your own build variants with a conda_build_config,yaml unless you are adding variants that are not defined in the conda-forge pinnings.

recipes/mfront/conda-forge.yml
```diff
@@ -0,0 +1,5 @@
+conda_forge_output_validation: true
+os_version:
+  linux_64: cos7
+win:
+  enabled: False
```
carterbox, line 5:
```suggestion
```
You shouldn't need this. You are already using the skip statement.

recipes/mfront/build.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# https://docs.conda.io/projects/conda-build/en/latest/resources/compiler-tools.html#an-aside-on-cmake-and-sysroots
+
+#export LDFLAGS="--sysroot ${CONDA_BUILD_SYSROOT} -L${CONDA_BUILD_SYSROOT}/lib64 -lm -lpthread -L${CONDA_BUILD_SYSROOT}/usr/lib64 -lrt -ldl -L${PREFIX}/lib -lz -lgomp"
+#export CXXFLAGS="${CXXFLAGS} -isysroot ${CONDA_BUILD_SYSROOT} -I${PREFIX}/include"
+mkdir build -p
+cd build
+#echo "boost include dir : "
+#ls $PREFIX/include/boost 
+cmake .. -DCMAKE_TOOLCHAIN_FILE="${RECIPE_DIR}/cross-linux.cmake" -DCMAKE_BUILD_TYPE=Release -Dlocal-castem-header=ON -Denable-fortran=ON -Denable-aster=ON -Denable-cyrano=ON -DPython_ADDITIONAL_VERSIONS=${CONDA_PY} -Denable-python=ON -Denable-python-bindings=OFF -Denable-broken-boost-python-module-visibility-handling=ON -DCMAKE_INSTALL_PREFIX=$PREFIX
```
carterbox, line 14:
```suggestion
cmake .. -DCMAKE_TOOLCHAIN_FILE="${RECIPE_DIR}/cross-linux.cmake" -DCMAKE_BUILD_TYPE=Release -Dlocal-castem-header=ON -Denable-fortran=ON -Denable-aster=ON -Denable-cyrano=ON -DPython_ADDITIONAL_VERSIONS=${CONDA_PY} -Denable-python=ON -Denable-python-bindings=OFF -Denable-broken-boost-python-module-visibility-handling=ON -DCMAKE_INSTALL_PREFIX=$PREFIX ${CMAKE_ARGS}
```

recipes/code-aster/conda-forge.yml
```diff
@@ -0,0 +1,5 @@
+conda_forge_output_validation: true
+os_version:
+  linux_64: cos7
+win:
+  enabled: False
```
carterbox, line 5:
```suggestion
```

recipes/code-aster/build.sh
```diff
@@ -0,0 +1,78 @@
...
+#cp $PREFIX/lib/libboost_python${CONDA_PY}.so.* $PREFIX/lib/libboost_python3.so
+#ls -l $PREFIX/lib/libboost*.so
+
+#export FC=/usr/bin/gfortran
+#export CC=/usr/bin/gcc
+#export CXX=/usr/bin/g++
+
+#export CFLAGS="${CFLAGS} -i sysroot ${CONDA_BUILD_SYSROOT}"
+#echo "CFLAGS=\"${CFLAGS} -i sysroot ${CONDA_BUILD_SYSROOT}\"" >> setup.cfg
+sed -i 's@cmake\ ..\ @cmake\ ..\ -DCMAKE_TOOLCHAIN_FILE='"${RECIPE_DIR}"'/cross-linux.cmake\ @g' products.py
```
carterbox, line 18:
```suggestion
sed -i 's@cmake\ ..\ @cmake\ ..\ ${CMAKE_ARGS} -DCMAKE_TOOLCHAIN_FILE='"${RECIPE_DIR}"'/cross-linux.cmake\ @g' products.py
```

recipes/code-aster/LICENSE.txt
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2015-2022, conda-forge
```
carterbox, line 1: This is supposed to be the license of the package (which is not owned by conda-forge). Often, the license is included in the source tarball, so you usually don't need to add it manually because conda-build looks for the license AFTER unpacking the source.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+{% set version = "14.6.0" %}
+#{% set sha256 = "2afa430a588dc690c7797e3b87de67742cd19a92c82bcb715b007709b0983131" %}
+{% set sha256 = "dcb3900de1e5c0624060253660a63512ab415cbe143cdd87867a0276ef6bf233" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}_{{ version }}.tar.gz
```
carterbox, line 11:
```suggestion
```
Usually specifying the file name is not necessary for direct downloads.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - flex 
+    - swig 
+    - gettext  
+    - tk   
+    - boost {{ boost }}
+    - python {{ python }}
+    - numpy {{ numpy }}
+    - mfront 3.2.1
+    #- metis=5.1.0    
+    #- mumps-seq=5.1.2
```
carterbox, line 45: Please clean unexplained commented out dependencies and other lines from your recipes.

carterbox (changes requested):

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+{% set sha256 = "dcb3900de1e5c0624060253660a63512ab415cbe143cdd87867a0276ef6bf233" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}_{{ version }}.tar.gz
+  #url: https://www.code-aster.org/FICHIERS/aster-full-src-14.6.0-1.noarch.tar.gz # Downloaded bytes did not match Content-Length
+  url: https://collegamenti.s3.eu-west-1.amazonaws.com/aster-full-src-{{ version }}-1.noarch.tar.gz
```
carterbox, line 13: How did you find this url? It doesn't seem like an official release.

carterbox (changes requested):

recipes/code-aster/meta.yaml
ldallolio: I could have used (and would have liked to) the official release url (the commented one), this is where I downloaded the software. The main issue is that when using build-locally it did not download in a stable way : the package is quite big, the download timeouts are a bit too strict. So I downloaded it and uploaded it into s3 to make it stable. If you have any way to suggest that I can use to make sure that it is downloaded safely I gladly get rid of the s3 link ?

recipes/code-aster/LICENSE.txt
ldallolio: I agree with that, I also prefer to use the one inside the package. In this case, the official archive is an "archive of archives" (the license file is inside a second layer of archive) so the magic to find the license inside will not work.

recipes/code-aster/conda_build_config.yaml
ldallolio: Thanks, I did not know about that. I am wondering if this can work as expected : I know that this complex software will only work using a gcc/gfortran 8 version (and not yet on later ones), only with some hdf5 versions... Should I put these constraints in the meta.yaml file ?

recipes/code-aster/LICENSE.txt
ldallolio: I just tested it again : the exact error message is "conda.CondaError: Downloaded bytes did not match Content-Length". Of course, when downloaded manually, the hash is the same as provided in alternative link.

recipes/code-aster/meta.yaml
ldallolio: I just tested it again : the exact error message is "conda.CondaError: Downloaded bytes did not match Content-Length". Of course, when downloaded manually, the hash is the same as provided in alternative link.

carterbox: I provided the URL for direct download from their official sourceforge

recipes/code-aster/LICENSE.txt
carterbox: I deleted this license file because it is not the correct type of license file and the copyright holder is listed as conda-forge.

carterbox (changes requested):
Please use the snapshot from the project's official sourceforge which I have located for you. If you need to update to future versions you can update this url using the commit hash of the new release.

recipes/code-aster/meta.yaml
ldallolio: Thanks for this :-) I found the new official link (the project moved from sourceforge to gitlab) and used that one

ldallolio:
Hi, I think I have integrated most if not all remarks (thanks for those : the recipe is much better now), could you please review this again? Thank you very much!

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - {{ pin_subpackage('code-aster', max_pin='x.x') }}
+    - {{ pin_compatible('python', max_pin='x.x') }}
+requirements:
+  build:
+    - gcc 8  # [linux]
+    - gxx 8  # [linux]
+    - gfortran 8  # [linux]
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+    - sysroot_linux-64 =2.17  # [linux]
```
carterbox, line 46: Please see comments on mfront about compilers and sysroot.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+  host:
+    - cmake  
+    - make
+    - bison 
+    - flex 
+    - swig 
+    - python 3.8
+    - numpy {{ numpy }}
```
carterbox, line 54:
```suggestion
    - numpy
```

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - gfortran 8  # [linux]
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+    - sysroot_linux-64 =2.17  # [linux]
+  host:
+    - cmake  
+    - make
+    - bison 
+    - flex 
+    - swig
```
carterbox, line 52: cmake, make, bison, swig?, flex? Should be listed in the requirements/build section because code-aster doesn't link to any libraries in them (they should match the build platfrom not a target platform when cross-compiling).

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+build:
+  number: 0
+  skip: true  # [py>=39 or osx or win or (python_impl == 'pypy')]
+  detect_binary_files_with_prefix: True
+  run_exports:
+    - gcc 8  # [linux]
+    - gxx 8  # [linux]
+    - gfortran 8  # [linux]
+  ignore_run_exports:
+    - numpy
```
carterbox, line 35: Why would numpy be needed at build time, but not also run time? If you're linking against the NumPy C API, then wouldn't you want numpy available at run time?

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+    - sysroot_linux-64 =2.17  # [linux]
+  host:
+    - cmake  
+    - make
+    - bison 
+    - flex 
+    - swig 
+    - python 3.8
```
carterbox, line 53:
```suggestion
    - python
```
Use the skip statement to specify a python version.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+  detect_binary_files_with_prefix: True
+  run_exports:
+    - gcc 8  # [linux]
+    - gxx 8  # [linux]
+    - gfortran 8  # [linux]
+  ignore_run_exports:
+    - numpy
+  run_exports:
+    - {{ pin_subpackage('code-aster', max_pin='x.x') }}
+    - {{ pin_compatible('python', max_pin='x.x') }}
```
carterbox, line 38:
```suggestion
```
This should not be needed for reasons explained in the mfront recipe comments.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+    - as_run --info 
+    - homard || true  # [unix]
+    - as_run --test forma01a
+    - as_run --test sslp114a
+    - as_run --test mumps02a
+    - as_run --test mumps01a
+    - as_run --test umat001a
+    - as_run --test zzzz121a
+    #- as_run --test mfron01a
+    - test -f $PREFIX/include/aster/aster.h
```
carterbox, line 93: There's lot of python scripts and a shared library being installed in `$PREFIX/lib64/aster`. We don't use `$PREFIX/lib64` because it is not added to the conda library path. Probably you want to move these to `$PREFIX/lib`.

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+    folder: homard
+    
+build:
+  number: 0
+  skip: true  # [py>=39 or osx or win or (python_impl == 'pypy')]
+  detect_binary_files_with_prefix: True
+  run_exports:
+    - gcc 8  # [linux]
+    - gxx 8  # [linux]
+    - gfortran 8  # [linux]
```
carterbox, line 33: If code-aster needs compilers at runtime, just add the compilers to the run requirements. I can't think of an example where a package would export a dependency on compilers.

recipes/code-aster/build.sh
```diff
@@ -0,0 +1,75 @@
...
+mkdir -p $PREFIX/etc/codeaster/
+cp -Rf $RECIPE_DIR/contrib/etc/* $PREFIX/etc/codeaster/
+
+echo "**************** A S T E R  B U I L D  E N D S  H E R E ****************"
+
+echo "**************** C L E A N U P  S T A R T S  H E R E ****************"
+
+rm -Rf $PREFIX/metis-aster
+rm -Rf $PREFIX/mumps-aster
+rm -Rf $PREFIX/homard
```
carterbox, line 73: Metis, mumps, and homard are just build dependencies?

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+{% set minor_version = "11" %}
+{% set sha256 = "107ed98d10bc05866bdb3053f088fb0f51276ea14f7b615ec0d410a4b147ef12" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - fn: {{ name }}_{{ version }}.{{ minor_version }}.tar.gz
+    url: https://gitlab.com/codeaster/src/-/archive/{{ version }}.{{ minor_version }}/src-{{ version }}.{{ minor_version }}.tar.gz
```
carterbox, line 12:
```suggestion
{% set name = "code-aster" %}
{% set version = "14.6.11" %}
{% set sha256 = "107ed98d10bc05866bdb3053f088fb0f51276ea14f7b615ec0d410a4b147ef12" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  - fn: {{ name }}_{{ version }}.tar.gz
    url: https://gitlab.com/codeaster/src/-/archive/{{ version }}/src-{{ version }}.tar.gz
```
Why would you split the version into two variables? The package version should be the full version.

carterbox (changes requested):
There should not be `.circleci` or `.scripts` folders in your recipes directories.

I have also been having trouble with conda-build 3.24, so I understand why you have edited outside the recipes folder.

recipes/code-aster/meta.yaml
ldallolio: I am still working on this but I think I have one real life case : mfront can behave like a code generator "on the fly", the user provides the mathematical model written in a domain specific language and it will generate a high performance code that will be compiled at runtime and dynamically linked. This model can be used to expand Code_Aster (and other simulation tools) with user defined material laws.

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+  url: https://github.com/thelfer/tfel/archive/refs/tags/TFEL-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  skip: true  # [py2k or osx or win or (python_impl == 'pypy')]
+  detect_binary_files_with_prefix: True
+  run_exports:
+    - {{ pin_subpackage('mfront', max_pin='x.x.x') }}
+    - {{ pin_compatible('python', max_pin='x.x') }}
```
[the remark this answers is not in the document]
ldallolio: I can double check but I had some issues locally when downstream packages switch to a different python version : there are some boost::python bindings that must be linked and are built for a specific python version (same thing for aster) so the result is python specific if I am not wrong.

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
+{% set name = "mfront" %}
```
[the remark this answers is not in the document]
ldallolio: tfel is the initial name (from the first letters of the author's name) but is widely known as mfront in the community (I suppose that T. Helfer did not bother renaming the package), you can see this change in the site name and title if you want : https://tfel.sourceforge.net/ I am not against renaming it but I think people would look up in conda using "mfront" as a keyword (I would)

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - $PREFIX/bin/tfel-check --version
+    - $PREFIX/bin/mtest --version
+    - $PREFIX/bin/mfront --list-stress-criteria
+    - $PREFIX/bin/mfront-doc --version
+    - $PREFIX/bin/mfront-query --version
+    - $PREFIX/bin/mfm --version
+    - test -f $PREFIX/lib/libTFELConfig.so  # [linux]
+    - test -f $PREFIX/lib/libTFELPhysicalConstants.so  # [linux]
+    - test -f $PREFIX/lib/libTFELException.so  # [linux]
+    - test -f $PREFIX/lib/libTFELUtilities.so  # [linux]
```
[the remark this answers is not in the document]
ldallolio: Yes the python bindings are optional (as an example, Code_Aster would not need them since it binds directly to the C++ / fortran interface)

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+  detect_binary_files_with_prefix: True
+  run_exports:
+    - {{ pin_subpackage('mfront', max_pin='x.x.x') }}
+    - {{ pin_compatible('python', max_pin='x.x') }}
+
+requirements:
+  build:
+    - gcc 8  # [linux]
+    - gxx 8  # [linux]
+    - gfortran 8  # [linux]
```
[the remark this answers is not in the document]
ldallolio: Indeed I could not compile this aster version using more recent compilers, so I kept these ones for mfront since aster will link at mfront. I am confident that newer versions will work on more recent compilers and I will remove this constraint (some projects I plan to submit need this specific version)

recipes/code-aster/meta.yaml
ldallolio: I wrote something about this in the mfront comment too : Code_Aster will build python-related libraries (.so) and in the past I had some trouble when changing python version, since the libraries would stay in ../lib/python3.8/site-packages/libXXX.so and that would be incompatible with the (let's say) runtime python 3.9 chosen by downstream and also not in the library path. I will try to understand if it works without that pin but I think it would be better for downstream to pick the aster library built for the runtime python

recipes/mfront/build.sh
```diff
@@ -0,0 +1,31 @@
...
+         -Denable-fortran=ON \
+         -Denable-aster=ON \
+         -Denable-cyrano=ON \
+         -DPython_ADDITIONAL_VERSIONS=${CONDA_PY} \
+         -DPYTHON_INCLUDE_DIRS=${PREFIX}/include \
+         -DCOMPILER_CXXFLAGS="-I${PREFIX}/include" \
+         -Denable-python=ON \
+         -Denable-python-bindings=ON \
+         -Denable-portable-build=ON \
+         -DCMAKE_INSTALL_PREFIX=$PREFIX \
```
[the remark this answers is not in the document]
ldallolio: I totally agree, it was bothering me too. Added in a project specific cmake variable (the only one that seems to have an effect), thanks for pointing that out!

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+  host:
+    - cmake  
+    - make  # [not win]
+    - ninja  # [win]
+    - boost
+    - python {{ python }}
+  run:
+    - python
+    - {{ pin_compatible('boost', max_pin='x.x') }}
+    - libzlib  # [linux]
```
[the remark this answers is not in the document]
ldallolio: I think I added because conda was complaining of a missing run dependency (overlinking error if I am not mistaken), I will try to remove and see what happens. It might also be caused by some "standard" build (using "mamba build" directly to save time instead of "build-locally"), I will see what I can do about this :)

recipes/code-aster/meta.yaml
ldallolio: I think I fixed this one (I found the magic option and I do not see any lib64 in the output anymore :)

recipes/mfront/meta.yaml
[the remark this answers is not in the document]
ldallolio: Yes I tried and it breaks because of overlinking. I think (might be wrong) that it picks zlib from sysroot then complains because of the missing dependency, maybe I should also add it to the host dependencies ?

recipes/code-aster/meta.yaml
ldallolio: Fixed :-)

recipes/mfront/meta.yaml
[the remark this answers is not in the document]
ldallolio: Fixed :-)

recipes/mfront/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+    - gfortran 8  # [linux]
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]    
+    - sysroot_linux-64 =2.17  # [linux]
+  host:
+    - cmake  
+    - make  # [not win]
+    - ninja  # [win]
+    - boost
+    - python {{ python }}
```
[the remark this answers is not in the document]
ldallolio: Fixed :-)

recipes/mfront/meta.yaml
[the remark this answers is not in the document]
ldallolio: Added to host dependencies too

recipes/code-aster/meta.yaml
ldallolio: Fixed :-)

recipes/code-aster/meta.yaml
ldallolio: Fixed, thanks for pointing that out

recipes/code-aster/meta.yaml
ldallolio: I am not totally sure that it links against the numpy C API, it does check for it at build time, it might be used by the python code only ? In any case, I removed the ignore, I agree with you

recipes/code-aster/meta.yaml
ldallolio: Fixed, I had an hard time with skip because py < 38 and py >=38 gave me an empty build but since py = 37 is no longer there it works anyway so I removed the version statement

recipes/code-aster/meta.yaml
ldallolio: I tried and failed : both mfront and code aster need this old version to build. I will try to upgrade newer versions in the future

recipes/miss3d/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  skip: true  # [osx or win]
+  detect_binary_files_with_prefix: True   
+  ignore_run_exports:
+    - liblapack
+    - libblas
+    - libopenblas
```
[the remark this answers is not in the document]
ldallolio: You are right, it was an half baked test with libopenblas, I cleaned it up now :)

ldallolio:
> There should not be `.circleci` or `.scripts` folders in your recipes directories.
> 
> I have also been having trouble with conda-build 3.24, so I understand why you have edited outside the recipes folder.

Fixed this too, as you wrote : I had to use an older conda-build to make it work, I will try to upgrade in the future

ldallolio:

recipes/code-aster/meta.yaml
```diff
@@ -0,0 +1,102 @@
...
+    url: https://gitlab.com/codeaster/src/-/archive/{{ version }}/src-{{ version }}.tar.gz
+    sha256: {{ sha256 }}
+  - url: https://downloads.sourceforge.net/project/prereq/metis-5.1.0_aster4.tar.gz
+    sha256: 1b295f9ce260c19d2d93fb2430be4139496c56f58f74ef7956ccd26c854fcc2e
+    folder: metis-aster
+  - url: http://graal.ens-lyon.fr/MUMPS/MUMPS_5.1.2.tar.gz
+    sha256: eb345cda145da9aea01b851d17e54e7eef08e16bfa148100ac1f7f046cd42ae9
+    folder: mumps-aster
+    patches:
+      - contrib/mumps-aster-diff.patch
```
carterbox, line 20: I see that you have correctly included the licenses for these dependencies, but can you please also write two comments explaining how the aster versions of MUMPS and METIS have been modified such that you cannot use the MUMPS and METIS packages already published on conda-forge? Thank you.

carterbox (changes requested):

carterbox:
Assuming my change to the zlib dependency works, I believe these packages are all setup correctly. I will merge as soon as the comments are added about the need for special builds of MUMPS and Metis.

P.S. The code_aster package is quite large. I encourage you to double check that all of the files which are included in the package are needed and not just development files that are being shipped accidentally.

ldallolio:
I think I removed all files outside the recipe directories, about the package size Code_Aster has always been quite large, I did not see any build files  in the package. The only thing I could remove are the examples but they are really used as a documentation and starting points by engineers using the software so I believe they should stay into the package...

carterbox:

carterbox:
LGTM! Thanks for you patience while we got this sorted, @ldallolio. 🚀

PR #19296, opened 2022-06-14
Add libhdbpp recipe

beenje:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/libhdbpp/meta.yaml
+++ b/recipes/libhdbpp/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "libhdbpp" %}
+{% set version = "2.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.com/tango-controls/hdbpp/libhdbpp/-/archive/v{{ version }}/libhdbpp-v{{ version }}.tar.gz
+  sha256: 53e1647d0106787adaed5d9e82c3eb53c5d5b1a4892387b1e52a183fadde4b93
+
+build:
+  number: 0
+  skip: true  # [not linux]
+
+requirements:
+  build:
+    - cmake
+    - make
+    - {{ compiler('cxx') }}
+    - libtool
+    - pkg-config
+  host:
+    - cppzmq
+    - cpptango
+    - omniorb
+
+test:
+  commands:
+    - test -f ${PREFIX}/lib/libhdb++${SHLIB_EXT}
+    - test -f ${PREFIX}/lib/libhdb++${SHLIB_EXT}.${PKG_VERSION}
+    - test -f ${PREFIX}/include/hdb++/HdbClient.h
+    - test -f ${PREFIX}/include/hdb++/AbstractDB.h
+
+about:
+  home: https://www.tango-controls.org
+  license: LGPL-3.0-or-later
+  license_file: LICENSE
+  summary: Interface library for HDB++ archiving system.
+  description: |
+    HDB++ is an archiving system for Tango-Controls.
+    libhdbpp provides an abstract means to archive data events to a Cassandra database,
+    MySQL database, or potentially other database backends.
+  dev_url: https://gitlab.com/tango-controls/hdbpp/libhdbpp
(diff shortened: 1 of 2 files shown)


beenje:

recipes/libhdbpp/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.com/tango-controls/hdbpp/libhdbpp/-/archive/v{{ version }}/libhdbpp-v{{ version }}.tar.gz
+  sha256: 53e1647d0106787adaed5d9e82c3eb53c5d5b1a4892387b1e52a183fadde4b93
+
+build:
+  number: 0
+  skip: true  # [not linux]
```
carterbox, line 14: Looks like upstream supports linux and windows.

```suggestion
  skip: true  # [osx]
```

recipes/libhdbpp/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+requirements:
+  build:
+    - cmake
+    - make
+    - {{ compiler('cxx') }}
+    - libtool
+    - pkg-config
+  host:
+    - cppzmq
+    - cpptango
```
carterbox, line 25: The [logs shows](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=521716&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1341) that you are probably statically linking against cpptango.

carterbox (changes requested):

recipes/libhdbpp/meta.yaml
beenje: The `cpptango` package only contains a shared library and some headers.
Looks like `libhdbpp` is only creating an interface library: https://gitlab.com/tango-controls/hdbpp/libhdbpp/-/blob/master/cmake/FindTango.cmake#L15, so it's not linking against it.

`cpptango` is needed to compile this library and it defines a `run_exports`. Should I ignore this `run_exports`?
`libhdbpp` is an interface library and will be used to compile other packages.

beenje:
@bourtemb do you have additional comments?

recipes/libhdbpp/meta.yaml
carterbox: If there are no linkage/symbols from libhdb++.so to/in libcpptango.so, then it is not a runtime dependency of this library and the run_export should be ignored.

I'm not sure what you mean by interface library. If downstream packages need to link against cpptango, they should explicitly add it to their recipe host requirements.

recipes/libhdbpp/meta.yaml
carterbox: pkg-config is available for Windows? CMake does not require platform specific directions. Does pkg-config require platform specific directions? I don't know; I've never used pkg-config.

recipes/libhdbpp/meta.yaml
beenje: OK, thanks. `run_exports` ignored.

recipes/libhdbpp/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  url: https://gitlab.com/tango-controls/hdbpp/libhdbpp/-/archive/v{{ version }}/libhdbpp-v{{ version }}.tar.gz
+  sha256: 53e1647d0106787adaed5d9e82c3eb53c5d5b1a4892387b1e52a183fadde4b93
+
+build:
+  number: 0
+  skip: true  # [not linux]
+  # Library doesn't link against cpptango (only headers used)
+  ignore_run_exports:
+    - cpptango
+
```
carterbox, line 18:
```suggestion
  run_exports:
    - {{ pin_subpackage('libhdbpp') }}

```
People that build against this library may not need cpptango, but they definitely need this library? I assume this project uses SEMVER?

carterbox (changes requested):

recipes/libhdbpp/meta.yaml
beenje: Good point. Thanks! Added.

PR #19298, opened 2022-06-14
Release Pycrostates

vferat:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pycrostates/meta.yaml
+++ b/recipes/pycrostates/meta.yaml
@@ -5,19 +5,21 @@ package:
   name: {{ name|lower }}
   version: {{ version }}
 
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pycrostates-{{ version }}.tar.gz
+  sha256: 8a9292d5f7cae08ac7aaf492460577305fb41ae05a08d0f79f5aaabbae411a43
+
 build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
-source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pycrostates-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


vferat:
@mscheltienne

vferat:
Willing to maintain this package

mscheltienne:
I am willing to maintain this package.

vferat:
@conda-forge/help-python ready for review, not sure what is going wrong, looks like a problem with the mffpy feedstock.

recipes/pycrostates/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - setuptools >=61.0.0
+    - wheel
+  run:
+    - decorator
+    - importlib-metadata
+    - importlib-resources
+    - jinja2
+    - joblib
+    - matplotlib-base
+    - mne >=1.0.0
```
mscheltienne, line 30:
```suggestion
    - mne-base >=1.0.0
```

The issue is definitely coming from `mffpy` dependencies.. and we don't need it.
So let's pin to `mne-base` instead and simplify the installation since we anyway pin the extra mne dependencies that we are using: `joblib`, `scikit-learn`; in our req.

mscheltienne:

hoechenberger:
> @vferat It's up to you if you want to depend on mne or mne-base.

Why is that optional though? Do or do you not need the "full" MNE functionality? :)

The upstream package depends on "mne" from PyPI, which is identical to "mne-base" here.

So why would you want to pull in all the additional stuff of the "mne" conda-forge package?

hoechenberger:
> Ok, good! Everything that your package imports itself should be explicitly listed as a dependency, even if it's already a dependency of another dependency.

I just read this sentence again and found it confusing.

What I mean is if your package does e.g.
```shell
import numpy as np
import mne
```
then both, `numpy` and `mne`, should be listed as dependencies, even though we know that `mne` already depends on `numpy` itself.

Hope this is a bit clearer now

PR #19300, opened 2022-06-14
DOC: Add note about static libraries [ci skip]

carterbox:
Add a note to the PR checklist about explicitly excluding static libraries. CFEPs are located separately from maintainer documentation, so they are less likely to be noticed.


beenje (approved):
LGTM

carterbox:
Thanks, @beenje!

PR #19324, opened 2022-06-18
Adding mmg, iscd tools

ldallolio:
Checklist
- [X] Build number is 0.


ldallolio:
@conda-forge/staged-recipes : could you please review ? Thank you !

bryanloz-xilinx:
Just a note from stuff I've learned in CMake.

Instead of directly calling a platform's Make or NMake command, you can use cmake platform independent commands.

i.e. instead of:
https://github.com/conda-forge/staged-recipes/pull/19324/files#diff-8685d56953fcb189f1586b6ea5a55849672e0dac80945a3aa4d360028f4e46f8R4

You can do:
https://github.com/Xilinx/XRT/blob/master/build/build.bat#L186

ldallolio:
> Just a note from stuff I've learned in CMake.
> 
> Instead of directly calling a platform's Make or NMake command, you can use cmake platform independent commands.
> 
> i.e. instead of: https://github.com/conda-forge/staged-recipes/pull/19324/files#diff-8685d56953fcb189f1586b6ea5a55849672e0dac80945a3aa4d360028f4e46f8R4
> 
> You can do: https://github.com/Xilinx/XRT/blob/master/build/build.bat#L186
> 
> I think... unless there is something special I don't know about NMake.

Thanks for the advice : I did not know about that and I really like these generalizations, I just switched to the cmake --build and cmake --install syntax :)

recipes/Advection/conda_build_config.yaml
```diff
@@ -0,0 +1,35 @@
+cross_compiler_target_platform:  # [win]
```
carterbox, line 1: `conda_build_config.yaml` is loaded on our CI servers from [this repository](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml), so you don't need to provide it yourself unless you want to add build variants for packages that are not already included in there.

It seems like you are adding many redundant version pinnings and also trying to use older compilers and add packages from outside channels (which is not allowed). Can you explain why you need to overrride so many of our pinnings and use old compilers?

recipes/Advection/meta.yaml
```diff
@@ -0,0 +1,42 @@
+{% set name = "iscd-advection" %}
+{% set version = "1.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}_{{ version }}.tar.gz
+  url: https://github.com/ISCDtoolbox/Advection/archive/refs/heads/master.zip
```
carterbox, line 10: Do not use a git branch (master) for a release. This zip file will change if more commits are added to the branch, so it is not an absolute point of reference. You should not create your own release version from this repository. Please ask the maintainers of these projects to create a tagged release for you.

carterbox (changes requested):

carterbox:
How does `mmgsuite` get the source of / link to iscd tools? git sub_modules? download during the build process?

recipes/Advection/conda_build_config.yaml
ldallolio: I removed the `conda_build_config.yaml` files. The reason behind this is because other packages (such as code_aster) need this libraries and need to be compiled with a specific version. After a lot of trial and error we found out the right "combination" and tried to enforce it for all our dependencies (since conda shares libraries). We did not think of conda-forge and its pinning system, so thanks for the hint.

recipes/Advection/meta.yaml
ldallolio: I totally agree about this. We had a business meeting with the maintainers (university researchers) and tried to explain why it was important to tag the software. They agreed  but... it might take some time. I could fork and tag myself but I suppose that it wouldn't be acceptable since it would not be the official source... Still, we need these dependencies for other packages so, could we go for the "master" package until there is an official tag ? Maybe I could try to get a zip package of a specific commit, would that be better ?

carterbox: I believe it is possible to download a tar/zip via commit hash from GitHub; that would be better. As for the version number of this package, you should choose something like `0.0.0rc0` where you would increment rc0, rc1, rc2 whenever you move to a new commit until they release something official.

Also, write a comment in the recipe explaining that there are no tagged releases which is why you are using archives of specific commits.

ldallolio:
Hi Daniel,

I am unsure to fully understand this question but, in my mind :
- the only relationship between mmgsuite and iscd tools is by conda
dependency mechanism, I am not aware of a download during the build process
or git submodules being used.
- I do not understand why I have to whitelist some libraries (they should
be taken from the conda dependencies...)

So it is not very clear to me what happens : is it related to the conda
forge build system or the mmgsuite code ?

Any help or hint is more than welcome! :)

Il giorno sab 16 lug 2022 alle ore 01:53 Daniel Ching <
***@***.***> ha scritto:

recipes/Advection/meta.yaml
ldallolio: I did as you suggested (thank you for the advice!). The code maintainer provided a few tags that were integrated, for the rest I fixed the 0.0.0rc0 and added the comment in the recipe

ldallolio:
Hi,
After a full-weekend work, I think I fixed the issues with this MR. Could you please review and tell me if I need to do something else?
Thanks! All the best

recipes/NavierStokes/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+source:
+  # there are no tagged releases which so using archives of specific commits
+  url: https://github.com/ISCDtoolbox/NavierStokes/archive/2e14ef5fa4884674b7bbc2212af256652f02912c.zip
+  sha256: 7efe022f2b3ff88001b99c9cb83da5bdd15fb2b2084201662a55e4d666157116
+
+build:
+  number: 0
+  skip: true  # [py<3 or osx or win] # Cannot open include file: 'pthread.h': No such file or directory  
+  ignore_run_exports_from:
+    - {{ compiler('cxx') }}
```
carterbox, line 17:
```suggestion
build:
  number: 0
  skip: true  # [py<3 or osx or win] # Cannot open include file: 'pthread.h': No such file or directory  
  ignore_run_exports_from:
    - {{ compiler('cxx') }}
  run_exports:
    - {{ pin_subpackage('iscd-nstokes', exact=True) }}
```
For the packages which ship libraries and headers (which is all?), you need to include run_exports of the same package. This ensures that downstream packages always depend on the same version that was used to build them. For your packages which do not have official releases, use `exact=True` and for the official releases use `max_pin='x.x.x'`.

recipes/NavierStokes/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  description: |
+    Nstokes is a finite element solver for Stokes and Navier-Stokes incompressible flow problems in two and three dimensions.
+  doc_url: https://github.com/ISCDtoolbox/NavierStokes/wiki
+  dev_url: https://github.com/ISCDtoolbox/NavierStokes 
+
+extra:
+  recipe-maintainers:
+    - ldallolio
+    - cnrdn
+    - fbordeu
```
carterbox, line 50: All maintainers must comment on this PR before it is merged.

recipes/NavierStokes/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - cmake
+    - make  # [not win]
+    - jom  # [win]    
+    - iscd-commons
+  run:
+    - iscd-commons
+
+test:
+  commands:
+    - test -f $PREFIX/bin/nstokes  # [unix]
```
carterbox, line 33: Please add a test for the shared libraries and headers. e.g.

```
- test -f $PREFIX/lib/libNavierStoves.so
- test -f $PREFIX/include/navier.h
```

Because most of your packages are not stable releases, these tests will help prevent releasing future versions of the packages that break their ABI/API.

carterbox (changes requested):
It looks like your packages are building/linking correctly which is good. However, we need to fix up the metadata in the recipes. I have added comments to the `iscd-nstokes` package. Please apply these fixes to all of your packages.

recipes/Advection/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  description: |
+    Advect is a program for solving linear advection problems in two and three dimensions.
+  doc_url: https://github.com/ISCDtoolbox/Advection/tree/master/Documentation
+  dev_url: https://github.com/ISCDtoolbox/Advection 
+
+extra:
+  recipe-maintainers:
+    - ldallolio
+    - cnrdn
+    - fbordeu
```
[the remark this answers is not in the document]
ldallolio: You are right of course, I wrote them to remember to write to agree, in the meantime I will remove them as suggested

recipes/NavierStokes/meta.yaml
ldallolio: Thanks for taking the time to explain the test reason, I added them wherever I could

recipes/NavierStokes/meta.yaml
ldallolio: I wrote them asking to write in the PR, in the meantime I will remove them as suggested, you are right.

recipes/NavierStokes/meta.yaml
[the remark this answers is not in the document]
ldallolio: I did not think of the "or true" strategy, quite useful indeed :)

recipes/NavierStokes/meta.yaml
ldallolio: Hi, I added run_exports to every packages as you suggested. I tried very hard (a few hours of tests) to use `exact=True` but I couldn't : the package could not be found when later I used it as a dependency in `mmg`. The issue seems similar to this one : https://github.com/conda/conda-build/issues/4415 but no workaround seemed to work in my case. It looks like the build will look for a `0` build package number when it should use the hash instead. The only solution I could find was to used the `max_pin='x.x.x'` for all cases. I think (I might be wrong) that the risk of regression is limited. If you have other tests that you wish me to do please let me know and I will do my best :-)

fbordeu:
ok to be recipe-maintainer

recipes/Commons/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('iscd-commons', max_pin='x.x.x') }}
+  skip: true  # [py<3 or win] # Cannot open include file: 'pthread.h': No such file or directory
+  ignore_run_exports_from:
+    - {{ compiler('cxx') }}
+  run_exports:
+    - {{ pin_subpackage('iscd-commons', max_pin='x.x.x') }}
```
carterbox, line 21:
```suggestion
build:
  number: 0
  skip: true  # [py<3 or win] # Cannot open include file: 'pthread.h': No such file or directory
  ignore_run_exports_from:
    - {{ compiler('cxx') }}
  run_exports:
    - {{ pin_subpackage('iscd-commons', max_pin='x.x.x') }}
```
Remove duplicate run_export

recipes/NavierStokes/meta.yaml
carterbox: Thanks for trying to debug that issue.

In this case, we will use `{{ version }}` as the pinning. This will pin the version, but not the build number. Otherwise, by using `max_pin='x.x.x'` we get a `>=0.0.0rc0,<0.1.0` constraint which is not correct because it would allow `0.0.0rc1` at runtime.

recipes/mmg/meta.yaml
```diff
@@ -28,11 +28,11 @@ requirements:
     - jom  # [win]  
     - libscotch
     - iscd-commons
-    - {{ pin_subpackage("iscd-linearelasticity", exact=True) }}    
+    - iscd-linearelasticity  
   run:
     - libscotch
     - iscd-commons
-    - {{ pin_subpackage("iscd-linearelasticity", exact=True) }}
+    - iscd-linearelasticity
```
carterbox: You cannot use `pin_subpackage` here because "iscd-linearelasticity" is not a package that is defined in this recipe. i.e. it is not a subpackage in this file.

You would use `pin_compatible` in `requirements/run` if you want an externally defined dependency to be compatible in the run and host sections.

carterbox:
Also, please ping the team for review, not individuals.

carterbox (approved):
Merge when builds succeed.

PR #19337, opened 2022-06-19
Add xilinx-runtime

bryanloz-xilinx:
Checklist
- [x] Build number is 0.


bryanloz-xilinx:
@conda-forge/help-c-cpp this is ready for review.

bryanloz-xilinx:
I am @bryanloz-xilinx and I am willing to be listed as a maintainer.
In the future I would like to recruit more helpers.

bryanloz-xilinx:
@xhochy I changed source branch to get rid of the history of the 175K patch file. Hence I had to make a new PR.
I listened to your advice regarding the run requirements. With the current run requirements the tests will fail. I also don't expect the build to go through in the CI because of the underlinking/overlinking error.

bryanloz-xilinx:
@xhochy I keep getting these errors in the CI
```
["  ERROR (xilinx-runtime,lib/libxrt_hwemu.so.2.13.0): .. but ['libprotobuf'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)"
```

timsnyder:
I'm happy to be a maintainer on this recipe.  We will eventually use this in https://fires.im/

bryanloz-xilinx:
> I'm happy to be a maintainer on this recipe. We will eventually use this in https://fires.im/

Very impressive work there @timsnyder

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - cmake
+    - make  # [unix]
+    - pkg-config
+    - git
+
+  host:
+    - python
+    - pybind11
+    - boost
+    - protobuf
```
xhochy, line 34: You need to list all native dependencies you link to here. `libprotobuf` is the C++ part of protobuf whereas `protobuf` is the python package.
```suggestion
    - protobuf
    - libprotobuf
```

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - libdrm
+    - clhpp
+    - libcurl
+
+  run:
+    - python
+    - {{ pin_compatible('boost-cpp', max_pin='x.x') }}
+    - {{ pin_compatible('openssl', max_pin='x.x') }}
+    - {{ pin_compatible('ncurses', max_pin='x.x') }}
+    - {{ pin_compatible('libprotobuf', max_pin='x.x') }}
```
xhochy, line 46: These should not be needed. As said, `run_exports` takes care of it.
```suggestion
```

xhochy (changes requested):
See comments.

bryanloz-xilinx:
@xhochy Appreciate your time, and help.
This package takes about an hour to build in CI. Hopefully the package builds okay with the current build requirements.

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - rapidjson
+    - libuuid
+    - libdrm
+    - clhpp
+    - libcurl
+    - openssl
+    - ncurses
+
+  run:
+    - python
```
xhochy, line 46: Ok, `boost-cpp` is a special package and has no run_exports as in most cases, only the headers are used. As it has a `pin_run_as_build` though, adding it as a plain dependency here is sufficient. If you also need Boost.Python, you will also need `boost` listed here.
```suggestion
    - python
    - boost
    - boost-cpp
```

recipes/xilinx-runtime/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/env/bash
+set -e
+
+mkdir -p $SRC_DIR/build-release
+
+cmake \
+  -S $SRC_DIR/src  \
```
xhochy, line 7: Add `$CMAKE_ARGS` to pick up our default arguments.
```suggestion
cmake \
  $CMAKE_ARGS \
  -S $SRC_DIR/src  \
```

recipes/xilinx-runtime/build.sh
```diff
@@ -0,0 +1,20 @@
...
+  -B $SRC_DIR/build-release  \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_PREFIX_PATH=${PREFIX} \
+  -DCMAKE_INSTALL_PREFIX=${PREFIX} \
```
xhochy, line 11: Part of `CMAKE_ARGS`
```suggestion
```

recipes/xilinx-runtime/conda_build_config.yaml
```diff
@@ -0,0 +1,10 @@
+c_compiler:
```
xhochy, line 1: Why is this file needed? Can you add a comment or omit it?

bryanloz-xilinx: Our C++ code has only been verified on gcc9. By default gcc will default to gcc12 in conda-forge. So the purpose of this file is to constrain the compiler version on linux. In the future I imagine it would be used to constrain the compiler version for Windows as well. Is this not the correct approach?

xhochy: We try to build all packages with the same compiler on conda-forge. Sticking to older compilers increases maintenance (not only for the package maintainer) and should only be done if things don't work. In this case, I would suspect that compiling with the default compilers should work out of the box.

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - boost
+    - boost-cpp
+
+test:
+  commands:
```
xhochy, line 51: As this is a python package, is there also a module we could import?

timsnyder:
```suggestion
  imports:
    - pyxrt
  commands:
```

timsnyder (commented):
@bryanloz-xilinx it is going to be great to have this!  I think the only really broken thing I can see is that the `CL/cl_ext_xilinx.h` header needs to be patched or the package won't work if you are trying to use the Khronos' headers from the `ocl-icd` conda package.  The `CL/cl_ext_xilinx.h` header assumes that XRT is installed in it's own separate PREFIX an that the Khronos OpenCL headers are in a later include search path.  The comment I made in the patches section of your meta.yaml points to how I patched things.

I understand that upstream will be reluctant to remove the old `CL/ext.h` header that conflicts with `ocl-icd` due to tests that still use it (see https://github.com/Xilinx/XRT/issues/1339) but unless there is something different in your build than what I had, conda-build isn't going to include the XRT version of that header because it will be seen as already installed by one of the dependencies and there's no way to cleanly support the 'feature' of having `#include <CL/cl_ext.h>` mean:
* really include `CL/cl_ext_xilinx.h`
* which in turn includes the actual `CL/cl_ext.h` from Khronos.

Patching `#include_next` to `#include` may be the only portion that needs to remain in the conda recipe if upstream can't accept it.  I would expect that they could fix the missing `#include <CL/cl_ext_xilinx.h>` patches that I also found.

recipes/xilinx-runtime/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/env/bash
+set -e
+
+mkdir -p $SRC_DIR/build-release
+
```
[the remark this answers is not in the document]
timsnyder: It would be really nice if upstream could be patched to correctly include the generated version artifacts in the release tarballs so that the version information is correct when building from tarball.

recipes/xilinx-runtime/conda.patch
```diff
@@ -0,0 +1,223 @@
...
++#endif()
+diff -Naur XRT-202210.2.13.466/src/CMakeLists.txt work/src/CMakeLists.txt
+--- XRT-202210.2.13.466/src/CMakeLists.txt	2022-04-14 12:29:02.000000000 -0700
++++ work/src/CMakeLists.txt	2022-06-19 10:07:59.471327068 -0700
+@@ -48,8 +48,10 @@
+   -Wno-overloaded-virtual
+    )
+ else()
+-  set(XRT_WARN_OPTS -Wall -Werror )
++  set(XRT_WARN_OPTS -Wall)
```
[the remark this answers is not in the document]
bryanloz-xilinx: @timsnyder yes this gets picked up in the build. CMake will not treat warnings as errors. XRT team was trying to force all warnings to be errors, so that developers would actually fix their warnings. However, this is too draconian when you consider that you need the world to compile your stuff in different environments. In the latest greatest XRT, there is now an option to turn this off, however here, I have to hack it in the patch.

bryanloz-xilinx:
@timsnyder really loving your feedback. I typically never use XRT's python interface, so its really helpful to have your feedback.

recipes/xilinx-runtime/build.sh
```diff
@@ -0,0 +1,20 @@
...
+  -S $SRC_DIR/src  \
+  -B $SRC_DIR/build-release  \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DCMAKE_PREFIX_PATH=${PREFIX} \
+  -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+  -DCMAKE_SKIP_INSTALL_RPATH=ON
+
+cmake --build $SRC_DIR/build-release --parallel ${CPU_COUNT}
+
+cmake --install $SRC_DIR/build-release
```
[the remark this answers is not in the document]
bryanloz-xilinx: @timsnyder I do not intend to include the static libraries. Will double check.

recipes/xilinx-runtime/build.sh
```diff
@@ -0,0 +1,20 @@
...
+
+# XILINX_XRT
+mkdir -p $PREFIX/etc/conda/activate.d
+echo -e "#!/bin/env/bash\nexport XILINX_XRT=\$CONDA_PREFIX\n" > $PREFIX/etc/conda/activate.d/env_vars.sh
```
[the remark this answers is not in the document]
bryanloz-xilinx: @timsnyder ok so what name should I use? `xilinx_runtime_xilinx_xrt.sh`

I like this 1 liner approach, and I'm sure it can be duplicated for the deactivate.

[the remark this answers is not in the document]
timsnyder:
```suggestion
echo -e "#!/bin/env/bash\nexport XILINX_XRT=\$CONDA_PREFIX\n" > $PREFIX/etc/conda/activate.d/${PKG_NAME}_activate.sh
```

recipes/xilinx-runtime/build.sh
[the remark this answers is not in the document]
bryanloz-xilinx: > It would be really nice if upstream could be patched to correctly include the generated version artifacts in the release tarballs so that the version information is correct when building from tarball.

I opened this issue on your behalf: https://github.com/Xilinx/XRT/issues/6794

recipes/xilinx-runtime/build.sh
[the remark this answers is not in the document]
bryanloz-xilinx: @timsnyder When I look at the CMake Infra...

I don't see any option for controlling static or dynamic libs... for example:
https://github.com/Xilinx/XRT/blob/master/src/runtime_src/core/pcie/emulation/cpu_em/CMakeLists.txt#L34

We explicitly build both... and I think they get installed together.

Again, Xilinx does not follow the convention of using -DBUILD_SHARED_LIBS to easily control this.

https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

Do you want to open an issue against XRT for this?

timsnyder:
Looks like the linux_64 build is still failing in CI.  I'll take a look and see if I can suggest anything.  I'll also try to install it and test it some more.

bryanloz-xilinx:
> Looks like the linux_64 build is still failing in CI. I'll take a look and see if I can suggest anything. I'll also try to install it and test it some more.

I fixed it at one point, my last cleanup push where I removed a few installations must have broken it.

timsnyder:
@bryanloz-xilinx , I also remembered this morning that if you don't patch `xbutil2` like I did in https://github.com/timsnyder/staged-recipes/blob/xilinx-xrt/recipes/xilinx-xrt/build.sh#L32 it will fail to find the testcases for `xbutil validate -d <device>`.  The larger issue at hand is that there are places in the CPP where `/opt/xilinx/xrt` is hardcoded as the prefix and this should probably be cleaned up upstream so that they use a generated file that contains `CMAKE_INSTALL_PREFIX` instead.

bryanloz-xilinx:
> @bryanloz-xilinx , I also remembered this morning that if you don't patch `xbutil2` like I did in https://github.com/timsnyder/staged-recipes/blob/xilinx-xrt/recipes/xilinx-xrt/build.sh#L32 it will fail to find the testcases for `xbutil validate -d <device>`. The larger issue at hand is that there are places in the CPP where `/opt/xilinx/xrt` is hardcoded as the prefix and this should probably be cleaned up upstream so that they use a generated file that contains `CMAKE_INSTALL_PREFIX` instead.

Before I do the next recipe... I am going to drop a nuclear bomb on `/opt/xilinx` in the upstream. Its going to be painful. We have already waited 4 years or more to make this fix... but the longer we wait... the more atrocious it will be to fix.

timsnyder:
It might be even better for the executables to search relative to where they are installed instead of needing to have a hardcoded absolute path that needs to be patched by conda at install time.  The binary string patching works a lot of the time but I know there are also pitfalls and some conda-forge recipes disable it completely.

> We have already waited 4 years or more to make this fix... but the longer we wait... the more atrocious it will be to fix.

I understand how people get used to certain assumptions and how hard it is to break out of those.  I honestly didn't think the patches I needed to do were too onerous and mostly could be cleaned up so that upstream could continue to release things the way they have and also have a conda package that works the same.

> Before I do the next recipe...

What other things are you going to create recipes for?

timsnyder:
@bryanloz-xilinx , given the reply at https://github.com/Xilinx/XRT/issues/6779#issuecomment-1161925441 it seems like the conda package shouldn't include `bin/xbflash`, but only include the `xbflash.qspi`.

recipes/xilinx-runtime/build.sh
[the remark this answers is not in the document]
timsnyder: https://github.com/Xilinx/XRT/issues/6799, in the meantime, you should add a manual remove.  This should cover things:
```suggestion
cmake --install $SRC_DIR/build-release

rm -f $PREFIX/lib/libxrt*.a $PREFIX/lib/libxilinx*.a
```

bryanloz-xilinx:
@xhochy hope you are doing well! When you have a minute, can you please let me know what the next step is? Thanks again for your guidance!

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - pybind11
+    - boost
+    - boost-cpp
+    - protobuf
+    - libprotobuf
+    - rapidjson
+    - libuuid
+    - libdrm
+    - clhpp
+    - libcurl
```
xhochy, line 41: The log output suggest that this isn't needed.
```suggestion
```

bryanloz-xilinx: @xhochy libcurl provides curl.h into the include/ directory which is required for building the C++. It may not be needed at runtime. Perhaps I should ignore run exports for libcurl just like I am for libdrm?

xhochy: Yes, sounds like a good approach!

bryanloz-xilinx:
@timsnyder do you know why this Warning pops up in the build?
```
WARNING: No ICDs were found. Either,
- Install a conda package providing a OpenCL implementation (pocl, oclgrind, intel-compute-runtime, beignet) or 
- Make your system-wide implementation visible by installing ocl-icd-system conda package.
```

recipes/xilinx-runtime/meta.yaml
bryanloz-xilinx: This should be resolved.

timsnyder:
> @timsnyder do you know why this Warning pops up in the build?

It happens when you activate the build environment with `ocl-icd` installed and it doesn't know that you're actually building a package that is going to provide an ICD.  It is safe to ignore it.

https://github.com/conda-forge/ocl-icd-feedstock/blob/main/recipe/activate.sh is where the message is coming from if you're curious.

timsnyder:
@bryanloz-xilinx, I know we're only targeting linux right now but I was reminded yesterday that the binary PREFIX replacement that conda does on windows is limited in it's capabilities and we currently depend on that feature of conda for a number of binaries in xilinx-runtime.  See https://github.com/conda-forge/cppcheck-feedstock/issues/19#issuecomment-1168071505 for more info.

bryanloz-xilinx:
@xhochy are you ready to merge?

bryanloz-xilinx:
xbutil validate works
```
$ xbutil validate --device 0000:3b:00.1
Starting validation for 1 devices

Validate Device           : [0000:3b:00.1]
    Platform              : xilinx_u250_gen3x16_base_3
    SC Version            : 4.6.18
    Platform ID           : 48810C9D-1786-0EF5-3E9E-529E8B14CE39
-------------------------------------------------------------------------------
Test 1 [0000:3b:00.1]     : aux-connection 
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Test 2 [0000:3b:00.1]     : pcie-link 
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Test 3 [0000:3b:00.1]     : sc-version 
    Test Status           : [PASSED]
-------------------------------------------------------------------------------
Test 4 [0000:3b:00.1]     : verify 
Test 5 [0000:3b:00.1]     : iops 
Test 6 [0000:3b:00.1]     : mem-bw 
Test 7 [0000:3b:00.1]     : vcu 
Validation completed. Please run the command '--verbose' option for more details
```

bryanloz-xilinx:
@timsnyder looks like I can't simply remove the static libraries from the build with rm -rf, because if I do... `find_package(XRT)` will fail, due to missing targets:

```
CMake Error at /scratch/bryanloz/mambaforge/envs/test/share/cmake/XRT/xrt-dev-targets.cmake:99 (message):
  The imported target "XRT::xilinxopencl_static" references the file

     "/scratch/bryanloz/mambaforge/envs/test/lib/libxilinxopencl_static.a"

  but this file does not exist.  Possible reasons include:

  * The file was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and contained

     "/scratch/bryanloz/mambaforge/envs/test/share/cmake/XRT/xrt-dev-targets.cmake"

  but not all the files it references.

  /scratch/bryanloz/mambaforge/envs/test/share/cmake/XRT/xrt-config.cmake:59 (include)
  /home/bryanloz/UPGRADE_MACHINE/XRT/tests/CMake/utils.cmake:24 (find_package)
  00_hello/CMakeLists.txt:8 (include)
```

timsnyder:
> I can't simply remove the static libraries from the build with rm

good catch.  I'd probably patch the cmake and pkg-config files.  The reasons for not shipping static libraries by default in https://github.com/conda-forge/cfep/blob/main/cfep-18.md make a lot of sense.  I don't mind taking a look at the patch.

bryanloz-xilinx:
> > I can't simply remove the static libraries from the build with rm
> 
> good catch. I'd probably patch the cmake and pkg-config files. The reasons for not shipping static libraries by default in https://github.com/conda-forge/cfep/blob/main/cfep-18.md make a lot of sense. I don't mind taking a look at the patch.

This would be awesome...

BTW... the way I am doing one massive conda.patch is pretty dumb right?

I do like how you had the patches organized. If we are going to continue waiting for a merge, maybe we can clean that up more.

timsnyder:
> BTW... the way I am doing one massive conda.patch is pretty dumb right?

It's not the greatest but I don't use patch based workflows for anything other than conda recipes and it took me awhile to grok how to manage them.  I basically just write commits in a clone of XRT and then use `git format-patch -N` to create the patches like I had them.  You can also apply the patches with `git am recipe/patches/*`.

bryanloz-xilinx:
Wow.. version.h is actually needed at times... 
```
/scratch/bryanloz/mambaforge/envs/xrt/include/xrt/detail/abi.h:8:11: fatal error: version.h: No such file or directory
    8 | # include "version.h"
```

Unless disabled with `-DDISABLE_ABI_CHECK`

timsnyder:
> version.h is actually needed at times...

I could have sworn that I grepped through the include files to see if it was used by them.  Oh, actually, that's one that I found by reviewing the files in your package and didn't remove in my recipe.  It would be really nice if the installed header had xrt in the name somewhere, either `include/xrtversion.h` or `include/xrt/version.h`...

bryanloz-xilinx:
@timsnyder I reorganized the patches.
I generate them with diff -Naur since this recipe isn't using git_url... (maybe there are other ways, but this worked best for me).
I also tested the latest recipe here on CentOS machine with Alveo U250.
I saw this test working on the board:
https://github.com/Xilinx/XRT/tree/master/tests/xrt/02_simple

The only problem now is that the recipe will include static libraries.
I looked through the CMake files, and there are many places that static libraries are referenced.
I would prefer not to fix this issue for now... but I gather that you would prefer that it is fixed.
Will you have time to look into it next week?

timsnyder:
@bryanloz-xilinx, I'm going to try to whip out a PR to upstream that will optionally split the static libraries into their own COMPONENT so that they can be packaged separately.  I thought a little more about the usecase for XRT static libs and I think there are probably many users that build their host app statically so that they can package it into the xclbin and have everything they need to deploy the app included in the xclbin.  That makes me think that splitting them into their own package is better than removing them and it looks like just as much work.

timsnyder:
This is the jist of what I'm going to try: https://stackoverflow.com/questions/21852817/cmake-how-to-create-alias-for-installing-different-targets/21853784#21853784

timsnyder:
@bryanloz-xilinx, I have an implementation pushed to https://github.com/timsnyder/XRT/tree/static_component.  It is split into two commits because I touched the existing XRT_DEV_COMPONENT in the second commit, exposing it as an OPTION so that one can set it on cmake cmdline without patching the cmake files.  Might want to roll that back separately.

To use my implementation, you will want to include `-DXRT_STATIC_COMPONENT=xrt-static` to your first cmake cmdline and then when installing add `-DCMAKE_INSTALL_COMPONENT=xrt` and it should include only the dynamic libraries and have working files for  CMake  `find_package` downstream.  If `cmake -DCMAKE_INSTALL_COMPONENT=xrt --install $SRC_DIR/build-release` doesn't work, you might have to do the the `-P` thing suggested in the SO answer I linked.  So `cmake -DCMAKE_INSTALL_COMPONENT=xrt -P $SRC_DIR/build-release/cmake_install.cmake`

I haven't tested this yet.  I wanted to share it before I get pulled away to something else.

timsnyder:
Opened https://github.com/Xilinx/XRT/pull/6830

bryanloz-xilinx:
@timsnyder I made a patch from your PR, and I see no static libs!
```
$ ls $CONDA_PREFIX/lib/*xrt*
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_core.so.2@             /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_noop.so.2@
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_core.so.2.13.466*      /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_noop.so.2.13.466*
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_coreutil.so.2@         /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt++.so.2@
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_coreutil.so.2.13.466*  /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt++.so.2.13.466*
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_hwemu.so.2@            /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_swemu.so.2@
/scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_hwemu.so.2.13.466*     /scratch/bryanloz/mambaforge/envs/xrt0/lib/libxrt_swemu.so.2.13.466*
```

But the xrt-config.cmake is missing:

```
$ ls $CONDA_PREFIX/share/cmake/XRT/
xrt-dev-targets.cmake  xrt-targets.cmake  xrt-targets-release.cmake
```

Old ENV:
```
$ ls /scratch/bryanloz/mambaforge/envs/XRT/share/cmake/XRT/
xrt-config.cmake  xrt-config-version.cmake  xrt-dev-targets.cmake  xrt-dev-targets-release.cmake  xrt-targets.cmake  xrt-targets-release.cmake
```

I don't get why, the installation of those files is handled here:
https://github.com/Xilinx/XRT/blob/master/src/CMake/findpackage.cmake#L37

I used: `-DXRT_STATIC_COMPONENT=xrt-static`

and: `cmake --install $SRC_DIR/build-release --component xrt`

bryanloz-xilinx:
@conda-forge/staged-recipes please merge this! Happy 4th!

timsnyder:
> But the xrt-config.cmake is missing:

 This is indeed strange.  Does this cause `find_package(XRT)` to fail downstream?  What are the consequences of not having `xrt-config.cmake`?  I'll look into why it isn't being included and double check that the `pkg-config` files are also correct with this change.

bryanloz-xilinx:
> > But the xrt-config.cmake is missing:
> 
> This is indeed strange. Does this cause `find_package(XRT)` to fail downstream? What are the consequences of not having `xrt-config.cmake`? I'll look into why it isn't being included and double check that the `pkg-config` files are also correct with this change.

Yes, I also noticed that XRT header files weren’t getting installed. I believe I “fixed” it, by making the change I suggested in your XRT PR. But, you didn’t like my change?

timsnyder:
> Yes, I also noticed that XRT header files weren’t getting installed. I believe I “fixed” it, by making the change I suggested in your XRT PR. But, you didn’t like my change?

Not sure what you mean about not liking your change.  I committed the suggestions you made.

bryanloz-xilinx:
> > Yes, I also noticed that XRT header files weren’t getting installed. I believe I “fixed” it, by making the change I suggested in your XRT PR. But, you didn’t like my change?
> 
> Not sure what you mean about not liking your change. I committed the suggestions you made.

I was still waking up when I was looking at the PR. I just got confused, please ignore. Thanks for the work you did. It will at least work here as a patch.

timsnyder:
After remembering that I need to clone aws-fpga and source the sdk-setup.sh to get XRT to build on my AWS VM, I built it on my branch using `build.sh -split_shared` (only because it is an easy way to separate my XRT_SHARED_COMPONENT changes from your patch and interpretation of the changes and because it installs everything into separate packages with cpack).  I see the xrt-config.cmake ending up in the main xrt package (for component xrt):
```
bash-4.2$ for f in *.tar.gz; do echo $f; tar tf $f | grep xrt-config.cmake; done
xrt_202220.2.14.0_7.9.2009-x86_64-aws.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-azure.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-container.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-xbflash2.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-xbflash.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-xrt-static.tar.gz
xrt_202220.2.14.0_7.9.2009-x86_64-xrt.tar.gz
opt/xilinx/xrt/share/cmake/XRT/xrt-config.cmake
```

Now that I'm confident that I didn't mess something up in XRT_STATIC_COMPONENT to prevent it from being installed, I'll clone your recipe and build it to dig into why `xrt-config.cmake` isn't being included in the conda package.

bryanloz-xilinx:
@timsnyder the file is there now, after converting from CMake option to CMake cache var. everything is good now from my perspective

timsnyder:
Oh.  Good.  I was confused then.  I'll grab the package out of CI and give it a whirl.

bryanloz-xilinx:
> Oh. Good. I was confused then. I'll grab the package out of CI and give it a whirl.

Here is what I have been doing to test. It may be naive, but just sharing in case it is useful to you, or you can educate me on a better way...

1. clone this fork
2. cd */recipes/
3. conda build ./xilinx-runtime --output_folder ./channel -c conda-forge
4. conda create -n xrt cmake cxx-compiler ocl-icd xilinx-runtime -c ./channel -c conda-forge

I am independent of any CI infrastructure, but I guess that is good and bad.

timsnyder:
> Here is what I have been doing to test. It may be naive, but just sharing in case it is useful to you, or you can educate me on a better way...

What you're doing is good and I'm glad you're testing it out.  However, if you want to [try the packages that are being built by CI](https://conda-forge.org/docs/maintainer/updating_pkgs.html#downloading-prebuilt-packages-from-ci), you can browse to the build results page on Azure Pipelines for the platform you care about (I usually click through the links in 'checks' and then more detail on Azure) https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_build/results?buildId=530459&view=logs&jobId=6f142865-96c3-535c-b7ea-873d86b887bd 

There, you should see a "N artifacts produced" link like this:
image

After clicking the artifacts link, you should get to a page like this:
image

Clicking  the download artifacts link will give you a conda_pkgs_linux.zip file that contains a bunch of stuff, including the built packages from the CI.  You can install those using the `-c <directory>` directory trick (because the index should be in there too).  Or, what I usually do is upload those packages to https://anaconda.org/tsnyder so that others can install them for testing by specifying `-c tsnyder`.  

If you don't have an anaconda.org account, just sign up for one, then you'll need to `conda install anaconda-client` to get the CLI.  Then:
```
anaconda login # and interactively give it your username and password, it will store an auth token in your homedir
anaconda upload <path to packages you want to upload> # it defaults to uploading to your personal channel
```

Also, if you want to build (a linux package) using the same docker container that CI uses, you can use the `build_locally.py` script at the root of staged_recipes (and there will be one in the feedstock as well) as documented at https://conda-forge.org/docs/maintainer/updating_pkgs.html#testing-changes-locally. That will create a `build_artifacts` directory that acts as your conda-build `--croot` or `CONDA_BLD_PATH`.  So, it will contain the working directories conda-build used to build the package as well as a `linux-64` 'channel' directory containing the built packages.

For building macos locally, I have some instructions that I got from members of conda-forge/core and tweaked that I could share.  For building Windows, I use Virtualbox with the Microsoft VM's from the link pointed to by https://conda-forge.org/docs/maintainer/knowledge_base.html#local-testing

bryanloz-xilinx:
@timsnyder what do you think about this recipe now? ready for merge?
I know you don't like how some of the installed header files have dangerous names like `version.h`.
I would like to fix that in upstream at a later date, when we remove dependence on `/opt/xilinx/xrt`.
I don't think it will be a problem, because most well planned projects install all of their header files into a subdirectory. i.e. `#include <glog/glog.h>`

carterbox (changes requested):
Optional: Downstream users may benefit from splitting the python components into a separate outputs/recipes. If you are interested in doing this, it would be better to do it here where staged-recipes can help.

recipes/xilinx-runtime/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - patches/13_fix_appdebug_install.patch
+    - patches/14_remove_static_libs.patch
+
+build:
+  number: 0
+  skip: true  # [not linux]
+  ignore_run_exports_from:
+    - libdrm
+    - libcurl
+
```
carterbox, line 35:
```suggestion
  run_exports:
    - {{ pin_subpackage('xilinx-runtime'), max_pin='x') }}

```
Because you are shipping headers, you need a run_export which describes the ABI compatibility of this package for downstream packages that link against it. Please adjust the max_pin according to the compatibility guarantees of this library.

carterbox (changes requested):

recipes/xilinx-runtime/meta.yaml
bryanloz-xilinx: Thanks for pointing this our, this is fixed now, I am just following the lead from libprotobuf and json-c. Resolving this.

bryanloz-xilinx:
> Optional: Downstream users may benefit from splitting the python components into a separate outputs/recipes. If you are interested in doing this, it would be better to do it here where staged-recipes can help.

The python components here are just pybind11 bindings around the C++, so I don't think I want to go this route.

bryanloz-xilinx:
@conda-forge/help-c-cpp this is ready for review.

bryanloz-xilinx:
I am super excited to see this merged!
I also learned a new acronym LGTM!
Thanks @xhochy @carterbox and @timsnyder for all your help with this!

PR #19344, opened 2022-06-20
Add rule-engine recipe

teake:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/rule-engine/meta.yaml
+++ b/recipes/rule-engine/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "rule-engine" %}
+{% set version = "3.4.0" %}
+
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/rule-engine-{{ version }}.tar.gz
+  sha256: d35186aac7e621676a92f7c88c6599bc167831d33cb72a9f3131b8f73300915a
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - pip
+    - python
+  run:
+    - ply >=3.9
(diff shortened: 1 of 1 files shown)


teake:

recipes/rule-engine/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - pip
+    - python >=3.4
+  run:
+    - ply >=3.9
+    - python >=3.4
+    - python-dateutil ~=2.7
```
BastianZim, line 25:
```suggestion
    - python-dateutil >=2.7,==2.*
```
Condo-build can't really understand `~=`(Only through some hacks) so it's better to use the conda-build equivalent.

BastianZim (changes requested):
Just one small thing, then this is ready to go 🚀

recipes/rule-engine/meta.yaml
teake: That didn't work. I've changed it to `>=2.7,<3` as in the [examples in the conda docs](https://conda.io/projects/conda/en/latest/user-guide/concepts/pkg-specs.html#package-match-specifications).

Btw, just so I understand, is it documented somewhere that conda-build cannot understand `~=` properly?

BastianZim: Hmm strange. @marcelotrevisani this is what's suggested by grayskull but it didn't work. Do you know why?

> Btw, just so I understand, is it documented somewhere that conda-build cannot understand ~= properly?

Conda-build can only understand the following: https://docs.conda.io/projects/conda-build/en/latest/resources/package-spec.html#build-version-spec
We are patching some other things in the background (like the one here) but it's better to be on the safe side.

teake: Ok, thanks for the clarification!

PR #19354, opened 2022-06-21
Add nutpie

aseyboldt:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/nutpie/meta.yaml
+++ b/recipes/nutpie/meta.yaml
@@ -12,14 +12,16 @@ source:
 build:
   number: 0
   skip: True  # [py<38]
-  script: |
-    maturin build --bindings pyo3 --no-sdist --release --interpreter {{ PYTHON }} --out wheels
-    {{ PYTHON }} -m pip install -vv --no-build-isolation --no-deps --no-index --find-links ./wheels nutpie
+  script:
+    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+    - maturin build --bindings pyo3 --no-sdist --release --interpreter {{ PYTHON }} --out wheels
+    - {{ PYTHON }} -m pip install -vv --no-build-isolation --no-deps --no-index --find-links ./wheels nutpie
 
 requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
+    - cargo-bundle-licenses
   host:
     - python
     - setuptools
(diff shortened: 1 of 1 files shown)


aseyboldt:
And I agree to be listed as maintainer :-)

BastianZim:
Not a rust reviewer so I cannot comment on the compiler but please follow the docs about adding the rust licenses: https://conda-forge.org/docs/maintainer/adding_pkgs.html#licenses-of-included-dependencies

aseyboldt:
@BastianZim Thanks, I fixed the missing licenses.

aseyboldt:
@conda-forge/staged-recipes Would be great to have it merged soon :-)

carterbox:
> I've got a small packaging question as well though: This is a mixed python/rust package using maturin to package.
I had to include a build dependency of {{ compiler("c") }} though to avoid an error about missing libgcc-ng as a dependency, even though I'm not using a C compiler, but only the rust compiler. Does this maybe indicate that {{ compiler('rust') }} is missing something?

I think some rust programs will link to libc depending on whether they make certain system calls. Not sure when rust will implement its own library to replace these calls.

carterbox (changes requested):
There are [some warnings during your test phase that you may want to addresss.](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=527413&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1832) It seems that your package may need a compiler at runtime and maybe blas?

Either way please respond explicitly.

aseyboldt:
> I think some rust programs will link to libc depending on whether they make certain system calls. Not sure when rust will implement its own library to replace these calls.

Makes sense.

The compiler warning and missing blas are due to the runtime dependency on aesara, which can call a compiler to speed up graph computations. This functionality isn't used in this package at all. I think there are some discussions to split aesara into two packages, one that doesn't use a compiler, and one that does. If this does indeed happen, nutpie would only depend on that base package.
Anyway, this failure occurs if aesara is imported, and it disables some optimizations in response to it. So I don't think there is much I can do about this from this package.

aseyboldt:
Although aesara itself does actually depend at runtime on a c and c++ compiler. (https://github.com/conda-forge/aesara-feedstock/blob/main/recipe/meta.yaml#L30). So maybe this actually is an issue in the aesara package, where it doesn't locate the compiler provided by conda. The same error also appears in the most recent [aesara build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=523596&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=841356e0-85bb-57d8-dbbc-852e683d1642&l=1060)

PR #19371, opened 2022-06-23
Add recipe for LBFGSpp

awvwgk:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/lbfgspp/meta.yaml
+++ b/recipes/lbfgspp/meta.yaml
@@ -33,7 +33,7 @@ test:
     - test -d $PREFIX/include/LBFGSpp  # [unix]
     - if not exist %LIBRARY_PREFIX%\include\LBFGS.h exit 1  # [win]
     - if not exist %LIBRARY_PREFIX%\include\LBFGSB.h exit 1  # [win]
-    - for i in examples/*.cpp; do $CXX $CXXFLAGS -I$PREFIX/include/eigen3 -o "$i.out" "$i" && "$i.out"; done  # [unix]
+    - set -ex; for i in examples/*.cpp; do $CXX $CXXFLAGS -I$PREFIX/include/eigen3 -o "$i.out" "$i" && "$i.out"; done  # [unix]
 
 
 about:
--- a/recipes/lbfgspp/meta.yaml
+++ b/recipes/lbfgspp/meta.yaml
@@ -1,0 +1,42 @@
+{% set name = "lbfgspp" %}
+{% set version = "0.2.0" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/yixuan/lbfgspp/archive/refs/tags/v{{ version }}.tar.gz
+    sha256: 7101744a538c3aff52e10c82267305847b0b5e9d39f9974b4b29812cd1398ff9
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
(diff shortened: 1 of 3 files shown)


recipes/lbfgspp/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+  host:
+    - eigen
+
```
carterbox, line 22:
```suggestion
  host:
    - eigen =3.0.*
  run:
    - eigen =3.0.*
```
Two things:

1. Upstream [explicitly requires eigen 3.0](https://github.com/yixuan/LBFGSpp/blob/00f0f1bd96689bd40e6c8b16cc135665eed5acc9/CMakeLists.txt#L41)
2. To be nice to users, we add eigen to the run section so when they add this package to requirements, eigen is also added (it's mentioned in the headers).

recipes/lbfgspp/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+test:
+  commands:
+    - test -f $PREFIX/include/LBFGS.h  # [unix]
+    - test -f $PREFIX/include/LBFGSB.h  # [unix]
+    - test -d $PREFIX/include/LBFGSpp  # [unix]
+    - if not exist %LIBRARY_PREFIX%\include\LBFGS.h exit 1  # [win]
+    - if not exist %LIBRARY_PREFIX%\include\LBFGSB.h exit 1  # [win]
```
carterbox, line 29: Could you please also compile and run a test program in the test section from the examples to check that this library is compatible with our compilers? Sometimes our compilers are too old and require source patches to silence syntax errors, etc.

carterbox (changes requested):

recipes/lbfgspp/meta.yaml
carterbox: I think the cmake version requirements is actually equivalent to `>=3.0,<4`. Otherwise, the builds would have failed already.

awvwgk:
@conda-forge/help-c-cpp Requested changes are addressed now.

carterbox (changes requested):

recipes/lbfgspp/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - examples/
+  requires:
+    - {{ compiler('cxx') }}
+  commands:
+    - test -f $PREFIX/include/LBFGS.h  # [unix]
+    - test -f $PREFIX/include/LBFGSB.h  # [unix]
+    - test -d $PREFIX/include/LBFGSpp  # [unix]
+    - if not exist %LIBRARY_PREFIX%\include\LBFGS.h exit 1  # [win]
+    - if not exist %LIBRARY_PREFIX%\include\LBFGSB.h exit 1  # [win]
+    - for i in examples/*.cpp; do $CXX $CXXFLAGS -o "$i.out" "$i" && "$i.out"; done  # [unix]
```
carterbox, line 36:
```suggestion
    - for i in examples/*.cpp; do $CXX $CXXFLAGS -I$PREFIX/include/eigen3 -o "$i.out" "$i" && "$i.out"; done  # [unix]
```
This is probably the issue with your script.

carterbox (changes requested):

awvwgk:
@conda-forge/help-c-cpp Fixed example compilation. Thanks @carterbox.

PR #19385, opened 2022-06-24
adding biofilm and dependencies

smautner:
checklist is checked

Checklist
- [x] Build number is 0.


smautner:

recipes/biofilm/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/biofilm-{{ version }}.tar.gz
+  sha256: 9cd7e1ab5e211618cfb7476d51b7a82045fa0292a3ea9cd0af685ede6ce3fa69
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: True  # [win or osx or py<38]
```
ocefpaf, line 16: Looks like this package declare some empty script, see this line in `setup.py`:

```
 scripts=[ ],
````

If we can get rid of that we can make it noarch. A patch and a PR upstream would do it ;-p

Comment (ocefpaf) [reply recipes/biofilm/meta.yaml:16]: Not sure I follow. auto-sklearn is available on Windows and OSX-64, it is a noarch package BTW.

File: recipes/biofilm/meta.yaml [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/biofilm-{{ version }}.tar.gz
+  sha256: b285d80da26aebdf41c7b4273c1b31717a035ff731169092a4ca1ac98b2cff3d
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: True  # [win or osx or py<38]
```
carterbox, line 16: This package should be `noarch: python`. It doesn't matter that it is uninstallable on Windows and OSX right now. If auto-sklearn or other packages become available for Windows and OSX later, it will become installable then without needing to rebuild your package.

https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=noarch#noarch-python

carterbox (changes requested):

recipes/eden-kernel/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  imports:
+    - eden
+    - eden.display
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/smautner/
```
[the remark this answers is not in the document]
smautner: fixed

recipes/biofilm/meta.yaml
smautner: @carterbox it looks like it does matter as the auto-testing complains now.

smautner:

msarahan:
cycling to get a fresh build since logs are no longer accessible here.

BastianZim:
Maintainer of `auto-sklearn` feedstock here. The `__linux` dep is because it is only available on Linux, but otherwise, noarch. Therefore, this package here should be able to ignore that. As long as it's available on Linux and builds there, we can merge.
If `auto-sklearn` one day becomes available on either OS, I'll remove that, and the package should be available everywhere.

In short, this is desired behaviour to make sure that it's not installed on WIN or OSX where it is not supported.

Ref: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html

recipes/biofilm/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/biofilm-{{ version }}.tar.gz
+  sha256: b285d80da26aebdf41c7b4273c1b31717a035ff731169092a4ca1ac98b2cff3d
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  noarch: python
```
msarahan, line 16:
```suggestion
  noarch: python
  skip: True   # [not linux]
```

Thanks @BastianZim for the info. @smautner, this change should make things happy.

BastianZim: Of course! This isn't actually needed, sorry, should've been more clear. The failure is fine and noarch can stay. We introduced this at some point to keep packages noarch, even if they aren't available on all platforms.

BastianZim: Since the package is built on Linux it's fine.

BastianZim: > Being encouraged to make recipes fail is a bizarre experience though.

That's because the CI here will always build on all OS so that we can see what is happening. In your feedstock, it will only build on Linux, and you won't (hopefully) see any errors.

PR #19400, opened 2022-06-24
Adding mmcv (Pytorch-based computer-vision lib)

apatsekin:
Hey guys,

This conda package exists under [Esri](https://anaconda.org/esri/mmcv-full) channel (commercial GIS software vendor). 

Adding `mmcv-full` package for now, since that's the main one. Vendor also publishes `mmcv` which is very light weight (pure python) and can be easily added later.

Checklist
- [V] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [V] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/5eddbd7fc9d1502169089da06c3688d9759be978/recipes/example/meta.yaml#L64-L73) for an example).
- [V] Source is from official source.
- [V] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).
- [V] If static libraries are linked in, the license of the static library is packaged.
- [V] Build number is 0.
- [V] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details).
- [V] GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there.
- [V] When in trouble, please check our [knowledge base documentation](https://conda-forge.org/docs/maintainer/knowledge_base.html) before pinging a team.


apatsekin:
hey @conda-forge/help-python, would you please review this one?

apatsekin:
Failing rn with:
```
['  ERROR (mmcv-full,lib/python3.7/site-packages/mmcv/_ext.cpython-37m-x86_64-linux-gnu.so): $RPATH/libc10.so not found in packages, sysroot(s) nor the missing_dso_whitelist.\n.. is this binary repackaging?', '  ERROR (mmcv-full,lib/python3.7/site-packages/mmcv/_ext.cpython-37m-x86_64-linux-gnu.so): $RPATH/libtorch_cpu.so not found in packages, sysroot(s) nor the missing_dso_whitelist.\n.. is this binary repackaging?', '  ERROR (mmcv-full,lib/python3.7/site-packages/mmcv/_ext.cpython-37m-x86_64-linux-gnu.so): $RPATH/libtorch_python.so not found in packages, sysroot(s) nor the missing_dso_whitelist.\n.. is this binary repackaging?']
```

apatsekin:
hey @conda-forge/help-python-c, could you take a look?

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,115 @@
...
+
+source:
+  url: https://github.com/open-mmlab/mmcv/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 650929f52f1d0b2bae1bdaaa874e1e6dfdfa15f40bab5c366330488a9949ddd0
+build:
+  number: {{ number }}
+  skip: true  # [win]
+
+outputs:
+  - name: mmcv-full
```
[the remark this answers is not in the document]
apatsekin: Yeah, that's how vendor distributes it:

```
mmcv-full: comprehensive, with full features and various CUDA ops out of box. It takes longer time to build.

mmcv: lite, without CUDA ops but all other features, similar to mmcv<1.0.0. It is useful when you do not need those CUDA ops.
```
https://mmcv.readthedocs.io/en/latest/get_started/installation.html

Pip packages are also `mmcv` and `mmcv-full`. The former one is very light and pure-python, so it will be easy to add it later.

recipes/mmcv/conda_build_config.yml
```diff
@@ -0,0 +1,2 @@
+pytorch:
+  - 1.11
```
hmaarrfk, line 2: Do you actually need this file?

hmaarrfk: this file will stop you from migrating to newer versions.

apatsekin: @hmaarrfk , as @ngam mentioned MMCV ([upstream](https://mmcv.readthedocs.io/en/latest/get_started/installation.html)) is built for all  combinations of python/cuda/torch. Assuming we should do the same having just `1.11` for now.
My reasoning was:
1. If we don't pin it during build - it's quite opaque what version it was built against.
2. Whenever new pytorch is released - any further re-build of this stock will pick it up without a guarantee of compatibility.

What do you suggest?

hmaarrfk: once you rerender, ci support files are generated losying the specific versions of the dependencies.

in my experience, getting the bots help is critical when you maintain mkre than about 5 packages.

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+      host:
+        - python
+        - pip
+        - requests
+        - future
+        - six
+        - typing
+        - pkg-config  # [unix]
+        - typing_extensions
+        - pytorch ={{ pytorch }}={{ proc_type }}*
```
hmaarrfk, line 56: I think you will have to add a run_constraint on pytorch. We don't export the exact GPU requirement. I'm not sure we want to. So for now, I think you will have to follow

https://github.com/conda-forge/torchvision-feedstock/blob/main/recipe/meta.yaml#L73

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - scipy
+        - libopencv
+        - py-opencv
+        - opencv
+        - python-lmdb
+        - libiconv
+        - tifffile
+        - python-dateutil
+        # mmcv provides binary wheels bound to specific pytorch version
+        - pytorch ={{ pytorch }}={{ proc_type }}*
```
hmaarrfk, line 81: Once you add the run_constraint, you won't need this anymore

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+
+about:
+  home: https://github.com/open-mmlab/mmcv
+  summary: 'Foundational library for computer vision research'
+  description: |
+    MMCV is a foundational library for computer vision research and supports many research projects.
+  license: Apache-2.0
+  license_family: Apache
+  license_file: LICENSE.txt
```
hmaarrfk, line 125: Why are you using the custom file? You should be able to pick it up from the source itself.
```suggestion
  license_file: LICENSE
```

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+      host:
+        - python
+        - pip
+        - requests
+        - future
+        - six
+        - typing
+        - pkg-config  # [unix]
+        - typing_extensions
+        - pytorch =*={{ proc_type }}*
```
hmaarrfk, line 56:
```suggestion
        - pytorch ={{ pytorch }}={{ proc_type }}*
```

I think you will want to keep this syntax here, so that it pick up the pin from your custom `conda_build_config.yaml` or the global one.

recipes/mmcv/LICENSE.txt
```diff
@@ -0,0 +1,13 @@
+Copyright 2015-2022, conda-forge contributors
```
hmaarrfk, line 1: you should be able to remove this file too.

apatsekin:
@hmaarrfk , thanks for the tips! I'm not sure I understand how the `test` pytorch will be pinned correctly?
currently we have:
1. version/device pin in `host`.
2. device pin in `run_constrained`
3. no pins in `test`.
^ that sounds to me like conda will follow `device` pin, but will not pick up `version` pin from `host`?

hmaarrfk:
The test will install the same package that you built, following all the constraints and the run_constraints. Therefore, if the final package is pinned correctly, the test will have all the required pinnings.

Otherwise, your users will have to manually specify the matching pytorch, which you don't want them to.

To be honest, it is good that the upstream repository tests with all different pytorch combinations. However, I don't think it is realistic for us to be releasing many different versions of pytorch at the moment.

I (one of the pytorch-feedstock maintainers) consider everything but the latest pytorch release essentially broken due to us learning how to package such a complex project. The advantage of testing with many different releases of pytorch is that sometimes downstream packages do not have the time to immediately release the latest version. This builds some buffer. However, it really does take 1-2 days to reliably build pytroch, even on a massive computer.

recipes/mmcv/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - future
+        - six
+        - typing
+        - pkg-config  # [unix]
+        - typing_extensions
+        - pytorch ={{ pytorch }}={{ proc_type }}*
+        - torchvision
+        - setuptools
+      run:
+        # intentionally skipping opencv and pytorch (following upstream vendor)
```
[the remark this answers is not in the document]
apatsekin: `mmcv-full` depends on those two, but upstream vendor doesn't put them as runtime dependencies in pip packaging. I would imagine that it's because:
1) you can use parts of the library without those heavy deps (especially without opencv)
2) opencv with 4 distribution options under Pip (contrib/not-contrib, headless/not-headless) makes it a mess to depend on a single one. I.e. if your package depends on `headless` version and user has already normal one installed - pip will still install `headless` on top of that. I think the same will happen with conda, i.e. if users uses FastAI's headless version which is quite popular.

apatsekin:
hey @hmaarrfk , anything else I can address? Or could you merge it ?

hmaarrfk:
Sorry. I missed the fact that you did take action on the suggested points.

Make sure that the pttorch pin appears in the ci support files

apatsekin:
@hmaarrfk thanks! Quick q: can I push a PR to the feedstock repo to rename it ? :) I messed up the name itself - `mmcv-recipe` instead of `mmcv`. Doesn't affect package name, but looks inconsistent with other repos..

hmaarrfk:
If you want, we could archive that repo, but i don't really think it is worth it.

We have https://conda-forge.org/feedstock-outputs/

I'll remember to check feedstock name in the future

PR #19427, opened 2022-06-26
Add arrow-c-glib

xmnlab:
Checklist
- [x] Build number is 0.


recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    Arrow GLib provides C API.
+  license: Apache-2.0
+  license_family: Apache
+  license_file: LICENSE.txt
+  doc_url: https://arrow.apache.org/docs/c_glib/
+  dev_url: https://github.com/apache/arrow/tree/master/c_glib
+
+extra:
+  recipe-maintainers:
+    - xmnlab
```
xhochy, line 42: Can I also be a maintainer here? I take care of all the other Arrow stuff so I would be happy to help

xmnlab: Sure thing. I will do it in a bit :) thanks

xmnlab:
> Found no recipes to build

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=527309&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=539

not sure what is happening ... @xhochy do you have any idea why it is not finding the recipe?
locally it works fine.

PS: I will add tests soon

recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  url: https://dist.apache.org/repos/dist/release/arrow/arrow-{{ version }}/apache-arrow-{{ version }}.tar.gz
+  sha256: ad9a05705117c989c116bae9ac70492fe015050e1b80fb0e38fde4b5d863aaa3
+
+build:
+  number: 0
+  noarch: generic
+  skip: true  # [win]
+  missing_dso_whitelist:
```
xmnlab, line 16: @xhochy not sure if this is the right way to go .. just added it in order to fix the issues on CI 
do you have any tips about this?

xmnlab:
Hi @@conda-forge/help-c-cpp this PR is ready for review. thanks!

recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+build:
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - autoconf
```
kou, line 20: I think that we don't need this because we use Meson.

recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - cmake
+    - gnuconfig
```
kou, line 22: ditto.

recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - gobject-introspection
+    - make
```
kou, line 24: I think that we don't need this because we use Ninja.

recipes/arrow-c-glib/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+if [[ ${OSTYPE} == 'darwin'* ]]; then
+  export XML_CATALOG_FILES="${PREFIX}/etc/xml/catalog"
+fi
```
kou, line 8: I think that we don't need this because GTK-Doc isn't enabled by `-Dgtk_doc=true`.

xmnlab:
thanks for the review @kou ! recommendations applied :)

kou (approved):
+1 as a Apache Arrow C GLib developer.

xmnlab:
thanks for the review @kou

carterbox (changes requested):
This recipe needs an appropriate run_export.

xmnlab:
> This recipe needs an appropriate run_export.

could you give more explanation about this? what I need to add there?  arrow-cpp and glib? 
how can I know that it is missing something in the run_export? can I add a test that will break if the run_export is not correct? 
maybe I should add a test to try to compile a code using arrow_c_glib?

@carterbox thanks for the review.

carterbox:
https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

This is metadata which tells conda whether packages compiling against this package also need it at runtime (we don't allow static linking, so the answer is almost always yes). You should add this package only to this section.

Please use the pin_subpackage jinja template to pin an appropriate version of this package. As the maintainer of this package, you should know its API/ABI compatibility.

There is no way to test if run_exports are correct from inside the recipe. Usually, we figure out if run exports are incorrect when downstream packages are broken by updates to this package.

recipes/arrow-c-glib/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  version: {{ version }}
+
+source:
+  url: https://dist.apache.org/repos/dist/release/arrow/arrow-{{ version }}/apache-arrow-{{ version }}.tar.gz
+  sha256: ad9a05705117c989c116bae9ac70492fe015050e1b80fb0e38fde4b5d863aaa3
+
+build:
+  number: 0
+  skip: true  # [win]
+
```
carterbox, line 15:
```suggestion
  skip: true  # [win]
  run_exports:
    - {{ pin_subpackage('arrow-c-glib') }}
```
Should be something *like* this, but with your ABI compatibility guarantees in mind.

xmnlab:
thanks @carterbox for all the explanation and for the example. very helpful.
changes applied!

jorisvandenbossche:
Something I am wondering: would it make sense to have this as another output of the arrow-cpp recipe? (like pyarrow is done)

xmnlab:
@jorisvandenbossche that is a good question. 
I had this doubt too .. but @kou mentioned that it would be better to have it in a different repo.
@kou could you answer @jorisvandenbossche question pls? (I am not finding the explanation you sent in my history).
thanks a lot!!

kou:
Oh, sorry. I didn't notice that pyarrow is included in arrow-cpp.
@xhochy Is it OK that we add arrow-c-glib to https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/meta.yaml#L29 ? Or should we keep separated repository?

kou:
FYI: It seems that we have a separated repository for R: https://github.com/conda-forge/r-arrow-feedstock

xmnlab:
is it ok to keep it separated?

kou:
It seems OK because we have no more responses.

xmnlab:
great! thanks @kou !

PR #19438, opened 2022-06-27
Knoten recipe via Grayskull 

Kelvinrr:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/knoten/meta.yaml
+++ b/recipes/knoten/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - pip
   host:
     - python
+    - python_abi
     - pip
     - ale >=0.3.2
     - coveralls
--- a/recipes/knoten/meta.yaml
+++ b/recipes/knoten/meta.yaml
@@ -15,13 +15,12 @@ build:
   number: 0
 
 requirements:
-  build:
   host:
-    - python >=3.7,<3.8
+    - python >=3.7
     - pip 
   run:
     - ale >=0.3.2
-    - python >=3.7,<3.8 
+    - python >=3.7
     - csmapi >=1.0
     - gdal >=3.0.0
     - jinja2
--- a/recipes/knoten/meta.yaml
+++ b/recipes/knoten/meta.yaml
@@ -11,18 +11,19 @@ source:
 
 build:
   script: {{ PYTHON }} -m pip install . -vv
+  skip: true # [ win ]
   noarch: python 
   number: 0
 
 requirements:
   build:
   host:
-    - python >=3.7,<3.8 
+    - python
     - setuptools 
     - pip 
   run:
     - ale >=0.3.2
-    - python >=3.7,<3.8
+    - python 
     - csmapi >=1.0
     - gdal >=3.0.0
     - jinja2
--- a/recipes/knoten/meta.yaml
+++ b/recipes/knoten/meta.yaml
@@ -21,11 +21,10 @@ requirements:
     - pip
   host:
     - python
-    - python_abi
     - pip
     - ale >=0.3.2
     - coveralls
-    - csmapi =1.0.1
+    - csmapi
     - gdal >=3.0.0
     - holoviews
     - jinja2
@@ -48,7 +47,7 @@ requirements:
     - pip
     - ale >=0.3.2
     - coveralls
-    - csmapi =1.0.1
+    - csmapi
     - gdal >=3.0.0
     - holoviews
     - jinja2
--- a/recipes/knoten/meta.yaml
+++ b/recipes/knoten/meta.yaml
@@ -16,53 +16,23 @@ build:
 
 requirements:
   build:
-    - python
-    - setuptools
-    - pip
-    - ale >=0.3.2
-    - coveralls
-    - csmapi
-    - gdal >=3.0.0
-    - holoviews
-    - jinja2
-    - matplotlib-base
-    - numpy
-    - plio
-    - plotly
-    - plotly-orca 
-    - psutil 
-    - pvl
-    - pyproj
-    - pysis
-    - pytest
-    - requests
-    - scipy
-    - shapely
-    - usgscsm >=1.3.1
+  host:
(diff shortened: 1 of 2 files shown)


recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
+{% set name = "knoten" %}
+{% set version = "0.2.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/USGS-Astrogeology/knoten/archive/{{ version }}.tar.gz
+  sha256: d671f1c8c32107cf1ef1d2dadf236ba98f9f19dcfb744e302cb20860748164b0
```
ladoramkershner, line 10: Is this a git hash? Is this in place of tagging for releases?

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - knoten
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/USGS-Astrogeology/knoten
+  summary: 'A library to leverage python wrapped Community Sensor Models (CSMs) for common spatial/sensor operations and testing.'
+  description: Automated control network generation.
```
ladoramkershner, line 48: I think this need to be updated

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  doc_url: https://github.com/USGS-Astrogeology/knoten#readme
+  dev_url: https://github.com/USGS-Astrogeology/knoten
+  license: Unlicense
+  license_file: LICENSE.md
```
ladoramkershner, line 52: Is this under a CC0-1.0 and should we specify that or does CC0-1.0 technically fall under an Unlicense?

recipes/knoten/meta.yaml
beenje: This is the sha to check the integrity of the GitHub archive.

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/USGS-Astrogeology/knoten/archive/{{ version }}.tar.gz
+  sha256: d671f1c8c32107cf1ef1d2dadf236ba98f9f19dcfb744e302cb20860748164b0
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [ win or py2k ]
```
beenje, line 14: This package is pure python. It should be `noarch: python`.
Please remove the `skip`. It doesn't matter if the package can't be installed on windows due to missing dependencies. See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  number: 0
+
+requirements:
+  build:
+  host:
+    - python >=3.7,<3.8
```
beenje, line 20: This seems quite restrictive. Are you sure it doesn't work with 3.8 and above?

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - setuptools
```
beenje, line 21: No need to specify `setuptools`. `pip` is enough here.

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - python >=3.7,<3.8
+    - setuptools 
+    - pip 
+  run:
+    - ale >=0.3.2
+    - python >=3.7,<3.8 
+    - csmapi >=1.0
+    - gdal >=3.0.0
+    - jinja2
+    - {{ pin_compatible('numpy') }}
```
beenje, line 29: Just use `numpy`.  `pin_compatible` is only needed when numpy is used to compile the package.

recipes/knoten/meta.yaml
beenje: Yes, please use `CC0-1.0`.

beenje (changes requested):

recipes/knoten/meta.yaml
Kelvinrr: yeah, I've only spent a short time looking into it, but on of it's dependencies is generated via swig wrappers and something about how it's configured only allows for 3.7. Maybe lower, but definitely not higher. I wasn't able to get an environment with anything higher anyways.

recipes/knoten/meta.yaml
Kelvinrr: @ladoramkershner I think because knoten wasn't chosen for github dissemination it was never updated to CC0 and was still unlicense. I updated it a while back but forgot to update it here so thanks for pointing that out.

Kelvinrr:
@beenje Thanks for the review, should be good for re-review.

recipes/knoten/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  url: https://github.com/USGS-Astrogeology/knoten/archive/{{ version }}.tar.gz
+  sha256: d671f1c8c32107cf1ef1d2dadf236ba98f9f19dcfb744e302cb20860748164b0
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  noarch: python 
+  number: 0
+
+requirements:
+  build:
```
beenje, line 18: Please remove this empty `build:` section.

recipes/knoten/meta.yaml
beenje: If this project itself doesn't have any code not compatible with Python >=3.7, you shouldn't explicitely set an upper bound.
If a dependency requires Python 3.7, that dependency will constrain the Python version. And if a new version is released for another Python version, your package will just work without you having to update it.

You can remove the `<3.8`.

beenje (changes requested):

PR #19440, opened 2022-06-27
Add cWB

rhyspoulton:
Checklist
- [x] Build number is 0.


recipes/cWB/build.sh
```diff
@@ -0,0 +1,26 @@
...
+set -ex
+
+mkdir -pv build
+pushd build
+
+cmake \
+  -DCMAKE_BUILD_TYPE:STRING=Release \
+  -DCMAKE_INSTALL_LIBDIR:PATH="lib" \
+  -DCMAKE_INSTALL_PREFIX:PATH=${PREFIX} \
+  ${SRC_DIR}
```
duncanmmacleod, line 12: You should use `${CMAKE_ARGS}` which is automatically set by the compiler activation and handles most of the conda-specific arguments you might need:

```suggestion
cmake \
  ${CMAKE_ARGS} \
  -DCMAKE_BUILD_TYPE:STRING=Release \
  ${SRC_DIR}
```

recipes/cWB/build.sh
```diff
@@ -0,0 +1,26 @@
...
+# build
+cmake --build . --parallel ${CPU_COUNT} --verbose > build.log
+
+# install
+cmake --build . --parallel ${CPU_COUNT} --verbose --target install
+
+# install activate/deactivate scripts
+for action in activate deactivate; do
+	mkdir -p ${PREFIX}/etc/conda/${action}.d
+        _target="${PREFIX}/etc/conda/${action}.d/activate-${PKG_NAME}.sh"
```
duncanmmacleod, line 23: The deactivate script shouldn't be called `activate` (you probably copied this from me, which I think I have fixed in most places now):

```suggestion
        _target="${PREFIX}/etc/conda/${action}.d/${action}-${PKG_NAME}.sh"
```

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,69 @@
+{% set name = "cWB" %}
+{% set version = "6.4.2" %}
+
+package:
+   name: "{{ name|lower }}"
+   version: "{{ version }}"
+
+source:
+   url: https://gitlab.com/gwburst/public/library/-/archive/cWB-{{ version }}_test_conda_v2/library-cWB-{{ version }}_test_conda_v2.tar.gz
```
duncanmmacleod, line 9: This URL doesn't look like it should be used for real releases, is this a vestige of testing?

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+extra:
+  recipe-maintainers:
+      - rhyspoulton
```
duncanmmacleod, line 69:
```suggestion
    - rhyspoulton
```

recipes/cWB/meta.yaml
rhyspoulton: Yes at the moment, since they haven't done a new release of cWB. This is just an intermediate naming to get cWB working in Conda and after this link will be moved to the official releases.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - liblalsimulation
+    - tcsh
+    - sundials
+
+test:
+  commands:
+    - cwb_help
+    - cwb_watenv verbose
+    - ComputeWDMFourier
+    - root ${CWB_ROOTLOGON_FILE} -e '.q'
```
carterbox, line 57: This package ships header files and library files. Please add an existence test for one of each of these types. Also, what is a ".pcm" file? There are a few of these installed in `$PREFIX/lib`.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+   version: "{{ version }}"
+
+source:
+   url: https://gitlab.com/gwburst/public/library/-/archive/cWB-{{ version }}_test_conda_v2/library-cWB-{{ version }}_test_conda_v2.tar.gz
+   sha256: b5318a2aba3e029a3b4942d7a51d8878d6aa89dee88b34a5096a8de37dce7bad
+   patches:
+     # use rootcling from build env when cross-compiling
+     - rootcling-cross.patch  # [build_platform != target_platform]
+
+build:
```
carterbox, line 15:
```suggestion
build:
  run_exports:
    - {{ pin_subpackage( name|lower, max_pin='x.x.x' ) }}
```
Because this package is shipping shared libraries and headers, it needs a run_export. Above is a *suggestion*; please adjust this pinning expression according to the ABI compatibility guarantees of cWB.

recipes/cWB/meta.yaml
carterbox: This issue should be marked as resolved *after* the URL is moved to the official release.

carterbox (changes requested):

recipes/cWB/meta.yaml
rhyspoulton: The url will be updated when there is a new release of `cWB`, but this will be independent of this package being put into Conda.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make  # [unix]
+    - pkg-config  # [unix]
+    - root_base  # [build_platform != target_platform]
+    - sysroot_linux-64 2.17  # [linux64]
```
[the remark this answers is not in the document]
rhyspoulton: Done

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - healpix_cxx >=3.80
+    - root_base
+    - root
+    - libframel
+    - liblal
+    - liblalinspiral
+    - liblalburst
+    - liblalmetaio
+    - liblalsimulation
+    - tcsh
```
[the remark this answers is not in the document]
rhyspoulton: Yes, a lot of the scripts use `tcsh`.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+  run:
+    - healpix_cxx >=3.80
+    - root_base
+    - root
+    - libframel
+    - liblal
+    - liblalinspiral
+    - liblalburst
+    - liblalmetaio
+    - liblalsimulation
```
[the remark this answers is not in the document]
rhyspoulton: These packages do indeed have a run export so they have been removed.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - tcsh
+    - sundials
```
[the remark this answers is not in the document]
rhyspoulton: So it was added because this code uses root to compile code at runtime and some part of the code links against the sundials library. But after confirming with the developer it seems that `sundials` is not used since the code that uses is not currently supported so this can be safely removed.

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - root_base
+    - libframel
+    - liblal
+    - liblalinspiral
+    - liblalburst
+    - liblalmetaio
+    - liblalsimulation
+  run:
+    - healpix_cxx >=3.80
+    - root_base
```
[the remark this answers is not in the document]
rhyspoulton: I removed `healpix_cxx >=3.80`, but I kept `root_base` as it is required as runtime package

recipes/cWB/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - root
```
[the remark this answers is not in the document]
rhyspoulton: I have removed `root` and instead I am just using `root_base` as the python wrapper is not required.

recipes/cWB/meta.yaml
rhyspoulton: I have added tests for the most used library and header file. I am checking with the developer if the `.pcm` produced by root is used.

rhyspoulton: These `.pcm` files are required by root since they contain information about the cWB classes when calling functions. (https://root-forum.cern.ch/t/cannot-find-my-rdict-pcm-files/21901/2)

rhyspoulton:
@conda-forge/staged-recipes I have addressed the comments and this is ready for another review

conda-forge.yml
```diff
@@ -2,4 +2,6 @@ channel_priority: strict
 azure:
   settings_win:
     pool:
-      vmImage: windows-2019
\ No newline at end of file
+      vmImage: windows-2019
+os_version:
```
carterbox, line 6: Sorry, I was unclear. Do not edit anything outside the recipes directory.

You need to add a new `conda-forge.yml` inside `recipes/cWB` with the os_version stuff. This new `conda-forge.yml` will be copied to the correct place when your recipe is moved to its own feedstock.

carterbox (changes requested):
After conda-forge.yml additions moved into correct place, this recipe is good to go.

recipes/cWB/meta.yaml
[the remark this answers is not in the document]
carterbox: If root_base contains binaries/other files in addition to shared libraries, then this is correct. Otherwise, you may want to depend on "root" at runtime. I'm not sure what the additional things are packaged in "root" vs "root_base".

recipes/cWB/meta.yaml
carterbox: We expect the conda package releases to match the releases of the upstream project, so we cannot release a "6.4.2" package that was not build from a "6.4.2" release. This avoids ambiguity about what was in the source code. In the future, if you want to make pre-releases for this package, we have a mechanism for that, but it is not available for the first release.


https://conda-forge.org/docs/orga/guidelines.html#intended-usage

rhyspoulton: The URL has been updated to the `6.4.2` official release.

rhyspoulton:
@carterbox I have addressed your comment and this is ready to be reviewed

PR #19455, opened 2022-06-27
add UGrid c-lib

ocefpaf:
Adding https://github.com/Deltares/UGrid

@rsignell-usgs this is the first step but UGridpy will be more challenging b/c that is not ready to use an external c-lib for UGrid yet. They are bundling a pre-built .so/.dll files it in their source code and that is a no-no is most packaging ecosystems.

TODO:
- [ ] patch UGridpy to use the c-lib
- [ ] package the `meshkernel` dependency

### DIFF
--- a/recipes/UGrid/build.sh
+++ b/recipes/UGrid/build.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+mkdir build && pushd build
+cmake \
+    -D CMAKE_INSTALL_PREFIX=${PREFIX} \
+    -D netCDFCxx_DIR=${PREFIX}/lib/cmake/netCDF \
+    ${SRC_DIR}
+
+make
+ctest
+make install
--- a/recipes/UGrid/build.sh
+++ b/recipes/UGrid/build.sh
@@ -3,6 +3,7 @@
 mkdir build && pushd build
 cmake \
     -D CMAKE_INSTALL_PREFIX=${PREFIX} \
+    -D CXX_STANDARD=17 \
     -D netCDFCxx_DIR=${PREFIX}/lib/cmake/netCDF \
     ${SRC_DIR}
 
--- a/recipes/UGrid/meta.yaml
+++ b/recipes/UGrid/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "UGrid" %}
+{% set version = "0.0.1.dev" %}
+{% set commit = "79841fe3af2cf60eece683e6cff2d5dc31690698" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Deltares/UGrid/archive/{{ commit }}.zip
+  sha256: a8c42f8d23c947de342dcb3a82102487c9af207a3e10da0d0228b8e9c1df7575
+  patches:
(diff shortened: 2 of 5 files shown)


ocefpaf:
@wolfv sorry to bug you but you are the only C++ expert I know. This one passes locally but it is failing in our CIs, any advice for me here? All of my patches here are "Google-SO" based and I have no idea of that I'm doing.

wolfv:
indeed, weird error (can't easily reproduce the compiler error on Godbolt, either).

I have some ideas on what to try:

Easiest: just remove the unique_ptr in these tests, and use:

```
    const double[] node_y{0, 0, 1, 1, 2, 2, 3, 3, 0, 1, 2, 3, 0, 1, 2, 3};
    mesh2d.node_y = node_y;
...
```


Or, instead of `std::unique_ptr<double> const node_y(new double[]{...});` you could try `std::unique_ptr<double[]> const node_y(new double[]{...});` (not sure if that would make a difference ...).

recipes/UGrid/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+mkdir build && pushd build
+cmake \
+    -D CMAKE_INSTALL_PREFIX=${PREFIX} \
```
marcelotrevisani, line 5:
```suggestion
    -D CMAKE_INSTALL_PREFIX=${PREFIX} \
    -D CXX_STANDARD=14 \
```

recipes/UGrid/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
+    - make
+  host:
+    - boost
+    - boost-cpp
+    - netcdf-cxx4
```
marcelotrevisani, line 28: you might need limb here as well
```suggestion
    - netcdf-cxx4
    - openlibm
```

marcelotrevisani, line 28:
```suggestion
    - netcdf-cxx4 >=4.3.1
```

recipes/UGrid/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  run:
+    - boost
+    - boost-cpp
+    - netcdf-cxx4
```
marcelotrevisani, line 32:
```suggestion
    - netcdf-cxx4 >=4.3.1
```

recipes/UGrid/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - patches/use_runtime_error.patch
+
+build:
+  skip: True  # [win]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
```
marcelotrevisani, line 23:
```suggestion
    - cmake >=3.19
```

ocefpaf:
@rsignell-usgs both experts on cpp I know tried and failed here. I don't see the point on pursuing this one b/c upstream does not have an official release, proper CI, and transparent wheel builds. Also, both xugrid and uxarray, the former is also from Deltares btw, looks more promising and easier to use/package.

PR #19465, opened 2022-06-28
Add pyneqsys to conda-forge

ProfLeao:
Adds pyneqsys package to recipes.


ProfLeao:
Need to fix file location

PR #19467, opened 2022-06-28
Add conda recipe for sed-plume

mcflugen:
Checklist
- [x] Build number is 0.


mcflugen:
@conda-forge/help-python This is ready for review. Thanks!

mcflugen:
@conda-forge/help-python Sorry for the false alarm. This ready for review now—for reals. Thanks!

mcflugen:
@conda-forge/help-python Please let me know if there is anything I can do to move this along. Thanks!

recipes/sed-plume/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - python
+    - pip
+    - cython
+    - cythongsl
+    - numpy
+    - gsl
+  run:
+    - python
+    - gsl
+    - libopenblas
```
carterbox, line 34:
```suggestion
```
BLAS is not a python library or a binary. It should not be in the run requirements without also being in the host section. If your package links against BLAS, please follow the BLAS directions in the docs.

https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=blas#blas

But also, does this package link to BLAS? The linker doesn't detect links to BLAS on any platform.
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=529045&view=logs&j=e35d4f76-8ff2-5536-d795-df91e63eb9f7&t=fa7b4b17-b6ff-5c9c-8cfc-15f888c92310&l=4853

recipes/sed-plume/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - tomlkit
+
+test:
+  imports:
+    - plume
+  commands:
+    - plume --version
+    - plume --help
+    - plume generate --help
+    - plume run --help
```
carterbox, line 49:
```suggestion
test:
  requires:
    - pip
  imports:
    - plume
  commands:
    - plume --version
    - plume --help
    - plume generate --help
    - plume run --help
    - pip check
```

carterbox (changes requested):

recipes/sed-plume/meta.yaml
mcflugen: @carterbox I had added *libopenblas* as a requirement to solve a linker error I was getting but that error seems to have gone away!

mcflugen:
@carterbox I added *netcdf4* as a requirement to fix the *pip check* error. *landlab* no longer uses *netcdf* directly but instead uses it through *xarray*. I'm not sure why the *netcdf* library is no longer coming with *xarray*.

recipes/sed-plume/meta.yaml
carterbox: Looks like your project includes the following headers

```
#include "gsl/gsl_cblas.h"
#include "gsl/gsl_blas_types.h"
#include "gsl/gsl_blas.h"
```
And the logs show that `centerline.cpython-39-x86_64-linux-gnu.so` links to libopenblas.

However, I don't know enough about your package to tell whether this link to libopenblas is because there are actually libopenblas symbols included in those headers or if there is just a mistake in your build process that accidentally links to libopenblas when it is not needed.

carterbox (changes requested):
This package is installing a "tests" module directly to site-packages. This is probably an error. You will need to patch the upstream package to not include tests in the global namespace.

Probably, you should add an exclude statement to this find_package config.

https://github.com/mcflugen/plume/blob/66b4c80517891a14abe1bed781adf8415d4cfb2e/pyproject.toml#L80

recipes/sed-plume/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - cython
+    - cythongsl
+    - numpy
+    - gsl
+    - libblas
+  run:
+    - python
+    - gsl
+    - landlab >=2
+    - netcdf4
```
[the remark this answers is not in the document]
mcflugen: @carterbox *landlab* no longer uses *netcdf4* directly (i.e. it doesn't include anything like `import netcdf4`). Instead, it uses *xarray* for reading and writing to *netCDF* files. However, as you say, *xarray* may in the future change its implementation of its netcdf io functions to not use the *netcdf4* package so I don't think it should include it in its dependencies. In fact, I'm now not certain *xarray* does use *netcdf4*.

recipes/sed-plume/meta.yaml
mcflugen: Interesting. I also would have thought that if *libopenblas* was required it would have been listed as a dependency for *gsl*.

recipes/sed-plume/meta.yaml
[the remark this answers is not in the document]
mcflugen: Ah, never mind. I see that *netcdf4* is an optional dependency for *xarray that's required only if you want to read/write netcdf files.

mcflugen:
@carterbox Do you think this is ready to go now? I've patched *landlab* so that it doesn't install its tests in *site-packages*, fixed its dependencies to include *netcdf4*, and added `pip check` to *sed-plume*'s tests.

mcflugen:
@conda-forge/help-python I think this is ready to go now. Thanks!

mcflugen:
@conda-forge/staged-recipes I fear my pull request may have gotten lost in the shuffle. Anyway, I think it's is ready to go now. Thanks!

PR #19499, opened 2022-06-30
Add recipe for ringdecomposerlib

awvwgk:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/ringdecomposerlib/meta.yaml
+++ b/recipes/ringdecomposerlib/meta.yaml
@@ -38,6 +38,7 @@ about:
     Calculation of Unique Ring Families (URFs) and other cycle bases
     (Relevant Cycles (RC), Smallest Set of Smallest Rings (SSSR)).
   home: https://github.com/rareylab/RingDecomposerLib
+  doc_url: https://ringdecomposerlib.readthedocs.io/en/latest/
 
 extra:
   recipe-maintainers:
--- a/recipes/ringdecomposerlib/meta.yaml
+++ b/recipes/ringdecomposerlib/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "RingDecomposerLib" %}
+{% set version = "1.1.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/rareylab/RingDecomposerLib/archive/refs/tags/v{{ version }}_rdkit.tar.gz
+  sha256: 944b5816712a48bbf88aa25d4300ce11871ddf6e971218eac08f90ed2192f715
+  patches:
+    - language_c.patch
+
+build:
+  number: 0
+
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin='x') }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - cmake
+    - ninja
+
+test:
+  commands:
+    - test -f $PREFIX/lib/lib{{ name }}${SHLIB_EXT}  # [unix]
+    - if not exist %LIBRARY_PREFIX%\\bin\\{{ name }}.dll exit 1  # [win]
+    - if not exist %LIBRARY_PREFIX%\\lib\\{{ name }}.lib exit 1  # [win]
+    - test -f $PREFIX/include/{{ name }}.h  # [unix]
+    - if not exist %LIBRARY_PREFIX%\\include\\{{ name }}.h exit 1  # [win]
+
+about:
+  license: BSD-3-Clause
+  license_file: LICENSE
+  summary: |
+    Calculation of Unique Ring Families (URFs) and other cycle bases
+    (Relevant Cycles (RC), Smallest Set of Smallest Rings (SSSR)).
+  home: https://github.com/rareylab/RingDecomposerLib
+
+extra:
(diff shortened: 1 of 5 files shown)


recipes/ringdecomposerlib/meta.yaml
```diff
@@ -0,0 +1,44 @@
+{% set name = "RingDecomposerLib" %}
+{% set version = "1.1.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/rareylab/RingDecomposerLib/archive/refs/tags/v{{ version }}_rdkit.tar.gz
```
saraedum, line 9: What is the meaning of `rdkit` here? It's part of the version number it seems.

recipes/ringdecomposerlib/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - test -f $PREFIX/include/{{ name }}.h  # [unix]
+    - if not exist %LIBRARY_PREFIX%\\include\\{{ name }}.h exit 1  # [win]
+
+about:
+  license: BSD-3-Clause
+  license_file: LICENSE
+  summary: |
+    Calculation of Unique Ring Families (URFs) and other cycle bases
+    (Relevant Cycles (RC), Smallest Set of Smallest Rings (SSSR)).
+  home: https://github.com/rareylab/RingDecomposerLib
```
saraedum, line 40:
```suggestion
  home: https://github.com/rareylab/RingDecomposerLib
  doc_url: https://ringdecomposerlib.readthedocs.io/en/latest/
```

saraedum:
I see that you decided not to package the Python wrapper. What was the problem here? (Maybe you could mention that very briefly in the `meta.yaml`.)

recipes/ringdecomposerlib/meta.yaml
awvwgk: I guess this [RDKit](https://www.rdkit.org/) is meant, not sure why the label was included in tag. The CMake build file sets for all available releases (1.1, 1.1.1, 1.1.2, and 1.1.3(\_rdkit)) the version 1.0

awvwgk: There is no 1.1.3 release without the label and I think including the label in the version will confuse the conda/mamba.

awvwgk:
> I see that you decided not to package the Python wrapper. What was the problem here? (Maybe you could mention that very briefly in the `meta.yaml`.)

It's a Cython based wrapper, I gave it a quick try to see whether it works out of the box but decided not to investigate the issue further since I only need the compiled library and I don't have much experience with Cython on conda-forge yet.

awvwgk:
@conda-forge/help-c-cpp This recipe is ready for review again

recipes/ringdecomposerlib/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin='x') }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - cmake
+    - ninja
+
```
carterbox, line 25:
```suggestion

# NOTE: Python wrapper not built because it statically links to library,
# so the build steps needs patching. (static linking not allowed on conda-forge)

```
It looks like the python wrapper statically links to the library, which is not allowed. I think it's OK to wait for someone who wants the python wrapper to request it. Building the python wrapper with require some patching.


https://github.com/rareylab/RingDecomposerLib/blob/3a7ff93de0d9c4f6a5661508549c6063573f39c7/src/python/CMakeLists.txt#L21

recipes/ringdecomposerlib/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+test:
+  commands:
+    - test -f $PREFIX/lib/lib{{ name }}${SHLIB_EXT}  # [unix]
+    - if not exist %LIBRARY_PREFIX%\\bin\\{{ name }}.dll exit 1  # [win]
+    - if not exist %LIBRARY_PREFIX%\\lib\\{{ name }}.lib exit 1  # [win]
```
carterbox, line 30:
```suggestion
    - test -f $PREFIX/lib/lib{{ name }}${SHLIB_EXT}  # [unix]
    - ! test -f $PREFIX/lib/lib{{ name }}Static.a  # [unix]
    - if not exist %LIBRARY_PREFIX%\\bin\\{{ name }}.dll exit 1  # [win]
    - if not exist %LIBRARY_PREFIX%\\lib\\{{ name }}.lib exit 1  # [win]
    - if exist %LIBRARY_PREFIX%\\lib\\{{ name }}Static.lib exit 1  # [win]
```
Your outputs contain static libraries; these are not allowed. Please test that the static libs are excluded from the outputs. You can either add a call to `rm` as the last line in the build scripts or patch the CMake files/options so these are not generated. Hopefully, there is already a CMake option to skip building the static libs in the first place.

https://github.com/conda-forge/cfep/blob/main/cfep-18.md

carterbox (changes requested):

awvwgk:
@conda-forge/help-c-cpp I removed the static libraries from the distribution.

PR #19501, opened 2022-06-30
Add inputs via grayskull

Tobias-Fischer:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/inputs/meta.yaml
+++ b/recipes/inputs/meta.yaml
@@ -10,16 +10,16 @@ source:
   sha256: a31d5b96a3525f1232f326be9e7ce8ccaf873c6b1fb84d9f3c9bc3d79b23eae4
 
 build:
-  noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
 requirements:
   host:
-    - python >=3.6
+    - python
     - pip
   run:
-    - python >=3.6
+    - python
+    - pyobjc-framework-quartz  # [osx]
 
 test:
   imports:
--- a/recipes/inputs/meta.yaml
+++ b/recipes/inputs/meta.yaml
@@ -1,0 +1,40 @@
+{% set name = "inputs" %}
+{% set version = "0.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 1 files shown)


Tobias-Fischer:
@conda-forge/help-python - this is ready for review. As it's a noarch package, the osx-64 error does not matter.

Tobias-Fischer:
Kind reminder @conda-forge/help-python @conda-forge/staged-recipes. Thanks :)!

ocefpaf:
> @conda-forge/help-python - this is ready for review. As it's a noarch package, the osx-64 error does not matter.

Not necessarily, this package won't work on osx-64 without `pyobjc-framework-quartz`. See https://github.com/zeth/inputs/blob/a46681dbf77d6ab07834f550e5855c1f50701f99/setup.py#L31.

I'm on the fence between a better users experience and removing the noarch here or letting it "broken" on osx and hope the user knows that they need to install an extra dependency. I'm more inclined to the former though.

Tobias-Fischer:
Just for me to understand: Is the plan to convert `pyobjc-framework-quartz` to a virtual package which leads to an empty package on `win` and `linux`, so we can add it as a dependency for `inputs` (i.e. this PR)?

carterbox:
I'm also confused about how the `__osx` package helps here. Can someone explain? 🤔

This package doesn't qualify for noarch because it has a platform specific dependency. The simplest solution is to remove noarch and use a selector so the additional package is added for osx.

As was suggested, we could also add empty build variants of `pyobjc-framework-quartz`, so that selectors are not needed for this recipe, but this solution also doesn't need us to use `__osx`.

carterbox (changes requested):
Waiting for switch away from noarch or an open PR at https://github.com/conda-forge/pyobjc-framework-quartz-feedstock to add empty metapackages.

Tobias-Fischer:
Ok I moved away from `noarch` and added the osx-specific dependency. It would be great if you could have another look. We can always, later on, change the recipe if `pyobjc-framework-quartz` provides a virtual package in the future.

BastianZim:
By adding an impossible version on osx it couldn't have been installed there. But I just realised that that also obviously applies to the other OSs so this approach is definitely correct — my bad.

PR #19523, opened 2022-07-04
Pyreports

thewchan:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/pyreports/meta.yaml
+++ b/recipes/pyreports/meta.yaml
@@ -1,6 +1,5 @@
 {% set name = "pyreports" %}
-{% set version = "1.2.1" %}
-
+{% set version = "1.4.0" %}
 
 package:
   name: {{ name|lower }}
@@ -8,23 +7,24 @@ package:
 
 source:
   url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyreports-{{ version }}.tar.gz
-  sha256: b7e39ea93ea438e9af339f30af1d65cddbda30c8c12666c63cfc8c379cf44beb
+  sha256: 59873c391096609c103232fe7617760810824e81c9bef366b26f1944d05f2acb
 
 build:
-  number: 0
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
+  number: 0
 
 requirements:
   host:
     - pip
-    - python >=3.6
+    - python >=3.7
   run:
     - ldap3
     - mysql-connector-python
-    - psycopg2-binary
+    - nosqlapi
+    - psycopg2
     - pymssql
-    - python >=3.6
+    - python >=3.7
     - tablib
 
 test:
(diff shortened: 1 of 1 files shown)


thewchan:
@conda-forge/staged-recipes ready for review!

recipes/pyreports/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pyreports-{{ version }}.tar.gz
+  sha256: 59873c391096609c103232fe7617760810824e81c9bef366b26f1944d05f2acb
+
+build:
+  skip: true  # [win or py<37]
```
ocefpaf, line 13: What is the reason for skipping it on Windows? I'm just wondering if we can make it noarch.

thewchan: I don't remember why exactly but there's something in the source code that made this fail to build for win.

thewchan: done, I've reverted it

thewchan:
@ocefpaf The error is now logged

recipes/pyreports/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - psycopg2
+    - pymssql
+    - python >=3.7
+    - tablib
+
+test:
+  imports:
+    - pyreports
+  commands:
+    - pip list | grep -E "{{ name }}\s+{{ version }}"
```
ocefpaf, line 34: This is probably not necessary.

recipes/pyreports/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+requirements:
+  host:
+    - pip
+    - python >=3.7
+  run:
+    - ldap3
+    - mysql-connector-python
+    - nosqlapi
+    - psycopg2
+    - pymssql
```
ocefpaf, line 26: B/c it is a missing dependency on Windows we can keep this as noarch instead of skipping Windows.

recipes/pyreports/meta.yaml
thewchan: Do you mean the grep test? I must have changed it for some reason... Although I can change back to pip check instead

recipes/pyreports/meta.yaml
thewchan: Ok

recipes/pyreports/meta.yaml
thewchan: Okay, I've reverted it

thewchan: yes looks like running pip check fails the tests thats why I switched to grep

thewchan: oh wait, or maybe it's something else?

ocefpaf:
It is OK to drop the pip check test in this case. You already depend on `psycopg2` and that seems to be enough for this library to function.

thewchan:
@ocefpaf no need for grep check? I usually do that if pip check isn't working

ocefpaf:
> @ocefpaf no need for grep check? I usually do that if pip check isn't working

No need. It would never fail, right? I don't see the point in adding it.

PR #19564, opened 2022-07-07
Add coin-or-ipopt

jschueller:
Renames the existing ipopt package (https://github.com/conda-forge/ipopt-feedstock) for consistency with other coin-or packages 
@conda-forge/help-c-cpp @conda-forge/ipopt


moorepants:
This seems like an unnecessary change and will mostly cause issues because there are recipes, instructions, etc. downstream of this package that would all have to be modified. I'm -1 to changing the name.

jschueller:
fine

PR #19570, opened 2022-07-07
Add skia-pathops

naveen521kk:
Checklist
- [ ] ~~Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).~~ vends a portion of libskia, but justification is discussed in comments below
- [x] Build number is 0.


naveen521kk:
The Linux error is because pre-build `gn` binaries require at least `GLIBC_2.18`, there's a version of `gn` in the source which doesn't require that, I'll patch the source to use that, see https://github.com/fonttools/skia-builder/commit/8aca8a5148f472a0e6a17e87f2680ccd53b82a1a

naveen521kk:
This PR is, ready for review. Windows builds are failing because `gn` isn't there on Windows, I've made a PR to the feedstock https://github.com/conda-forge/gn-feedstock/pull/10

carterbox (commented):
My understanding is that skia-pathops is a python wrapper around the CXX skia library (libskia), and the libskia is being linked in statically to skia-pathops.

Because conda is capable of delivering non-python packages, we have typically require that CXX libraries are linked dynamically into python modules.

Can you justify why libskia should be statically linked in this case? Are there other python wrappers of the skia CXX library? Do other non-python projects use the skia library?

naveen521kk:
> My understanding is that skia-pathops is a python wrapper around the CXX skia library (libskia)

Yes. It only wraps a part of `skia` (only the pathops module).

> Can you justify why libskia should be statically linked in this case?

AFAIK, Skia doesn't provide any ABI/API guarantees and asks to use the "tip-of-tree" of the skia development branch https://skia.org/docs/user/release/. So every time skia is updated all the extension needs to be rebuilt, not just that, if they change the API some extensions won't be able to build. That's why I thought having it statically linked built from the source that was tested by the authors is better. Also, I'm not sure how rebuilds are done here, if it's done automatically (that is a bot opens a PR or something like that), then I think having a separate package should mostly work for this package (as the pathops API is pretty stable, AFAIK).

> Are there other python wrappers of the skia CXX library?

There exists [skia-python](https://pypi.org/project/skia-python/). I don't think there exists any other project that wraps skia.

> Do other non-python projects use the skia library?

Skia was built for Chromium and there is even a [page is wikipedia](https://en.wikipedia.org/wiki/Skia_Graphics_Engine#Application) describing the users of this library. Other than that I know there's a wrapper around the skia library for Rust called [tiny-skia](https://crates.io/crates/tiny-skia) and for .NET called [SkiaSharp](https://github.com/mono/SkiaSharp).

carterbox:
> AFAIK, Skia doesn't provide any ABI/API guarantees and asks to use the "tip-of-tree" of the skia development branch https://skia.org/docs/user/release/.

Because of no API/ABI guarantees and because the submodule commit location for libskia is not one of the "stable" canvaskit or chrome releases, I agree that it's appropriate to not package libskia separately.

There also seems to be few other packages that would reuse libskia if it was published in its entirety.

recipes/skia-pathops/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  patches:
+    # use ninja from conda instead
+    - disable-virtualenv-creation.patch
+    # use gn from conda
+    - use-gn-from-conda.patch
+    # use cc, cxx set by the environment
+    - inherit-cc-cxx-from-environment.patch
+
+build:
+  skip: true  # [py<37]
```
carterbox, line 20:
```suggestion
  skip: true  # [py<37]
  # FIXME: Unskip Windows when gn becomes available for Windows
  skip: true  # [win]
```

naveen521kk:
Thanks!

PR #19599, opened 2022-07-08
add functorch

### LINKED ISSUE #19569: Package request: functorch
### Package name

functorch

### Package version

Newest

### Package website

https://github.com/pytorch/functorch

### Package availability

https://github.com/pytorch/functorch/releases/tag/v0.2.0

### Additional comments

_No response_

### Package is not available

- [X] The package is not available on conda-forge.

### No previous issues or open PRs

- [X] No previous issue exists and no PR has been opened.

ngam:
fixes #19569 

Checklist
- [x] Build number is 0.


recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - "$RPATH/libc10.dylib"
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+  host:
+    - pip
+    - python
+    - pytorch
```
carterbox, line 35:
```suggestion
    - pytorch >=1.10.0,<2
```
There is a minimum pytorch version requirement.

https://github.com/pytorch/functorch/blob/9b96f14e65ffdf64a28416054fe536ed14297fe9/setup.py#L48

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - python
+    - pytorch 
+
+test:
+  imports:
+    - functorch
+  commands:
+    - pip check
+  requires:
+    - pip
```
carterbox, line 46:
```suggestion
test:
  files:
    - test/
  imports:
    - functorch
  commands:
    - pip check
    - pytest test/test_vmap.py -v
    - pytest test/test_eager_transforms.py -v
  requires:
    - pip
    - pytest
```
Since this is a compiled module and "under heavy development", could you please add one of the short sanity tests from the README to check that it runs correctly?

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/pytorch/functorch/archive/v{{ version }}.tar.gz
+  sha256: 3b297f4acf1d61fd95b86bd2ec1805bc9c236bb0fb341a42c2300a4b3d497fbd
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
```
carterbox, line 15:
```suggestion
  skip: true  # [win]
  skip: true  # [cuda_compiler_version != None]
```
Doesn't seem like this package needs the cuda compiler, but apparently our CI things any mention of "cuda" should activate the cuda builds.

carterbox (changes requested):

recipes/functorch/meta.yaml
hmaarrfk: I don't think this is correct. If it does link to the pytorch ABI the migrator will skip this. I think you can leave it in a comment, but the pinnings will take care of this for you.

carterbox: I expected the global pinning a specified pinning to be combined. i.e.

```
pytorch >=1.10,<2,1.11
```

Where 1.11 is the global pinning for pytorch. Because you can set a python version range on a package and the incompatible python builds are skipped? Are you saying the migrator won't migrate this recipe to pytorch =1.12 or that it won't migrate to pytorch=2.0 because I added the major version requirement?

hmaarrfk: Lets see what happens. https://github.com/conda-forge/torchvision-feedstock/pull/56

hmaarrfk: last i remember, the migrator looks for pinnings that aren't set. It will provide a pinning for those. I had to go through some gymnasitcs where a requirement is listed twice, once without any constraints, the other with the constraints.

hmaarrfk: https://github.com/conda-forge/torchvision-feedstock/pull/56#pullrequestreview-1033456278

hmaarrfk: to be specific, i don't think it will migrate to 1.12 AND it will ignore the pin of 1.11.

The problem is that it might solve, at compile time, to pytorch 1.10, and build a version of the package that is incompatible with the rest of the ecosystem.

carterbox: This is good to know. Thanks for taking the time to research that for me @hmaarrfk. 😄 

This is a problem because if you don't check the global pinning for every dependency, you could be overriding it without knowing!

hmaarrfk: Yes. This is true.

carterbox (commented):
My only remaining concern is that it looks like this library is using pybind11 in the build logs, but pybind11 is not listed in the requirements of this recipe because it is using the pybind11 headers from the pytorch/includes directory instead of from the pybind11 package.

@conda-forge/core, do you have advice on how to handle this?

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "functorch" %}
+{% set version = "0.1.1" %}
```
zou3519, line 2: for the future: functorch 0.2.x and PyTorch 1.12.x have some weird c++ ABI interactions (see warning in https://pytorch.org/functorch/stable/install.html#pip) that we're still working on resolving. But this recipe for 0.1.1 seems reasonable to me.

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - pip
+    - python
+    # minimum supported pytorch version is 1.10
+    - pytorch
+  run:
+    - python
+    - pytorch
```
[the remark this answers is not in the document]
hmaarrfk: yes.

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - "$RPATH/libc10.dylib"
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - pip
+    - python
+    # minimum supported pytorch version is 1.10
+    - pytorch
```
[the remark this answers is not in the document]
hmaarrfk: Yeah.... this will be set by the global pinnings for conda-forge. today, they are set to pytorch 1.11.0

recipes/functorch/meta.yaml
hmaarrfk: understood. I think this will be caught by some tests of ours. Generally speaking, that type of ABI incompatibility is what we hope to solve at conda-forge. 

This recipe will be built, and tested, with the same pytorch libraries.

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  source_files:
+    - test
+  requires:
+    - pip
+  imports:
+    - functorch
+  commands:
+    - pip check
+    # sanity check from readme
+    - python -c "import torch; from functorch import vmap; x = torch.randn(3); y = vmap(torch.sin)(x); assert torch.allclose(y, x.sin())"
```
hmaarrfk, line 52: Can you add this to a run_test.py file? it just makes it easier to read.

hmaarrfk: https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html

recipes/functorch/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "functorch" %}
```
hmaarrfk, line 1: its probably just easier to get rid of this `name` variable.

recipes/functorch/meta.yaml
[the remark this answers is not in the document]
carterbox: This is my mistake. I didn't realize that the main branch lists different requirements than each of the release tags.

carterbox:
Devendoring of pytorch may occur in the future which will require rebuilding this package.

PR #19613, opened 2022-07-11
Add oso package recipe

alexpilot11:
Checklist
- [x] Build number is 0.


recipes/oso/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  commands:
+    - pip check
+    - make python-test
+  requires:
+    - pip
+    - pytest-cov
+    - ruamel_yaml
+    # fails pip check: https://github.com/conda-forge/importlib_metadata-feedstock/issues/88
+    - importlib-metadata !=4.7.1
+    - cffi >=1.0.0
```
bollwyvl, line 52: as `cffi` appears in host, and is needed in `test` to pass `pip check`, it probably belongs in `run` (aside from the `importlib-metadata` snafu). Basically anything in `install_requires` which they load from:

https://github.com/osohq/oso/blob/main/languages/python/oso/requirements.txt#L1

recipes/oso/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/osohq/oso/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 412b62ca46b649cd2b0ddfd9027180eeda9aa8d4f059da44978fa56ac103e90a
+
+build:
+  number: {{ build_number }}
+  script:
+    - bash {{ RECIPE_DIR }}/build_oso.sh
```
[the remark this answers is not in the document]
alexpilot11: checked in now

recipes/oso/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}                    # [unix]
+    - {{ compiler('cxx') }}                  # [unix]
+    - {{ compiler('m2w64_c') }}              # [win]
```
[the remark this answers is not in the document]
alexpilot11: I swapped to `msvc_c` instead of `m2w64_c`, etc and I got this: `conda_build.exceptions.DependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: {MatchSpec("msvc_cxx_win-64"), MatchSpec("msvc_c_win-64")}`

Not sure if that's the wrong string to use there or what though but I've swapped back for now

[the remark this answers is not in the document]
bollwyvl: just `c`.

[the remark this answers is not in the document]
bollwyvl: example: https://github.com/conda-forge/python-wasmer-split-feedstock/blob/main/recipe/meta.yaml

alexpilot11:
currently my tests are failing locally because their tests are doing a string compare on an escaped character. it seems when I run with `conda mambabuild recipes/oso` it escapes the string differently. I cloned the `oso` repo and ran the same tests with the same command, and it passes. Any thoughts on why this might be?

bollwyvl:
```
>       assert str(e.value).startswith(
            "'\\u{0}' is not a valid character. Found in this is not allowed\\u{0} at line 2, column 17"
        )
E       assert False
E        +  where False = <built-in method startswith of str object at 0x7f662acb3160>("'\\u{0}' is not a valid character. Found in this is not allowed\\u{0} at line 2, column 17")
E        +    where <built-in method startswith of str object at 0x7f662acb3160> = '\'\\0\' is not a valid character. Found in this is not allowed\\0 at line 2, column 17:\n\t002:     f(a) if a = "this is not allowed\x00\n\t                     ^\n\n\tGet help with Oso from our engineers: https://help.osohq.com/error/InvalidTokenCharacter'.startswith
E        +      where '\'\\0\' is not a valid character. Found in this is not allowed\\0 at line 2, column 17:\n\t002:     f(a) if a = "this is not allowed\x00\n\t                     ^\n\n\tGet help with Oso from our engineers: https://help.osohq.com/error/InvalidTokenCharacter' = str(InvalidTokenCharacter('\'\\0\' is not a valid character. Found in this is not allowed\\0 at line 2, column 17:\n\t002:     f(a) if a = "this is not allowed\x00\n\t                     ^\n\n\tGet help with Oso from our engineers: https://help.osohq.com/error/InvalidTokenCharacter'))
E        +        where InvalidTokenCharacter('\'\\0\' is not a valid character. Found in this is not allowed\\0 at line 2, column 17:\n\t002:     f(a) if a = "this is not allowed\x00\n\t                     ^\n\n\tGet help with Oso from our engineers: https://help.osohq.com/error/InvalidTokenCharacter') = <ExceptionInfo InvalidTokenCharacter('\'\\0\' is not a valid character. Found in this is not allowed\\0 at line 2, column 17:\n\t002:...t                     ^\n\n\tGet help with Oso from our engineers: https://help.osohq.com/error/InvalidTokenCharacter') tblen=6>.value
```

As it looks like this is comparing error value text, it may not take all possible python versions into account.

recommend skipping with `-k`, e.g.

```yaml
commands:
  - pytest ... -k "not parser_errors"
```

If you only need to skip tests on some platform/pythons, you can:
- use jinja template vars to build up a list of skips, then squash them into a filter
  - while this works, it can get messy
- use the well-known (by `conda-build`) `run_test.py`, and build up the arguments dynamically ([example](https://github.com/conda-forge/ipykernel-feedstock/blob/main/recipe/run_test.py)

recipes/oso/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - ruamel_yaml
+    - jq
+    - sqlalchemy
+
+about:
+  home: https://github.com/osohq/oso
+  summary: 'Oso is a batteries-included framework for building authorization in your application.'
+  license: Apache
+  license_family: Apache
+  license_file: LICENSE
```
bollwyvl, line 60: As this bundles a bunch of third-party stuff in the rust static binary, you'll want to use [cargo-bundle-licenses](https://github.com/sstadick/cargo-bundle-licenses) to capture all of the built static binary licenses. There are many [examples in other feedstocks](https://github.com/search?q=org%3Aconda-forge+cargo-bundle-licenses&type=code).

As annoying as this is, it's _far far better_ than the bad old days of curating them by hand.

alexpilot11:
looks like my local was on `main` while I had pointed the recipe to `v0.26.1`

[this commit](https://github.com/osohq/oso/pull/1576/commits/d04670abfaaf6da5e3df242fe61ac0a6d293cc27) seemed to fix the test (that was failing when I checked out the tag locally). question is, should I bump to the `sqlalchemy-v0.26.1` release or just skip the test?

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+    - cffi >=1.0.0
+  run:
+    - python >=3.6
+    - packaging ~=20.9
```
bollwyvl, line 25: don't use the `~` operator: different python build tools have it map to different things, and you don't awlayas get the same thing from conda.

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - pip
+    - pytest-cov
+    - oso
+    - sqlalchemy
+    - flask-sqlalchemy
+
+about:
+  home: https://github.com/osohq/oso/tree/main/languages/python/sqlalchemy-oso
+  summary: 'sqlalchemy-oso the SqlAlchemy integration for oso, the batteries-included framework for building authorization in your application.'
+  license: Apache
```
bollwyvl, line 53: Apache-2.0

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  license_family: Apache
```
bollwyvl, line 54: you don't even need to specify this (gets filled in with a proper SPDX above)

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "sqlalchemy-oso" %}
+{% set build_number = 0 %}
```
bollwyvl, line 2: remove this (not being used)

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+{% set version = "0.26.1" %}
+
+package:
+  name: {{ name|lower }}
```
bollwyvl, line 6: remove `name` variable: it only gets used here, and probably isn't going to change.

recipes/oso/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+  doc_url: https://docs.osohq.com/
+  dev_url: https://github.com/osohq/oso
+
+extra:
+  feedstock-name: oso
```
bollwyvl, line 66: only need this with multi-outputs

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  home: https://github.com/osohq/oso/tree/main/languages/python/sqlalchemy-oso
+  summary: 'sqlalchemy-oso the SqlAlchemy integration for oso, the batteries-included framework for building authorization in your application.'
+  license: Apache
+  license_family: Apache
+  license_file: LICENSE
+  doc_url: https://docs.osohq.com/python/reference/api/sqlalchemy.html
+  dev_url: https://github.com/osohq/oso
+
+extra:
+  feedstock-name: oso
```
bollwyvl, line 60: can remove this line

alexpilot11:
it looks like `jq` is needed for the windows build but the feedstock doesn't support a windows build. issues [here](https://github.com/conda-forge/jq-feedstock/pull/16) and [here](https://github.com/conda-forge/jq-feedstock/pull/15) that haven't been touched in a while. I also think that if those got in, I'd probably need to revert to the mingw compilers. thoughts on that?

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - python >=3.6
+    - pip
+    - cffi >=1.0.0
+  run:
+    - python >=3.6
+    - packaging ==20.9
+    # fails pip check: https://github.com/conda-forge/importlib_metadata-feedstock/issues/88
+    - importlib-metadata !=4.7.1
+    - cffi >=1.0.0
+    - oso >=0.26.1
```
bollwyvl, line 27: this is _probably_ going to be `- oso =={{ version }}.*`

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+test:
+  source_files:
+    - languages/python/sqlalchemy-oso/tests
+  imports:
+    - oso
+    - polar
+    - sqlalchemy
+    - flask_sqlalchemy
+    - sqlalchemy_oso
```
bollwyvl, line 39: this is the only one you're actually testing, probably remove the others

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+requirements:
+  host:
+    - python >=3.6
+    - pip
+    - cffi >=1.0.0
+  run:
+    - python >=3.6
+    - packaging ==20.9
+    # fails pip check: https://github.com/conda-forge/importlib_metadata-feedstock/issues/88
+    - importlib-metadata !=4.7.1
```
bollwyvl, line 25: probably move this to a `test/requires` unless it's actually called out

bollwyvl:
> `jq` is needed for the windows build

welcome to the wonderful world of even turning on a non-trivial windows build chain.

the dude who made the first failing pr was a _giant_ in the mingw ecosystem (sometimes usernames are accurate), so unless things have gotten a **lot** better, i can't recommend taking that up. you also can't ask him for help, as he passed a few years back.

in the meantime: a cursory inspection suggests it's probably worth avoiding `jq` in the makefile, as it only appears to use it for some cosmetic tasks. You can do it by:
- generating an actual patch and [applying it in source](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#patches) or using  
- patching it out "hot" with `sed` or a `python` script
- ignoring the root makfile all together, as it _looks_ like it's just calling


```bash
cargo build -p polar-c-api
cd languages/python/oso 
make build
```

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+    - cffi >=1.0.0
+  run:
+    - python >=3.6
+    - packaging ==20.9
```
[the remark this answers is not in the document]
alexpilot11: they use `packaging~=20.9` but you said not to use it [here](https://github.com/conda-forge/staged-recipes/pull/19613#discussion_r920347784)

should I just do `packaging >=20.9,<21.0`?

recipes/sqlalchemy-oso/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    # fails pip check: https://github.com/conda-forge/importlib_metadata-feedstock/issues/88
+    - importlib-metadata !=4.7.1
+    - cffi >=1.0.0
+    - oso >=0.26.1
+    - sqlalchemy >=1.3.17,<1.5
+    - flask-sqlalchemy
```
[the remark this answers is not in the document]
alexpilot11: I think because one of the tests imports [here](https://github.com/osohq/oso/blob/bc0c5b77e150df8fb273489525b1ff51fb53fe4b/languages/python/sqlalchemy-oso/tests/test_flask.py#L5) which raises an error [here](https://github.com/osohq/oso/blob/bc0c5b77e150df8fb273489525b1ff51fb53fe4b/languages/python/sqlalchemy-oso/sqlalchemy_oso/flask.py#L14)  before skipping the flask tests it has become a hard dependency? 

might be misunderstanding something here

recipes/sqlalchemy-oso/meta.yaml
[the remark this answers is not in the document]
bollwyvl: yeah, `>=` ranges are much more precise. indeed, poetry folk creates so many issues with these that it just generates a `setup.py` that translates them into these operators.

recipes/sqlalchemy-oso/meta.yaml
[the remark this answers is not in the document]
bollwyvl: right. just test. the rule of thumb is libraries want as few, and as loosely-pinned, runtime dependencies as possible. full-on applications will have more, and sometimes much more-tightly-pinned, dependencies.

recipes/oso/bld.bat
```diff
@@ -2,10 +2,10 @@
 
 cargo build -p polar-c-api
 
-cd $SRC_DIR\languages\python\oso
+cd %SRC_DIR%\languages\python\oso
```
bollwyvl, line 5: stricly speaking, the slashes would be `/` as well... though it generally _will_ work this way

recipes/oso/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - importlib-metadata !=4.7.1
+    - cffi >=1.0.0
+
+test:
+  source_files:
+    - languages/python/oso/tests
+  imports:
+    - oso
+    - polar
+    - sqlalchemy
```
bollwyvl, line 42: probably remove this import test

bollwyvl:
Even though this failed:

```
================================== FAILURES ===================================
___________________________ test_repr_when_logging ____________________________

polar = <polar.polar.Polar object at 0x000002AA29816700>
capsys = <_pytest.capture.CaptureFixture object at 0x000002AA29845F70>

    def test_repr_when_logging(polar, capsys):
        old_polar_log = os.getenv("POLAR_LOG")
        os.environ["POLAR_LOG"] = "1"
        polar.register_class(Foo)
        polar.load_str("f(_foo: Foo) if 1 = 1;")
        list(polar.query_rule("f", Foo()))
        captured = capsys.readouterr()
        assert f"QUERY RULE: f({FOO_REPR} TYPE `Foo`)" in captured.out
        if not old_polar_log:
>           os.unsetenv("POLAR_LOG")
E           AttributeError: module 'os' has no attribute 'unsetenv'

tests\test_repr.py:25: AttributeError
=========================== short test summary info ===========================
FAILED tests/test_repr.py::test_repr_when_logging - AttributeError: module 'o..
```

It's still quite encouraging. As this was fixed in py3.9, with the aforementioned jinja-based skips, to be _fully_ accurate, you'd want something like:

```jinja
{% skips = skips + ["repr_when_logging"] %} # [win and py<39]
```

however, it's probably fine to just add it to the general skips.

This may or may  not be worth reporting upstream (probably not, unless you figure out some way to fix it with e.g. `mock`).

alexpilot11:
ah, didn't see your comment before pushing the lock. will undo

alexpilot11:
@bollwyvl with a green build, I think this is close, if not there. Do you mind if I add you as a maintainer as well? After that I'll mark as ready for review

recipes/oso/meta.yaml
```diff
@@ -39,11 +39,10 @@ test:
   imports:
     - oso
     - polar
-    - sqlalchemy
   commands:
     - cd languages/python/oso
     - pip check
-    - pytest -vv tests -k "not test_parser_errors"
+    - pytest -vv tests -k "not test_parser_errors and not test_repr_when_logging"
```
bollwyvl, line 45: Now that basic testing is established, consider adding coverage:

```bash
pytest -vv tests --cov=oso --cov=polar --no-cov-on-fail --cov-report=term-missing:skip-covered --cov-fail-under=100
```

This can help find serious regressions (e.g. missing test dependencies). Alternately: remove `pytest-cov` (and any other unused test dependencies): e.g. is `sqlalchemy` really needed? This is important, as sometimes spurious heavyweight test deps bring along transient deps that shadow real runtime deps.

bollwyvl:
> add you as a maintainer as well?

Sure: you can add me as a maintainer...

recipes/oso/meta.yaml
alexpilot11: fail under 100? cov is at 85.62%. should I just drop to like "fail under 80"?

bollwyvl:
Right. So you'd tune down to the worst case, minus one or so... Losing five
percent coverage off 85 would be kind of a thing.

bollwyvl:
@conda-forge/staged-recipes @conda-forge/help-python @conda-forge/help-rust ready for review

bollwyvl (approved):
seems pretty clean now

alexpilot11:
Thank you @ocefpaf !

PR #19652, opened 2022-07-14
Add oeel recipe

giswqs:
Checklist
- [x] Build number is 0.


mgravey:
I want to be listed as maintainer

giswqs:
@conda-forge/help-python The recipe has passed all checks and ready for your review.

giswqs:
@conda-forge/help-python The recipe has passed all checks and is ready for your review. Thanks.

giswqs:
@conda-forge/help-python The recipe has passed all checks and is ready for your review. Thanks.

giswqs:
@conda-forge/help-python It has been more than three weeks since we submitted the recipe. We would greatly appreciate it if you can review it soon. Thanks.

recipes/oeel/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+requirements:
+  host:
+    - pip
+    - python >=3.7
+  run:
+    - python >=3.7
+    - earthengine-api
+    - pyzmq
+    - git
+    - nodejs >=16
```
BastianZim, line 26: If you require nodejs at run is there nothing that needs to be compiled to JavaScript before?

giswqs:
@BastianZim I moved `nodejs` from `run` to `build`, the recipe failed to build. I think `nodejs` is only needed at run.

giswqs:
I just added `nodejs` back to `run`. The recipe has passed all checks.

BastianZim:
Found why. The recipe downloads something during run: https://github.com/open-geocomputing/OpenEarthEngineLibrary/blob/88da34d586159620f033bc253d7404d0ddd1e877/oeel/__init__.py#L38-L58

This is not really ideal as we prefer to have these kind of dependencies packaged. Is it possible to add them here?

giswqs:
@BastianZim Do you have an example for including npm packages and git repos in a recipe? I am not sure how to do that.

BastianZim:
I don't really know of anything at the top of my head but this is how it should work:

1. Add the git repo as a separate package to conda-forge. 
2. Add a patch here so that the package fails instead of downloading: https://github.com/open-geocomputing/OpenEarthEngineLibrary/blob/88da34d586159620f033bc253d7404d0ddd1e877/oeel/__init__.py#L40
3. Add the npm packages here
4. Patch the package so that it doesn't install from NPM but instead fails if the packages are not available.

I know this is quite a bit of work but that would be the standard process. You can also ask core if they can make an exception but I don't know if that works.

giswqs:
What do you mean by `Add the npm packages here`? We can probably include the files from the git repo directly in the `oeel` PyPI package rather than cloning the repo at runtime. My question is how to include the npm packages in the recipe?

BastianZim:
Ahh sorry, I meant add them as packages to conda-forge and then add them as dependencies in the meta.yaml.

For the git repo, I also meant packaging it for conda-forge and then adding it as a requirement in the meta.yaml.

giswqs:
I understood, but my question is how to add npm packages in a recipe so that it becomes a conda package that can be used dependencies by other recipes? An example recipe that has done similar things would be really helpful to learn from.

BastianZim:
Ahh I see. I'm only in python so I don't know a recipe that you could use but the gitter chat is normally really helpful: https://gitter.im/conda-forge/conda-forge.github.io

mgravey:
https://github.com/conda-forge/staged-recipes/pull/19652#issuecomment-1224444077
This is some js code that can be assimilated to some data, and it's needed to the same point that an other package can download an image (that change each day)

https://github.com/conda-forge/staged-recipes/pull/19652#issuecomment-1224444077
This sound terrible because then users need to install this manually. If the proposed solution maybe makes sense from a conda point of view, it's clearly not from PyPI. And will require to split conda code from PyPI code.

https://github.com/conda-forge/staged-recipes/pull/19652#issuecomment-1224464847
This will require an update of the package each time an oeel file is modified. Generating lot of uploads and useless data transfer. Why to download everything when you can simply update what you already have?

BastianZim:
> This is some js code that can be assimilated to some data, and it's needed to the same point that an other package can download an image (that change each day)

JS code is fine, it just needs to be compiled here and packaged, not downloaded afterwards. Does the code change every day as well? You can either include it in the repo or add it as a dependency, but we prefer not vendoring stuff so adding as a dependency would be the best way.

> This sound terrible because then users need to install this manually. If the proposed solution maybe makes sense from a conda point of view, it's clearly not from PyPI. And will require to split conda code from PyPI code.

I think you're misunderstanding me. The other dependencies will be added to conda-forge and the listed under requirements here. Unlike PyPI, conda allows you to specify dependencies that are not python. 

> This will require an update of the package each time an oeel file is modified. Generating lot of uploads and useless data transfer. Why to download everything when you can simply update what you already have?

Do you mean every time the software is updated, you need to release a new version?

By the way, this is a package for NPM packages: https://github.com/conda-forge/staged-recipes/pull/20181/files

mgravey:
1) You can assume that the JS (not Python) code change on a regular basis (I would say currently once in the week, but expected to happen more often in the future)

2) you propose to properly put rpm dependencies in conda. But then how do we do in PyPI, where it's not possible ?
We will need two python versions one PyPI the other conda, right ?

3)If with pack the JS code in a python package, each time the JS code evolve (regularly, but unrelated to the Python interface that is much much more stable in time) we will need to release the complete package, that everyone download, even if a single character changed in the complete JS code. When the current solution relying on Git will just check and download the difference.

BastianZim:
1. Ideally that means that you create a new release for that. If you really don't want to you'll have to contact core as they are the only ones that can make an exception. 
2. You create an environment variable that dictates if already available packages are used or downloaded. This can then be injected when building.
3. Yes but conda-forge works on the basis of releases and versions. We do not allow packages that are not versioned.

mgravey:
Imagine I make package that allow each day to get the bing image ? I need to upload a new package each day ?

BastianZim:
You are downloading content that is different from changing code. Additionally, we have a bot that does version updates: https://conda-forge.org/docs/maintainer/updating_pkgs.html#how-does-regro-cf-autotick-bot-create-automatic-version-updates

PR #19683, opened 2022-07-16
Add Package: bigdft-suite

william-dawson:
Checklist
- [x] Build number is 0.


william-dawson:
@jan-janssen I see you have made various attempts at building a BigDFT package with conda. I think that this PR should take care of things. If you have any suggestions from your experience, let me know. We still need to make a new release (1.9.4) to finalize the package. Also, if you have some use of BigDFT we're always happy to collaborate.

recipes/bigdft/build.sh
```diff
@@ -0,0 +1,18 @@
+# Files
+cp $RECIPE_DIR/conda.rc .
+cp $RECIPE_DIR/patches/* .
+
+# Patches
+patch futile/flib/utils.c utils.patch
+patch bigdft/src/output.f90 output.patch
```
jan-janssen, line 7: You can patch the sources directly in the `meta.yaml` file.

jan-janssen:
> @jan-janssen I see you have made various attempts at building a BigDFT package with conda. I think that this PR should take care of things. If you have any suggestions from your experience, let me know. We still need to make a new release (1.9.4) to finalize the package. Also, if you have some use of BigDFT we're always happy to collaborate.

Thanks a lot. @chrisjsewell maybe you want to have a look as well

chrisjsewell:
This is great cheers @william-dawson 🎉 
It is the last outstanding code, that we were looking to use via conda (https://github.com/marvel-nccr/quantum-mobile/issues/161)

I can have a quick look, but @jan-janssen has a lot more experience of these recipes than me, so mainly I will leave it in he's capable hands to review 😄

Just an FYI, this is the [ansible](https://github.com/ansible/ansible) role I have currently been using to build it with: https://github.com/marvel-nccr/ansible-role-bigdft/blob/master/tasks/main.yml

recipes/bigdft/build.sh
william-dawson: Thanks, I've made this modification!

luigigenovese:
Hello,
thanks for the efforts and replies! I would be very happy to be part of the maintainers for this package (although I did not participate since now to the package recipe creation). Let me know if this is possible. Many thanks! Luigi Genovese

william-dawson:
@chrisjsewell That is very interesting. With BigDFT, you will have the same problem as abinit in that it depends on libxc version 4. I suppose this is why the spack people work so hard to mix and match library versions.

chrisjsewell:
> That is very interesting.

If you are particularly interested lol, then this is essentially what I was working towards lol, to have all the codes running in a "common" way: https://github.com/marvel-nccr/ansible-role-aiida-cws (and also big thanks to @jan-janssen for getting all the others on conda 🙏 )

william-dawson:
@jan-janssen I believe all the tasks are completed. Using `channel_targets` in `conda_build_config.yaml`, I've designated this as a release candidate. If everything looks good to you, I'd like to merge this in, and then start testing out workflows based on installing bigdft from conda. If everything is working well, I'll make a new merge request sometime in September with the official 1.9.3 release of BigDFT.

jan-janssen:

recipes/bigdft/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+{% set version = "1.9.3_dev" %}
+{% set build = 0 %}
+{% set mpi = mpi %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.com/l_sim/bigdft-suite/-/archive/Prerelease1.9.3/bigdft-suite-Prerelease1.9.3.tar.gz
```
chrisburr, line 11: The URL should depend on `version` to avoid issues with accidentally deploying a version with the wrong sources.

Also is this a pre-release?

william-dawson: > The URL should depend on version to avoid issues with accidentally deploying a version with the wrong sources.

Done. Thanks for the correction.

> Also is this a pre-release?

That is my plan. We will release version 1.9.3 of BigDFT in about a month and a half, and I'd like to have a pre-release available so all developers can try out the conda approach. I believe that my `conda_build_config.yaml` should properly label this as a pre-release, but let me know if there are fixes I should make.

recipes/bigdft/build.sh
```diff
@@ -0,0 +1,13 @@
...
+cp $RECIPE_DIR/conda.rc .
+
+# Build
+export JHBUILD_RUN_AS_ROOT="please do it"
+mkdir build
+cd build
+python ../Installer.py -y autogen
+python ../Installer.py -y build -f ../conda.rc
+
+# Activate script
```
jaimergp, line 11: Ideally, this should be paired with a `deactivate.sh` script that undoes the effect of the activation script, so the env vars set do not permeate into the user session once the env is deactivated.

william-dawson: Thanks. Do you know of any projects that have well written `deactivate.sh` scripts that unset environment variables? Because it seems like I'd first need to store all of the original environment variable values somewhere (noting that multiple sessions with different initial conditions could be inside the environment at the same time) for restoring. I can write a script to do all this but I'd like to avoid any pitfalls.

jan-janssen: You can search Github for projects in the conda-forge organization which use a `deactivate.sh` script https://github.com/search?q=org%3Aconda-forge+deactivate.sh&type=code

jaimergp: [This is a good example](https://github.com/conda-forge/intel-compiler-repack-feedstock/tree/main/recipe). The idea is to backup potential values that were set for that env var _before_ you activate. Then on deactivation, restore to the backup _or_, if empty, unset.

You will need to take notes of what the `source`d script is doing, though :)

william-dawson: Thank you for the suggestions! Storing in environment variables with the `conda` suffix seems like a good strategy. I've implemented a deactivate script based on this.

jan-janssen:
@chrisburr Can you take another look?

recipes/bigdft/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+source:
+  url: https://gitlab.com/l_sim/bigdft-suite/-/archive/{{ version }}/bigdft-suite-{{ version }}.tar.gz
+  md5: a756c1673c45e7e7b7be3cdf6aef81b1
+  patches:
+    - patches/output.patch
+    - patches/utils.patch
+
+build:
+  number: {{ build }}
+  skip: true  # [win or osx]
```
carterbox, line 19:
```suggestion
build:
  number: {{ build }}
  skip: true  # [win or osx]
  # add build string so packages can depend on
  # mpi or nompi variants explicitly:
  # `pkg * mpi_mpich_*` for mpich
  # `pkg * mpi_*` for any mpi
  # `pkg * nompi_*` for no mpi
  {% if mpi != 'nompi' %}
  {% set mpi_prefix = "mpi_" + mpi %}
  {% else %}
  {% set mpi_prefix = "nompi" %}
  {% endif %}
  string: "{{ mpi_prefix }}_py{{ py }}h{{ PKG_HASH }}_{{ build }}"
```
Please indicate the MPI variant in the build string

https://conda-forge.org/docs/maintainer/knowledge_base.html#preferring-a-provider-usually-nompi

recipes/bigdft/test.py
```diff
@@ -0,0 +1,19 @@
...
+code = SystemCalculator()
+
+from BigDFT.Inputfiles import Inputfile
+inp = Inputfile()
+inp.set_xc("PBE")
+inp.set_hgrid(0.4)
+
+log = code.run(sys=sys, input=inp)
+print(log.energy)
+
```
carterbox, line 19:
```suggestion

import CheSS
import futile
import gi
```
Test imports of all your packages.

recipes/bigdft/run_test.sh
```diff
@@ -0,0 +1,2 @@
+source bigdftvars.sh
+python test.py
```
carterbox, line 2: Maybe a few of the many CLI binaries being shipped? Do any of them accept `--help` or `--version` and return?

recipes/bigdft/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+source:
+  url: https://gitlab.com/l_sim/bigdft-suite/-/archive/{{ version }}/bigdft-suite-{{ version }}.tar.gz
+  md5: a756c1673c45e7e7b7be3cdf6aef81b1
+  patches:
+    - patches/output.patch
+    - patches/utils.patch
+
+build:
+  number: {{ build }}
```
carterbox, line 18:
```suggestion
build:
  run_exports:
     -  {{ pin_subpackage(name|lower) }}
  number: {{ build }}
```
Shipping header files means that you should pin this package according to ABI compatability guarantees.

recipes/bigdft/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+test:
+  files:
+    - test.py
+
+about:
+  home: https://bigdft.org
+  license: GPL-3.0-only
+  license_family: GPL
+  license_file: LICENSE
```
carterbox, line 60: The files in licenses are not being included, but I think globbing works? otherwise, you need to be explicit.

```suggestion
  license_file:
    - LICENSE
    - licenses/*
```

carterbox (changes requested):
Is `libfmalloc` part of this package or are you redistributing?

Shipping static libraries is not allowed. The following libs are being shipped statically: abinit, atlab, bigdft, CheSS, dics, futile, Galn, orbs, Psolver.

_jhbuild/ is in included in the artifacts, is the supposed to be shipped?

recipes/bigdft/meta.yaml
william-dawson: There is no `nompi` version of our package, it must be installed with mpi. So in our case this seems unnecessary. Do you recommend instead just putting `{% set mpi_prefix = "mpi_" + mpi %}`? Or to just leave it all out?

carterbox: This is just a suggestion from the docs page. Please modify to fit your use case. You have separate variants for `mpich` and `openmpi` and they should be noted in the build string.

william-dawson: Thanks. I've made the modification accordingly, and will continue to work on the other suggestions.

recipes/bigdft/test.py
william-dawson: CheSS is depreciated (I'm put in a pull request to remove it from our main repository) and gi is too experimental to test, but testing the imports explicitly of futile and BigDFT sounds like a good idea to me (better than the error coming from deep in a run). I've made a modification based on this suggestion.

recipes/bigdft/run_test.sh
william-dawson: Thanks. I've added some tests for the CLI apps I know are more regularly used with a `-h` option. I might talk to the other developers about whether all of the helper binaries should be built by default in the future.

william-dawson:
> Is `libfmalloc` part of this package or are you redistributing?
> 
> Shipping static libraries is not allowed. The following libs are being shipped statically: abinit, atlab, bigdft, CheSS, dics, futile, Galn, orbs, Psolver.
> 
> _jhbuild/ is in included in the artifacts, is the supposed to be shipped?

I've set my install script to remove the `_jhbuild` directory as it seems to be only used at build time. I've also modified the configuration lines to build dynamic versions of the libraries (and remove the static versions). This indeed includes `libfmalloc`, which is our Fortran module for managing dynamic memory. 

I think I have addressed all your suggestions, but let me know if I missed or misunderstood something. We appreciate the feedback!

recipes/bigdft/run_test.sh
```diff
@@ -0,0 +1,10 @@
+# Setup the environment
```
carterbox, line 1:
```suggestion
set -ex

# Check for expected libraries ABI
test -f ${PREFIX}/lib/bigdft/libvol.so
test -f ${PREFIX}/lib/libatlab-1.so.0
test -f ${PREFIX}/lib/libbigdft-1.so.9
test -f ${PREFIX}/lib/libCheSS-1.so.2
test -f ${PREFIX}/lib/libfmalloc-1.so.9
test -f ${PREFIX}/lib/libfutile-1.so.9
test -f ${PREFIX}/lib/liborbs.so.0
test -f ${PREFIX}/lib/libPSolver-1.so.9

# Check for header
test -f ${PREFIX}/include/bigdft.h
test -f ${PREFIX}/include/futile/tree.h
test -f ${PREFIX}/include/atlab/box.h

# Setup the environment
```
It's good practice to do existence checks for libraries with major version. This will cause tests to fail and prevent from breaking downstream packages if the SONAME of these libraries changed unexpectedly.

recipes/bigdft/test.py
carterbox: The tests in this conda recipe are for *packaging* validation, so their purpose is to check that the artifacts were installed correctly. They are not necessarily a comment on whether a feature of the package is deprecated or considered experimental. If you expect that an end user of this package would still be importing the CheSS module, then you should test it.

carterbox (changes requested):

recipes/bigdft/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+source:
+  url: https://gitlab.com/l_sim/bigdft-suite/-/archive/{{ version }}/bigdft-suite-{{ version }}.tar.gz
+  md5: a756c1673c45e7e7b7be3cdf6aef81b1
+  patches:
+    - patches/output.patch
+    - patches/utils.patch
+
+build:
+  run_exports:
+     -  {{ pin_subpackage(name|lower) }}
```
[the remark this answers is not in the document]
william-dawson: I see. I didn't understand that point at the time. I think it would be difficult to guarantee that with the current release cycle, so I've updated it to include the minor version as well.

recipes/bigdft/test.py
```diff
@@ -0,0 +1,24 @@
+# Test that we can import everything
+import futile
```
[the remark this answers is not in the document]
william-dawson: I asked about this upstream, and they said they will fix it as you suggest in a future release.

william-dawson:
@carterbox I've gone through the most recent suggestions. For the testing of the dynamic libraries you suggested, I had to remove `liborbs` from the list, as there is an issue with the symbolic link. I will report that upstream. For the CheSS and gi modules, thank you for the clarification. There is essentially a 0% chance someone would include it in their code, so I think it is safe to leave that test out.

william-dawson (commented):
(see individual comments)

carterbox:
@conda-forge/help-python-c, please review

carterbox (approved):
Apparently, I'm the only reviewer of compiled packages right now. 😆

william-dawson:
> Apparently, I'm the only reviewer of compiled packages right now. 😆

Be sure to add that you to CV 😄

PR #19771, opened 2022-07-23
add beaker-py

h-vetinari:
Needed for [tango](https://github.com/conda-forge/tango-feedstock/pull/12).

The name reflects the upstream name for the python bindings, though we couldn't package the underlying `beaker` (written in Go apparently) as-is because there's https://github.com/conda-forge/beaker-feedstock already.

### DIFF
--- a/recipes/beaker-py/meta.yaml
+++ b/recipes/beaker-py/meta.yaml
@@ -9,10 +9,9 @@ source:
   sha256: d1445637651e9a6ddbcce609108284a4d6b4ad3603e579a96dbdc1e77d9e0447
 
 build:
-  skip: true  # [win]
+  noarch: python
   number: 0
   script:
-    - rm pyproject.toml
     - {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
--- a/recipes/beaker-py/meta.yaml
+++ b/recipes/beaker-py/meta.yaml
@@ -36,6 +36,7 @@ test:
     - beaker
   source_files:
     - tests/
+    - test_fixtures/
   commands:
     - pip check
     - pytest -v tests/
--- a/recipes/beaker-py/meta.yaml
+++ b/recipes/beaker-py/meta.yaml
@@ -11,15 +11,14 @@ source:
 build:
   noarch: python
   number: 0
-  script:
-    - {{ PYTHON }} -m pip install . --no-deps -vv
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   host:
-    - python
+    - python >=3.7
     - pip
   run:
-    - python
+    - python >=3.7
     - docker-py >=5.0,<6
     - packaging
     - pydantic >=1.8.2,<1.10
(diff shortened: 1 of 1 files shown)


h-vetinari:
@dirkgr @epwalsh

Could you comment on the failing test suite here? AFAICT it makes little sense to try to run docker-based tests here (not least since conda-forge's [docker-py](https://github.com/conda-forge/docker-py-feedstock) actually doesn't install a docker-daemon).

epwalsh:
The test failures are because many of the tests require a Beaker personal access token. I hate to say it but the best option might be to just not run any of the unit tests.

h-vetinari:
> The test failures are because many of the tests require a Beaker personal access token. I hate to say it but the best option might be to just not run any of the unit tests.

Would it not make sense (medium-term & upstream) to make those tests conditional on the presence of the token, and skip them if it's not present?

epwalsh:
> Would it not make sense (medium-term & upstream) to make those tests conditional on the presence of the token, and skip them if it's not present?

That's a possibility, but I understated how many tests require a token... it's pretty much 100% of the tests. Though I've proposed an idea internally (https://github.com/allenai/beaker/issues/2239 - repo is private so you won't be able to view the issue, but I'm linking here for my own records) that would allow the Beaker server to issue temporary tokens that could be used by GitHub runners via approved repositories. If that gets implemented then we wouldn't have to skip any tests.

h-vetinari:
> If that gets implemented then we wouldn't have to skip any tests.

That sounds great, thanks! I like not skipping tests. :)

h-vetinari:
@conda-forge/help-python

This should be ready for review. :)

recipes/beaker-py/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+source:
+  url: https://github.com/allenai/beaker-py/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d1445637651e9a6ddbcce609108284a4d6b4ad3603e579a96dbdc1e77d9e0447
+
+build:
+  skip: true  # [win]
+  number: 0
+  script:
+    - rm pyproject.toml
```
jaimergp, line 15: Why is this needed? It only lists the build backend for `pip install`.

```suggestion
```

recipes/beaker-py/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+package:
+  name: beaker-py
+  version: {{ version }}
+
+source:
+  url: https://github.com/allenai/beaker-py/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d1445637651e9a6ddbcce609108284a4d6b4ad3603e579a96dbdc1e77d9e0447
+
+build:
+  skip: true  # [win]
```
jaimergp, line 12: Doesn't this package qualify for `noarch: python`?

```suggestion
  noarch: python
```

jaimergp (changes requested):
Just a couple questions!

h-vetinari:
Thanks for the review @jaimergp, this should be ready now I hope.

PR #19813, opened 2022-07-26
Add mitmproxy and dependencies

AdrianFreundQC:
Additional notes:
* mitmproxy is downloaded from github releases, as they don't have an sdist on pypi, only a wheel.
* kaitaistruct is version 0.9, despite version 0.10 being released, as the current mitmproxy release depends on <0.10.
* kaitaistruct 0.9 doesn't contain a license file, so I manually added it. Once it is updated to 0.10 this is no longer needed, as 0.10 contains the file.

Checklist
- [x] Build number is 0.


AdrianFreundQC:
@conda-forge/help-python I'm having some weird issues here: In the CI `pip check` complains about brotli not being installed (`mitmproxy 8.1.1 requires brotli, which is not installed.`), but it is listed in the dependencies and the CI log shows that it has been installed. Any help would be appreciated.

recipes/mitmproxy/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - ldap3 >=2.8,<2.10
+    - msgpack-python >=1.0.0,<1.1.0
+    - passlib >=1.6.5,<1.8
+    - protobuf >=3.14,<5
+    - publicsuffix2 >=2.20190812,<3
+    - pydivert >=2.0.3,<2.2          # [win]
+    - pyopenssl >=21.0,<22.1
+    - pyparsing >=2.4.2,<3.1
+    - pyperclip >=1.6.0,<1.9
+    - python
```
jonashaag, line 45:
```suggestion
    - python >=3.9
```

jonashaag: See https://github.com/mitmproxy/mitmproxy/blob/main/setup.py#L71

AdrianFreundQC: I think that needs to be like that. Look at the linter messages in this PR:
1. Because this recipe has a windows only dependency it can't be `noarch: python`, as noarch packages have to be the same on all architectures: https://github.com/conda-forge/staged-recipes/pull/19813#issuecomment-1195692656
2.Non `noarch` packages can't specify python version constraints: https://github.com/conda-forge/staged-recipes/pull/19813#issuecomment-1195703937

Because of that I used the solution I found here: https://groups.google.com/g/conda-forge/c/vZj02SDKcO0 and added the python version as a selector to the skip instead.
This also seems to be what other Non-`noarch` python packages do. For example numpy: https://github.com/conda-forge/numpy-feedstock/blob/main/recipe/meta.yaml#L23

jonashaag: You are right, sorry for the confusing suggestion!

jonashaag:
@conda-forge/staged-recipes

recipes/pydivert/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pydivert-{{ version }}.tar.gz
+  sha256: f0e150f4ff591b78e35f514e319561dadff7f24a82186a171dd4d465483de5b4
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: True  # [not win]
```
ocefpaf, line 15: Just curious, what is the issue that limits this on Windows?

jonashaag: pydivert is a Windows-only package

PR #19855, opened 2022-07-28
Add shared-memory-dict

maxb2:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/shared-memory-dict/meta.yaml
+++ b/recipes/shared-memory-dict/meta.yaml
@@ -15,8 +15,10 @@ build:
   number: 0
 
 requirements:
-  build:
+  host:
+    - pip
     - python >=3.6,<4.0
+    - poetry
 
 test:
   imports:
@@ -32,6 +34,7 @@ outputs:
       host: &id001
         - pip
         - python >=3.6,<4.0
+        - poetry
       run:
         - python >=3.6,<4.0
   - name: shared-memory-dict-aiocache
--- a/recipes/shared-memory-dict/meta.yaml
+++ b/recipes/shared-memory-dict/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - python >=3.6,<4.0
     - poetry
   run:
-    - python
+    - python >=3.6,<4.0
 
 test:
   imports:
@@ -44,20 +44,20 @@ outputs:
       host: *id001
       run:
         - shared-memory-dict =={{ version }}
-        - aiocache>=0.11.1,<0.12.0
+        - aiocache >=0.11.1,<0.12.0
   - name: shared-memory-dict-all
     requirements:
       host: *id001
       run:
         - shared-memory-dict =={{ version }}
-        - aiocache>=0.11.1,<0.12.0
-        - django>=3.0.8,<4.0.0
+        - aiocache >=0.11.1,<0.12.0
+        - django >=3.0.8,<4.0.0
   - name: shared-memory-dict-django
     requirements:
       host: *id001
       run:
         - shared-memory-dict =={{ version }}
-        - django>=3.0.8,<4.0.0
+        - django >=3.0.8,<4.0.0
 
 about:
   home: https://github.com/luizalabs/shared-memory-dict
--- a/recipes/shared-memory-dict/meta.yaml
+++ b/recipes/shared-memory-dict/meta.yaml
@@ -10,22 +10,13 @@ source:
   sha256: 8294257e1a2466afb5e19a878a8c446325b8913d57d6d9578ccba905564a974d
 
 build:
-  noarch: python
   number: 0
-
-requirements:
-  host:
-    - pip
-    - python >=3.6,<4.0
-
-test:
-  requires:
-    - pip
 
 outputs:
   - name: shared-memory-dict
     build:
-      script: {{ PYTHON }} -m pip install . -vv
+      noarch: python
+    script: build_base.sh
     requirements:
       host:
         - pip
@@ -37,24 +28,24 @@ outputs:
         - shared_memory_dict
       commands:
         - pip check
+      requires:
+        - pip
+
   - name: shared-memory-dict-aiocache
+    build:
+      noarch: generic
     requirements:
-      host:
-        - pip
-        - python >=3.6,<4.0
       run:
         - {{ pin_subpackage('shared-memory-dict', max_pin="x.x.x") }}
         - aiocache >=0.11.1,<0.12.0
     test:
       imports:
         - shared_memory_dict
-      commands:
-        - pip check
+
   - name: shared-memory-dict-all
+    build:
+      noarch: generic
     requirements:
-      host:
-        - pip
-        - python >=3.6,<4.0
       run:
         - {{ pin_subpackage('shared-memory-dict', max_pin="x.x.x") }}
         - aiocache >=0.11.1,<0.12.0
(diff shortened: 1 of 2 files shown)


maxb2:
I'm willing to be listed as a maintainer

maxb2:

maxb2:

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.6,<4.0
+    - poetry
```
jakirkham, line 21: Need to include `python` in `run`

```suggestion
    - poetry
  run:
    - python
```

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+      run:
+        - shared-memory-dict =={{ version }}
+        - aiocache >=0.11.1,<0.12.0
+        - django >=3.0.8,<4.0.0
+  - name: shared-memory-dict-django
+    requirements:
+      host: *id001
+      run:
+        - shared-memory-dict =={{ version }}
+        - django >=3.0.8,<4.0.0
```
BastianZim, line 60: This section needs to be formatted in a different way. Please have a look at the mathplotlib feedstock as an example on how to do it: https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml

BastianZim (changes requested):

recipes/shared-memory-dict/meta.yaml
maxb2: See [778a669](https://github.com/conda-forge/staged-recipes/pull/19855/commits/778a66946d8d34b23308b0f2008d9758f1981e48). FYI, I used the new grayskull feature to [automatically add optional dependencies](https://github.com/conda-incubator/grayskull/pull/341) to make the original recipe.

BastianZim: Great! Seems though like there is a bug in Grayskull as the build script needs to be added like this: https://github.com/conda-forge/sagemaker-inference-toolkit-feedstock/blob/main/recipe/build_base.sh

can you also please open an issue in Grayskull so that this gets incorporated?

maxb2:
@BastianZim, I updated the recipe to follow [sagemaker-inference-toolkit-feedstock](https://github.com/conda-forge/sagemaker-inference-toolkit-feedstock/blob/main/recipe/meta.yaml) like you suggested, but the import test in subsequent packages keeps [failing](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=557878&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1265). I don't know what else to try.

BastianZim:
I'm currently on my phone but I'll have a look in an hour or so.

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+about:
+  home: https://github.com/luizalabs/shared-memory-dict
+  summary: A very simple shared memory dict implementation
+  license: MIT
+  license_file:
+    - LICENSE
+
+extra:
+  recipe-maintainers:
+    - maxb2
```
BastianZim, line 77:
```suggestion
    - maxb2
  feedstock-name: shared-memory-dict
```
Unrelated to the error, this just makes sure that the feedstock name is correct.

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,77 @@
+{% set name = "shared-memory-dict" %}
+{% set version = "0.7.2" %}
+
+package:
+  name: {{ name|lower }}
```
BastianZim, line 5:
```suggestion
  name: {{ name|lower }}-meta
```
This package will actually not be built, it is just a meta package.

BastianZim (commented):
I can't see the error either but let's make sure that everything is in order and then I can try tracking it down.

recipes/shared-memory-dict/meta.yaml
maxb2: Will do once this gets merged!

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+outputs:
+  - name: shared-memory-dict
+    build:
+      noarch: python
+    script: build_base.sh
+    requirements:
+      host:
+        - pip
+        - python >=3.6,<4.0
```
BastianZim, line 23:
```suggestion
        - python >=3.8,<4.0
```
Upstream lists 3.8: https://github.com/luizalabs/shared-memory-dict/blob/ff8726973064afaec23eceb09b031abf535a8b4b/pyproject.toml#L21

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - shared_memory_dict
+
+  - name: shared-memory-dict-all
+    build:
+      noarch: generic
+    requirements:
+      run:
+        - {{ pin_subpackage('shared-memory-dict', exact=True) }}
+        - aiocache >=0.11.1,<0.12.0
+        - django >=3.0.8,<4.0.0
```
BastianZim, line 52: To make your life easier, you can also just install `shared-memory-dict-aiocache` and `shared-memory-dict-django` here. Then you only need to update dependencies once.

BastianZim (commented):
I have added two small comments.

For the overall problem: I don't know unfortunately what is causing it but one thing I noticed in the [logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=558009&view=logs&j=e35d4f76-8ff2-5536-d795-df91e63eb9f7&t=fa7b4b17-b6ff-5c9c-8cfc-15f888c92310&l=1385) is that pip installs the build tools (poetry etc) since they are not specified here. Upstream doesn't do that either though so I would first recommend asking them to add the build tools to their [pyproject.toml](https://python-poetry.org/docs/pyproject/#poetry-and-pep-517).

You can also add `--no-index` to the build script which will maybe make it fail sooner.

Unfortunately, multi-output scripts are very finicky, so the error might be somewhere else. 

@ocefpaf Have you seen this before and maybe have an idea?

recipes/shared-memory-dict/meta.yaml
maxb2: Hmm, I guess `grayskull` missed that.

maxb2:
> For the overall problem: I don't know unfortunately what is causing it but one thing I noticed in the [logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=558009&view=logs&j=e35d4f76-8ff2-5536-d795-df91e63eb9f7&t=fa7b4b17-b6ff-5c9c-8cfc-15f888c92310&l=1385) is that pip installs the build tools (poetry etc) since they are not specified here. Upstream doesn't do that either though so I would first recommend asking them to add the build tools to their [pyproject.toml](https://python-poetry.org/docs/pyproject/#poetry-and-pep-517).

Upstream does have `poetry` set as the [build backend](https://github.com/luizalabs/shared-memory-dict/blob/main/pyproject.toml#L50-L52), but it is using the old way to do it. I'll submit a PR to update it.

BastianZim:
Ahh true — must've missed that. Please add poetry to host then.

recipes/shared-memory-dict/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    build:
+      noarch: python
+    script: build_base.sh
+    requirements:
+      host:
+        - pip
+        - python >=3.8,<4.0
+        - poetry
+      run:
+        - python >=3.8,<4.0
```
carterbox, line 26:
```suggestion
      run:
        - python >=3.8,<4.0
      run_constrained:
        - aiocache >=0.11.1,<0.12.0
        - django >=3.0.8,<4.0.0
```
You actually need to specify these version constraints here and not in the metapackages because if these packages are installed manually, they still need to meet the constraints.

carterbox (changes requested):

carterbox:
Needed to add all of the special pip args that are set automatically in the top package.

maxb2:
Thanks for all the help getting this merged!

PR #19876, opened 2022-07-29
Adds tethys-platform

swainn:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/tethys-platform/bld.bat
+++ b/recipes/tethys-platform/bld.bat
@@ -1,2 +1,2 @@
-set PBR_VERSION=3.4.3
+set PBR_VERSION=%PKG_VERSION%
 %PYTHON% -m pip install . -vv
--- a/recipes/tethys-platform/bld.bat
+++ b/recipes/tethys-platform/bld.bat
@@ -1,0 +1,2 @@
+set PBR_VERSION=3.4.3
+%PYTHON% -m pip install . -vv
--- a/recipes/tethys-platform/build.sh
+++ b/recipes/tethys-platform/build.sh
@@ -1,1 +1,1 @@
-PBR_VERSION=3.4.3 $PYTHON -m pip install . -vv
+PBR_VERSION=$PKG_VERSION $PYTHON -m pip install . -vv
--- a/recipes/tethys-platform/build.sh
+++ b/recipes/tethys-platform/build.sh
@@ -1,0 +1,1 @@
+PBR_VERSION=3.4.3 {{ PYTHON }} -m pip install . -vv
--- a/recipes/tethys-platform/build.sh
+++ b/recipes/tethys-platform/build.sh
@@ -1,1 +1,1 @@
-PBR_VERSION=3.4.3 {{ PYTHON }} -m pip install . -vv
+PBR_VERSION=3.4.3 $PYTHON -m pip install . -vv
--- a/recipes/tethys-platform/meta.yaml
+++ b/recipes/tethys-platform/meta.yaml
@@ -12,9 +12,6 @@ source:
 build:
   noarch: python
   number: 0
-  ignore_run_exports:
-    - pip
-    - pbr
   entry_points:
     - tethys = tethys_cli:tethys_command
 
--- a/recipes/tethys-platform/meta.yaml
+++ b/recipes/tethys-platform/meta.yaml
@@ -1,0 +1,110 @@
+{% set name = "tethys-platform" %}
+{% set version = "3.4.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/tethysplatform/tethys/archive/refs/tags/{{ version }}.tar.gz
+  sha256: 3775d5599e756206204e09ca9a2114c8866fad7fd29999b256343c847ada3b1f
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python>=3.6
+    - pip
+    - pbr
+  run:
+    - python>=3.6
+    - pycrypto=2.6.*
+    - pyopenssl=22.0.*
+    - docker-py=5.0.*
+    - distro=1.6.*
+    - postgresql
+    - psycopg2<2.9
+    - sqlalchemy=1.4.*
+    - geoalchemy2=0.6.*
+    - plotly=5.9.*
+    - bokeh>=2.0.0*
+    - tethys_dataset_services>=2.0.0
+    - hs_restclient=1.3.*
+    - owslib=0.26.*
+    - requests=2.28.*
+    - social-auth-app-django=4.0.*
+    - dask=1.2.*
+    - tethys_dask_scheduler>=1.0.2
+    - channels=2.*
+    - daphne=2.5.*
+    - service_identity=18.1.*
+    - condorpy=0.6.*
+    - siphon=0.9
+    - python-jose=3.3.*
+    - pyjwt<2.0.0
+    - arrow=1.2.*
+    - isodate=0.6.*
+    - django=2.2.*
+    - django-axes=5.35.*
+    - django-filter=22.1
+    - djangorestframework=3.13.*
+    - django-bootstrap3=11.1.*
+    - django-model-utils=4.0.*
+    - django-guardian=2.4.*
+    - django-gravatar2=1.4.*
+    - django-termsandconditions=2.0.*
+    - django-session-security=2.6.*
+    - django-analytical=3.1.*
+    - django-simple-captcha=0.5.*
+    - django-recaptcha2=1.4.*
+    - django-mfa2<2.5
+    - social-auth-app-django=4.0.*
+    - requests-mock=1.9.*
+    - selenium=4.3.*
+    - coverage=6.4.*
+    - factory_boy=3.2.*
+    - pillow=9.2.*
+    - pip=22.1.*
+    - future=0.18.*
+    - flake8=4.0.*
+    - pbr=5.9.*
+    - git
+    - conda
+    - fido2<1.0.0
+
+test:
+  imports:
+    - tethys_apps
+    - tethys_cli
+    - tethys_compute
+    - tethys_config
+    - tethys_gizmos
+    - tethys_portal
+    - tethys_quotas
+    - tethys_sdk
+    - tethys_services
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: http://www.tethysplatform.org/
+  summary: 'Geospatial and scientific web applications for the 21st century.'
+  description: |
+    Tethys Platform has been designed to lower the barrier to geospatial and 
+    scientific web app development. Use it to deliver your models and data as 
+    interactive web apps.
+  license: BSD-2-Clause
+  license_family: BSD
+  license_file: LICENSE
+  doc_url: http://docs.tethysplatform.org/en/stable/
+  dev_url: https://github.com/tethysplatform/tethys
+
+extra:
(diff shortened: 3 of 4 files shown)


swainn:
I agree to be a maintainer

sdc50:
I agree to be a maintainer

swainn:
@conda-forge/help-python this is good to go.

swainn:
> @conda-forge/help-python this is good to go.

@conda-forge/help-python any feedback on this yet?

recipes/tethys-platform/bld.bat
```diff
@@ -0,0 +1,2 @@
+set PBR_VERSION=3.4.3
```
carterbox, line 1:
```suggestion
set PBR_VERSION=$PKG_VERSION
```
https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

recipes/tethys-platform/build.sh
```diff
@@ -0,0 +1 @@
+PBR_VERSION=3.4.3 $PYTHON -m pip install . -vv
```
carterbox, line 1:
```suggestion
PBR_VERSION=%PKG_VERSION% $PYTHON -m pip install . -vv
```
https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

recipes/tethys-platform/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - tethys_cli
+    - tethys_compute
+    - tethys_config
+    - tethys_gizmos
+    - tethys_portal
+    - tethys_quotas
+    - tethys_sdk
+    - tethys_services
+  commands:
+    - tethys version
```
carterbox, line 91:
```suggestion
test:
  requires:
    - pip
  imports:
    - tethys_apps
    - tethys_cli
    - tethys_compute
    - tethys_config
    - tethys_gizmos
    - tethys_portal
    - tethys_quotas
    - tethys_sdk
    - tethys_services
  commands:
    - pip check
    - tethys version
```

recipes/tethys-platform/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - django-simple-captcha =0.5.*
+    - django-recaptcha2 =1.4.*
+    - django-mfa2 <2.5
+    - social-auth-app-django =4.0.*
+    - requests-mock =1.9.*
+    - factory_boy =3.2.*
+    - pillow =9.2.*
+    - pip =22.1.*
+    - future =0.18.*
+    - pbr =5.9.*
```
carterbox, line 75: Why are pip and pbr required but also ignored from run_exports? Why are they required at runtime?

recipes/tethys-platform/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+    - git
+    - conda
```
carterbox, line 77: Why are git and conda required at runtime?

carterbox (changes requested):

recipes/tethys-platform/meta.yaml
swainn: pip and conda are required because Tethys uses their APIs in the `tethys install` command to install dependencies of apps that it is installing.

recipes/tethys-platform/meta.yaml
swainn: pip and conda are required because Tethys uses their APIs in the `tethys install` command to install dependencies of apps that it is installing.

recipes/tethys-platform/bld.bat
swainn: Done.

recipes/tethys-platform/meta.yaml
swainn: Done.

recipes/tethys-platform/meta.yaml
swainn: I'm not sure I understand your comment about them being ignored from run_exports. Doc link?

recipes/tethys-platform/meta.yaml
swainn: Removed git

swainn:
@conda-forge/help-python and @carterbox I've resolved your comments to the best of my understanding. Please help me to better under stand the comment about ignoring them from run_exports. Thank you.

recipes/tethys-platform/meta.yaml
carterbox:
```yaml
  ignore_run_exports:
    - pip
    - pbr
```
Why does your recipe include this section?

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

carterbox: run_exports are packages which are implicitly added to the run requirements of your recipe when you explicity add a related package to your host/build requirements section. It is used for compiled libraries to ensure that linked shared libraries are included. Why do your recipe want to prevent `pip` and `pbr` from being exported in this manner? I'm not even sure that any packages would export `pip` or `pbr`.

carterbox: > Removed pbr.

Looks like pbr has a module which is also imported at runtime.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=559916&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=2664

swainn: Tests failed when pbr was removed as runtime dependency. It is required for the version command.

See: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=559916&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823

swainn: > ```yaml
>   ignore_run_exports:
>     - pip
>     - pbr
> ```
> 
> Why does your recipe include this section?
> 
> https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

Ah yes. That was added due to this error on the Windows build check: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=546753&view=logs&j=240f1fee-52bc-5498-a14a-8361bde76ba0&t=2fb5b0a7-737c-5f5c-8f3f-f6db6174bacf&l=2151

carterbox: > Ah yes. That was added due to this error on the Windows build check

The linked error message is as follows:

> [noarch_python] Noarch package contains binary script: tethys.exe

This error says that conda found a platform specific binary in a noarch package. No arch means it runs on any platform. A windows binary cannot run on Linux. This should be unrelated to whether pip or pbr are in the requirements/run section, and prevents a package from being valid noarch. Is your package [noarch or not?](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python)

Also, we do not allow pre-compiled binaries. You must use our compiler toolchain to compile binaries from source.

dopplershift: Specifically, I think it's due to the script listed `pyproject.toml`. Those are [not allowed for noarch](https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=noarch#noarch-python). If you converted that to a proper `console_script` entrypoint, that would allow setuptools to handle creating the proper executable on both Windows and *nix. I think most of the effort would be in properly specifying the metadata.

BastianZim: Poetry scripts are converted to entry points according to https://github.com/python-poetry/poetry/issues/927
They just need to be explicitly listed as entry points in the meta.yaml. Was that the case when the error occurred?

carterbox: Maybe your console_scipts entry point is configured incorrectly?

dopplershift: I don't see poetry being used here? So I think the problem is that it's declared as a setuptools "script", while it needs to be a "console_script".

BastianZim: Ahh true you're right :)

carterbox: I think that error message is from an old commit before the entry point was registered with the recipe. I expect it would not occur now.

recipes/tethys-platform/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+source:
+  url: https://github.com/tethysplatform/tethys/archive/refs/tags/{{ version }}.tar.gz
+  sha256: 3775d5599e756206204e09ca9a2114c8866fad7fd29999b256343c847ada3b1f
+
+build:
+  noarch: python
+  number: 0
+  ignore_run_exports:
+    - pip
+    - pbr
```
carterbox, line 17:
```suggestion
```
I believe this is unrelated to failed Windows builds from earlier. Regardless, it shouldn't be necessary.

carterbox (changes requested):

recipes/tethys-platform/meta.yaml
carterbox: @BastianZim @dopplershift, the "console_script" syntax is removed with the shift to `pyproject.toml`

https://setuptools.pypa.io/en/latest/userguide/quickstart.html?highlight=console_scripts#entry-points-and-automatic-script-creation

swainn: >This error says that conda found a platform specific binary in a noarch package. No arch means it runs on any platform. A windows binary cannot run on Linux. This should be unrelated to whether pip or pbr are in the requirements/run section, and prevents a package from being valid noarch. Is your package [noarch or not?](https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python)

As far as I can tell, the project meets the requirements listed in that link for noarch. We are not vendoring any pre-compiled binaries. `tethys.exe` is generated by setuptools as the entrypoint for our console script, which is specified in the setup.cfg (see below).

>Specifically, I think it's due to the script listed pyproject.toml. Those are [not allowed for noarch](https://conda-forge.org/docs/maintainer/knowledge_base.html?highlight=noarch#noarch-python). 

Tethys Platform 3.X doesn't have a pyproject.toml (you probably looked at the master branch)? Tethys 3.4 is on the tethys_3.4 branch.

> If you converted that to a proper console_script entrypoint, that would allow setuptools to handle creating the proper executable on both Windows and *nix. I think most of the effort would be in properly specifying the metadata.

> Maybe your console_scipts entry point is configured incorrectly?

 The console script is configured in the [setup.cfg](https://github.com/tethysplatform/tethys/blob/release-3.4/setup.cfg#L34) as follows:

```ini
[entry_points]
console_scripts =
    tethys = tethys_cli:tethys_command
```

How would this need to be changed to make it a "proper" `console_script`?

> Poetry scripts are converted to entry points according to https://github.com/python-poetry/poetry/issues/927
They just need to be explicitly listed as entry points in the meta.yaml. Was that the case when the error occurred?

We are not using poetry as our packaging / build system. We use setuptools.

dopplershift: Ah I missed that, but it looks like the missing entry point was, as suggested by @carterbox, just outdated.

swainn: I think I know what happened now. I added `pip` and `pbr` to the `ignore_run_exports` section thinking that the warnings were related to the `tethys.exe` error. But I also added the `entry_points` section to the recipe in the same commit thinking both changes were necessary and the build passed.

I've now removed the `entry_points` section as @carterbox suggested and the Windows build still works because I've specified the `entry_point`. Sorry for the confusion.

PR #19909, opened 2022-08-02
Adding rapid-pe package

dwysocki:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/rapid-pe/meta.yaml
+++ b/recipes/rapid-pe/meta.yaml
@@ -41,7 +41,7 @@ test:
 about:
   home: https://git.ligo.org/rapidpe-rift/rapidpe/
   summary: 'RapidPE: The original low-latency gravitational wave parameter estimation code.'
-  license: GPL-2.0
+  license: GPL-2.0-or-later
   license_file:
     - COPYING
 
--- a/recipes/rapid-pe/meta.yaml
+++ b/recipes/rapid-pe/meta.yaml
@@ -20,8 +20,8 @@ requirements:
     - setuptools
   run:
     - h5py
-    - lalsuite
-    - lscsoft-glue
+    - lalsuite==7.5
+    - lscsoft-glue==3.0.1
     - matplotlib-base
     - numpy
     - python
--- a/recipes/rapid-pe/meta.yaml
+++ b/recipes/rapid-pe/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
+  skip: true  # [win]
 
 requirements:
   host:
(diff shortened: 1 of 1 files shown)


dwysocki:

recipes/rapid-pe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - rapid_pe
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://git.ligo.org/rapidpe-rift/rapidpe/
+  summary: 'RapidPE: The original low-latency gravitational wave parameter estimation code.'
+  license: GPL-2.0
```
duncanmmacleod, line 44: @dwysocki, this is ambiguous, can you change this to one of

```suggestion
  license: GPL-2.0-only
```

or

```suggestion
  license: GPL-2.0-or-later
```

dwysocki: Thanks for catching this, I've taken your latter suggestion in the latest commit

dwysocki:

recipes/rapid-pe/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/rapid_pe-{{ version }}.tar.gz
+  sha256: 53739bd328779f8484cfc1887ba18aa7b8a0b0ea82b064bb4866b2cbc6eaa518
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
```
ocefpaf, line 15: If you are skipping Windows just b/c the lack of pytorch then you should still make this package `noarch` and let the solver fail on Windows. Then when pytorch is available on Windows, this package will be ready.

dwysocki: It's not PyTorch, but rather `lal` which doesn't support Windows (see [here](https://github.com/conda-forge/lal-feedstock/blob/df3c97bc502f2d24e1b3fdf8f4ef9f393abbd757/recipe/meta.yaml#L37)).    I simply copied the `skip` line from `lal`'s `meta.yaml`, though I'm happy to just let the pipeline fail if that's the preferred approach.

Also it looks like the executables would not be made accessible on Windows conda/conda-build#4385, which would make our package not very useful.

ocefpaf: `lal` is also unavailable on Windows but that one cannot be noarch, this one can. So the situation is the same. However, if this package uses the old and depreacated scripts, then you are correct, this cannot be noarch :-/

PR #19920, opened 2022-08-03
Add TIES20 package

adw62:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -1,13 +1,13 @@
 {% set name = "TIES20" %}
-{% set version = "20.10" %}
+{% set version = "2.0.1" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 source:
-  url: https://github.com/UCL-CCS/TIES/archive/refs/tags/{{ version }}.tar.gz
-  sha256: a8b74cd61e66d4fbb8b4565cc18eaa6ac2280743b8764c57bacd67ec4158882a
+  url: https://github.com/UCL-CCS/TIES/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: b3c6f7b1cb60415dc63f86ccc8a231c74bc1bc133746a3cde4b494f6fa686261
 
 build:
   number: 0
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -17,9 +17,6 @@ build:
     - ties = ties:cli.command_line_script
 
 requirements:
-  build:
-    - {{ compiler('c') }}
-    - {{ compiler('cxx') }}
   host:
     - python
     - pip
@@ -46,7 +43,7 @@ test:
   commands:
     - pip check
     - ties --help
-    - python -m ties.testing.run
+    - pytest -m ties.testing.run
 
 about:
   home: https://github.com/UCL-CCS/TIES20
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -17,6 +17,8 @@ build:
     - ties = ties:cli.command_line_script
 
 requirements:
+  build:
+    - {{ compiler('c') }}
   host:
     - python
     - pip
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -45,7 +45,7 @@ test:
   commands:
     - pip check
     - ties --help
-    - pytest -m ties.testing.run
+    - python -m ties.testing.run
 
 about:
   home: https://github.com/UCL-CCS/TIES20
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "TIES20" %}
-{% set version = "2.0.1" %}
+{% set version = "2.0.2" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/UCL-CCS/TIES/archive/refs/tags/v{{ version }}.tar.gz
-  sha256: b3c6f7b1cb60415dc63f86ccc8a231c74bc1bc133746a3cde4b494f6fa686261
+  sha256: 1225060a1a7ef81235a51d97e51e38dac58bb6892cdbd14b8de55b2cdfd5443d
 
 build:
   number: 0
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -45,10 +45,7 @@ test:
     - pytest
   commands:
     - pip check
-    - ls $SRC_DIR/
-    - ls $SRC_DIR/..
-    - ls $SRC_DIR/../..
-    - ls $SRC_DIR/../../..
+    - ls $PREFIX
     - pytest $SRC_DIR/../../../testing
 
 about:
--- a/recipes/TIES20/meta.yaml
+++ b/recipes/TIES20/meta.yaml
@@ -6,8 +6,8 @@ package:
   version: {{ version }}
 
 source:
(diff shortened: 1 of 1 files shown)


adw62:
@conda-forge/help-python it's ready for the review.

We are failing on Windows as there is not ambertools build for Windows. Open to any ideas for a work around or happy to proceed if Windows fail is okay in this case.

Thanks,
Alex

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  noarch: python
+  entry_points:
+    - ties = ties:cli.command_line_script
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
carterbox, line 22: Noarch implies no compiled code. You need to choose between noarch OR compiled cython modules.

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  host:
+    - python >=3.7
+    - pip
+    - numpy
+    - cython
+  run:
+    - python >=3.7
+    - ambertools >=20.0,<=22.0
+    - libgfortran
```
carterbox, line 31: libgfortran should not be manually added in the run section. If you need to link to a fortran library you probably should have had the fortran compiler in the build requirements.

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - ties_md
+    - numpy
+    - cython
+    - setuptools
```
carterbox, line 35: These should probably be in the host section.

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+test:
+  imports:
+    - ties
+  requires:
+    - pip
+    - pytest
+  commands:
+    - pip check
+    - python -c "import pytest; pytest.main(['./TIES20-{{ version }}/testing'])"
```
carterbox, line 49: Where is `./TIES20-{{ version }}/testing` located? I don't see it in the recipe directory.

carterbox (changes requested):
Fails for noarch are fine if it is due to missing dependencies.

carterbox:
> Also any ideas for best way to add dwave_networkx with out using pip like this:

If you need another dependency that is not on conda-forge, then you should add it to the PR in its own folder. You cannot depend on PYPI packages.

BastianZim:
> Also any ideas for best way to add dwave_networkx with out using pip like this:

The package is called dwave-networkx on conda-forge.

recipes/TIES20/meta.yaml
bieniekmateusz: Thanks, we'll remove the "noarch: python".

recipes/TIES20/meta.yaml
bieniekmateusz: We can remove it. I believe we were fixing an issue in a dependency which is not relevant anymore. Thanks

recipes/TIES20/meta.yaml
bieniekmateusz: It's in the repository under `testing`. Can we use the env variable `SRC_DIR` to access it? Ie 
```
- pytest $SRC_DIR/testing
```

adw62:
@conda-forge/help-python it's ready for the review.

We see OSX fail for Python == 3.10 so we would like to limit the python version. I notice the linter does not like Non noarch packages with python version constraints so if there is a more correct way to achieve this please let me know.

Thanks,
Alex

carterbox:
Add a skip statement that skips the python version of your choosing.

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors

adw62:
@conda-forge/help-python it's ready for the review.

Thanks,
Alex

recipes/TIES20/meta.yaml
carterbox: If you want these files available for testing, then you should add them to `test/source_files`

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#source-files

carterbox (changes requested):

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+test:
+  imports:
+    - ties
+  requires:
+    - pip
+    - pytest
+  commands:
+    - pip check
+    - ties --help
+    - python -m ties.testing.run
```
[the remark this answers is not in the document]
adw62: Thanks for your input it is very helpful. Please may I ask, if we don't run the tests in the site-packages directory what is the alternative location for best practice?

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - tabulate
+
+test:
+  imports:
+    - ties
+  requires:
+    - pip
+    - pytest
+  commands:
+    - pip check
```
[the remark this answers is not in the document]
adw62: Added these

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "TIES20" %}
+{% set version = "20.9" %}
```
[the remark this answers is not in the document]
adw62: Corrected version

recipes/TIES20/meta.yaml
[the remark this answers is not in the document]
adw62: On more reflection I think I understand that `python -m ties.testing.run` is a much nicer way to run the tests. It should work provided the test paths are correct which they are since we copy to a tmp dir. Therefore I think things are okay as they are? Please correct me if I'm wrong.

adw62:
@conda-forge/help-python it's ready for the review.

A little unsure on if our use of a tmp dir to run the tests is okay but everything else is resolved.

Cheers,
Alex

recipes/TIES20/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  number: 0
+  skip: True  # [win or osx and py==310]
+  script: {{ PYTHON }} -m pip install . -vv
+  entry_points:
+    - ties = ties:cli.command_line_script
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
carterbox, line 22: Is the CXX compiler actually used? If not, remove it. Cython compiles to c code.

recipes/TIES20/meta.yaml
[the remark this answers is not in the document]
carterbox: [Your tests are failing](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=612349&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=7563), but you are returning exit code 0, so conda doesn't know they are failing.

If the tests don't require generating any new files, then it's not necessary to copy all the tests to a new directory. Just run them in place. Use [importlib](https://docs.python.org/3/library/importlib.resources.html) if you need to access files or maybe `os.path.dir(__file__)`.

carterbox (changes requested):

recipes/TIES20/meta.yaml
bieniekmateusz: Thank you, resolved.

bieniekmateusz:
Thank you @carterbox, I also resolved the test cases. If it's ok, I left it with the copying for now as we're intending to add more test cases that include generating files.

recipes/TIES20/meta.yaml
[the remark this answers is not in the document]
bieniekmateusz: The tests have been addressed now.

bieniekmateusz (approved):
Tests & minimal build requirements.

PR #19973, opened 2022-08-08
Added recipe for `deepspeed`

sarthakpati:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/deepspeed/meta.yaml
+++ b/recipes/deepspeed/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - pytorch >=1.7
   run:
     - python >=3.6
+    - pytorch >=1.7
 
 test:
   imports:


recipes/deepspeed/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.6
+    - pytorch >=1.7
```
h-vetinari, line 21: AFAICT you're going to need more build tools here? Previously the python-wrappers for ninja were missing, but it seems that upstream now [asks for](https://github.com/microsoft/DeepSpeed/blob/master/setup.py#L8) "visual cpp build tool", which sounds windows-only, though there is an [`install.sh`](https://github.com/microsoft/DeepSpeed/blob/master/install.sh) file.

recipes/deepspeed/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 0ee39e1b01fe4ecbd89e7c4358db6327cee6fe52a201157999e1c995985453a4
+
+build:
+  noarch: python
```
h-vetinari, line 13: It sounds very much like this package will not be noarch (which also only can be built on linux AFAIK).

recipes/deepspeed/meta.yaml
h-vetinari: Please remove the lower bound for pytorch. Conda-forge currently builds everything against 1.10 (and soon 1.11 & 1.12), because the pytorch ABI can essentially change with every version. In particular, each build with a host-dep of `pytorch` will only be installable with a specific pytorch-version (and the migrators will make sure that - by and large - this is the latest one).

If you insist on keeping a lower bound, you need to do something like:

```suggestion
  script: {{ PYTHON }} -m pip install . -vv
  number: 0

requirements:
  host:
    - pip
    - python >=3.6
    # unpinned for migrator to pick it up
    - pytorch
    # lower bound 
    - pytorch >=1.7
```

h-vetinari (commented):
Previous attempt for reference: #19021

h-vetinari:
> My bad, it seems like I had missed a few things. Closing this out.

Um, that's completely normal? 😅 

The review was not meant to be discouraging. It's not an easy library to package, completely normal that this needs a few iterations.

PR #19995, opened 2022-08-09
add gdptools

ocefpaf:
This one falls under the exception in CFEP-05 b/c it won't see a "stable" release anytime soon, even though it is a stable release.

See https://github.com/conda-forge/cfep/blob/main/cfep-05.md#exceptions


ocefpaf:

ocefpaf:

recipes/gdptools/tests/test_RWE_Class.py
```diff
@@ -0,0 +1,138 @@
...
+from gdptools.ancillary import _generate_weights_pershp
+from gdptools.ancillary import _get_cells_poly
+from gdptools.ancillary import _get_crs
+from gdptools.ancillary import _get_data_via_catalog
+from gdptools.ancillary import _get_shp_file
+from gdptools.gdp_data_class import CatGrids
+from gdptools.gdp_data_class import CatParams
+
+
+@pytest.mark.parametrize(
```
msarahan, line 23: in the longer-term, these should probably move to the project instead of this recipe, right?

ocefpaf: Yep. I'm planning on sending a PR upstream so we can remove this from here.

recipes/gdptools/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+test:
+  files:
+   - tests/
+  requires:
+    - pip
+    - pytest
+  imports:
+    - gdptools
+  commands:
+    # Some of the pins are too aggressive making the package impossible to install.
```
msarahan, line 63: The pins are terribly tight. Would it be ok to remove some of the upper bounds? I guess you really already have on several of them.

I've been advising people to not have upper bounds unless they really know they need them. Maybe I'm wrong on that. I feel like here at conda-forge, we have repodata patching to fix/impose upper bounds when needed, and we should not prematurely introduce conflicts that don't really exist.

recipes/gdptools/tests/test_RWE_Class.py
ocefpaf: That is why I added the tests. I relaxed all those that I could and ensure a passing build. The others will need some upstream modifications.

Yeah, folks usually confuse apps with libraries.

recipes/gdptools/meta.yaml
ocefpaf: That is why I added the tests. I relaxed all those that I could and ensure a passing build. The others will need some upstream modifications.

Yeah, folks usually confuse apps with libraries.

PR #20168, opened 2022-08-23
add pyhdk

leshikus:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pyhdk/meta.yaml
+++ b/recipes/pyhdk/meta.yaml
@@ -1,0 +1,150 @@
+{% set version = "0.1.0" %}  # PEP 386
+{% set source_version = "0.1" %}
+{% set base_version = "0.1.0" %}
+{% set number = "0" %}
+{% set cuda_enabled = cuda_compiler_version is not undefined and cuda_compiler_version == '11.0' %}
+{% set build_ext_version = "1.0.0" %}
+{% set build_ext = "cuda" if cuda_enabled else "cpu" %}
+{% set build_string = "h{}_{}".format(PKG_HASH, number) %}
+{% set build_string_ext = "{}_{}".format(build_string, build_ext) %}
+{% set py_build_string_ext = "py{}{}_{}".format(CONDA_PY, build_string, build_ext) %}
+{% set py_build_string = "py{}{}".format(CONDA_PY, build_string) %}
+{% set install_base = "opt/hdk" %}
+{% set arrow_version = "6.*" %}
+{% set arrow_proc_version = "3.*" %}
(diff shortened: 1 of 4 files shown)


leshikus:

recipes/pyhdk/meta.yaml
```diff
@@ -0,0 +1,150 @@
...
+      string: {{ py_build_string_ext }}
+      skip: true  # [cuda_compiler_version not in (undefined, "None", "11.0") or not linux64]
+      missing_dso_whitelist:
+        - '*/libcuda.*'  # [cuda_compiler_version not in (undefined, "None")]
+      track_features:
+        {{ "- arrow-cuda" if cuda_enabled else "" }}
+      missing_dso_whitelist:
+        - '*/libjvm.so'
+      ignore_run_exports:
+        - arrow-cpp {{ arrow_version }}
```
msarahan, line 44: Could you please add a comment about why these are necessary? many of these are not in your host section, and thus wouldn't be introducing run_exports anyway.

recipes/pyhdk/meta.yaml
```diff
@@ -0,0 +1,150 @@
...
+        - tbb
+        - zlib
+      rpaths:
+        - lib
+        # path for libjvm.so searching
+        - lib/server
+        # fixes "not found" in `ldd $PREFIX/lib64/libQueryEngine.so` output
+        - {{ install_base }}/lib64
+      run_exports:
+        - {{ pin_subpackage('pyhdk',  max_pin='x.x.x') }}
```
msarahan, line 62: This means that anyone who uses pyhdk as part of their build process (in the host section) needs the exact same version at runtime. Is that the intent? That is very unusual for Python packages. I understand if your tool is compiling something and leaving that binary artifact in the pyhdk-consuming package, but otherwise this constraint should be unnecessary. If it doesn't need to generate some binary artifact that gets stored in consumer packages, then I'd expect people to use this package only as a run requirement, and then run_exports isn't relevant anyway.

leshikus: Hello Mike,
thank you for your comments.

The recipe initially came from omniscidb project, and this is the reason of some outdated blocks you mentioned.

I've removed both run_exports and ignore_run_exports. The checks seem to pass. 
Thank you for putting your effort into this.

recipes/pyhdk/meta.yaml
leshikus: Hi Mike, I've deleted the section, see the full answer below

Garra1980:
@msarahan any other comments here? thanks in advance

PR #20215, opened 2022-08-26
Adding omsl - OpenModelica Modelica Standard Language

joewa:
Adding Modelica Standard Language that can be used with the package omcompiler

Checklist
- [x] Build number is 0.


joewa:
I am happy to provide my first PR and I suppose its ready for a review.

Can you please have a look @conda-forge/staged-recipes or `@conda-forge/help-python-c`

Many thanks!

recipes/omsl/conda_build_config.yaml
```diff
@@ -0,0 +1,13 @@
+mslversion:
+  - 3.2.3
+  - 4.0.0
```
carterbox, line 3: Do not use the conda_build_config to build multiple versions of osml. Start with the older version, then once this PR is merged, open a PR on the feedstock to update to the newer version. If you need to maintain v3.x and v4.x simultaneously, then create a branch for each in the feedstock. Every commit on the feedstock is published.

recipes/omsl/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+build:
+  script_env:
+   - MSLVERSION={{ mslversion }}
+  number: 0
+  skip: true  # [not linux]
+
+requirements:
+  host:
+    - python
+    - pip
```
carterbox, line 20: Why is python listed in the host requirements, the install script is only copying files, it doesn't look like python is used at all?

carterbox (changes requested):
@conda-forge/omcompiler, I'm not familiar with this Modelica language. Is it standard practice to install uncompiled libraries to `/lib/omlibrary/`? Are `mo` files compiled or are they interpreted like python files?

recipes/omsl/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  run:
+    - python
+    - omcompiler
+    - libiconv           # required by omcompiler (if not already present on the host)
+    - gcc_linux-64       # required by omcompiler (if not already present on the host)
+    - gxx_linux-64       # required by omcompiler (if not already present on the host)
+    - gfortran_linux-64  # required by omcompiler (if not already present on the host)
+    - make               # required by omcompiler (if not already present on the host)
+    - cmake              # required by omcompiler (if not already present on the host)
```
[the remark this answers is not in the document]
joewa: Agreed, I will reach out to the maintainer of @conda-forge/omcompiler-feedstock asking him to add these packages in the run section.
These packages are installed on most many hosts by default. I found that they are required when installing omcompiler on a fresh container in the cloud. Can you please have a look @jschueller ?

In the first step, `omcompiler` builds C/C++ code from the mo-files (and the libraries). In the 2nd step, it calls gxx to build a binary that represents the model and returns the timeseries. The the "Modelica Standard Library" is used as a foundation for almost all Modelica models. This feedstock is intended to make it  installable through conda.

And yes, `omcompiler` expects to find the Modelica libraries either in `/lib/omlibrary/` or in `~/.openmodelica/libraries/`. The 1st is the default, e.g. when the omcompiler and the Modelica libraries are installed though `apt-get` in a Debian environment (of course, `conda` and `apt` will never be mixed). The 2nd is used by the [package manager](https://openmodelica.org/doc/OpenModelicaUsersGuide/latest/packagemanager.html) which is built in `omcompiler` enabling the installation of Modelica libraries in a mo-script via `installPackage`. Unfortunately, this cannot be used in our CI environment.

recipes/omsl/meta.yaml
joewa: Correct, I will remove it.

recipes/omsl/conda_build_config.yaml
joewa: OK, understood, thanks for explaining this. Indeed, I will have to maintain v3.x and v4.x because they are not compatible and v3.x is still widely used in production. So I removed conda_build_config.

joewa:
I hope its OK now.
Can you please have a look @conda-forge/staged-recipes @carterbox ?
Many thanks!

recipes/omsl/meta.yaml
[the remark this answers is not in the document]
carterbox: Why aren't the mo-files precompiled and shipped and dynamically linked into end-users' models? Is this a JIT compiler-type framework where mo-files are compiled as-needed?

recipes/omsl/meta.yaml
carterbox: Same question for pip. Why is pip needed to install these libraries?

recipes/omsl/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  script_env:
+   - MSLVERSION={{ mslversion }}
+  number: 0
+  skip: true  # [not linux]
+
+requirements:
+  host:
+    - pip
+  run:
+    - omcompiler
```
carterbox, line 22:
```suggestion
```
omsl is a library right? It should not depend on a compiler package at runtime. Users should know that they need a compiler to use these files. It seems that this library is very similar to a C/CXX header-only library i.e. Eigen. The Eigen package is a collection of source files, and it does not depend on a compiler.

carterbox (changes requested):

carterbox (changes requested):
If this package is really just a source distribution, then we could probably mark it as `noarch: generic` as long as the install directory is the same for linux/unix/windows?

recipes/omsl/meta.yaml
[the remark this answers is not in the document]
jschueller: feel free to send a PR against the omcompiler package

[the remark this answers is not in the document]
joewa: This is indeed a JIT compiler-type framework where the mo-files are compiled as needed. In general "a model" consists of a set of (mostly) acausal "differential algebraic equations" (DAE) where the number of variables and equations is always equal. Most user models are composed from (sub-) models from the MSL (this library). The mo-files cannot be precompiled because the causality (i.e. the direction of a signal flow) depends on how a model is composed (as far as I understood). omcompiler creates the C code which has causality.

joewa:
I agree, we could mark this package (omsl) as `noarch: generic` because this is really just a source distribution. But please mind that its probably useless without `omcompiler`, which is currently available for Linux only. There might be a Windows release in the future but [MaxOSX](https://openmodelica.org/download/download-mac) will never be supported.

The current tests are invoking `omcompiler` to build a model from this library which ensures that the library is installed in the correct location and OK.
So if we would mark it as `noarch: generic`, I suppose we would have to remove the tests because `omcompiler` is available for Linux only.

My proposal would be to keep it as it is. Of course `noarch: generic` will work, but I will have to remove the tests then.

What do you think @carterbox ?

joewa:
Can you please have a look @conda-forge/staged-recipes @carterbox ?

carterbox (approved):
Thanks for answering all my questions. Consider noarch:generic if future releases of the compiler are available for other platforms.

joewa:
Thanks you too for all your explanations!
I learned a lot!

PR #20231, opened 2022-08-29
Add parallelio

xylar:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -5,7 +5,7 @@ export NETCDF_FORTRAN_PATH=$(dirname $(dirname $(which nf-config)))
 
 mkdir build
 cd build
-FC=mpifort CC=mpicc CXX=mpicxx cmake \
+FC=mpifort CC=mpicc cmake \
     -DCMAKE_INSTALL_PREFIX=$PREFIX \
     -DBUILD_SHARED_LIBS:BOOL=ON \
     -DPIO_USE_MALLOC:BOOL=ON \
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -14,7 +14,9 @@ FC=mpifort CC=mpicc CXX=mpicxx cmake \
     -DPIO_ENABLE_EXAMPLES:BOOL=OFF \
     -DPIO_ENABLE_TIMING:BOOL=OFF \
     -DNetCDF_C_PATH=$NETCDF_C_PATH \
-    -DNetCDF_Fortran_PATH=$NETCDF_FORTRAN_PATH ..
+    -DNetCDF_Fortran_PATH=$NETCDF_FORTRAN_PATH \
+    -DWITH_PNETCDF:BOOL=OFF \
+    ..
 
 make
 
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+export NETCDF_C_PATH=$(dirname $(dirname $(which nc-config)))
+export NETCDF_FORTRAN_PATH=$(dirname $(dirname $(which nf-config)))
+
+mkdir build
+cd build
+FC=mpifort CC=mpicc CXX=mpicxx cmake \
+    -DCMAKE_INSTALL_PREFIX=$PREFIX \
+    -DBUILD_SHARED_LIBS:BOOL=ON \
+    -DPIO_USE_MALLOC:BOOL=ON \
+    -DPIO_ENABLE_TOOLS:BOOL=OFF \
+    -DPIO_ENABLE_TESTS:BOOL=OFF \
+    -DPIO_ENABLE_EXAMPLES:BOOL=OFF \
+    -DPIO_ENABLE_TIMING:BOOL=OFF \
+    -DNetCDF_C_PATH=$NETCDF_C_PATH \
+    -DNetCDF_Fortran_PATH=$NETCDF_FORTRAN_PATH ..
+
+make
+
+# make tests
+# ctest
+
+make install
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -18,9 +18,9 @@ FC=mpifort CC=mpicc cmake \
     -DWITH_PNETCDF:BOOL=OFF \
     ..
 
-make
+cmake --build .
 
 # make tests
 # ctest
 
-make install
+cmake --install .
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -6,6 +6,8 @@ export NETCDF_FORTRAN_PATH=$(dirname $(dirname $(which nf-config)))
 mkdir build
 cd build
 FC=mpifort CC=mpicc cmake \
+    ${CMAKE_ARGS} \
+    -DCMAKE_BUILD_TYPE=Release \
     -DCMAKE_INSTALL_PREFIX=$PREFIX \
     -DBUILD_SHARED_LIBS:BOOL=ON \
     -DPIO_USE_MALLOC:BOOL=ON \
--- a/recipes/parallelio/build.sh
+++ b/recipes/parallelio/build.sh
@@ -1,12 +1,14 @@
 #!/bin/bash
 
+# Remove some files that may cause static linking?
+rm -rfv $PREFIX/lib/objects*
+
 export NETCDF_C_PATH=$(dirname $(dirname $(which nc-config)))
 export NETCDF_FORTRAN_PATH=$(dirname $(dirname $(which nf-config)))
 
 mkdir build
 cd build
 FC=mpifort CC=mpicc cmake \
-    ${CMAKE_ARGS} \
     -DCMAKE_BUILD_TYPE=Release \
     -DCMAKE_INSTALL_PREFIX=$PREFIX \
     -DBUILD_SHARED_LIBS:BOOL=ON \
--- a/recipes/parallelio/conda_build_config.yaml
+++ b/recipes/parallelio/conda_build_config.yaml
@@ -1,0 +1,4 @@
+mpi:
+  - openmpi
+  - mpich
+
--- a/recipes/parallelio/conda_build_config.yaml
+++ b/recipes/parallelio/conda_build_config.yaml
@@ -2,3 +2,10 @@ mpi:
   - openmpi
   - mpich
 
+# temporary because mambabuild seems to be getting confused
+hdf5:
+  - 1.12.1
+libnetcdf:
+  - 4.8.1
+netcdf_fortran:
+  - 4.5
--- a/recipes/parallelio/conda_build_config.yaml
+++ b/recipes/parallelio/conda_build_config.yaml
@@ -1,11 +1,3 @@
 mpi:
   - openmpi
   - mpich
-
-# temporary because mambabuild seems to be getting confused
-hdf5:
-  - 1.12.1
-libnetcdf:
-  - 4.8.1
-netcdf_fortran:
-  - 4.5
--- a/recipes/parallelio/meta.yaml
+++ b/recipes/parallelio/meta.yaml
@@ -5,9 +5,11 @@
 # recipe-lint fails if mpi is undefined
 {% set mpi = mpi or 'mpich' %}
 {% if mpi == "mpich" %}
-# prioritize mpich via build number
-{% set build = build + 100 %}
+  # prioritize mpich via build number
+  {% set build = build + 100 %}
 {% endif %}
+
+{% set mpi_prefix = "mpi_" + mpi %}
 
 package:
   name: parallelio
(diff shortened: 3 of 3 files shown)


xylar:
This appears to be needed for https://github.com/conda-forge/esmf-feedstock/pull/78.  Without an external `parallelio`, ESMF attempts to vendor its own internal version (which is not building successfully).

recipes/parallelio/conda_build_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+  - openmpi
+  - mpich
+
+# temporary because mambabuild seems to be getting confused
+hdf5:
+  - 1.12.1
+libnetcdf:
+  - 4.8.1
+netcdf_fortran:
+  - 4.5
```
xylar, line 11: As commented, here, I wasn't able to get `conda mambabuild` to run locally without these pins, taken from `conda-forge-pinning`.  I kept seeing errors about not finding:
```
netcdf-fortran mpi_openmpi.*
```
I could not figure out why it was getting confused about a constraint on a build string that it seems to think is on the version.

xylar: I will remove these as soon as this is in its own feedstock.

xylar:
@edwardhartnett and @jedwards4b, I'm hoping to add a recipe for ParallelIO on conda-forge.  I have a lot of experience maintaining recipes here but I'm better with python packages than c/c++/fortran libraries.  Still, I'm a developer with E3SM and can get help from that team if need be.

I'm wondering if you have input on this recipe, particularly the build script (`build.sh`).  You would both be welcome to become maintainers of the package but I'm assuming that's not something you're interested in doing.

recipes/parallelio/build.sh
```diff
@@ -0,0 +1,24 @@
...
+FC=mpifort CC=mpicc CXX=mpicxx cmake \
+    -DCMAKE_INSTALL_PREFIX=$PREFIX \
+    -DBUILD_SHARED_LIBS:BOOL=ON \
+    -DPIO_USE_MALLOC:BOOL=ON \
+    -DPIO_ENABLE_TOOLS:BOOL=OFF \
+    -DPIO_ENABLE_TESTS:BOOL=OFF \
+    -DPIO_ENABLE_EXAMPLES:BOOL=OFF \
+    -DPIO_ENABLE_TIMING:BOOL=OFF \
+    -DNetCDF_C_PATH=$NETCDF_C_PATH \
+    -DNetCDF_Fortran_PATH=$NETCDF_FORTRAN_PATH ..
```
xylar, line 17: For now, I am only including NetCDF and not PNetCDF because there isn't a conda-forge package for PNetCDF.  I hope to follow up soon with a PNetCDF package at which point I will make a build of ParallelIO that includes PNetCDF support.

recipes/parallelio/build.sh
```diff
@@ -0,0 +1,24 @@
...
+
+make
+
+# make tests
+# ctest
```
xylar, line 22: I had trouble with some tests when I ran locally:
```
[ 71%] Linking C executable piodecomptest
/home/xylar/miniconda3/conda-bld/parallelio_1661760167351/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/12.1.0/../../../../x86_64-conda-linux-gnu/bin/ld: CMakeFiles/piodecomptest.dir/mpi_argp.c.o: in function `argp_usage':
mpi_argp.c:(.text.argp_usage+0x0): multiple definition of `argp_usage'; CMakeFiles/piodecomptest.dir/piodecomptest.c.o:piodecomptest.c:(.text.argp_usage+0x0): first defined here
/home/xylar/miniconda3/conda-bld/parallelio_1661760167351/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/12.1.0/../../../../x86_64-conda-linux-gnu/bin/ld: CMakeFiles/piodecomptest.dir/mpi_argp.c.o: in function `_option_is_short':
mpi_argp.c:(.text._option_is_short+0x0): multiple definition of `_option_is_short'; CMakeFiles/piodecomptest.dir/piodecomptest.c.o:piodecomptest.c:(.text._option_is_short+0x0): first defined here
/home/xylar/miniconda3/conda-bld/parallelio_1661760167351/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/12.1.0/../../../../x86_64-conda-linux-gnu/bin/ld: CMakeFiles/piodecomptest.dir/mpi_argp.c.o: in function `_option_is_end':
mpi_argp.c:(.text._option_is_end+0x0): multiple definition of `_option_is_end'; CMakeFiles/piodecomptest.dir/piodecomptest.c.o:piodecomptest.c:(.text._option_is_end+0x0): first defined here
```
I don't yet feel like I have the expertise to debug this but I hope to figure out whether I have something misconfigured here (most likely a missing cmake flag) or if there is a problem with the test.

jedwards4b: I've not seen this error before, but the problem appears to be in <argp.h>

jedwards4b:
Yes you can list me as a maintainer - my github id is jedwards4b.
   
I think that it is worth mentioning that we have a [spack package](https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/parallelio) for parallelio and e3sm has the [scorpio fork](https://github.com/E3SM-Project/scorpio) of the repository.

xylar:
Thanks @jedwards4b, I've added you.  Your insights as a fellow maintainer will be really valuable!

> I think that it is worth mentioning that we have a [spack package](https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages/parallelio) for parallelio and e3sm has the [scorpio fork](https://github.com/E3SM-Project/scorpio) of the repository.

Yes, I'm aware of ParallelIO being on Spack as well.  On HPC systems in most contexts, that's likely to be the preferred way to build.  In cases where folks want a conda environment with a bunch of python packages for analysis, including ESMPy, conda will continue to be an appealing alternative.  The latest ESMF seems like it will be happier with an external PIO.

recipes/parallelio/build.sh
xylar: I have commented out that test and can get the rest to build but most (all?) MPI tests fail on run because I don't have enough cores on my laptop.  I think this is going to be even worse for CI.  So I think we should let `ctest` be for now.

xylar:
@conda-forge/help-c-cpp, this is ready for review.  Your recommendations to improve the recipe are most welcome.  My expertise is more with python than with c/c++ packages.  But I use the ESMF package frequently, and that package would very much benefit from this one.

recipes/parallelio/conda_build_config.yaml
carterbox:
```
netcdf-fortran mpi_openmpi.*
```
Above is an invalid package spec. It should be something like this:

```
netcdf-fortran * mpi_openmpi_*
```
or
```
netcdf-fortran =*=mpi_openmpi_*
```

carterbox: I assume this problem is because you weren't using the `build-locally.py` script? This script does environment/recipe augmentation stuff in addition to starting a docker container.

recipes/parallelio/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+    - libgomp      # [linux]
+  host:
+    - {{ mpi }}
+    # these need to be listed twice so conda-smithy build picks up the pins
+    - hdf5
+    - hdf5 * {{ mpi_prefix }}_*
+    - libnetcdf
+    - libnetcdf * {{ mpi_prefix }}_*
+    - netcdf-fortran
+    - netcdf-fortran * {{ mpi_prefix }}_*
```
carterbox, line 48: Check the logs. The link inspector thinks you are overdepending on netcdf-fortran and hdf5 because it can't find any dynamic links to these libraries. Are y'all doing static linking?

carterbox (changes requested):

recipes/parallelio/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('fortran') }}
+    - cmake
+    - make
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
```
carterbox, line 39: The openmp packages are also unused according to the link inspector.

recipes/parallelio/conda_build_config.yaml
xylar: I can't install docker on my laptop because it interferes with my ability to use wifi on my laptop at work.  I can take these out and see if CI does okay.

xylar: I see you went ahead and took these out, and things built.  Thanks.  Thanks for the tip about `build-locally.py`. I'll see if I can find a solution where I can install Docker and still use LANL's wifi.

recipes/parallelio/meta.yaml
xylar: Okay, I'm fine with removing these.  It may be that I am not enabling OpenMP support properly but this can be something we add down the road.

recipes/parallelio/meta.yaml
xylar: @carterbox, you may be right that `hdf5` isn't needed.  It isn't explicitly referenced in the `cmake` command.

Regarding `netcdf-fortran`, this is clearly a dependency.  The problem could easily lie with that library rather than here.  It would appear the static and shared libraries are both being included in that package:
https://github.com/conda-forge/netcdf-fortran-feedstock/blob/main/recipe/build.sh#L32-L58
As a result, presumably the static library is being linked here instead of the dynamic?

I won't have time for awhile but I can look into splitting that package into only the dynamic library and a separate `netcdf-fortran-static` package.

xylar:
@carterbox, I really appreciate your careful review and the changes you pushed.  I'm going to try to fix `netcdf-fortran` and then I will remove the other dependencies you flagged as unneeded and ask for a re-review.

recipes/parallelio/conda_build_config.yaml
carterbox: Pretty it works with podman too in case that helps.

xylar: That could be worth a try, thanks!

recipes/parallelio/meta.yaml
xylar: This is turning out to be more complicated than I hoped.  The c library has not yet been split into separate dynamic and static packages:
https://github.com/conda-forge/libnetcdf-feedstock/pull/135
That's presumably a prerequisite for splitting the fortran library, since it has the c library as a dependency.

xylar:
@carterbox, I could use some help here.  I believe my fellow maintainers and I have made the changes needed so that the latest `libnetcdf` and `netcdf-fortran` package no longer include static libraries.

I have removed unneeded HDF5 and OpenMP dependencies.

Now, we are seeing the same weird glitch here that I was seeing in local (non-docker) builds before:
```
 - libgfortran5 >=11.3.0
 - netcdf-fortran mpi_openmpi_.*
 - libgfortran 5.*
 - libnetcdf >=4.8.1,<4.8.2.0a0 mpi_openmpi_*
 - openmpi >=4.1.4,<5.0a0

with channels:
 - local
 - conda-forge
 - conda-forge

The reported errors are:
-   - nothing provides requested netcdf-fortran mpi_openmpi_.*

```
`mamba` is seeing the build constraint as a version constraint.  Any clue?

carterbox:
I think the problem is not with this recipe, but with a broken build of netcdf-fortran that had an incorrect run_export which caused downstream packages to be broken.

xylar:
@carterbox, thanks, I'll rebuild that release of netcdf-fortran to fix it.  I think you're likely right.

carterbox:
No. You should wait.

xylar:
> No. You should wait.

Can you be more specific?

carterbox:
Please ping the help team again when the repodata has been patched and the build environment is solvable. Thanks!

xylar:

xylar:
@conda-forge/help-c-cpp, this is ready for another look when you have time.

it seems that the issues with `netcdf-fortran` have been fixed by a new build.  I have also requested that the builds with bad `run_export` get marked as broken, see https://github.com/conda-forge/admin-requests/pull/497.  We can wait for that PR to get merged if necessary.

xylar:
@conda-forge/help-c-cpp and @carterbox, I'm keen to have this re-reviewed when you can.

recipes/parallelio/meta.yaml
carterbox: Looks like the static libs were removed after the latest build on this PR. I will restart the build to see if it builds successfully without static libs for netcdf-fortran.

https://github.com/conda-forge/netcdf-fortran-feedstock/pull/76

carterbox:
Restarting build to check compatibility with latest netcdf builds.

recipes/parallelio/meta.yaml
carterbox: [Logs are still showing no dynamic linkage to netcdf-fortran.
](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=598878&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1486) We should be able to detect a link to `libnetcdff.so.7` if it is actually being used.

carterbox:
I'm stumped about the non-detectable links / overdepending warning for netcdf-fortran. I've added this library to the `build/ignore_run_exports_from` section. Since it looks like we're building against the version with static libs removed, it seems to me that maybe only the headers are used(?) or the build system for this library requests packages that aren't actually needed.

PR #20238, opened 2022-08-29
Add uvw

asafkahlon:
Checklist
- [x] Build number is 0.


asafkahlon:
@conda-forge/help-c-cpp @conda-forge/staged-recipes Please review :)

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+test:
+  commands:
+    - test -f ${PREFIX}/include/uvw.hpp
+
+about:
+  home: https://github.com/skypjack/uvw
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+  summary: A C++ wrapper for libuv
```
saraedum, line 36:
```suggestion
  summary: A C++ wrapper for libuv, a multi-platform support library with a focus on asynchronous I/O
```

It's maybe nice to give people an idea what this is about.

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,40 @@
+{% set name = "uvw" %}
+{% set version = "v2.12.1_libuv_v1.44" %}
+{% set sha256 = "3460842778e91e7d2fae4201e49e4521e9d94fbbf3891ae6c52d3c9fc0673598" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
```
saraedum, line 7: The version suffix seems to track which (minimal?) version of libuv gets vendored into uvw. This versioning scheme likely does not play nice with conda. Could it just be `2.12.1`?

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+build:
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  build:
+    - cmake
+    - make  # [unix]
+    - {{ compiler('cxx') }}
+    - libuv
```
saraedum, line 23: Why is libuv required as a build dependency? That's very uncommon.

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/skypjack/uvw/archive/{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  skip: true  # [win]
```
saraedum, line 16: Why are we not building for Windows?

recipes/uvw/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cmake -Bbuild ${CMAKE_ARGS} -D BUILD_TESTING=OFF -D CMAKE_BUILD_TYPE=Release -D CMAKE_INSTALL_PREFIX=$PREFIX .
```
saraedum, line 3: uvw seems to provide a shared library. Sholudn't we try to build that one as well?

saraedum (changes requested):

recipes/uvw/meta.yaml
asafkahlon: You're right. Removed.

recipes/uvw/meta.yaml
asafkahlon: Added

recipes/uvw/meta.yaml
asafkahlon: Added

recipes/uvw/build.sh
asafkahlon: Added

recipes/uvw/meta.yaml
asafkahlon: Yes, it can be `2.12.1` but we still have to save the currect compatible `libuv` version, so I added it with another variable and constructed the full filename with both `uvw` and `libuv` version variables.

asafkahlon:
@saraedum @conda-forge/help-c-cpp @conda-forge/staged-recipes ready for a second review 😄

asafkahlon:
@saraedum @conda-forge/help-c-cpp @conda-forge/staged-recipes
Ping?

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,45 @@
+{% set name = "uvw" %}
+{% set uvw_version = "2.12.1" %}
```
saraedum, line 2: I am not entirely sure how our automatic updates work and whether they might get confused by this naming. I think it's more standard to call this `version`.

saraedum:
The build seems to vendor libuv:

```
-- Installing: $PREFIX/include/uvw/uv/include
-- Installing: $PREFIX/include/uvw/uv/include/uv.h
-- Installing: $PREFIX/include/uvw/uv/include/uv
-- Installing: $PREFIX/include/uvw/uv/include/uv/bsd.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/sunos.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/stdint-msvc2008.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/version.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/linux.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/posix.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/threadpool.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/os390.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/unix.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/tree.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/win.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/errno.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/aix.h
-- Installing: $PREFIX/include/uvw/uv/include/uv/darwin.h
-- Installing: $PREFIX/lib/uvw/libuv.so.1.0.0
-- Installing: $PREFIX/lib/uvw/libuv.so.1
-- Installing: $PREFIX/lib/uvw/libuv.so
```

Usually, dependencies should not be vendored.

recipes/uvw/meta.yaml
asafkahlon: OK, changed

asafkahlon:
@saraedum I fixed the `libuv` vendoring issue by finding the package via `pkg-config`.

Now it's all set.
I'll appreciate another (hopefully final) round of review 😄

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - make  # [unix]
+    - pkg-config
+  host:
+    - libuv
```
saraedum, line 26:
```suggestion
    # Since libuv is not in the conda-forge-pinning-feedstock yet, we need to select a libuv version explicitly
    - libuv {{ libuv_version }}
```

recipes/uvw/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  run:
+    - libuv
```
saraedum, line 28:
```suggestion
    # Since libuv is lacking a run_exports, we need to pin the version used at runtime; see https://github.com/conda-forge/libuv-feedstock/pull/58
    - libuv
```

saraedum (changes requested):

saraedum:
I am not entirely sure about the naming of this package. This is a relatively popular package knows as uvw and several distributions. However, there's already something completely different knows as uvw on pypi: https://pypi.org/project/uvw/

So, assuming that both of these want to be in conda-forge eventually, one of the two needs a different name. I am not 100% sure about the latest naming scheme but I think we try to follow pypi names when possible. There seems also to be some consensus to name C/C++ libraries as `libwhatever`, https://github.com/conda-forge/conda-forge.github.io/issues/1073.

So, I guess this one should be called `libuvw` after all.

saraedum:
Also, since this is now a shared library, it needs a `run_exports` section.

saraedum:
What is the versioning scheme of uvw? They say that they use semantic versioning but does that mean that all 2.x.x versions are ABI compatible? (This is almost never the case in my experience.)

If that's not the case, then the pin needs to be stronger, x.x or even x.x.x.

saraedum:
Without knowing anything about this library, this would seem like a breaking API & ABI change: https://github.com/skypjack/uvw/commit/9f8047d6d4f71abff46021232645a9f0daa2f748 (just randomly clicked something from the changelog.)

saraedum:
So, it should probably be `x.x`. They don't seem to be using patch releases much.

asafkahlon:
> So, it should probably be `x.x`. They don't seem to be using patch releases much.

I agree

PR #20280, opened 2022-09-02
Add Neovim and related libraries

AntoinePrv:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/luvit/build.sh
+++ b/recipes/luvit/build.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
+cmake -B build -S luv \
+	-D CMAKE_INSTALL_PREFIX="${PREFIX}" \
+	-D CMAKE_BUILD_TYPE=Release \
+	-D BUILD_SHARED_LIBS=ON \
+	-D BUILD_STATIC_LIBS=OFF \
+	-D WITH_SHARED_LIBUV=ON \
+	-D WITH_LUA_ENGINE=LuaJIT \
+	-D LUA_BUILD_TYPE=System \
+	-D LUA_COMPAT53_DIR="${PWD}/compat53" \
+	"${CMAKE_ARGS[@]}"
+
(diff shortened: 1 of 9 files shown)


AntoinePrv:
@tdejager

tdejager:
> @tdejager

Hey @AntoinePrv can you maybe add me as a collaborator in your forked repo then I can commit the unibilium stuff. :)

AntoinePrv:
Hi @tdejager,

```
    - test -f "${PREFIX}/usr/local/lib/libunibilium.${SHLIB_EXT}"
    - test -f "${PREFIX}/usr/local/include/libunibilium.h"
```
We'd typically want to have lib installed under `"${PREFIX}/lib"` and headers under `"${PREFIX}/include"`. Do you know if there is a way to achieve that with unibilium? It's surprising that specifying `DESTDIR` add `usr/local` to the path.
If needed we could make a PR there.

tdejager:
Yeah either we could patch the Makefile or send a pull request to make the
make file potentially use the ?= operator for PREFIX instead. What do you
think is best?

AntoinePrv:
A dirty fix could be to install in a local directory, then `mv` the folder to `"${PREFIX}"` (if the install folder is not hard-coded anywhere)

tdejager:
Yeah, that would be great. However `/usr/local/lib` is hardcoded in the
makefile. So I could use the DESTDIR to do something like `../../../foo`,
but that seems a bit dirty.

AntoinePrv:
Looks like it could be done with `PREFIX` rather than `DESTDIR`.
It looks like there is also a `CMakeLists.txt` but `brew` seems to not be using it either so let's roll without for now.

- [ ] There is also a `TERMINFO_DIRS` that default to using `ncurses-config` on Linux. While adding `ncurses` to the build would point to the environment (not on MacOS!), I doubt that this should be the only lookup location (at least until we ship Alacritty on conda-forge!).

tdejager:
Hi @AntoinePrv should I have a look why the osx build fails?

AntoinePrv:
If you can that would be nice!

PR #20351, opened 2022-09-08
Cosmosis standard library

vitenti:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/cosmosis-standard-library/build.sh
+++ b/recipes/cosmosis-standard-library/build.sh
@@ -1,10 +1,27 @@
 #!/usr/bin/env bash
 
-
 mkdir -p "${PREFIX}/lib/cosmosis-standard-library"
 cp -a -r * "${PREFIX}/lib/cosmosis-standard-library"
 cd "${PREFIX}/lib/cosmosis-standard-library"
 
+function DoSource() { source cosmosis-configure ; } ; DoSource
+
+export USER_CXXFLAGS="$USER_CXXFLAGS -I${PREFIX}/include"
+export USER_LDFLAGS="$USER_LDFLAGS -L${PREFIX}/lib"
+
+# This next line is a hack to workaround cosmosis makefiles
+export GSL_LIB="$GSL_LIB -L${PREFIX}/lib"
+export GSL_INC="$GSL_INC -I${PREFIX}/include"
+export LAPACK_LINK="$LAPACK_LINK -lblas"
+export CFITSIO_LIB="${PREFIX}/lib"
+export CFITSIO_INC="${PREFIX}/include"
+
 make
+
+cp "${COSMOSIS_SRC_DIR}/datablock/libcosmosis.so" "${PREFIX}/lib"
+find . -name '.a' 
+
+#find . -print0 -name '.a' | xargs -0 rm 
+#find . -print0 -name '.o' | xargs -0 rm 
 
 
--- a/recipes/cosmosis-standard-library/build.sh
+++ b/recipes/cosmosis-standard-library/build.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 mkdir -p "${PREFIX}/lib/cosmosis-standard-library"
-cp -a -r * "${PREFIX}/lib/cosmosis-standard-library"
+cp -a * "${PREFIX}/lib/cosmosis-standard-library"
 cd "${PREFIX}/lib/cosmosis-standard-library"
 
 function DoSource() { source cosmosis-configure ; } ; DoSource
--- a/recipes/cosmosis-standard-library/build.sh
+++ b/recipes/cosmosis-standard-library/build.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+
+mkdir -p "${PREFIX}/lib/cosmosis-standard-library"
+cp -a -r * "${PREFIX}/lib/cosmosis-standard-library"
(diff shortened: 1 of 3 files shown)


vitenti:

recipes/cosmosis-standard-library/build.sh
```diff
@@ -0,0 +1,27 @@
...
+export CFITSIO_LIB="${PREFIX}/lib"
+export CFITSIO_INC="${PREFIX}/include"
+
+make
+
+cp "${COSMOSIS_SRC_DIR}/datablock/libcosmosis.so" "${PREFIX}/lib"
+find . -name '.a' 
+
+#find . -print0 -name '.a' | xargs -0 rm 
+#find . -print0 -name '.o' | xargs -0 rm
```
beckermr, line 25: We have a policy of removing all `*.a` files (ie static libs) and making special packages for them. Is there a reason this is commented out? 

This might relate to some of our previous conversations. I don't recall and my apologies if it does!

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/joezuntz/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: fc815e5068a1e212c1af8af2106c42e14881f466b462c0a6c918a144e5fbb30c
+
+build:
+  number: 0
+  skip: True  # [win or (py < 36)]
+  detect_binary_files_with_prefix: True
```
beckermr, line 15: The default here is true and so this line should not be needed.

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  missing_dso_whitelist:
+    - "*/liblimber.so"
```
beckermr, line 17: What's going on here?

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - {{ compiler('cxx') }}
+    - make
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
+  host:
+    - python
+    - cosmosis
+    - cython
+    - mpich          # [mpi == "mpich"]
+    - openmpi        # [mpi == "openmpi"]
```
beckermr, line 32: These mpi selectors don't do anything here since they are only evaluated at run time. Please see the conda-forge docs for how to build MPI variants for your packages. https://conda-forge.org/docs/maintainer/knowledge_base.html#message-passing-interface-mpi

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - gsl
+    - fftw
+    - scipy
+    - astropy
+    - numpy
+    - matplotlib
+    - fitsio
+    - pyyaml
+    - scikit-learn
+    - fast-pt
```
beckermr, line 46: This host section cannot be right. Only things that cosmosis links to via code need to be here. For numpy specifically, see the docs on how to link to numpy properly. https://conda-forge.org/docs/maintainer/knowledge_base.html#building-against-numpy

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  description: |
+    It is a framework for structuring cosmological parameter estimation with a
+    focus on flexibility, re-usability, debugging, verifiability, and code 
+    sharing in the form of calculation modules. It consolidates and connects 
+    together existing code for predicting cosmic observables, and makes mapping
+    out experimental likelihoods with a range of different techniques much more
+    accessible.
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.txt
```
beckermr, line 67: The license file here is incorrect. We cannot simply apply random licenses to other people's code and then distribute it on conda-forge. As far as I can tell, the upstream package has no license. @joezuntz will need to figure out the correct license for the upstream code and apply the license himself there before we can merge, this.

recipes/cosmosis-standard-library/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  doc_url: https://cosmosis.readthedocs.io/en/latest/
+
+extra:
+  recipe-maintainers:
+    - vitenti
+    - marcpaterno
```
beckermr, line 73: All maintainers need to comment saying they are ok to maintain the recipe.

beckermr (changes requested):

recipes/cosmosis-standard-library/meta.yaml
joezuntz: This is very complicated, and is one reason I have avoided trying to share a compiled version of the CSL.

The different bits of included code are under different licenses. Some are standard licenses, but there are many different ones, GPL, LGPL, BSD, MIT. 

Some are vague, like  "You can use X freely, provided that in your publications, you cite at least the paper ..." or similar generic statements.

I understand it is permitted or acceptable to share the source code of all of them, but I would not recommend trying to sort through this to the level needed here.

There is already a pacakge `cosmosis-build-standard-library` which downloads and builds the CSL for the user. I would strongly recommend using that instead, especially as most users of CSL are also developers.  It supplies a one-line command to download and build the CSL.  If more is needed then we could add something to that command to install the result to a central location.

PR #20396, opened 2022-09-12
add klayout

curtisma:
Checklist
- [ ] Build number is 0.


curtisma:
I need you both to comment on this PR stating that you agree to be a maintainer.

@klayoutmatthias
@proppy

recipes/klayout/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+"${SRC_DIR}/build.sh"
+echo "bin-release Contents"
+echo "--------------------"
+ls bin-release
```
proppy, line 5: find? (if you want to make sure to list everything recursively?)

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+  host:
+  # All Platforms
+    - python>3
+    - pip
+    - ruby>2
+    - Qt>=5
+    - make
+    - zlib
+  # Windows
+  - expat # [not win]
```
proppy, line 51: fix indent?

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - curl
+  - pthreads-win32
```
proppy, line 53: [win]?

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+
+test:
+  # For python packages, it is useful to run pip check. However, sometimes the
+  # metadata used by pip is out of date. Thus this section is optional if it is
+  # failing.
+  commands:
+    - pip check
```
proppy, line 60:
```
klayout -v
```

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+  license_file: LICENSE
+  doc_url: https://www.klayout.de/doc.html
+  dev_url: https://github.com/KLayout/klayout
+
+extra:
+  recipe-maintainers:
+    # GitHub IDs for maintainers of the recipe.
+    # Always check with the people listed below if they are OK becoming maintainers of the recipe. (There will be spam!)
+    - curtisma
+    - klayoutmatthias
```
proppy, line 81: I agree to be co-maintainer.

recipes/klayout/build.sh
curtisma: I'll update that if necessary for debug.

recipes/klayout/meta.yaml
curtisma: done

recipes/klayout/meta.yaml
curtisma: done

recipes/klayout/meta.yaml
curtisma: Thanks!

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+# when copying and pasting from another recipe, but not really needed.
+{% set name = "klayout" %}
+{% set version = "0.27.11" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/simplejson/simplejson/releases/download/{{ version }}/klayout-{{ version }}.tar.gz
```
proppy, line 16: shouldn't that point to https://github.com/KLayout/klayout/releases ?

curtisma: yep, fixed

curtisma: done

recipes/klayout/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+# when copying and pasting from another recipe, but not really needed.
+{% set name = "klayout" %}
+{% set version = "0.27.11" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/KLayout/klayout/releases/download/{{ version }}/klayout-{{ version }}.tar.gz
```
proppy, line 16: Looks like this needs to be:
```
https://github.com/KLayout/klayout/archive/refs/tags/v{{ version }}.tar.gz
```

curtisma:
@joamatab, are you willing to be added as a maintainer as well?

curtisma:
Well I was able to fix the error about the missing pythonlib ".a" file.

joamatab:
Hi Curtis, Im happy to also be a maintainer

Thomas made it available on PyPI
@thomaslima

and Floris has it on his conda environment
@flaport

flaport:
This looks great, if possible I'd like to help to maintain this too.

on another note... do you think we could package the klayout-gui too? Something like this: https://anaconda.org/flaport/klayout-gui

curtisma:
> This looks great, if possible I'd like to help to maintain this too.

Ok, sure, I'll add you also.


> on another note... do you think we could package the klayout-gui too? Something like this: https://anaconda.org/flaport/klayout-gui

Nice!  I'll have to take a look at your work there  I was intending this to include the gui version of klayout.  Then we could possibly create a second recipe called "klayout-python" with just the python package as a lighter weight option for Python users.  

Should the "klayout" package include just the gui version or also include the python package?

At the moment I'm still working through getting it to compile.   It'd be great if you are able to help since you have experience with compiling it previously.  I've submitted several Conda recipes in the past but this is the first time I've compiled klayout.  Using conda-forge to handle the CI and CD should help make it easier to host the package and update it.  It will also make it easier for users to find it compared to having it on your own channel.

curtisma:
> Thomas made it available on PyPI

@thomaslima, should I add you as a maintainer also?  haha

flaport:
@curtisma ,

I think just calling the python-only version `klayout` and the gui-version `klayout-gui` is probably the best solution.

That way the python-only package [matches the name on pypi](https://pypi.org/project/klayout/), which will avoid confusion.

curtisma:
> @curtisma ,
> 
> I think just calling the python-only version `klayout` and the gui-version `klayout-gui` is probably the best solution.
> 
> That way the python-only package [matches the name on pypi](https://pypi.org/project/klayout/), which will avoid confusion.

ok, that makes sense.  I'll change the current recipe package name to "klayout-layout" and add a "klayout" recipe based on the wheel.

thomaslima:
> > Thomas made it available on PyPI
> 
> @thomaslima, should I add you as a maintainer also? haha

I consent. Thanks for this effort!

curtisma:
Anybody have a suggestion for getting the Linux_64 build to work?

I posted more details about it on the [conda-forge gitter forum](https://gitter.im/conda-forge/conda-forge.github.io?at=637049a4abfb5860bb1349dd)

I'm going to start with linux_64 build first and then work on the others.  I will also rename it as mentioned above once we have it working.

@proppy @flaport

proppy:
> @curtisma curtisma changed the title add klayout add klayout-gui 1 hour ago

I'd be in favor to keep `klayout` as the default name as I think most people would expect the gui to be present when installing the package from conda-forge (just like the klayout packages from other general purpose distribution).

Maybe we could have a different package (`klayout-python`?) for the Python API only version, wdyt?

curtisma:
Yeah, @proppy, sounds good.  I'll change that back.  Any suggestion on the build error?

curtisma:
I created a seperate [klayout-python PR](https://github.com/conda-forge/staged-recipes/pull/20396)

proppy:
Have you tried setting `BUILD_EXPERT=1`, my understanding was that it was added to allow forwarding compiler / flags options thru environment variables, see
https://github.com/KLayout/klayout/blob/master/build.sh#L659

But @klayoutmatthias should be able to confirm if that's helpfull.

klayoutmatthias:
I tried to debug the issue with qmake.

For short: I do not fully understand what's going on, but I have a patch that should make linux64 work. Here is my patched build.sh for "klayout" package:

```
#!/bin/bash

# even though we specify QMAKE_CXX in -expert mode, qmake still
# needs "g++" while obtaining gcc paths (while bootstrapping?)
# So we create a temporary link called "g++"
mkdir tmp_exe
cd tmp_exe
ln -s $GXX g++
cd ..
export PATH=$(pwd)/tmp_exe:$PATH 

${SRC_DIR}/build.sh -python ${PYTHON} -bin ${PREFIX}/bin -expert

echo "bin Contents"
echo "--------------------"
ls "${PREFIX}/bin"
```

First I added `-expert` (BUILD_EXPERT=1) which takes all tools from the conda environment. Still I get the `Cannot run compiler 'g++'` error. Looking at the qmake source code, it should be taken from QMAKE_CXX (hence from the conda environment). Yet it is using the mkspecs default (`g++`). I assume this is a kind of bootstrap step and the configuration is not evaluated yet at this point. So the hack is to create a g++ symbolic link pointing to the actual g++ tool.

Maybe there is a more elegant solution, but the above hack seems to work.

klayoutmatthias:
Update: the above patch makes qmake work, but the build fails because the OpenGL libraries are not found .. :(

curtisma:
Thanks for the help @klayoutmatthias!

> Update: the above patch makes qmake work, but the build fails because the OpenGL libraries are not found .. :(

I found the following section in the conda-forge docs about handling libGL dependencies:
https://conda-forge.org/docs/maintainer/knowledge_base.html#libgl

Maybe I'll just start by adding all of those items to the requirements?  Unless we only need a subset of them, what do you think?
I'll also add the `yum_requirements.txt` in case it is needed for testing.  I noticed they have that in [the pyopengl recipe](https://github.com/conda-forge/pyopengl-feedstock/blob/main/recipe/yum_requirements.txt)

thomaslima:
> > @curtisma curtisma changed the title add klayout add klayout-gui 1 hour ago
> 
> I'd be in favor to keep `klayout` as the default name as I think most people would expect the gui to be present when installing the package from conda-forge (just like the klayout packages from other general purpose distribution).
> 
> Maybe we could have a different package (`klayout-python`?) for the Python API only version, wdyt?

Hard disagree on the default name. I think klayout as default for conda shouldn't come with GUI features. I think the gui should be an optional install. Surely there is a way to resolve this without going to two different names. Doesn't conda support optional build options? like `conda install klayout --with-gui`?

If two different names are required, I strongly suggest the one with more features have a longer name. Like `klayout-gui` superseding `klayout` (python only). Could it be possible to have `klayout-gui` have `klayout` as dependency?

Thanks,
Thomas

proppy:
> I think klayout as default for conda shouldn't come with GUI features.

@thomaslima see https://github.com/hdl/conda-eda/issues/193#issuecomment-1084035801 for a bit more background about my suggestion to have the conda-forge package be full featured: we also plan to have a recipe as part of `conda-eda` that provide a klayout package with minimal dependencies (and just depending on the `main` channel).

> Doesn't conda support optional build options? like conda install klayout --with-gui?

`conda-build` has a feature called "outputs" that allow defining subpackages: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#outputs-section

Maybe we could leverage it here:
- `klayout.cli`: klayout cli without the gui
- `klayout.gui`: klayout cli with gui
- `klayout.python`: klayout pythoon bindings (no cli, and qtless feature as part of 0.28)

Then we could make the `klayout` one be a meta packages for `klayout.python` + one of the other two (if we people most people would be interested by having a functioning klayout binary + some python bindings).

What do you think?

curtisma:
@proppy, The outputs feature is interesting.  Hoever I don't think it's well-known though.  At least I haven't seen it in use in any other recipes.  So I'd prefer to keep it simple and seperate them into different packages.  Hopefully there are no conflicts between the Python and gui packages so both can be installed.  We should probably test that out though.

Another option is we we call them "klayout-python" and "klayout-gui" to keep them descriptive and not have a "klayout" package.

curtisma:
A larger goal of mine is to provide a more integrated open-source design environment.  We can do this by having the [Viper IC Design Environment](http://www.cascode-labs.org/viper/) (IDE) provide the "library / project manager" gui along with an easy project and environment setup CLI.  Then other Conda packages, including Klayout and xschem, will provide individual tools as extensions to work on cells.  It will should remain light-weight but extensible and customisable, similar to VS Code.

Another goal would be to have a full design environment setup and configured in less than 15mins based on a simple set of instructions with less than 10 steps.
This could be done using an IC-design-specific Conda distribution, with a single installer, using Conda [Constructor](https://github.com/conda/constructor), similar to the data science Anaconda distribution and miniforge.  

There could be different flavors of the distribution, one for analog design, digital design, and integrated photonics.  Each would be a different set of packages.  Each could have both an installer and a simple environment.yaml to create a new environment in an existing conda installation.  Some could also support using proprietary tools as-necessary.  This was the idea behind [viper-forge](https://www.cascode-labs.org/viper-forge/).

What do you guys think?  Do you agree with these goals?  Are there others that should be added?
@proppy, @mithro, @joamatab @thomaslima @klayoutmatthias @flaport

proppy:
> Another goal would be to have a full design environment setup and configured in less than 15mins based on a simple set of instructions with less than 10 steps.
This could be done using an IC-design-specific Conda distribution, with a single installer, using Conda [Constructor](https://github.com/conda/constructor), similar to the data science Anaconda distribution and miniforge.

This is really aligned with what we discussed here:
https://github.com/hdl/conda-eda/issues/245#issuecomment-1303053749
And there is a proof of concept here for digital design:
https://github.com/hdl/conda-eda/pull/247

Let's continue the discussion there? (and keep this issue focused on klayout?)

proppy:
> The outputs feature is interesting. Hoever I don't think it's well-known though

From the point of view of the user there is no different with a standalone packages: the `.` in the package name is just a convention, other package like https://github.com/conda-forge/opencv-feedstock/blob/main/recipe/meta.yaml#L205 use totally different names; so this is more of a convenience to be able to  maintain multiple packages from the same feedstock which I believe is what we want to do here.

curtisma:
hmm, seems like we may need to rerun the build.  It's throwing the following error:
`conda.CondaMultiError: Downloaded bytes did not match Content-Length`

proppy:

klayoutmatthias:
Thanks for this wonderful discussion! :)

Just my 2 cents:

`klayout.gui` and `klayout.python` make sense to me as package names. `klayout.cli` makes sense if you like to provide a standalone runtime for DRC for example. However, `klayout.gui` contains a buddy tool called `strmrun` which essentially does the same. 

If `klayout.cli` is about reducing dependencies I need to say I hardly used KLayout without Qt. I think it works (I have one CI build for that), but I'm not entirely sure.

Thanks a lot for all your efforts!

Matthias

thomasdorch:
Hi all, I have made some progress on getting a [conda build for Windows](https://github.com/thomasdorch/klayout-conda) . I had to go around the `built.bat` script since it requires klayout-bits, which isn't necessary to build the GUI on Windows. It's not extensively tested but it works on my machine. Let me know if/where I should submit a PR if it's helpful!

klayoutmatthias:
I have also made some progress with the conda package for Linux (https://github.com/klayoutmatthias/staged-recipes/tree/klayout). It builds on the main channel, but not yet with conda-forge. I had to change a few things compared to this PR.

It produces a conda package file which tests successfully, but trying to install it with "--use-local" makes conda stall in an endless loop to fulfill requirements. No idea if that is a bug in the package or in conda.

I can file a PR if you like.

Matthias

thomasdorch:
@klayoutmatthias I think that may have to do with the way dependencies are versioned in the `meta.yaml`. According to the [conda package match spec](https://conda.io/projects/conda-build/en/latest/resources/package-spec.html), there needs to be a space in-between the package name and any relational operators:

> Remember that the version specification cannot contain spaces, as spaces are used to delimit the package, version, and build string in the whole match specification. python >= 2.7 is an invalid match specification. Furthermore, python>=2.7 is matched as any version of a package named python>=2.7.

i.e. instead of `Qt>=5`, it needs to be `Qt >=5`

thomaslima:
Thanks for the discussion. I have to say I am not too familiar with the conda packaging process. I thought one recipe meant one install process. Sounds like this sub packaging would be the way to go.

Moreover, I was also confused about the three versions `cli`, `gui` and `python`. I don't know what `cli` does. Can anyone give an example?  I still vote for `.python` being the default, since at least to my understanding conda is python package manager. The full featured could be `.gui`.

One quick clarification: would installing `klayout.gui` be a functional replacement for installing from a binary downloaded from klayout.de?

proppy:
>  I still vote for .python being the default, since at least to my understanding conda is python package manager.

conda has strong tie in the python ecosystem initially but is now more of a general cross platform packaging system and distribute a lot of packages with standalone binaries and native libraries.

proppy:
> One quick clarification: would installing klayout.gui be a functional replacement for installing from a binary downloaded from klayout.de?

That would be an alternative for the conda-forge audience (similar to how klayout.de lists packages for different Linux distributions).

PR #20413, opened 2022-09-13
Initial Commit for Nyxus Recipe

sameeul:
Checklist
- [x] Build number is 0.

@conda-forge//help-c-cpp, ready for review!


recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,67 @@
+{% set name = "nyxus" %}
+{% set version = "0.0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/sameeul/nyxus/archive/refs/heads/remove_submodules.zip
```
bollwyvl, line 9: one of the things from the deleted checklist in the issue template is about using official sources... you should try to do so if possible.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  number: 0
+  skip: true  # [py<36]
+
+requirements:
+  build:
+    - cmake
+    - make  # [not win]
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - setuptools
```
bollwyvl, line 22: anything python-related should _probably_ be in `host` rather than build

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - blosc 
+    - pybind11 
+    - xtensor 
+    - xsimd 
+    - numpy
+    - pandas 
+    - m2-patch  # [win]
+
+  host:
+    - python {{ PY_VER }}
```
bollwyvl, line 37: should just be `python`, `conda-build` takes care of matrixing-out version stuff

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+  run:
+    - python {{ PY_VER }}
+    - libtiff 
+    - blosc 
+    - numpy
```
bollwyvl, line 43: if it actually builds against numpy, there are some tricks:

https://conda-forge.org/docs/maintainer/knowledge_base.html#building-against-numpy

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - git
+    - z5py 
+    - libtiff 
+    - boost 
+    - nlohmann_json 
+    - blosc 
+    - pybind11 
+    - xtensor 
+    - xsimd 
+    - numpy
```
bollwyvl, line 32: `numpy` and `pandas` have pins: it's good to respect them:

https://github.com/PolusAI/nyxus/blob/0.2.3/setup.py#L110

recipes/nyxus/bld.bat
```diff
@@ -0,0 +1,2 @@
+set CMAKE_ARGS="-DCMAKE_PREFIX_PATH=$PREFIX -DCMAKE_INSTALL_PREFIX=$PREFIX -DUSEGPU=ON"
```
bollwyvl, line 1: if it's possible to build without GPU support, it can often be worth it... it can be a bit tricky, though, to handle the metapackages properly...

recipes/nyxus/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e # Abort on error.
+
+PY_VER=$(python -c "import sys; print('{}.{}'.format(*sys.version_info[:2]))")
+PY_ABIFLAGS=$(python -c "import sys; print('' if sys.version_info.major == 2 else sys.abiflags)")
+PY_ABI=${PY_VER}${PY_ABIFLAGS}
```
bollwyvl, line 7: some of these are available from https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html#environment-variables-set-during-the-build-process

recipes/nyxus/LICENSE.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
bollwyvl, line 1: this should come from the upstream tarball, if at all possible

recipes/nyxus/build.sh
```diff
@@ -0,0 +1,9 @@
...
+
+CMAKE_ARGS="-DCMAKE_PREFIX_PATH=$PREFIX -DCMAKE_INSTALL_PREFIX=$PREFIX -DUSEGPU=ON -DCMAKE_OSX_DEPLOYMENT_TARGET=10.15" python setup.py install
```
bollwyvl, line 9: if possible, you should emulate the example invocation, e.g


```
python -m pip install .
```

this creates better metadata than calling `setup.py` directly

recipes/nyxus/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e # Abort on error.
```
bollwyvl, line 3: `set -eux` is a bit stricter and will give you more feedback

recipes/nyxus/meta.yaml
sameeul: Yes, we will move to official source as soon as one of my PR gets merged.

recipes/nyxus/meta.yaml
sameeul: Got it!

recipes/nyxus/bld.bat
sameeul: So, we want to build with CUDA support eventually. What is the best way to do that. I am thinking about adding cudatoolkit-dev as a build dependency and cudatoolkit as run dependency.

recipes/nyxus/LICENSE.txt
sameeul: The upstream actually has the license. I will remove it from here.

recipes/nyxus/meta.yaml
carterbox: Are your changes required for this recipe to build on conda-forge? Use a patch instead of publishing from an unofficial repository.

sameeul: My changes will get merged to the official repo today or tomorrow. As soon as that gets merged, I plan to update this. I resorted to this quick hack since I need to change both the build process in the actual repo and test it out with the recipe.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - {{ pin_compatible('numpy') }}
+    - pandas
+
+test:
+  requires:
+    - pytest
+  imports:
+    - nyxus
+  commands:
+    - python -m pytest run_test.py
```
bollwyvl, line 56: probably add `pip` to `requires` and `pip check` to `commands`.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - numpy
+    - pandas 
+    - m2-patch  # [win]
+
+  run:
+    - python
+    - libtiff 
+    - blosc 
+    - {{ pin_compatible('numpy') }}
+    - pandas
```
bollwyvl, line 48: could also do with a minimum `pandas` bottom pin matching upstream

recipes/nyxus/meta.yaml
sameeul: so, like the following?
```
  requires:
    - pytest
    - pip
  imports:
    - nyxus
  commands:
    - pip check
    - python -m pytest run_test.py
```

recipes/nyxus/bld.bat
carterbox: Do not use cudatoolkit-dev in a recipe. Just add `compiler('cuda')` to the build requirements section. Add the cudatoolkit to your host requirements if you need the CUDA libraries.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+    - xtensor 
+    - xsimd 
+    - numpy
+    - pandas 
+    - m2-patch  # [win]
+
+  run:
+    - python
+    - libtiff 
+    - blosc
```
carterbox, line 49:
```suggestion
```

recipes/nyxus/run_test.py
```diff
@@ -0,0 +1,5 @@
+import nyxus
```
carterbox, line 1: https://docs.conda.io/projects/conda-build/en/latest/concepts/recipe.html

This file is run twice because it is named run_test.[py] and also you call it in the test/commands section. Is it even needed? The import of nyxus is also tested in test/imports. Remove this file if it is redundant.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+  host:
+    - python
+    - setuptools
+    - pip
+    - z5py 
+    - libtiff 
+    - boost 
+    - nlohmann_json 
+    - blosc
```
carterbox, line 38:
```suggestion
    - blosc  # [cuda_compiler_version == "None"]
```
It looks like blosc is unused in CUDA builds because the link inspector says there are no links to it.

recipes/nyxus/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+  sha256: e82f349f6ff481e017c3ff2060642e66e81dbda5910272fba1b3c5a4b5d9b2fd
+  patches:
+    - 0001-fix-filesystem-issue.patch
+    - 0002-Fix-CUDA-Kernel-Issue.patch
+
+build:
+  number: 0
+  skip: true  # [py<36]
+  missing_dso_whitelist:
+    - "*/nvcuda.dll"   # [win]
```
carterbox, line 19:
```suggestion
build:
  number: 0
  skip: true  # [py<36]
  missing_dso_whitelist:
    - "*/nvcuda.dll"   # [win]
  ignore_run_exports_from:
  # The CXX standard library is unused on linux, but used on osx
    - {{ compiler('cxx') }}  # [linux]
```
According to the link inspector the cxx standard library is unused on linux.

carterbox (changes requested):

recipes/nyxus/meta.yaml
sameeul: This I believe only happens for CUDA 10.2. The build machine uses GCC 7.5 which fails the minimum GCC requirement for `xtensor`. There is a open issue in the `xtensor` Github about this. In this particular situation, we build Nyxus w/o zarr file support. CUDA 11.x uses blosc. So, should I use `# [cuda_compiler_version == "None" or cuda_compiler_version == "10.2.0"] ` ?

recipes/nyxus/run_test.py
sameeul: Thanks for the explanation. Removed the file.

recipes/nyxus/meta.yaml
sameeul: @carterbox : I updated meta.yaml and CMakeLists.txt so that it does not used bring in blosc in CUDA 10.2 build (where it is not used). I think the link inspector now does not find it anymore.

sameeul:
@conda-forge-admin, please ping conda-forge/help-c-cpp @conda-forge//help-python
@conda-forge//help-c-cpp, ready for review!
@conda-forge//help-python, ready for review!

recipes/nyxus/meta.yaml
carterbox: > So, should I use # [cuda_compiler_version != "10.2.0"] ?

Yes, something like that is fine. Could have also dropped builds before 11.

PR #20478, opened 2022-09-19
added parmoo to staged recipes

thchang:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/parmoo/meta.yaml
+++ b/recipes/parmoo/meta.yaml
@@ -27,7 +27,7 @@ requirements:
     - libensemble
     - plotly
     - dash
-    - kaleido
+    - python-kaleido
 
 test:
   imports:
--- a/recipes/parmoo/meta.yaml
+++ b/recipes/parmoo/meta.yaml
@@ -17,21 +17,24 @@ build:
 requirements:
   host:
     - python >=3.6
-    - tox
     - pip
   run:
     - python >=3.6
     - numpy
     - scipy
     - pydoe
+    - pandas
+    - libensemble
+    - plotly
+    - dash
+    - kaleido
 
 test:
   imports:
     - parmoo
   commands:
-    - py.test --pyargs parmoo
+    - py.test --pyargs parmoo.tests.unit_tests
   requires:
-    - libensemble
     - pytest
 
 about:
--- a/recipes/parmoo/meta.yaml
+++ b/recipes/parmoo/meta.yaml
@@ -31,6 +31,7 @@ test:
   commands:
     - py.test --pyargs parmoo
   requires:
+    - libensemble
     - pytest
 
 about:
--- a/recipes/parmoo/meta.yaml
+++ b/recipes/parmoo/meta.yaml
@@ -41,4 +41,4 @@ about:
 
 extra:
   recipe-maintainers:
-    - tyler
+    - thchang
--- a/recipes/parmoo/meta.yaml
+++ b/recipes/parmoo/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "parmoo" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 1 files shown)


thchang:
@conda-forge/help-python I have attempted to add our Python package ``parmoo`` to the staged recipes, by generating the recipe for pypi using grayskull, as instructed on: https://conda-forge.org/docs/maintainer/adding_pkgs.html

I have the following concerns about what I did:
 - How to handle optional requirements?  Currently, parmoo has one optional dependency https://github.com/libensemble/libensemble (also on conda-forge and pypi).  I don't see how to add optional dependencies here (e.g., on PyPI I had an ``extras_requires`` field in the ``setup.py`` file), but grayskull didn't generate anything here.  Similarly, to build the docs one needs ``sphinx``, ``sphinx_rtd_theme``, and ``sphinxcontribut.bibtex``, and to run tests one needs ``pytest``, ``pytest-cov``, and ``flake8``, but I don't see a way to add those either.
 - How to update requirements in the future?  We have another release coming out soon.  When that happens should I open a PR to update the ``version`` field in my yaml file?  Also, the update will add several new dependencies, anything extra I need to do for that?
 - Running tests:  I tried to copy advice from here:  https://github.com/conda-forge/conda-forge.github.io/issues/345 on how to have conda automatically run unit tests using pytest, but I don't know how to check whether I did it correctly.

Many thanks for advice (I am a conda noob),
---Tyler

recipes/parmoo/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - scipy
+    - pydoe
+
+test:
+  imports:
+    - parmoo
+  commands:
+    - py.test --pyargs parmoo
+  requires:
+    - libensemble
```
dopplershift, line 34: Looks like a test is failing due to missing `mpi4py`:
```suggestion
    - libensemble
    - mpi4py
```

recipes/parmoo/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.6
+    - tox
```
dopplershift, line 20: Is tox actually necessary to build the package?
```suggestion
```

dopplershift (changes requested):
Just some minor changes, one of which should help get everything green.

If you want to see the results of the builds yourself, you can click the "Details" link below, and follow the links to the build logs on Azure.

Once this recipe is merged, it will create a new "feedstock", where you'll have the ability to make and merge Pull Requests to maintain the recipe. There, in addition to bumping the version/hash for new releases, you can adjust the dependencies, build steps, etc. See [the docs](https://conda-forge.org/docs/maintainer/updating_pkgs.html) for more info.

Conda doesn't have a concept of optional dependencies, so you can either rely on users to install them manually as needed, or just add them as required dependencies. It depends what makes the most sense for this particular package.

recipes/parmoo/meta.yaml
thchang: no, tox is not required.  I missed that.  I think it was mistakenly grabbed by grayskull

thchang: removed it in the latest push

recipes/parmoo/meta.yaml
thchang: So this is an optional dependency of libensemble that I am not using for my tests.  It thinks I am trying to use it because libensemble runs cannot be started through pytest, which causes an error.  Instead, I've updated my pytest recipe to only run unit tests (not regression tests), this should solve the problem.  To run the full test script, we'd need to run a special test script that I wrote, contained in the parmoo/tests directory

thchang:
everything appears to be passing now (running unit tests only).  FYI, I have preemptively added a couple of optional dependencies to the requirements list, which are not needed yet but will be needed in the next release (in a couple weeks), so that we won't need to immediately revise the recipe

thchang:
Thanks @dopplershift for your helpful advice and explanation of the process

dopplershift (approved):

PR #20527, opened 2022-09-22
Add recipe and dependencies for the ufmt code formatter / import sorter

dstoeckel:
ufmt is a package combining black and usort to be used in CI pipelines. This PR adds the necessary packages from PyPI. Recipes were generated using grayskull.

Checklist
- [x] Build number is 0.


nicornk:
Okay to be maintainer

dstoeckel:
Okay to be maintainer

nicornk:
@conda-forge/core this is ready to be reviewed and merged. thank you!

nicornk:
pinging the correct @@conda-forge/staged-recipes

recipes/trailrunner/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+requirements:
+  host:
+    - python
+    - pip
+    - flit >=3,<4
+    - pathspec >=0.8.1
+  run:
+    - python
+    - pathspec >=0.8.1
```
ocefpaf, line 24: This package has a dependency on `futures3` for py<37. You can either add that, or maybe try to make it `noarch` with a skip for `py<37`

recipes/ufmt/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+requirements:
+  host:
+    - python
+    - pip
+    - flit >=3,<4
+    - black >=20.8b0
+    - moreorless >=0.4.0
+    - trailrunner >=1.2.1
+    - usort >=1.0
+    - tomlkit >=0.7.2
```
ocefpaf, line 25: These are probably not needed. If they are the package imports itself on build time and that is, most of the time, a bug upstream.

```suggestion
```

recipes/ufmt/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ufmt-{{ version }}.tar.gz
+  sha256: 7e4518e13507984998f5c7c9a0d3f59fe014ddc213bdb5cae8d11a6a8935db8c
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
```
ocefpaf, line 14:
```suggestion
  number: 0
  noarch: python
```

dstoeckel: I needed to make this a non-noarch package as the PyPI wheel for Windows contains ufmt.exe which breaks the CI for windows. Happy to add back and skip Windows.

recipes/trailrunner/meta.yaml
dstoeckel: I will go for the noarch + skip option. Thanks for the suggestion.

recipes/ufmt/meta.yaml
dstoeckel: I assume this is caused by flit (I iteratively added packages until build was succeeding ...). Agreed that this seems like an upstream bug to me. Does this need to get fixed before we can submit a package?

recipes/ufmt/meta.yaml
ocefpaf: It is probably the lack of the entrypoint.

recipes/trailrunner/meta.yaml
dstoeckel: Fixed

recipes/ufmt/meta.yaml
dstoeckel: Indeed flit attempts to load the modules in order to extract the version from the __init__.py field. There is an AST parsing based extractor, however the __version__ field is imported from a separate file. I am working around this by inlining the version string into the toplevel directory.

recipes/ufmt/meta.yaml
dstoeckel: Specifying an entrypoint fixed the issues. Moved to noarch.

dstoeckel:
Resolved all comments. I had to use patches to avoid having to specify dependencies twice: in run and host. This is ready for another review round @conda-forge/staged-recipies

nicornk:
@conda-forge/staged-recipes thanks!

recipes/trailrunner/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/trailrunner-{{ version }}.tar.gz
+  sha256: 157de2b2d82139d9583ec1dbc1981c33b39e24c420bb5294f86083f1f4b44311
+  patches:
+    # Building of the package fails as FLIT tries to determine the package version
+    # by importing the module. This fails as dependencies are not yet installed.
+    # Include the version directly in the top-level module to avoid this check.
+    - 0001-fix_flit_inline_package_version.patch
```
ocefpaf, line 15: Oh! You don't need to patch it like this b/c it will be a nightmare to maintain. In pathological cases where the package imports itself to get the version we are OK having the run dependencies in host/build. Better to just open an issue upstream and wait for them to fix it.

dstoeckel: I reverted to adding the required runtime dependencies to build time and created an upstream bug: https://github.com/omnilib/ufmt/issues/123

PR #20535, opened 2022-09-23
add snowline Laplace approximation and Importance Sampling python package for Bayesian inference

JohannesBuchner:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/snowline/meta.yaml
+++ b/recipes/snowline/meta.yaml
@@ -12,18 +12,13 @@ source:
 build:
   number: 0
   # may need to later add "skip: True  # [win]" to disable Windows.
-  skip: true  # [py<30 or python_impl == "pypy"]
+  skip: true  # [py<30 or python_impl == "pypy" or win]
   script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
-  build:
-    # cython project, so need a compiler
-    - {{ compiler('c') }}
-    - cython
   host:
     - python
     - pip
-    - cython
     - pypmc
     - iminuit
     - numpy
@@ -43,20 +38,19 @@ test:
   source_files:
     - tests/*
   requires:
-    - cython
     - pytest
-    - pytest-html
-    - pip
 
 about:
   home: https://johannesbuchner.github.io/snowline/
   license: GPL-3.0-or-later
   license_family: GPL
   license_file: LICENSE
-  summary: Fit and compare complex models reliably and rapidly. Advanced Nested Sampling.
+  summary: Fit and compare models very quickly. MCMC-free.
 
   description: |
     Fit and compare models very quickly. MCMC-free.
+    
+    Posterior distributions and corner plots without MCMC? No dealing with convergence criteria?
     
     Yes!
     
--- a/recipes/snowline/meta.yaml
+++ b/recipes/snowline/meta.yaml
@@ -38,7 +38,7 @@ test:
     - pip
   commands:
     - pip check
-    - pytest tests/
+    - pytest -sxv tests/
 
 about:
   home: https://johannesbuchner.github.io/snowline/
--- a/recipes/snowline/meta.yaml
+++ b/recipes/snowline/meta.yaml
@@ -29,7 +29,6 @@ requirements:
     # the following two are for (optional) plot functionality
     - matplotlib-base
     - corner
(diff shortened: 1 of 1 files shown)


JohannesBuchner:
I am seeing this error:

```
Reloading output folder: /home/conda/staged-recipes/build_artifacts
Mamba failed to solve:
 - pypmc
 - python 3.9.* *_cpython
 - pytest-runner
 - numpy 1.20.*
 - pip
 - iminuit

with channels:
 - local
 - conda-forge
 - conda-forge

The reported errors are:
-   - package pypmc-1.2-py37hc105733_0 requires python_abi 3.7.* *_cp37m, but none of the providers can be installed
```

pypmc is a cpython package that snowline depends on.

I don't understand why "python 3.9.* *_cpython" tries to pull in the py37h version of pypmc.

iminuit is another cpython package that snowline depends on. Its recipe is here: https://github.com/conda-forge/iminuit-feedstock/blob/main/recipe/meta.yaml

This [stackoverflow question](https://stackoverflow.com/questions/73433762/why-does-conda-package-install-complain-about-both-python-abi-3-7-and-3-9) has the same error but from a user perspective.

snowline is pure python, but requires the two cpython packages.

@conda-forge/help-python-c, can you please help?

recipes/snowline/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - pypmc
+    - iminuit
+
+test:
+  imports:
+    - snowline
+  source_files:
+    - tests/*
+  requires:
+    - pytest
```
carterbox, line 41:
```suggestion
test:
  imports:
    - snowline
  source_files:
    - tests/*
  requires:
    - pytest >=3
    - pip
  commands:
    - pip check
    - pytest tests/
```

recipes/snowline/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - numpy
+    - pytest-runner
+
+  run:
+    - python
+    - numpy
+    - matplotlib-base
+    - corner
+    - pypmc
+    - iminuit
```
carterbox, line 33: These do not match the upstream metadata

https://github.com/JohannesBuchner/snowline/blob/6f9b5f767903ebf0537569ff85752a8533664b8a/setup.py#L23

recipes/snowline/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python
+    - pip
+    - pypmc
+    - iminuit
+    - numpy
+    - pytest-runner
```
carterbox, line 25:
```suggestion
  host:
    - python
    - pytest-runner
    - setuptools
```

carterbox (changes requested):
If this package is pure python, then it can be `noarch: python`.

recipes/snowline/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python
+    - pytest-runner
+    - setuptools
+
+  run:
+    - python
```
carterbox, line 25:
```suggestion
build:
  number: 0
  noarch: python
  script: {{ PYTHON }} -m pip install . -vv

requirements:
  host:
    - python >=3.5
    - pytest-runner
    - setuptools
  run:
    - python >=3.5
```

carterbox (changes requested):
Figure out the tests and fix the incorrect runtime requirements.

carterbox:
> Tests are fixed by installing openssh to make openmpi work.

Looks like openssh is only required because these packages are being built inside a container? It's not actually a direct dependency of your package. Only direct dependencies of your package go in the requirements section. Packages that are only required for the tests go in the test/requires section, not the requirements/run section.

JohannesBuchner:
> > Tests are fixed by installing openssh to make openmpi work.
> 
> Looks like openssh is only required because these packages are being built inside a container? It's not actually a direct dependency of your package. Only direct dependencies of your package go in the requirements section. Packages that are only required for the tests go in the test/requires section, not the requirements/run section.

Thank you for pointing this out! I made the change.

JohannesBuchner:
I think this is ready to be merged

PR #20545, opened 2022-09-23
Added recipe for 'genv' and 'jupyterlab_genv'

razrotenberg:
Checklist
- [x] Build number is 0.


razrotenberg:
@conda-forge-admin please ping conda-forge/staged-recipes

razrotenberg:
@conda-forge-admin conda-forge/staged-recipes

razrotenberg:
@conda-forge-admin please ping conda-forge/staged-recipes

razrotenberg:
@conda-forge-admin please ping conda-forge/staged-recipes

recipes/genv/meta.yaml
```diff
@@ -0,0 +1,35 @@
...
+  description: |
+    genv lets you easily control, configure and monitor the GPU resources that you are using.
+    It is intendend to improve the life of individual GPU users as well as teams that share multi-GPU machines.
+    This project was highly inspired by pyenv and other version, package and environment management software like Conda, nvm, rbenv.
+  license: AGPL-3.0-or-later
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - razrotenberg
```
BastianZim, line 35: This whole recipe is quite non-standard. What is the reason for that?

razrotenberg: hi @BastianZim,
first of, thanks for reviewing this PR! 🙏 

I took inspiration from other recipes out there, not really sure which one exactly so I'm having trouble with pointing out some. I might misunderstood some things and implemented this recipe wrongly, but I'll try to explain why I did it this way.

Genv is not a package managed by some package manager like pip, npm, etc.
It is a standalone software and it is installed simply by cloning a Git repository from GitHub.
As Genv is a shell extension (just like Conda, PyEnv, etc.), it should be loaded into the current shell to work, and its `bin/` directory should be added to `PATH` so that `genv` commands would work.

More information about how to install it can be found in the project docs site [here](https://docs.genv.dev/usage/installation.html#terminal).

Because of all this, I found installing it under `/opt` as the most native way of installing such software in Linux.

This is why it is installed under `/opt/genv` in the Conda root directory, and the `genv` executable is symlink-ed from the `bin/` directory of Conda.

Because Genv should be loaded into the current shell, I needed `activate.sh` and `deactivate.sh` so that it would load and unload from the current shell when a Conda environment is activated or deactivated.

Please lmk if there's any more information that I can share here 👌

BastianZim: My pleasure!

Ahh ok that makes sense, thanks for the explanation. That is quite an unusual recipe so I'm going to ping some experts here because of the way it is built but that should already be enough info to get us started.

recipes/jupyterlab_genv/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7
+    - pip
+    - hatchling >=1.3.1
+    - jupyterlab ~=3.1
+    - hatch-jupyter-builder >=0.5
```
BastianZim, line 23: This is not listed upstream. Is it required?

recipes/jupyterlab_genv/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  run:
```
BastianZim, line 24: Since this depends on genv, it should probably be listed here?

BastianZim (changes requested):
Couple of small things

recipes/genv/meta.yaml
BastianZim: @ocefpaf Do you have any experience with these kinds of packages?

ocefpaf: Not really. But I'm OK merging this one.

recipes/jupyterlab_genv/meta.yaml
razrotenberg: yes it didn't work without it because of this https://github.com/run-ai/jupyterlab_genv/blob/main/pyproject.toml#L2

recipes/jupyterlab_genv/meta.yaml
razrotenberg: it's a very good point!
I think you are right. 
added it!

recipes/genv/LICENSE
```diff
@@ -0,0 +1,661 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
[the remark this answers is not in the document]
razrotenberg: oooh no I got the linter error:

> 
> I wanted to let you know that I linted all conda-recipes in your PR (recipes/genv, recipes/jupyterlab_genv) and found some lint.
> 
> Here's what I've got...
> 
> For recipes/genv:
> 
> license_file entry is missing, but is required.
> 

do you think I should restore the license file?

recipes/jupyterlab_genv/meta.yaml
razrotenberg: ok after adding `genv` as a dependency, the build failed a few times:

at first it didn't work on Windows as `genv` does not support Windows

when I skipped build on Windows, it said that:
> noarch packages can't have skips with selectors. If the selectors are necessary, please remove noarch: python.

when I removed the `noarch:python`, it said that:
> Non noarch packages should have python requirement without any version constraints

what do you think is the best solution here, should I specify the minimum version of Python?

BastianZim: Just ignore the windows failure, that's fine – the package will just not be available there, since genv isn't either. The package is only built on linux but can be installed on all OS so it doesn't matter.

BastianZim: > I'm also asking because what if genv is installed out of Conda?
specifying it here would require it to be installed in Conda even though it might be installed out of it?

Depends, if it's properly installed, conda will find it and just not install it.

recipes/jupyterlab_genv/meta.yaml
BastianZim: Yes, but `hatch-jupyter-builder` is not listed there. It's fine now, but ideally upstream and here are the same so it should be listed upstream as well.

recipes/jupyterlab_genv/meta.yaml
razrotenberg: ok great so I restored the `noarch: python`, set `genv` as a dependency, and removed the `skip: true # [win]`
I'm pretty sure that the check would fail because of build on Windows fails, so I hope it is not a stopper for merging 👌

razrotenberg: hi @BastianZim 
pls let me know if it's fine now or any further changes should be done

thanks!

PR #20652, opened 2022-10-05
Adding CyRSoXS

pbeaucage:
Adding a recipe for [CyRSoXS](https://GitHub.com/usnistgov/cyrsoxs), a simulation package with options for standalone binary use or Python library use via pybind11, built in C++/CUDA.

Build is skipped on non-CUDA platforms as the package requires CUDA.

Checklist
- [x] Build number is 0.


pbeaucage:
@pdudenas could you confirm here that you agree to co-maintain the cyrsoxs conda package?

pdudenas:
@pbeaucage yes, I agree to co-maintain the cyrsoxs conda package.

pbeaucage:
Ready for review!  @conda-forge-admin, please ping conda-forge/help-c-cpp

pbeaucage:
I've fixed the meta.yaml so that build now reports success (skipped build) on linux64, this was previously failing because this package will only build on CUDA-capable platforms.  This is now ready for review with all CI builds reporting success.

recipes/cyrsoxs/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  skip: True  # [cuda_compiler_version == "None"]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cmake
+    - make
+    - cudatoolkit
```
hmaarrfk, line 25:
```suggestion
```
I don't think you need to specify this. it will find the right one.

Instead we often specifcy other libraries
https://github.com/conda-forge/tensorflow-feedstock/blob/main/recipe/meta.yaml#L56

```
  host:
    # GPU requirements for CUDA
    - cudnn      # [cuda_compiler_version != "None"]
    - nccl       # [cuda_compiler_version != "None"]
```

recipes/cyrsoxs/build.sh
```diff
@@ -0,0 +1,12 @@
+
+mkdir build
+cd build
+cmake ${SRC_DIR} ${CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release 
+make install
+cd ../
+
+mkdir build-pybind
+cd build-pybind
+cmake ${SRC_DIR} ${CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release -DPYBIND=Yes -DUSE_SUBMODULE_PYBIND=No -DPython_EXECUTABLE="$PYTHON"  -DCMAKE_INSTALL_PREFIX=${PREFIX} \
```
hmaarrfk, line 10: its often really useful to just have one option per line once they get this big. this doesn't even fit on my screen.

recipes/cyrsoxs/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cmake
+    - make
+    - openmp
+    - python >3.7
+    - pip
```
hmaarrfk, line 27:
```suggestion
```

why do you need python in build. can you try to remove python and pip from here

recipes/cyrsoxs/build.sh
pbeaucage: Done, thanks.

recipes/cyrsoxs/meta.yaml
pbeaucage: I believe it needs Python so that pybind11 can link against it to produce the python module output... pip is definitely not needed, and I'm giving it a try in the local docker container without python also just to check.

pbeaucage: I believe pip in host is also just leftover from the template, now that I look at it.  Removing that as well.

pbeaucage: Works without python specified in build, updated.  Sorry!  I think pybind11 pulls it in as a dependency as it does wind up in the solved env.

recipes/cyrsoxs/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  skip: True  # [cuda_compiler_version == "None"]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cmake
+    - make
+    - openmp
```
hmaarrfk, line 25:
```suggestion
```
i also don't think you need in build. build should have, under normal circumstances, only build tools, not compilation libraries.

hmaarrfk: Maybe see https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp

recipes/cyrsoxs/meta.yaml
pbeaucage: Removed, build still works.  The CUDA dependencies for this are entirely within the toolkit.

recipes/cyrsoxs/meta.yaml
pbeaucage: The KB article suggests putting `libgomp` in `build`, just pushed this change.

recipes/cyrsoxs/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+{% set version = "1.1.5.0" %}
+{% set sha = "7ce6e64d5e794751476e6e1b6dbb1cd940b2dd26796b8d416fd82b6c20302ff9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/usnistgov/cyrsoxs/archive/refs/tags/V{{ version }}.tar.gz
+  sha256: {{ sha }}
```
hmaarrfk, line 11:
```suggestion
{% set version = "1.1.5.0" %}

package:
  name: cyrsoxs
  version: {{ version }}

source:
  url: https://github.com/usnistgov/cyrsoxs/archive/refs/tags/V{{ version }}.tar.gz
  sha256: 7ce6e64d5e794751476e6e1b6dbb1cd940b2dd26796b8d416fd82b6c20302ff9
```

Final cleanup. Many of us think that the `name` and `sha` variables is overdoing it. They aren't much more usefull than simply using the name of your package everywhere.

Anyway, this is something you can fix yourself in your feedstock. I'll come back around in an hour or so an merge whatever the status.

hmaarrfk (approved):

recipes/cyrsoxs/meta.yaml
pbeaucage: Sounds good, I agree, it's a little silly...the name isn't exactly going to change and sha is only in one place.  I'll commit this.  Really appreciate the review!

PR #20673, opened 2022-10-07
Add PySide6

jschueller:
cc @hoechenberger


hoechenberger:
cc @larsoner @cbrnr

ccordoba12:
This is moving in the direction of creating different packages/feedstocks for versions 2 and 6 of PySide, which (I think) hasn't been decided yet, right?

jschueller:
yes, I'm okay to move this to the pyside2 feedstock too

ccordoba12:
I think we have to decide first the migration path for Qt6, so please hold on before doing that.

jschueller:
ok

jschueller:
@ccordoba12 I think the Pyside6 package can be made available independently of a broader migration strategy for qt6
the question of whether to use a new feedstock or not remains but I have no objections to reuse @conda-forge/pyside2

jschueller:
goto https://github.com/conda-forge/pyside2-feedstock/pull/136

ccordoba12:
Ok, no problem with that.

PR #20674, opened 2022-10-07
Add qt6-webengine

jschueller:
cc @conda-forge/qt-webengine


ccordoba12:
Quick question: is it necessary to add a new feedstock for this? Why not add it as a new branch of `qt-webengine`, like it was done for `qt6-main`?

jschueller:
@ccordoba12 this makes sense, could you create a branch for me to target ?

PR #20682, opened 2022-10-07
added grape to the recipes

zommiommy:
Hi there, I'm one of the authors of [grape](https://github.com/AnacletoLAB/grape). 
I'm trying to make it available on conda, this is the basic recipe.

Our library depends on [ensmallen](https://github.com/AnacletoLAB/ensmallen) which is highly non-trivial to compile so we just upload the pre-compiled versions on PyPI for Linux, Mac, and Windows. Therefore we don't have a "source" distribution.

I'm not sure what's the best practice, should we try to package also the sub-libraries or could these just be retrieved through PyPi? 

Thank you!


recipes/grape/conda_build_config.yaml
```diff
@@ -0,0 +1,5 @@
+python:
```
xhochy, line 1: This file can be removed.

xhochy (commented):
Yes, you should be building the sub libraries from source here instead of downloading binaries from PyPI

PR #20716, opened 2022-10-10
PR: Add recipe for spyder-vim

dalthviz:
Checklist
- [x] Build number is 0.

Hi, adding the recipe for a Spyder IDE plugin: [spyder-vim](https://github.com/spyder-ide/spyder-vim)

Also, @CAM-Gerlach @ccordoba12 and @ok97465 do you agree to be maintainers of this package?

### DIFF
--- a/recipes/spyder-vim/meta.yaml
+++ b/recipes/spyder-vim/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "spyder-vim" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/spyder-vim-{{ version }}.tar.gz
+  sha256: 97c2ef31ae9bb24af1853d3741081cc27867f4c170e031a07daa213cc79577ca
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7
+    - pip
+  run:
+    - python >=3.7
+    - qtpy
+    - qtawesome
+    - spyder >=5.3.3
+
+test:
+  imports:
+    - spyder_vim
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/spyder-ide/spyder-vim
+  summary: A plugin to enable vim keybindings to the spyder editor
+  license: MIT
+  license_file: LICENSE.txt
+
+extra:
+  recipe-maintainers:
+    - CAM-Gerlach 
+    - ccordoba12
+    - dalthviz
+    - ok97465
--- a/recipes/spyder-vim/meta.yaml
+++ b/recipes/spyder-vim/meta.yaml
@@ -6,12 +6,12 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/spyder-vim-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0]|lower }}/{{ name|lower }}/{{ name|lower }}-{{ version }}.tar.gz
   sha256: 97c2ef31ae9bb24af1853d3741081cc27867f4c170e031a07daa213cc79577ca
 
 build:
   noarch: python
-  script: {{ PYTHON }} -m pip install . -vv
+  script: {{ PYTHON }} -m pip install . -vv --no-deps
   number: 0
 
 requirements:
@@ -34,9 +34,14 @@ test:
 
 about:
   home: https://github.com/spyder-ide/spyder-vim
-  summary: A plugin to enable vim keybindings to the spyder editor
+  summary: A plugin to enable Vim keybindings in the Spyder IDE's editor
   license: MIT
   license_file: LICENSE.txt
+  description: |
+    Spyder Plugin for executing Vim commands inside the editor.
+    Although it does not currently support many Vim commands, more will be added in the future.
+    After installing Spyder-Vim, you need to close and re-open Spyder if its running.
+    You should see the plugin at the bottom of the Editor.
 
 extra:
   recipe-maintainers:


ccordoba12:
+1 from me.

dopplershift (approved):
Recipe looks great. Just need confirmation from @CAM-Gerlach and @ok97465 that they are willing to be listed as maintainers for this recipe.

ok97465:
Ok. I will do it.

CAM-Gerlach:
Hey, sorry I've been massively backlogged due to the Python core dev summit/sprint I just got back from, and am still catching up on everything.

I'm happy to help if needed, as I'm familiar with the role on conda-forge, although to be clear I've never really used `spyder-vim` or been a significant contributor, I've just helped a bit with triaging and wrangling, as I sometimes do on other Spyder projects.

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/spyder-vim-{{ version }}.tar.gz
+  sha256: 97c2ef31ae9bb24af1853d3741081cc27867f4c170e031a07daa213cc79577ca
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
```
CAM-Gerlach, line 14:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps
```

To ensure nothing gets installed by `pip` aside from what `conda` installs

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7
+    - pip
+  run:
+    - python >=3.7
+    - qtpy
+    - qtawesome
```
CAM-Gerlach, line 24: Some reasonable lower bound here might be helpful...

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  imports:
+    - spyder_vim
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/spyder-ide/spyder-vim
+  summary: A plugin to enable vim keybindings to the spyder editor
```
CAM-Gerlach, line 37:
```suggestion
  summary: A plugin to enable Vim keybindings in the Spyder IDE's editor
```

Fix grammar, terminology and capitalization

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  license: MIT
+  license_file: LICENSE.txt
```
CAM-Gerlach, line 39:
```suggestion
  license_file: LICENSE.txt
  description: |
    Spyder Plugin for executing Vim commands inside the editor.
    Although it does not currently support many Vim commands, more will be added in the future.
    After installing Spyder-Vim, you need to close and re-open Spyder if its running.
    You should see the plugin at the bottom of the Editor.
```

Add a bit of a description field, pulled from the readme

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "spyder-vim" %}
```
CAM-Gerlach, line 1:
```suggestion
{% set name = "Spyder-Vim" %}
```

Proper-case

recipes/spyder-vim/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/spyder-vim-{{ version }}.tar.gz
```
CAM-Gerlach, line 9:
```suggestion
  url: https://pypi.io/packages/source/{{ name[0]|lower }}/{{ name|lower }}/{{ name|lower }}-{{ version }}.tar.gz
```

Missed a spot. Also, I don't think this is case sensitive, but just in case...

CAM-Gerlach (commented):
A few suggestions on the recipe...

recipes/spyder-vim/meta.yaml
dopplershift: Just FYI the [official conda-forge docs](https://conda-forge.org/docs/maintainer/adding_pkgs.html#use-pip) don't even suggest that :wink:

recipes/spyder-vim/meta.yaml
dopplershift: We generally don't capitalize the names for the conda-forge package.

recipes/spyder-vim/meta.yaml
CAM-Gerlach: Hmm. That's what I see in various other recipes I maintain, but the provenance is unclear; perhaps CF used to recommend/require it, but no longer does? I assume you would know better than I...

recipes/spyder-vim/meta.yaml
CAM-Gerlach: Right, and indeed it is under `package.name` below, and in most other uses. The intent here was to use the canonical display case in the Jinja `name` variable here, so it could be lowercased when used programmatically (with the `lower` filter) and left intact when used as a display name. However, I don't see it actually used as a display name anywhere currently, and the PyPI name is also lowercase, so indeed this isn't really necessary.

recipes/spyder-vim/meta.yaml
dalthviz: Lets keep it synched with upstream

recipes/spyder-vim/meta.yaml
dalthviz: I think this needs to also be changed on the plugin repo

recipes/spyder-vim/meta.yaml
dalthviz: I will skip this suggestion and leave the name as it is following the PyPI package name

recipes/spyder-vim/meta.yaml
dopplershift: I think it's changed over time. IIRC there are other protections against the problem you mentioned.

dopplershift (approved):

recipes/spyder-vim/meta.yaml
CAM-Gerlach: Yeah, only problem is if we do change it later, conda could backsolve to these old uncapped versions (which may or may not actually work).

PR #20882, opened 2022-10-25
Add graalpy package to make Graalpy available as a Python implementation

timfel:
Checklist
- [x] Build number is 0.


msimacek:
Hi, I confirm I want to be listed as a maintainer of graalpy packages.

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    sha256: 9b714747fbff0b9f477a102a22e6558f697322d8fa89b1ac9956938d3afb3b87
+    folder: graal
+  - url: https://github.com/oracle/graaljs/archive/refs/tags/vm-ce-{{ version }}.zip
+    sha256: 30c39db551ee1f1816fc32dd81506c3e297401ac780402952705e79ef770a567
+    folder: graaljs
+  - url: https://github.com/graalvm/labs-openjdk-17/releases/download/jvmci-23.0-b01/labsjdk-ce-17.0.5+6-jvmci-23.0-b01-linux-amd64.tar.gz  # [linux64]
+    sha1: e69e718eb8bbc8529f597a4bdd4c3dc53c514d20  # [linux64]
+    url: https://github.com/graalvm/labs-openjdk-17/releases/download/jvmci-23.0-b01/labsjdk-ce-17.0.5+6-jvmci-23.0-b01-linux-aarch64.tar.gz  # [aarch64]
+    sha1: f4b6019dc7571b1f118dc75d408c1e7bec98f07f  # [aarch64]
+    url: https://github.com/graalvm/labs-openjdk-17/releases/download/jvmci-23.0-b01/labsjdk-ce-17.0.5+6-jvmci-23.0-b01-darwin-amd64.tar.gz  # [osx]
```
isuruf, line 32:
```suggestion
    url: https://github.com/graalvm/labs-openjdk-17/releases/download/jvmci-23.0-b01/labsjdk-ce-17.0.5+6-jvmci-23.0-b01-darwin-amd64.tar.gz  # [osx and x86_64]
```

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,78 @@
+{% if name_suffix is not defined %}
+{% set name_suffix = "3.8" %}
+{% endif %}
+{% set name = "graalpy" ~ name_suffix %}
+{% set version = "22.3.0" %}
+{% set mx_version = "6.9.7" %}
+
+package:
+  name: {{ name }}
```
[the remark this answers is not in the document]
timfel: I wasn't sure what the meaning of it was for pypy. I'll change it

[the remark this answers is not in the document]
timfel: I tried to change the name to just graalpy, but playing around with it all week I couldn't get it to build then. One python_abi variant needs to come from graalpy, and graalpy should also provide a python variant. But graalpy also needs cpython to build, and whatever I tried, conda would either complain that there's a cyclic dependency between python<->graalpy or that it thinks graalpy depends on itself.

timfel:
It looks like the docker build runs out of memory building graalpy. Anything we can do there?

@isuruf I included the changes to python_abi in this PR, because the changes to the abi must be available to the `graalpy-meta` package during the build.

recipes/graalpy/meta.yaml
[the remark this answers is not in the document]
timfel: I could workaround the issue by not requiring `python` for the build of graalpy, but instead using `pypy3.8`. Unfortunately, pypy is not compatible enough and fails running our build script :/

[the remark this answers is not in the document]
isuruf: I think the issue is that locally you have a `python` build that depends on `graalpy`. What if you explicitly depend on the `cpython` version?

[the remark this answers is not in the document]
timfel: I tried to explicitly depend on `python 3.11 *_cp311` or even `python_abi 3.11` (which I think should pull in cpython by default), but in both cases conda then complained.
```
    raise exceptions.RecipeError("Mismatching hashes in recipe. Exact pins in dependencies "
conda_build.exceptions.RecipeError: Mismatching hashes in recipe. Exact pins in dependencies that contribute to the hash often cause this. Can you change one or more exact pins to version bound constraints?
Involved packages were:
Mismatching package: python (id 0_223_graalpy); dep: python 3.11.0 ha86cf86_0_cpython; consumer package: graalpy
```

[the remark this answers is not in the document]
timfel: Ok, I got it to work with a little patch and using `pypy3.8` as the build dependency

timfel:
I don't understand why I get these builds in the log
```
    graalpy-22.3.0-python_3.9.* *_cpython-on-linux-64
    graalpy-22.3.0-python_3.10.* *_cpython-on-linux-64
    graalpy-22.3.0-python_3.8.* *_cpython-on-linux-64
```

There shouldn't be any variants, why am I not simply getting a single build?

timfel:

timfel:
Some explanation:

There are two distributions for GraalPy: a) in a GraalVM, which means it runs the Java code on top of the Graal compiler and b) as a standalone native executable generated with GraalVM native image. The latter is created via a static analysis of the entire codebase, which takes about 11GB of RAM, too much for the Azure runners here. So I disabled the standalone distribution for now and only build the GraalVM-based distribution. The standalone should be preferred, which I've tried to express via the `track_features: - openjdk` for the GraalVM distribution build.

The python_abi maintained by @isuruf needs to be updated to be able to select GraalPy as a python implementation for other Python packages. Because there's a bit of a circular dependency here, with GraalPy providing the packages `graalpy` and `python` and the latter needing the `python_abi` build with GraalPy ABI, I've included the changes to `python_abi` in this PR so the CI run will work. What is the best approach to deal with this? My idea was:

1. Comment out the `python` package provided here and the references to `python_abi`
2. Remove the python_abi recipe from the PR
3. Merge this to make a `graalpy-feedstock` that for now only provides `graalpy` as a package
4. Propose my changes to `python_abi-feedstock`, because these can then reference the `graalpy` package newly available on conda-forge
5. Uncomment the `python` package provided in `graalpy-feedstock`, since now the `python_abi` build for GraalPy ABI should be available on conda-forge.

Are there better ways?

.scripts/build_steps.sh
```diff
@@ -5,6 +5,8 @@
 # changes to this script, consider a proposal to conda-smithy so that other feedstocks can also
 # benefit from the improvement.
 
+git config --global --add safe.directory '*'
```
timfel, line 8: I can revert this, but it also wouldn't hurt I think, and it fixes execution with podman instead of docker for me.

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,199 @@
...
+  - name: python
+    version: {{ gp_python_version }}
+    build:
+      number: {{ build_num }}
+      string: {{ build_num }}_{{ graalpy_abi }}_graalpy
+      track_features:
+        - graalpy
+      run_exports:
+        weak:
+          - {{ pin_subpackage("graalpy", max_pin="x.x", min_pin="x.x") }}
```
isuruf, line 135: You need a `python_abi` pin here like pypy.

timfel: I tried that at some point and then the build failed to resolve complaining about over constraining, but I've changed some other bits since then so I'll try again and see what the CI says.

timfel: Ok, it works like this.
```yaml
      run_exports:
        weak:
          - {{ pin_subpackage("graalpy", max_pin="x.x", min_pin="x.x") }}
          - {{ pin_subpackage("python_abi", exact=True) }}
```

timfel: I now used this:
````yaml
      run_exports:
        weak:
          - {{ pin_subpackage("graalpy", max_pin="x.x", min_pin="x.x") }}
          - {{ pin_subpackage("python_abi", max_pin="x.x", min_pin="x.x") }}
```

My reasoning is the following, please let me know if I am still misunderstanding something: The export for graalpy will imply the `run_constrained` spec `{{ pin_subpackage("python_abi", exact=True) }}` from there. So whichever graalpy version is chosen, it will constrain the python_abi to be the exact one that was built with it. Thus, even though we only have a version pin for the `python_abi` in the weak exports for `python` (which by itself may allow python_abi builds pulling in cpython), the `graalpy` weak export ensures that at runtime, users get a compatible python_abi provided via graalpy.

isuruf: I really don't like this complexity. How about you remove `python_abi` from this recipe and we build it in https://github.com/conda-forge/python_abi-feedstock ?

isuruf:
Some questions
1. Why is there no `python3.8` symlink?
2. What's the suffix for the wheel name when a wheel is made for graalpy?
3. Do you expect to add a short name to https://github.com/pypa/packaging/blob/main/packaging/tags.py#L31-L37 ?

isuruf:
4. Why is `sys.maxsize` equal to `2147483647` on a 64-bit system?

recipes/graalpy/meta.yaml
timfel: I would like to. The problem is what I described [above](https://github.com/conda-forge/staged-recipes/pull/20882#issuecomment-1304795244). As far as I can tell, we cannot pass the CI with this if I already have the `python` output in there, because that needs to export the python_abi with the graalpy_abi in the build string, which is not satisfiable unless we have it first in `python_abi`, but we cannot add it to `python_abi`, yet, because there it needs to depend on `graalpy` which is here. So the only alternative I see is I take all the outputs out of this recipe for now, merge it so we get `graalpy-feedstock`, then open a PR to add the graalpy abi to `python_abi-feedstock`, and then update the graalpy-feedstock to have the `python` output again. What I did here was my attempt at getting this into one PR instead of 3.

timfel:
> 4. Why is `sys.maxsize` equal to `2147483647` on a 64-bit system?

Because we are on Java, and this is the most compatible way to handle this for us.

timfel:
>     1. Why is there no `python3.8` symlink?

We just don't ship one in any of our other artifacts. If you think it would be useful, I can add it. So far nobody who used GraalPy seemed to expect it.

>     2. What's the suffix for the wheel name when a wheel is made for graalpy?

For Python C extensions on GraalPy, the PEP 427 `{python tag}-{abi tag}-{platform tag}` turn into `graalpy38-graalpy_38_native_x86_64_linux-linux_i686`

The verbosity of our abi tag comes from the fact that GraalPy does not execute native code for native extensions. The LLVM toolchain that comes with GraalPy is used to produce LLVM bitcode for all extensions, and the bitcode is interpreted, the native code in the shared objects is simply ignored by GraalPy. Our abi tag thus encodes the compilation settings for the bitcode (header versions, whether callouts to system libraries are allowed (native) or not (managed), and what underlying system our LLVM toolchain told the C extension it is building against - the latter does not have to match the runtime system, for example, we can just compile as if running on Linux x86_64 and then execute that same bitcode on Windows or Darwin, x86_64 or aarch64).

>     3. Do you expect to add a short name to https://github.com/pypa/packaging/blob/main/packaging/tags.py#L31-L37 ?

No, for now we expect to keep using `sys.implementation.name` (i.e., `graalpy`) as per PEP 425.

isuruf:
Thanks for the info. Why is the `platform_tag` for graalpy `linux_i686`? I guess that's because `packaging` looks at `sys.maxsize`? https://github.com/pypa/packaging/blob/main/packaging/tags.py#L40

timfel:
> Thanks for the info. Why is the `platform_tag` for graalpy `linux_i686`? I guess that's because `packaging` looks at `sys.maxsize`? https://github.com/pypa/packaging/blob/main/packaging/tags.py#L40

Yes :(

recipes/graalpy/meta.yaml
timfel: I pushed the variant where I removed python_abi and commented all references to it out (because they cannot be resolved). If you think this is better, I'm happy with this approach, too :) Then once there's a feedstock for graalpy, I will send the changes to python_abi-feedstock in a PR to that repo and then uncomment the constraints in the graalpy-feedstock and bump the build number there.

isuruf:
Would you be open to changing these in packaging? I can start a discussion in packaging mailing list.
From that I mean, is it okay for you to change the abi tag in graalpy?

timfel:
> Would you be open to changing these in packaging? I can start a discussion in packaging mailing list. From that I mean, is it okay for you to change the abi tag in graalpy?

I'd be happy to change it. I'll just have to get the contribution permission for the pypa/packaging project, but that shouldn't take more than a week.

isuruf:
I guess once https://github.com/pypa/packaging/pull/607/files#diff-e2a5f8198f944bfcd050534d68367fd50e3f9f28afe1bb4b54c72afccd8fc0f6R237-R238 lands the abi tag would change right?

isuruf:
> turn into graalpy38-graalpy_38_native_x86_64_linux-linux_i686 for example.

What's the value for macos?

timfel:
> I guess once https://github.com/pypa/packaging/pull/607/files#diff-e2a5f8198f944bfcd050534d68367fd50e3f9f28afe1bb4b54c72afccd8fc0f6R237-R238 lands the abi tag would change right?

Ah, thanks for pointing that out, I hadn't seen that PR. Just reading the code right now, I guess our wheels would have `graalpy38-graalpy_38_native-linux_i686` instead of `graalpy38-graalpy_38_native_x86_64_linux-linux_i686`. Not 100% sure, need to check on my work computer.

-------------------------

>> turn into graalpy38-graalpy_38_native_x86_64_linux-linux_i686 for example.

> What's the value for macos?

e.g. `graalpy38-graalpy_38_native_x86_64_darwin-darwin_19_6_0_x86_64`

isuruf:
> graalpy38-graalpy_38_native_x86_64_darwin-darwin_19_6_0_x86_64

That's weird. What do you get for `platform.system()`?

timfel:
> That's weird. What do you get for `platform.system()`?

Here's what I get for `platform.uname()` on GraalPy
* macOs: `uname_result(system='Darwin', node='tims-iMac-Pro.local', release='19.6.0', version='Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64', machine='x86_64', processor='i386')`
* Windows: `uname_result(system='win32', node='DESKTOP-UOQCHD2', release='10.0', version='', machine='x86_64', processor='Intel64 Family 6 Model 165 Stepping 2, GenuineIntel')`
* Linux: `uname_result(system='Linux', node='DESKTOP-UOQCHD2', release='5.15.68.1-microsoft-standard-WSL2', version='#1 SMP Mon Sep 19 19:14:52 UTC 2022', machine='x86_64', processor='x86_64')`

.scripts/build_steps.sh
jakirkham: Let's revert when we are done here

timfel:
@isuruf is there anything else to do here for me to move this along? Or do you feel we have to wait until the discussion around https://github.com/pypa/packaging/pull/607 and https://github.com/pypa/packaging/pull/614 resolves?

recipes/graalpy/meta.yaml
timfel: Since the python_abi PR is now merged and I can thus add the below to constrain to the proper `python_abi` version `python_abi 3.8.* *_graalpy_38_native`

```yaml
run_constrained:
  - python_abi {{ python_maj_min }}.* *_graalpy_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
```

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,182 @@
...
+      - patches/sulong-disable-tests.patch
+      - patches/sulong-relative-rpath.patch
+  - url: https://github.com/oracle/graalpython/archive/refs/tags/vm-ce-{{ version }}.zip
+    sha256: 72ad9a6e8d8283425768c2838d9f360a42eb1dc6096d0efe44464f4fd5c0982c
+    folder: graalpython
+    patches:
+      - patches/graalpy-honour-sysroot.patch
+      - patches/graalpy-reduce-image-build-xmx.patch
+      - patches/graalpy-fix-lzma.patch  # [osx]
+      - patches/graalpy-graalvm-version-tag.patch
```
isuruf, line 54: I forgot about this patch. Let me fix the python_abi-feedstock

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,182 @@
...
+    version: {{ gp_python_version }}
+    build:
+      number: {{ build_num }}
+      string: {{ build_num }}_{{ graalpy_abi }}{{ graalpy_maj_min.replace('.', '') }}_graalpy
+      track_features:
+        - graalpy
+      run_exports:
+        weak:
+          - {{ pin_subpackage("graalpy", max_pin="x.x", min_pin="x.x") }}
+          - python_abi {{ python_maj_min }}.* *_graalpy_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
```
isuruf, line 118:
```suggestion
          - python_abi {{ python_maj_min }}.* *_graalpy{{ graalpy_maj_min.replace('.', '') }}_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
```

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,182 @@
...
+        noarch:
+          - {{ pin_subpackage("python", max_pin="x.x", min_pin="x.x") }}
+    requirements:
+      host:
+        - {{ pin_subpackage("graalpy", max_pin="x.x.x", min_pin="x.x.x") }}
+      run:
+        - {{ pin_subpackage("graalpy", max_pin="x.x.x", min_pin="x.x.x") }}
+      run_constrained:
+        - python_abi {{ python_maj_min }}.* *_graalpy_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
```
isuruf, line 127:
```suggestion
        - python_abi {{ python_maj_min }}.* *_graalpy{{ graalpy_maj_min.replace('.', '') }}_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
```

recipes/graalpy/build_graalpy.sh
```diff
@@ -0,0 +1,157 @@
...
+    # match cpython include folder structure
+    mkdir -p $PREFIX/include/
+    ln -sf $GRAALVM_PREFIX/languages/python/include $PREFIX/include/python${PY_VERSION}
+
+    # license is packaged by the build process
+    cat $GRAALVM_PREFIX/*LICENSE*.txt $GRAALVM_PREFIX/3rd_party_license*.txt > $SRC_DIR/LICENSE_GRAALPY.txt
+fi
+
+# create the site-packages folder to match cpython
+mkdir -p $PREFIX/lib/python${PY_VERSION}/site-packages
```
isuruf, line 157: This is not enough. You need to do something like
```
site_packages=$(python -c 'import site; print(site.getsitepackages()[0])')
rm -rf ${site_packages}
ln -sf $PREFIX/lib/python${PY_VERSION}/site-packages ${site_packages}
```

recipes/graalpy/build_graalpy.sh
```diff
@@ -0,0 +1,157 @@
...
+    fi
+
+    # delete linux bits from sulong's llvm toolchain
+    if [ -n "${MACOS}" ]; then
+        rm -rf $GRAALVM_PREFIX/lib/llvm/lib/clang/*/lib/linux/
+    fi
+
+    # match cpython include folder structure
+    mkdir -p $PREFIX/include/
+    ln -sf $GRAALVM_PREFIX/languages/python/include $PREFIX/include/python${PY_VERSION}
```
isuruf, line 150: This needs to be the other way. Delete `$GRAALVM_PREFIX/languages/python/include` and link from `$PREFIX/include/python${PY_VERSION}` to that file.

That's because we don't want the graalvm version embedded in file names of downstream packages.

recipes/graalpy/build_graalpy.sh
timfel: `graalpy -c 'import site; print(site.getsitepackages()[0])'` gives `$PREFIX/lib/python${PY_VERSION}/site-packages` - graalpy already expects the site-packages folder right there where I'm creating it

recipes/graalpy/meta.yaml
```diff
@@ -0,0 +1,182 @@
...
+        noarch:
+          - {{ pin_subpackage("python", max_pin="x.x", min_pin="x.x") }}
+    requirements:
+      host:
+        - {{ pin_subpackage("graalpy", max_pin="x.x.x", min_pin="x.x.x") }}
+      run:
+        - {{ pin_subpackage("graalpy", max_pin="x.x.x", min_pin="x.x.x") }}
+      run_constrained:
+        - python_abi {{ python_maj_min }}.* *_graalpy{{ graalpy_maj_min.replace('.', '') }}_{{ python_maj_min.replace('.', '') }}_{{ graalpy_abi }}
+    test:
```
isuruf, line 128:
```suggestion
    test:
      requires:
        - setuptools
      imports:
        # check that noarch packages work
        - setuptools
```

recipes/graalpy/build_graalpy.sh
timfel: Done

recipes/graalpy/build_graalpy.sh
timfel: I added code to ensure this is as you asked, even if in the future graalpy expects it's site-packages directory to be somewhere else

timfel:
I think the test is missing a `requires: pip`, right? Or is that expected to just be there?

timfel:
@isuruf thanks for the help with updating it. it played with the current recipe locally with various noarch packages, and it seems to work fine, that's great :)

isuruf:
@conda-forge/openjdk, @conda-forge/python, we'll be adding variants of `openjdk` and `python` here? Are you okay with having these variants?

timfel:
Hi, and happy new year :) Was there anything from the Python or Openjdk teams? (I don't see @isuruf's comment linking those, so not sure if the notification worked for those teams)

PR #20892, opened 2022-10-25
Add recipe for google-weather-tools

blackvvine:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/google-weather-tools/meta.yaml
+++ b/recipes/google-weather-tools/meta.yaml
@@ -1,0 +1,89 @@
+{% set name = "google-weather-tools" %}
+{% set version = "0.3.2" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  url: "https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz"
+  sha256: db9e8566091b72a1f20813f043b356590dc6aefd9f5cc5c983c0cbc9d4faf0e5
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  build:
+    - python=3.8
+    - cdsapi
+    - ecmwf-api-client
+    - apache-beam[gcp]==2.40.0
+    - numpy
+    - pandas
+    - xarray
+    - requests
+    - google-cloud-firestore
+    - dataclasses
+    - cfgrib
+    - netcdf4
+    - geojson
+    - simplejson
+    - rioxarray
+    - pyproj
+    - gdal
+    - pygrib
+    - flake8
+    - pytest
+    - pytest-subtests
+    - xarray-beam
+    - absl-py
+    - metview-batch
+    - pip
+    - google-cloud-datastore
+    - google-cloud-bigquery
+    - urllib3==1.26.5
+    - earthengine-api >=0.1.263
+    - google-cloud-storage
+  host:
+    - pip
+    - python=3.8
+  run:
+    - pip
+    - apache-beam[gcp]==2.40.0
+    - python=3.8
+
+test:
+  imports:
+    - weather_dl
+    - weather_dl.download_pipeline
+    - weather_mv
+    - weather_mv.loader_pipeline
+    - weather_sp
+    - weather_sp.splitter_pipeline
+  requires:
+    - absl-py
+    - flake8
+    - netcdf4
+    - numpy
+    - pytest
+    - pytest-subtests
+    - pytype ==2021.11.29
+    - xarray
+    - xarray-beam
+
+about:
+  home: "https://weather-tools.readthedocs.io/"
+  license: Apache Software
+  license_family: APACHE
+  license_file: LICENSE
+  summary: "Apache Beam pipelines to make weather data accessible and useful."
+  doc_url: https://github.com/google/weather-tools/tree/main/docs
+  dev_url: https://github.com/google/weather-tools/blob/main/LICENSE
+
+extra:
+  recipe-maintainers:
+    - blackvvine
+    - alxmrs
+    - mahrsee1997


recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+  sha256: db9e8566091b72a1f20813f043b356590dc6aefd9f5cc5c983c0cbc9d4faf0e5
+
+build:
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  build:
+    - python=3.8.13
```
xhochy, line 19: These dependencies are too strict, should not be in `build` and generally don't match the format we use in recipes. I suggest you try generating this recipe using `grayskull`.

xhochy (changes requested):

recipes/google-weather-tools/meta.yaml
blackvvine: Hey @xhochy , thanks for looking into this. We have updated the meta.yaml to fix the format. Regarding the fixed library versions, this distribution is intended to replace PyPi as the official channel for production Dataflow pipelines and some of our 3rd party dependencies are very sensitive to dependency versions.

xhochy:
The recipe still looks very wrong. I strongly suggest you try out `grayskull` and read a bit on the conda-build documentation https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html about what you are doing here. This kind of package doesn't need a build section at all.

recipes/google-weather-tools/meta.yaml
blackvvine: We've fixed this in the new version which is based on `grayskull`

blackvvine:
@conda-forge/help-python The recipe has passed all checks and is ready for your review. Thanks.

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+
+test:
+  imports:
+    - weather_dl  # [unix]
+    - weather_mv  # [unix]
+    - weather_mv.loader_pipeline  # [unix]
+    - weather_sp  # [unix]
+    - weather_sp.splitter_pipeline  # [unix]
+  requires:
+    - absl-py
```
xhochy, line 47: If they are required, they should be listed in run

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  git_url: https://github.com/google/weather-tools.git
+  git_rev: "{{ git_rev }}"
+
+build:
+  skip: true  # [win]
```
xhochy, line 14:
```suggestion
  noarch: python 
```

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    {{ PYTHON }} -m pip install . -vv && 
+    {{ PYTHON }} -m pip install ./weather_dl -vv && 
+    {{ PYTHON }} -m pip install ./weather_mv -vv && 
+    {{ PYTHON }} -m pip install ./weather_sp -vv 
+
+requirements:
+  build:
+    - gdal =3.5.1
+    - pip
+    - python =3.8
```
xhochy, line 26:
```suggestion
```

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - setuptools_scm
+  run:
+    - apache-beam[gcp] ==2.40.0
+    - earthengine-api >=0.1.263
+    - gdal =3.5.1
+    - metview-batch
+    - python =3.8
+    - pip
```
xhochy, line 37:
```suggestion
```

xhochy (changes requested):

recipes/google-weather-tools/meta.yaml
blackvvine: @xhochy Thanks for your help and happy new year! Some of our runtime dependencies are not supported in Windows, what is the right way to release for Unix only if having **noarch** is mandatory?

blackvvine: According to [this answer](https://github.com/conda-forge/staged-recipes/issues/11393#issuecomment-638423909), I assume the recipe should not be `noarch` since it's not supported in Windows. Please correct me if I'm wrong.

blackvvine:
@conda-forge/help-python The recipe has passed all checks and is ready for your review. Thanks.

recipes/google-weather-tools/meta.yaml
dopplershift: No, if it's just that the dependencies are unavailable on Windows, that's fine for this to be noarch. It's only a problem with dependencies that are only install on Windows/not-Windows. noarch packages are generally only built on Linux systems, which will succeed. The missing dependencies will then keep it from being installable on Windows.

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+{% set git_rev = "conda-candidate-v0" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  git_url: https://github.com/google/weather-tools.git
+  git_rev: "{{ git_rev }}"
+
```
dopplershift, line 12: We generally only want to be using released versions.

alxmrs: +1  - is there a way to configure this to reflect our git release tags?

recipes/google-weather-tools/meta.yaml
alxmrs: I'm a bit confused by the discussion here. In my understanding, the solution to move forward is to configure this to be noarch. 

To add some general background context: We have not yet implemented windows support in the package we want to release (https://github.com/google/weather-tools/issues/64). If we set this to noarch, does that mean we have to fix this bug in order to proceed? 

Thanks for your inputs, @dopplershift and @xhochy.

dopplershift: Well since you said:
> Some of our runtime dependencies are not supported in Windows

That's why I said noarch was the right answer. This is the first time it's been mentioned that it doesn't work properly on Windows. In that case, we can proceed with non-noarch.

Having said that though, it looks like the problem is just that you're using `scripts` for the executables. That's pretty easy to move to [`console_scripts`](https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html) which work on Windows and are supported by our noarch builds.

recipes/google-weather-tools/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+{% set version = "0.3.2" %}
+{% set git_rev = "conda-candidate-v0" %}
+
+package:
+  name: "{{ name|lower }}"
+  version: "{{ version }}"
+
+source:
+  git_url: https://github.com/google/weather-tools.git
+  git_rev: "{{ git_rev }}"
```
dopplershift, line 11: If all you want is to just get a git tag, use:
```suggestion
  url: https://github.com/google/weather-tools/archive/refs/tags/v{{ version }}.tar.gz
  sha256: YOURSHAHERE
```

alxmrs: Thanks! Will TAL at this soon.

alxmrs:
Closing now; I think the project has gone with other distribution channels.

(edit: sorry -- I can't close -- @blackvvine would you mind?)

PR #20913, opened 2022-10-27
add pydistcheck

jameslamb:
**Description**

Adds a recipe for `pydistcheck` (https://pypi.org/project/pydistcheck/).

**Checklist**

- [x] Build number is 0.

### Notes for Reviewers

This recipe was generated with `grayskull`, following instructions at https://conda-forge.org/docs/maintainer/adding_pkgs.html.

### DIFF
--- a/recipes/pydistcheck/meta.yaml
+++ b/recipes/pydistcheck/meta.yaml
@@ -12,6 +12,8 @@ source:
 build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
+  entry_points:
+    - pydistcheck = pydistcheck.cli:check
   number: 0
 
 requirements:
@@ -28,6 +30,7 @@ test:
     - pydistcheck
   commands:
     - pip check
+    - pydistcheck --help
   requires:
     - pip
 
--- a/recipes/pydistcheck/meta.yaml
+++ b/recipes/pydistcheck/meta.yaml
@@ -10,18 +10,18 @@ source:
   sha256: 806b21446c4c544ec792e6313509d14735f8f93dd26f01d1ce8e5c556d79cb4d
 
 build:
-  skip: true  # [py<38]
+  noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - pip
   run:
-    - python
+    - python >=3.8
     - click
-    - tomli  # [py<311]
+    - tomli
 
 test:
   imports:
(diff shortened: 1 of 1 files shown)


jameslamb:

recipes/pydistcheck/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pydistcheck-{{ version }}.tar.gz
+  sha256: 806b21446c4c544ec792e6313509d14735f8f93dd26f01d1ce8e5c556d79cb4d
+
+build:
+  skip: true  # [py<38]
```
saraedum, line 13:
```suggestion
  noarch: python
```
Isn't this a pure python package?

If you have Python requirements, you could set `- python >=3.8` in the `host/run` sections below.

I see that you have a `tomli  # [py<311]` requirement. This won't work with `noarch: python`. But would it really hurt to install tomli always, even if python is 3.11?

saraedum:
This looks fine to me. I would prefer this to be a pure Python package so we don't have to build this for the full matrix of OSes and Python versions even if this adds an unused dependency in one case. But if you insist, I think it could also be merged as is.

recipes/pydistcheck/meta.yaml
jameslamb: Yes, this is a pure Python package.

I'd hoped to have `tomli` be conditional on the Python version so that that package would be lighter-weight in later versions of Python, but I didn't understand that doing that meant imposing a larger burden on the `conda-forge` infrastructure.

I'm ok with it being unconditional if that means my little CLI here will be a bit less expensive for `conda-forge` to build and distribute.

I just pushed https://github.com/conda-forge/staged-recipes/pull/20913/commits/81c4af9abfea7bdd9468e171f31329c734685ac3 which modifies the recipe based on your suggestions:

* makes `tomli` an unconditional dependency
* switches to `noarch: python`
* puts a floor of `python >= 3.8` in `requirements.host` and `requirements.run`

saraedum:
Sorry, you also need to list the entry points for this to work: https://conda.io/projects/conda-build/en/latest/resources/define-metadata.html#python-entry-points

jameslamb:
> you also need to list the entry points for this to work

ah ok, can do! I wasn't sure if that was necessary, since I have them in the project's `setup.cfg` ([code link](https://github.com/jameslamb/pydistcheck/blob/441c02713f593ea296fa9e82d4115108f1821808/setup.cfg#L43-L45)).

Just pushed https://github.com/conda-forge/staged-recipes/pull/20913/commits/9b5e22af06c1814c8be46992511f2dad13494169 which:

* adds the project's one entrypoint to the recipe
* adds `pydistcheck --help` to the commands in `test.commands:`, to confirm that the entrypoint is set up correctly when installed

PR #20921, opened 2022-10-27
Add recipe for `sphinxcontrib-jquery`

BSGalvan:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/sphinxcontrib-jquery/meta.yaml
+++ b/recipes/sphinxcontrib-jquery/meta.yaml
@@ -19,7 +19,7 @@ requirements:
   host:
     - python >=2.7
     - pip
-    - flit_core >=3.7,<4
+    - flit-core >=3.7,<4
   run:
     - python >=2.7
     - setuptools
--- a/recipes/sphinxcontrib-jquery/meta.yaml
+++ b/recipes/sphinxcontrib-jquery/meta.yaml
@@ -45,4 +45,5 @@ about:
 
 extra:
   recipe-maintainers:
+    - AA-Turner
     - BSGalvan
--- a/recipes/sphinxcontrib-jquery/meta.yaml
+++ b/recipes/sphinxcontrib-jquery/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "sphinxcontrib-jquery" %}
+{% set version = "2.0.0" %}
+{% set sha256 = "8fb65f6dba84bf7bcd1aea1f02ab3955ac34611d838bcc95d4983b805b234daa" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=2.7
+    - pip
+    - flit_core >=3.7,<4
+  run:
+    - python >=2.7
+    - setuptools
+    - sphinx
+
+test:
+  imports:
+    - sphinxcontrib.jquery
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://pypi.org/project/sphinxcontrib-jquery/
+  license: 0BSD
+  license_file: LICENCE
+  summary: A sphinx extension to include jQuery on newer sphinx releases
+  description: |
+    sphinxcontrib-jquery ensures that jQuery is always installed for use in
+    Sphinx themes or extensions.
+  dev_url: https://github.com/sphinx-contrib/jquery
+
+extra:
+  recipe-maintainers:
+    - BSGalvan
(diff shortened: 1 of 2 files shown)


recipes/sphinxcontrib-jquery/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  run:
+    - python >=2.7
+    - setuptools
+    - sphinx
+
+test:
+  imports:
+    - sphinxcontrib.jquery
+  commands:
+    - pip check
```
AA-Turner, line 32: Perhaps also run the package's tests here.

A

recipes/sphinxcontrib-jquery/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  license_file: LICENCE
+  summary: A sphinx extension to include jQuery on newer sphinx releases
+  description: |
+    sphinxcontrib-jquery ensures that jQuery is always installed for use in
+    Sphinx themes or extensions.
+  dev_url: https://github.com/sphinx-contrib/jquery
+
+extra:
+  recipe-maintainers:
+    - AA-Turner
```
AA-Turner, line 48: Confirmation comment.

A

recipes/sphinxcontrib-jquery/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  imports:
+    - sphinxcontrib.jquery
+  requires:
+    - pip
+    - pytest
+  source_files:
+    - tests
+  commands:
+    - pip check
+    - python -m pytest tests -vv
```
jaimergp, line 38: You could try with this, but first check that the PyPI tarball is actually packaging the tests.

```suggestion
  source_files:
    - tests/*
  commands:
    - pip check
    - ls -alh .
    - python -m pytest tests/ -vv
```

BSGalvan: Just confirmed (by downloading the [latest tarball from PyPI](https://files.pythonhosted.org/packages/38/e8/8ba91a585a8d0b6a1952eeeb8ea392e025e4efc7c860f4679794a776b116/sphinxcontrib-jquery-2.0.0.tar.gz)) that tests are not being packaged.

Pinging @AA-Turner for confirmation.

jaimergp: You might want to use the github tarball in that case. However, you are not required to run the full test suite on conda-forge as part of the packaging submission! It's ok if you want to run it once or twice before leaving a simple test section to ensure things are working, though.

bollwyvl:
Yep, the sdist doesn't ship the tests.

You can do multiple `source` entries...

```
source:
  - folder: dist
    sha256: 8fb65f6dba84bf7bcd1aea1f02ab3955ac34611d838bcc95d4983b805b234daa
  - folder: src
    url: https://github.com/sphinx-contrib/jquery/archive/refs/tags/v{{ version }}.tar.gz
    sha256: d3adb33f
```

...and then `cd` into all the appropriate folders. 

In this case this has the common advantage of removing the need for the vendored `LICENCE`, as you could refer to `- src/LICENCE`... I'd often do an upstream PR to add that _particular_ file to the sdist... and potentially the tests as well. As long as it doesn't appear in the `.whl`, and it's noarch, most projects seem fine with having a "fat" sdist. Not a big issue in this case, as it doesn't appear to do much!

recipes/sphinxcontrib-jquery/meta.yaml
BSGalvan: @jaimergp @bollwyvl Thanks for the insightful comments!

I've been thinking what would be the best way forward with this recipe, based on your suggestions.
I think that an upstream PR to add the `tests` into the sdist would be the most elegant way forward.
This way, whenever said PR is merged we can update this recipe to do the packaged tests.

Till that time however, since testing _is_ [being done upstream](https://github.com/sphinx-contrib/jquery/commit/08d948a4a22961c83f468fc5dd2200fc411c0c23), I feel that we can keep this recipe simple and sweet, doing only the (relatively) simple `import` and `pip check` tests.

@AA-Turner does that seem like a sensible course of action to you as well?

recipes/sphinxcontrib-jquery/meta.yaml
BSGalvan: @AA-Turner pinging you once more just in case the previous pings were lost in a sea of notifications.

I apologize for the large break since the last discussions, many things came up in real life here which pulled me away from GitHub.

To recap, previous discussions revealed that the package on PyPI was not packaging the tests, which made running those tests on `conda-forge`'s CI infrastructure [more complicated](https://github.com/conda-forge/staged-recipes/pull/20921#issuecomment-1294965232).

This led to [a thought](https://github.com/conda-forge/staged-recipes/pull/20921#discussion_r1008627170) that maybe it was best to leave testing to upstream (for now), and get tests running here if and when said tests were packaged into the sdist.

If that course of action sounds alright to you, we can move ahead with the simpler version of this recipe, and modify said recipe down the line. What say you?

BSGalvan:
@conda-forge/staged-recipes ready to go! Following feedback from all maintainers, only simple tests are now to be carried out on the `conda-forge` CI.

PR #20987, opened 2022-11-03
add professor

admorris:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/professor/0001-Fix-LC_ID_DYLIB-on-macOS.patch
+++ b/recipes/professor/0001-Fix-LC_ID_DYLIB-on-macOS.patch
@@ -1,0 +1,25 @@
+From 18480a41cfb23cdabf1241cb20ea9e793b4349b2 Mon Sep 17 00:00:00 2001
+From: Chris Burr <christopher.burr@cern.ch>
+Date: Fri, 4 Nov 2022 16:49:35 +0100
+Subject: [PATCH] Fix LC_ID_DYLIB on macOS
+
+---
+ Makefile | 2 +-
+ 1 file changed, 1 insertion(+), 1 deletion(-)
+
+diff --git a/Makefile b/Makefile
+index baf73e2..e8043f7 100644
+--- a/Makefile
++++ b/Makefile
+@@ -105,7 +105,7 @@ lib: lib/libProfessor2.so
+ 
+ lib/libProfessor2.so: $(LIBOBJECTS)
+ 	mkdir -p lib
+-	$(CXX) -shared -o $@ $(LIBOBJECTS)
++	cd lib && $(CXX) -shared -o $(notdir $@) $(addprefix ../,$(LIBOBJECTS))
+ 
+ obj/%.o: src/%.cc $(LIBHEADERS)
+ 	mkdir -p obj
+-- 
+2.37.1 (Apple Git-137.1)
+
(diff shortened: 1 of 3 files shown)


recipes/professor/meta.yaml
```diff
@@ -0,0 +1,85 @@
+{% set name = "Professor" %}
+{% set version = "2.3.4" %}
+{% set tag = name|lower ~ "-" ~ version %}
```
chrisburr, line 3: If upstream is amenable to it tagging dots instead of dashes will make the bots life easier for detecting updates.

If not, the trick is to define the version in terms of the tag rather than the opposite.

chrisburr: (doesn't matter for now, but if there is a new version)

admorris, line 3: > the trick is to define the version in terms of the tag rather than the opposite

@chrisburr like this?

```suggestion
{% set tag = "professor-2.3.4" %}
{% set version = tag | replace("professor-", "") %}
```

recipes/professor/0001-Fix-LC_ID_DYLIB-on-macOS.patch
```diff
@@ -0,0 +1,25 @@
...
+diff --git a/Makefile b/Makefile
+index baf73e2..e8043f7 100644
+--- a/Makefile
++++ b/Makefile
+@@ -105,7 +105,7 @@ lib: lib/libProfessor2.so
+ 
+ lib/libProfessor2.so: $(LIBOBJECTS)
+ 	mkdir -p lib
+-	$(CXX) -shared -o $@ $(LIBOBJECTS)
++	cd lib && $(CXX) -shared -o $(notdir $@) $(addprefix ../,$(LIBOBJECTS))
```
admorris, line 19: Is this macOS specific or should it go into upstream?

recipes/professor/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+{% set repo = "https://gitlab.com/hepcedar/" ~ name|lower %}
+{% set home = "https://" ~ name|lower ~ ".hepforge.org" %}
```
chrisburr, line 5: It's best to avoid being too clever with jinja2 as it makes the recipe hard to analyse:
```suggestion
```

recipes/professor/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: {{ repo }}/-/archive/{{ tag }}/{{ name|lower }}-{{ tag }}.tar.gz
```
chrisburr, line 12:
```suggestion
  url: https://gitlab.com/hepcedar/professor/-/archive/{{ tag }}/professor-{{ tag }}.tar.gz
```

recipes/professor/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - prof2-predict --help
+    - prof2-residuals --help
+    - prof2-runcombs --help
+    - prof2-sample --help
+    - prof2-sens --help
+    - prof2-tune --help
+    - prof2-tune-nest --help
+
+about:
+  home: {{ home }}
```
chrisburr, line 71:
```suggestion
  home: https://professor.hepforge.org
```

recipes/professor/0001-Fix-LC_ID_DYLIB-on-macOS.patch
chrisburr: This can go upstream. There are better ways to do it but this is the least invasive change which will make it work on most unix-like systems.

admorris: ok. I will include it in my MR targeting 2.3.5

PR #21036, opened 2022-11-05
adding parallelproj package

gschramm:
Checklist
- [x] Build number is 0.


gschramm:

gschramm:
Can somebody help me understanding why the cuda builds fail? I looked at the azure error logs, but don't really understand what the problem is.

jaimergp:
It can't seem to find `ninja` because it's looking for in the `host` environment instead of `build`... 🤔 This shouldn't happen so either:

* Call `ninja` manually (replace `cmake --build`)
* Add `ninja` to host (this is wrrrrrong but if the above doesn't work... 🤷 )

gschramm:
@jaimergp Thanks for the hints! Adding an empty host section to the requirements in the outputs helps. Unfortunately, the build still fails since cmake still fails to detect a cuda compiler. I am using ```check_language(CUDA)``` and the value of ```CMAKE_CUDA_COMPILER``` to test is cuda is there. I also make sure that I pass ```${CMAKE_ARGS}``` to the cmake call in the build script. Any idea why cuda is not found?

gschramm:
@jaimergp update: According to the last build error logs, it seems that when [cmake](https://github.com/gschramm/parallelproj/blob/master/CMakeLists.txt) is called from build_cuda.sh for the cuda output, some cmake options are not passed correctly. According to the log, ```SKIP_CUDA_LIB``` must have been ```TRUE``` (status message "skipping build of cuda lib" in the error log, but in [build_cuda.sh](https://github.com/gschramm/staged-recipes/blob/main/recipes/parallelproj/build_cuda.sh) it is not passed which means it should have the default value ```FALSE```. Any idea what is going wrong?

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+  # Uncomment the following line if the package is pure Python and the recipe
+  # is exactly the same for all platforms. It is okay if the dependencies are
+  # not built for all platforms/versions, although selectors are still not allowed.
+  # See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python
+  # for more details.
+  # noarch: python
+
+  number: {{ build_number }}
+  skip: True  # [win]
+  skip: True  # [osx]
```
h-vetinari, line 23:
```suggestion
  skip: True  # [osx or win]
  # only works for CUDA>=11.0
  skip: True  # [linux and cuda_compiler_version == "10.2"]
```

h-vetinari: Regarding your [question](https://gitter.im/conda-forge/conda-forge.github.io?at=637558744fee72663c7b6bac)

recipes/parallelproj/build_cuda.sh
```diff
@@ -0,0 +1,2 @@
+cmake -GNinja -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=lib ${CMAKE_ARGS} -DCMAKE_CUDA_ARCHITECTURES=all -DSKIP_DOCS=TRUE -DSKIP_OPENMP_LIB=TRUE ${SRC_DIR}
+cmake --build . --target install --verbose
```
h-vetinari, line 2: Good practice is generally to not build directly in-tree (use build folder), quit on error, and use linebreaks, i.e.

```
#!/bin/bash
set -ex

mkdir build
cd build

cmake -G Ninja \ 
    -DALSO \
    -DPLEASE \
    -DUSE \
    -DLINEBREAKS \
    -DFOR \
    -DREADABILITY \
    [other options] \
    ..

cmake --build . --target install --verbose 
```

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+        - cmake >=3.23
+      commands:
+        - ctest -VV
+
+  - name: parallelproj-gpu
+    build:
+      skip: True  # [cuda_compiler_version == "None"]
+      missing_dso_whitelist:
+        - $RPATH/libgomp.so.1
+        - $RPATH/libstdc++.so.6
```
h-vetinari, line 60: It seems to me like you should add libgomp / llvm-openmp to `host:`, then you shouldn't need this.

recipes/parallelproj/meta.yaml
gschramm: Thanks for the code review :). Changes were included

recipes/parallelproj/build_cuda.sh
gschramm: build files have been modified accordingly

recipes/parallelproj/meta.yaml
gschramm: @h-vetinari @jaimergp I added to libgomp / llvm-openmp to ```host:``` requirements, but the build with the Error: ```ERROR (parallelproj-gpu,lib/libparallelproj_cuda.so.1.2.5): $RPATH/libgomp.so.1 not found in packages, sysroot(s) nor the missing_dso_whitelist```. Any idea what I am doing wrong?

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    script: build_openmp.sh
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - cmake >=3.23
+        - ninja
+        - libgomp  # [linux]
+        - llvm-openmp  # [osx]
+      run:
```
h-vetinari, line 46: This should be `host:` as well

gschramm: Hm. The problem occurs when the parallelproj-gpu is build. And there the libs are in the requirements :/
So (without testing), I'd guess that adding it to the host req. for the parallelproj output would change that.
The parallelproj output get build fine.

h-vetinari:
Is it possible that your build is picking up the libgomp / libstdcxx from somewhere else? I'm a bit stumped what's going wrong here TBH.

recipes/parallelproj/meta.yaml
gschramm: Hm. Hard to say for me. Is there a way to test that?
Could it be related to this line here in the [CMakeList.txt](https://github.com/gschramm/parallelproj/blob/9483ad5122d076465e1511975657e89f7bec4122/cuda/CMakeLists.txt#L26)?

gschramm:
> .

Hm. Hard to say for me. Is there a way to test that?
Could it be related to this [line](https://github.com/gschramm/parallelproj/blob/9483ad5122d076465e1511975657e89f7bec4122/cuda/CMakeLists.txt#L26) or [this line](https://github.com/gschramm/parallelproj/blob/9483ad5122d076465e1511975657e89f7bec4122/cuda/CMakeLists.txt#L46) in the CMakeLists.txt?

gschramm:
@h-vetinari @jaimergp I had another detailed look into the raw error logs (see attachment below).
The libs mentioned in the overlinkling error (libstdc++.so.6 and libgomp.so.1) don't appear anywhere except for the error messages. 

Any ideas what else I can check / whom to ask for help?


[cuda_11_error_log.txt](https://github.com/conda-forge/staged-recipes/files/10045547/cuda_11_error_log.txt)

gschramm:
@conda-forge/help-c-cpp @conda-forge/staged-recipes From my point of view, this PR is ready to be merged. Could somebody do that? 

PS: Many thanks for all your efforts + help!

gschramm:
@conda-forge/help-c-cpp  please review

recipes/parallelproj/build.sh
```diff
@@ -0,0 +1,27 @@
...
+else
+  cmake -G Ninja \
+      ${CMAKE_ARGS} \
+      -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+      -DCMAKE_BUILD_TYPE=Release \
+      -DCMAKE_INSTALL_LIBDIR=lib \
+      -DSKIP_DOCS=TRUE \
+      -DSKIP_CUDA_LIB=TRUE \
+      ${SRC_DIR}
+fi
```
h-vetinari, line 25: This is a lot of duplication for only one option that differs, I suggest to condense as follows:
```suggestion
if [[ ${cuda_compiler_version:-None} != "None" ]]; then
  EXTRA_CMAKE_ARGS="-DCMAKE_CUDA_ARCHITECTURES=all"
else
  EXTRA_CMAKE_ARGS="-DSKIP_CUDA_LIB=TRUE"
fi

cmake -G Ninja \
    ${CMAKE_ARGS} \
    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_LIBDIR=lib \
    -DSKIP_DOCS=TRUE \
    ${EXTRA_CMAKE_ARGS} \
    ${SRC_DIR}
```

recipes/parallelproj/bld.bat
```diff
@@ -0,0 +1,20 @@
...
+    -DCMAKE_PREFIX_PATH:PATH="%LIBRARY_PREFIX%" ^
+    -DCMAKE_BUILD_TYPE:STRING=Release ^
+    -DSKIP_DOCS=TRUE ^
+    -DSKIP_CUDA_LIB=TRUE ^
+    %SRC_DIR%
+
+if errorlevel 1 exit 1
+
+cmake --build . --target install --verbose 
+if errorlevel 1 exit 1
```
h-vetinari, line 20: I prefer the less "magical" incantation (that makes it clear something is read from a variable), but that's just a nit.
```suggestion
    %SRC_DIR%
if %ERRORLEVEL% neq 0 exit 1

cmake --build . --target install --verbose 
if %ERRORLEVEL% neq 0 exit 1
```

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  url: https://github.com/gschramm/parallelproj/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 826a83e51db6459004d17c030c21fb32e7cb5205a45c28c8f50f87f235321196
+
+build:
+  # Uncomment the following line if the package is pure Python and the recipe
+  # is exactly the same for all platforms. It is okay if the dependencies are
+  # not built for all platforms/versions, although selectors are still not allowed.
+  # See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python
+  # for more details.
+  # noarch: python
```
h-vetinari, line 19: Remove this comment

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  number: {{ build_number }}
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cpu"   # [cuda_compiler_version == "None"]
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cuda"  # [cuda_compiler_version != "None"]
+
+  # If the installation is complex, or different between Unix and Windows,
+  # use separate bld.bat and build.sh files instead of this key. By default,
+  # the package will be built for the Python versions supported by conda-forge
+  # and for all major OSs. Add the line "skip: True  # [py<35]" (for example)
+  # to limit to Python 3.5 and newer, or "skip: True  # [not win]" to limit
+  # to Windows.
```
h-vetinari, line 30: Same here.

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [cuda_compiler_version != "None"]
+    - cmake >=3.23
+    - ninja
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
+  host:
+  run:
+    - libgomp  # [linux]
+    - llvm-openmp  # [osx]
```
h-vetinari, line 44: It's strange to me that this is not a host dependency, but some build/run hybrid. It still works because `llvm-openmp`'s  run_export is [marked](https://github.com/conda-forge/openmp-feedstock/blob/main/recipe/meta.yaml#L32) "strong" (e.g. it exports even from the build env), and `libgomp` does so [too](https://github.com/conda-forge/ctng-compilers-feedstock/blob/main/recipe/meta.yaml#L446) (see [here](https://github.com/conda-forge/ctng-compilers-feedstock/blob/main/recipe/meta.yaml#L519)), but I'm almost certain that what you actually want is:
```suggestion
  host:
    - libgomp      # [linux]
    - llvm-openmp  # [osx]
```

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - parallelproj_nontof_fwd_back_test.exe  # [win]
+
+about:
+  home: https://github.com/gschramm/parallelproj
+  license:  MIT
+  license_file: LICENSE
+  summary: 'parallelized TOF PET projectors'
+
+# Uncomment the following if this will be on a forge
+# Remove these lines if this is only be used for CI
```
h-vetinari, line 60: Another comment to remove

h-vetinari (commented):
LGTM in general, but some small improvements below

recipes/parallelproj/meta.yaml
gschramm: done

recipes/parallelproj/meta.yaml
gschramm: done

recipes/parallelproj/meta.yaml
gschramm: done

gschramm:
@h-vetinari I included all the suggested changes. Finally, two more questions for better understanding:
(1) Why are the openmp libs a host and not a run requirement? 
(2) Would it be good practice to export envirnmont variables pointing to the compiled libs using activate scripts?

h-vetinari:
> (1) Why are the openmp libs a host and not a run requirement?

It is _also_ a run requirement in practice, but that run requirement derives from being a host requirement (if you build against shared openmp, you would fail to load/link the package at runtime if openmp wasn't there). That's why the openmp package has a "run-export" that adds the run requirement wherever openmp is used as a host dependency.

In short: it's the cleaner and more accurate representation of what actually happens with the dependencies.

> (2) Would it be good practice to export envirnmont variables pointing to the compiled libs using activate scripts?

I'm not sure I understand. By default, everything that conda does happens in `$PREFIX`, so any downstream project wanting to consume parallelproj just needs to point there - no environment variables required

gschramm:
Ok. Thanks for the explanation. No need indeed for environment variables.
Is the PR now ready to be merged?

(ad 1) I am still confused about host / build req. I thought openmp should be a build and run req., but maybe my understanding of the "host" is wrong.

h-vetinari:
> Ok. Thanks for the explanation. No need indeed for environment variables. Is the PR now ready to be merged?
> 
> (ad 1) I am still confused about host / build req. I thought openmp should be a build and run req., but maybe my understanding of the "host" is wrong.

Don't worry, it happens to everyone 😅

Build requirements are specifically tools that are only necessary at build time, e.g. compilers.

Host requirements are (mostly) libraries you compile against - the code you compile will then depend on the existence of all the (usually shared) library artefacts you used _and their ABI_[^1]. These are the libraries for which we must run migrators that rebuild all dependent packages if a new (ABI-incompatible) version of that host dependency is released. If host deps are not static, then they export a run-requirement to ensure that the required artefacts will be there when the package is installed (there are host dependencies without run-exports, e.g. header-only libraries).

Run dependencies are things you'll just need at runtime (any python package you're just importing), but that you don't need to compile the package.

The distinction between build and host becomes clearer (perhaps; at least it did for me) in the context of cross-compilation. When we're compiling for osx-arm (e.g. apple M1), the build environment runs on a x86_64 architecture (i.e. on our CI agents), while the host (=target) environment will be set up with arm64 packages. Needless to say these cannot be installed together on any single machine (because we cannot mix CPU architectures).

[^1]: application binary interface. If it changes (e.g. a function in C/C++ changes its signature, or an argument changes type), any code compiled against the old ABI would crash (or worse) if it encounters a shared library that has a different ABI. This is why the run-export of libraries encodes an upper version bound.

gschramm:
(1) Ok. This explanation makes it a lot clearer to me. Would it makes sense to add a short paragraph about this in the knowledge base?
(2) Can the PR be merged now?

h-vetinari (approved):
> Can the PR be merged now?

I don't have the commit-bit here, but LGTM

gschramm:
@conda-forge/help-c-cpp Could somebody merge this PR?

recipes/.gitignore
```diff
@@ -0,0 +1,5 @@
+# conda smithy ci-skeleton start
+*.pyc
+
+build_artifacts
+# conda smithy ci-skeleton end
```
BastianZim, line 5: Please remove this and only change things in your recipe.

gschramm: Done.

BastianZim: Thanks. I don't review python-c but let me assign them to have this moved along.

gschramm: Thanks! Do I need to ping the c-cpp team directly, or will they see the assigned label?

gschramm:
@carterbox You could you maybe have a look at this PR or do you know somebody who could review it?

carterbox:
I'll have time this weekend.

carterbox:
If this is a shared library object with headers, you need a run export which pins a compatible version of this package.

gschramm:
> If this is a shared library object with headers, you need a run export which pins a compatible version of this package.

So sth like
```
run_export:
  -{{ pin_compatible('parallelproj', max_pin='x.x' }}
```

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/gschramm/parallelproj/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 826a83e51db6459004d17c030c21fb32e7cb5205a45c28c8f50f87f235321196
+
+build:
+  number: {{ build_number }}
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cpu"   # [cuda_compiler_version == "None"]
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cuda"  # [cuda_compiler_version != "None"]
```
carterbox, line 16:
```suggestion
  string: "cpuh{{ PKG_HASH }}_{{ build_number }}"   # [cuda_compiler_version == "None"]
  string: "cuda{{ cuda_compiler_version|string }}h{{ PKG_HASH }}_{{ build_number }}"  # [cuda_compiler_version != "None"]
```
Convention is to prefix the build string with `cuda + the toolkit version`. The build number should be the last thing in the string. You may use `{{ cuda_compiler_version | replace('.', '') }}` if you don't want the period.

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/gschramm/parallelproj/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 826a83e51db6459004d17c030c21fb32e7cb5205a45c28c8f50f87f235321196
+
+build:
```
carterbox, line 13:
```suggestion
build:
  run_exports:
    - {{ pin_subpackage( name|lower, max_pin='x.x') }}
```
Because this is a shared object with headers, you must include a run_export which pins according to the *ABI guarantees* of your library.

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - {{ compiler('cuda') }}  # [cuda_compiler_version != "None"]
+    - cmake >=3.23
+    - ninja
+  host:
+    - libgomp      # [linux]
+    - llvm-openmp  # [osx]
+test:
+  commands:
+    - test -f $PREFIX/lib/libparallelproj_c$SHLIB_EXT  # [unix]
+    - if not exist %LIBRARY_LIB%\parallelproj_c.lib exit 1  # [win]
```
carterbox, line 31: Typically win libraries have both .lib and .dll.

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - test -f $PREFIX/lib/libparallelproj_cuda$SHLIB_EXT  # [linux and cuda_compiler_version != "None"]
+    - parallelproj_nontof_fwd_back_test  # [not win]
+    - parallelproj_nontof_fwd_back_test.exe  # [win]
```
carterbox, line 34: Do end users need this test binary?

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [cuda_compiler_version != "None"]
+    - cmake >=3.23
+    - ninja
+  host:
+    - libgomp      # [linux]
+    - llvm-openmp  # [osx]
+test:
```
carterbox, line 28: Please also existence test for header and a Cmake config file.

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+build:
+  number: {{ build_number }}
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cpu"   # [cuda_compiler_version == "None"]
+  string: "h{{ PKG_HASH }}_{{ build_number }}_cuda"  # [cuda_compiler_version != "None"]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
carterbox, line 21: Is the CXX compiler actually used? If not please remove it or add it to the `ignore_run_exports_from` section. So that your library doesn't pick up the CXX standard library as a dependency.

EDIT: Looks like it is not used for non-cuda, so do this conditionally.

```suggestion
    - {{ compiler('cxx') }}  # [cuda_compiler_version != "None"]
```

recipes/parallelproj/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/gschramm/parallelproj/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 826a83e51db6459004d17c030c21fb32e7cb5205a45c28c8f50f87f235321196
+
+build:
+  number: {{ build_number }}
```
carterbox, line 14:
```suggestion
  number: {{ build_number }}
  ignore_run_exports:
    - cudatoolkit
```
It looks like [you aren't linking to the cudatoolkit](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=619558&view=logs&j=eb6ecf2b-0363-5c1a-baa7-2d771889d197&t=d8576026-8998-52ff-d027-1d631b0fb837&l=1257), so it's not needed.

recipes/parallelproj/build.sh
```diff
@@ -0,0 +1,22 @@
...
+cmake -G Ninja \
+    ${CMAKE_ARGS} \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DSKIP_DOCS=TRUE \
+    ${EXTRA_CMAKE_ARGS} \
+    ${SRC_DIR}
+
+cmake --build . --target install --verbose
```
carterbox, line 22: Why are you shipping the source file `projector_kernels.cu.1.2.6` in the lib directory?

carterbox (changes requested):

carterbox (changes requested):
Please also complete the checklist in the PR description. The checkboxes are interactive.

recipes/parallelproj/meta.yaml
gschramm: No, end users don't need those test binaries. I use those binaries as "ctest" to check whether certain functions of the lib behave as expected. Is there a better way to include e.g. "ctest -VV" in the test section?
I have "ctest -VV" in the build scripts in the beginning, but not sure if this really belongs there.

recipes/parallelproj/build.sh
gschramm: The reason is because using this file, python users can use cupy's RawKernel to directly execute the kernels on cupy GPU arrays (or also on pytorch tensors).
Is there a better way to ship the file?

recipes/parallelproj/meta.yaml
gschramm: On Windows the .dll gets installed in ```...Library/bin``` instead of  ```.../Library/lib```, what can I use instead of ```%LIBRARY_LIB%``` that gets expanded correctly?

carterbox: Yes. Check for both. `%LIBRARY_BIN%`

https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html

recipes/parallelproj/meta.yaml
carterbox: You can run test binaries during the build phase without installing them. You can run test binaries during the test phase without installing them by listing them as test source files.

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#source-files

recipes/parallelproj/build.sh
carterbox: If the end-use-case is CuPy RawKernel, put this file in python module and load it using `importlib.resources`. [That's what I do.](https://github.com/AdvancedPhotonSource/tike/blob/362463324d904806ab65c4ea425f9835d047f0e3/src/tike/operators/cupy/patch.py#L15) It works a lot better for python users because they don't need to hunt down the file in their system.

Also, the users who would want this source file are probably not the same ones who want the full pre-compiled library?

You could add the CuPy RawKernel Python module as a separate output to this recipe.

recipes/parallelproj/meta.yaml
gschramm: done

recipes/parallelproj/meta.yaml
gschramm: done

recipes/parallelproj/meta.yaml
gschramm: I added the ```source_file``` section, but the tests fail because the binaries are not found. Do you know why?

h-vetinari: Source files really have to be in the source, not created during the build phase

recipes/parallelproj/build.sh
gschramm: I think most python users will want both (the pre-compiled cuda lib and the kernel file) such they can either use ctypes to have an interface for the cuda lib for numpy arrays, or the cupy interface for gpu arrays.

I have a separate repository with a python package that uses the compiled libs. And I thought it would be a good idea to keep them separate, since there will be users that only want the compiled c/cuda libs.

recipes/parallelproj/meta.yaml
gschramm: And is there no way of getting the compiled test executables from the build phase?

recipes/parallelproj/build.sh
h-vetinari: How does the python project use the raw `.cu` file? Is it JIT-compiled somehow? And is there some usecase on the python side that couldn't be handled by correctly loading the precompiled lib?

recipes/parallelproj/meta.yaml
h-vetinari: Yeah you can test during the build phase. Everything that makes it to the test phase will also be installed for endusers

recipes/parallelproj/build.sh
gschramm: It is indeed JIT-compiled by cupy which allows the python end user to execute the kernel directly on cupy GPU arrays. The functions in the precompiled lib, take host arrays as input which are then transfered to the GPU where the kernel is executed. After executing the result is transfered back to the host. Using the cupy RawKernel, we can avoid memory transfer between host and GPU which is useful in iterative algorithms where we need to execute kernels many times.

h-vetinari: Thanks for the explanation. I'd be surprised if this were the first package where this situation arises.

@jakirkham @kkraus14 @jaimergp @isuruf 
Do we have experience/standards for how to distribute raw `.cu` kernels for JIT-compilation?

recipes/parallelproj/meta.yaml
gschramm: ```ctest -VV``` is now run during the build phase and the test executables are not installed anymore.

recipes/parallelproj/build.sh
gschramm: > If the end-use-case is CuPy RawKernel, put this file in python module and load it using `importlib.resources`. [That's what I do.](https://github.com/AdvancedPhotonSource/tike/blob/362463324d904806ab65c4ea425f9835d047f0e3/src/tike/operators/cupy/patch.py#L15) It works a lot better for python users because they don't need to hunt down the file in their system.
> 
> Also, the users who would want this source file are probably not the same ones who want the full pre-compiled library?
> 
> You could add the CuPy RawKernel Python module as a separate output to this recipe.

Putting the .cu kernel file in a python module is one option, but what is users of other high-level languages that support cuda JIT compilation want to use the kernels? For me it seems better to only ship the .cu kernel file independent of the high-level language that wants to use it.

h-vetinari: > For me, it seems better to only ship the .cu kernel file independent of the high-level language.

+1

gschramm: @carterbox @h-vetinari so shipping the .cu kernel file in the current way is ok?

carterbox: > shipping the .cu kernel file in the current way is ok?

Yes, It's just unusual.

> I have a separate repository with a python package that uses the compiled libs. And I thought it would be a good idea to keep them separate, since there will be users that only want the compiled c/cuda libs.

We prefer that you link that project to the libs in this project dynamically instead of static linking. If the APIs share a version number, then you can build them from the same recipe. If not, then keep them separate.

recipes/parallelproj/meta.yaml
carterbox: Folder structure is preserved, so you would have to call the tests this way:

```
  source_files:
    - build/c/tests/parallelproj_nontof_fwd_back_test  # [not win]
  commands:
    - build/c/tests/parallelproj_nontof_fwd_back_test  # [not win]
```
Note in the logs that the error was "parallelproj_nontof_fwd_back_test not found" during the test phase. Not that "build/c/tests/parallelproj_nontof_fwd_back_test not found" in the source.

gschramm: > Folder structure is preserved, so you would have to call the tests this way:
> 
> ```
>   source_files:
>     - build/c/tests/parallelproj_nontof_fwd_back_test  # [not win]
>   commands:
>     - build/c/tests/parallelproj_nontof_fwd_back_test  # [not win]
> ```
> 
> Note in the logs that the error was "parallelproj_nontof_fwd_back_test not found" during the test phase. Not that "build/c/tests/parallelproj_nontof_fwd_back_test not found" in the source.

Ah ok, sorry for having missed that. I corrected that, but now I get the error
```
build/c/tests/parallelproj_nontof_fwd_back_test: error while loading shared libraries: libparallelproj_c.so.1.2.8: cannot open shared object file: No such file or directory
```

But the lib got installed during the build phase:
```
-- Installing: $PREFIX/lib/libparallelproj_c.so.1.2.8
```

What am I missing?

recipes/parallelproj/build.sh
gschramm: The version numbers are different (I expect that the python version numbers will increase much quicker).
That is because the python package offers an interface to the functions defined in the libs, but also more things.
So we should keep them separated.

gschramm:
@jaimergp @carterbox @h-vetinari Are we ready to merge the PR now? As I wrote in the comments above, I am in favor of exectuing the test executables during build phase (as many other recipies on conda-forge do) and to ship the cuda kernel file the it is implemented right now.

gschramm:

gschramm:

recipes/parallelproj/meta.yaml
carterbox: I think sometimes test binaries aren't relocatable, so when you move them out of the same folder as the libraries they link to, they break. I think there's a way to ask conda-build to fix the breaks, but I don't know what it is without looking at the docs.

PR #21048, opened 2022-11-07
Add Nebari

viniciusdc:
Checklist
- [x] Build number is 0.


viniciusdc:

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - typer
+    - questionary
+
+test:
+  imports:
+    - nebari
+    - nebari.cli
+  commands:
+    - pip check
+    - nebari --help
```
viniciusdc, line 51:
```suggestion
    - nebari --version
```

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/nebari-{{ version }}.tar.gz
+  sha256: 02f59773c5b5035df53d156af3cf648bc88ce9384fc39ffa80291c8514d93980
+
+build:
+  number: 0
+  noarch: python
```
viniciusdc, line 15:
```suggestion
  noarch: python
  skip: true # [win]
```

viniciusdc:
@jaimergp could you have a look at this? I would like to know if this follows what was expected of the new noarch customizations builds... I think I did something wrong there

iameskild:
I agree to be a maintainer of this project 👍 

cc @viniciusdc

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - nebari --help
+    - if [[ $(nebari --version) != {{ version }} ]]; then exit 1; else exit 0; fi
+  requires:
+    - pip
+
+about:
+  home: https://github.com/nebari-dev/nebari
+  summary: A Jupyter and Dask-powered open source data science platform.
+  dev_url: https://github.com/nebari-dev/nebari
+  license: BSD-3-Clause
```
ndevenish, line 61:
```suggestion
  license: BSD-3-Clause
  license_family: BSD
```

ndevenish:
The nebari package on pypi includes the `LICENSE` file, so you don't need to include it in this feedstock (conda-build will pick it up from the package)

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - nebari = nebari.cli.main:app
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python >=3.8,<4.0
+    - pip
+    - setuptools_scm
+  run:
+    - __unix
```
ndevenish, line 26: I haven't seen this noarch-package restriction used before. What does this do on it's own? The page you linked to seems to suggest that it must have a selector e.g.
```
- __unix # [unix]
```
and that it is for using when you have an architecture-specific dependency. If it's just that one of the dependencies isn't available on windows - the old way was to just leave it in and so if that dependency ever did become available on windows, then it'd automatically qualify (since noarch is by default built on linux only, it would be fine).

If it explicitly doesn't work on windows, then maybe this is the approach to use...

BastianZim: It's from here: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html
Basically, it allows the package only to be installed in an environment where this is available and can be used if a package only works on Linux, for example. Keeps it noarch and reduces our resources.

I've been using this for at least a couple of months in a couple of packages, and it works really well! 

One example: https://github.com/conda-forge/smac-feedstock/blob/048bfd759b09f78f0d60f3c952e49261b48bf629/recipe/meta.yaml#L37

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - questionary
+
+test:
+  imports:
+    - nebari
+    - nebari.cli
+  commands:
+    - pip check
+    - nebari --help
+    - if [[ $(nebari --version) != {{ version }} ]]; then exit 1; else exit 0; fi
```
BastianZim, line 53: Nice! I'm wondering if it wouldn't make sense to include that in Grayskull, because `setuptools_scm`etc. is sometimes not included which this would catch.

recipes/nebari/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  noarch: python
+  entry_points:
+    - nebari = nebari.cli.main:app
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python >=3.8,<4.0
+    - pip
+    - setuptools_scm
```
goanpeca, line 24:
```suggestion
    - setuptools-scm
```

recipes/nebari/meta.yaml
dopplershift: Yes, my understanding is that you don't need (or even want for noarch) the selector.

BastianZim: Yes, the selector shouldn't be used. Otherwise, the selector will remove the virtual package on windows, making it available again on all platforms.

viniciusdc: Hi @BastianZim @dopplershift @ndevenish, thanks for the comments. Though I ended up being lost in conclusion... Regarding the use of `__unix`, based on the linked page and @BastianZim  comment above, it should work as a virtual dependency and restrict the package from being built on win, right?. Wich will not affect the package structure (maintaining its `noarch`)

On the other hand, using the `skip: True  # [win]` selectors are not advised as they are incompatible with `noarch`, and this is package is noarch. The only exception is that it does not support windows (it can be installed but can't be used...) 

..so I ended up losing track of which "selector" you all were referring to in this thread 😅

recipes/nebari/meta.yaml
viniciusdc: Good idea, that could be an extra custom setting that the Grayskull config could load -- e.g custom test class

recipes/nebari/meta.yaml
BastianZim: We really should add this to the docs, it is indeed becoming slightly confusing... 😄 
I'm just writing this down here once so that I can link later to it in case I need but maybe it's also helpful here.

There are two cases that are relevant here in which `arch` packages can be converted into `noarch` packages, both rely on virtual packages.

The first one is if there are different dependencies for different operating systems. In that case, https://github.com/conda-forge/conda-forge.github.io/pull/1839 can be used.

Let's assume that we have a package X that has a dependency Y but only on windows. In this example, OSX doesn't exist so I won't use Unix.

First, we instruct the CI to build the package on multiple different platforms, Linux and windows in this case (Through conda-forge.yml).
Then, we add the dependency using the `# [win]` [preprocessing](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors) selector. This makes sure that the dependency is only added when we are building on windows. 

```yml
requirements:
  run:
    - Y  # [win]
```

Now, we have two packages, one built on Linux with no extra dependency and one built on Windows with the extra dependency.
Finally, we need to make sure that a Linux user cannot accidentally install the package that has the extra windows dependency. 
For that, we include two [virtual packages](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html): `__linux` and `__win` in the run requirements. 
Virtual packages are only available on the platform they specify. Meaning if a package has a dependency on `__win`, the solver will check during the installation of the package if the solver is currently working on a windows system and only allows the installation to proceed if it really is on windows.
Since we can't have both in the same package, because then it couldn't be installed at all, we also add preprocessing selectors to both of them:
```yml
- __linux. # [linux]
- __win  # [win]
```
This, way we will have two packages. The first package is built on Linux and therefore has the `__linux` package and nothing else. The second one is built on Windows and therefore has the `__win` virtual package and the extra dependency.
When the end-user now installs the package, the solver will check that it gets the package that satisfies the virtual package, meaning the windows version on windows and the Linux version on Linux.
_Note:_ There are other virtual packages as well (Unix, OSX) but the approach is the same.

The meta.yaml is, therefore:

```yml
requirements:
  run:
    - Y  # [win]
    - __linux # [linux]
    - __win  # [win]
```
---
The second case is if we have a package that can only be used on a certain OS. This is commonly the case because it relies on system features that are not available on all OSs.

So far, this was done by using `skip: True` and the respective preprocessing selector: `# [win]`. That would mean that the package is only available on Linux and OSX.

Using the fact that virtual packages only allow the installation of a package if they exist, we can change this to a `noarch` solution through two steps.

First, the skip part needs to be removed and `noarch: python` added so that it is built as a noarch package.
Then, the virtual package is added to the run requirements.
This way, the package can only be installed on the OS that the virtual package allows. Here, the preprocessing selector shouldn't be used, otherwise, the virtual package will be removed again if it is built on a different platform, which we are trying to prevent.

```yml
requirements:
  run:
    - __linux
```

recipes/nebari/meta.yaml
BastianZim: True, that works as well, but I just meant that Grayskull should add that to the recipes it generates so that the CI can catch this. :)

recipes/nebari/meta.yaml
viniciusdc: That's wonderful docs here!! Thanks, @BastianZim. It could be added to the open docs @jaimergp created as supplementary information or a TLDR. What do you all think?

viniciusdc:
@conda-forge/core, this is ready for review. Please keep the above conversation in mind during the review

recipes/nebari/meta.yaml
BastianZim: My pleasure! :) Either one is fine for me. I can also wait until the other PR is merged and then add the info in a new PR.

iameskild:
@conda-forge/core this is ready for a final review. Thank you :)

cc @viniciusdc

BastianZim:
Do you need core approval? Otherwise, I'm happy to merge.

iameskild:
@BastianZim @viniciusdc since you are both more familiar with the processes and workings of conda-forge, I'll let you decide we should proceed :)

viniciusdc:
> Do you need core approval? Otherwise, I'm happy to merge.

Hi @BastianZim, if you have merge rights I would really appreciate if you can merge :smile:. thanks in advance. The CI tests for Linux are passing (wich is the target) and there are no problems with the recipe, thus I would say this is ready to go

recipes/nebari/LICENSE
```diff
@@ -0,0 +1,29 @@
+BSD 3-Clause License
```
BastianZim, line 1: This can be removed, it's already included in the source dist.

BastianZim (changes requested):
One small thing and I'll merge. :)

recipes/nebari/LICENSE
viniciusdc: Hi @BastianZim, thanks, I just removed the License file.

PR #21050, opened 2022-11-07
Add misc Qt6 modules

jschueller:
cc @conda-forge/qt-main


jschueller:
cc @hmaarrfk

hmaarrfk:
I would like to leave `6` out of the feedstock name of possible. That way we can have some kind of continuity with other qt packages

jschueller:
Is renaming the `recipe/qt6-*` folders enough ?

jschueller:
ok done using "feedstock-name: qt-xx"

PR #21056, opened 2022-11-08
feat: introduce cocotb support to conda-forge

tianrui-wei:
The recipe is generated with grayskull

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/cocotb/meta.yaml
+++ b/recipes/cocotb/meta.yaml
@@ -1,0 +1,51 @@
+{% set name = "cocotb" %}
+{% set version = "1.7.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cocotb-{{ version }}.tar.gz
+  sha256: 1f390bb6468d70e04acebeb9ff05b54be42cbbbbbb3bd6c75d1ec0a36cb97e23
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  missing_dso_whitelist:
+    - '*/libgcc_s.so.1'
+    - '*/libstdc++.so.6'
+  skip:
+    - true  # [win or osx]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - setuptools
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - cocotb
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://www.cocotb.org
+  summary: cocotb is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
+  license: BSD-4-Clause & BSD-3-Clause
+  license_file:
+    - LICENSE
+    - cocotb/_vendor/tcl/license.terms
+
+extra:
+  recipe-maintainers:
+    - tianrui-wei


tianrui-wei:

imphil:
Hi. Philipp from the cocotb project here.

Thanks for helping to make cocotb better. Can you explain what the benefits are of adding cocotb to conda-forge? I'm asking because the majority of support questions we get on the cocotb side are installation-related, and we are working hard to make our installer (as it appears on PyPi) robust and well-tested. Adding another installation variant to the mix is likely to increase our support burden significantly -- hence a closer look at the benefits is warranted. 

(Background: cocotb is a bit a "special" Python package, as it is loaded into a Python (libpython) embedded into a (typically closed-source) VHDL/Verilog simulator. Getting all the linking and library discovery right is a major challenge, and the end result is still rather fragile.)

tianrui-wei:
Hi Philipp,

Thanks for the reply. The reason I want to add cocotb to conda forge is to enable cocotb in a more reproducible build system like conda. Getting certain library and python versions with pip is still painful even with pip.

Regarding the maintenance burden, I'm happy to fix issues that's conda related should they arrive :)

Best regards,
Tianrui

imphil:
> The reason I want to add cocotb to conda forge is to enable cocotb in a more reproducible build system like conda. Getting certain library and python versions with pip is still painful even with pip.

That sounds interesting -- can you elaborate on that a bit more? Which kind of reproducibility is conda adding? Is it reproducing binaries from sources?

cocotb ships (since the 1.7 release) binary wheels and tests them in CI against a fair amount of simulators, so we know pretty well which builds users are running.

tianrui-wei:
@conda-forge-admin, please ping conda-forge/staged-recipe

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  requires:
+    - pip
+
+about:
+  home: https://www.cocotb.org
+  summary: cocotb is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
+  license: BSD-4-Clause & BSD-3-Clause
+  license_file:
+    - LICENSE
+    - cocotb/_vendor/tcl/license.terms
```
carterbox, line 47: This is incomplete. You need to include the license information for ALL of the revendored source code.

https://github.com/cocotb/cocotb/tree/master/cocotb/_vendor

carterbox:
Yes, conda is a general package manager. PYPI was not intended to distribute non-python packages, whereas conda supports package management for libraries/binaries that are non-python.

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - setuptools
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
```
carterbox, line 31: Docs say that this package [requires a simulator](https://docs.cocotb.org/en/stable/install.html). Where are the simulators?

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cocotb-{{ version }}.tar.gz
+  sha256: 1f390bb6468d70e04acebeb9ff05b54be42cbbbbbb3bd6c75d1ec0a36cb97e23
+
+build:
+  noarch: python
```
carterbox, line 13: This package is not noarch. [It contains compiled code.](https://github.com/cocotb/cocotb/blob/v1.7.2/setup.py#L90)

```suggestion
```

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - cocotb
+  commands:
```
carterbox, line 36: Please tests any CLIs that this package has.

https://github.com/cocotb/cocotb/blob/d2b8dab8f5baffbc6cc53a6c44068011e5532089/setup.py#L116

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cocotb-{{ version }}.tar.gz
+  sha256: 1f390bb6468d70e04acebeb9ff05b54be42cbbbbbb3bd6c75d1ec0a36cb97e23
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  missing_dso_whitelist:
+    - '*/libgcc_s.so.1'
+    - '*/libstdc++.so.6'
```
carterbox, line 18: This package contains compiled C code. Why would you whitelist missing C libraries?

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  skip:
+    - true  # [win or osx]
+
+requirements:
```
carterbox, line 22: find_libpython is also a dependency?

https://github.com/cocotb/cocotb/blob/d2b8dab8f5baffbc6cc53a6c44068011e5532089/setup.py#L101

carterbox (changes requested):

recipes/cocotb/meta.yaml
timkpaine: iverilog is already available via conda forge

timkpaine:
@carterbox / @tianrui-wei ive resolved all (?) except the license comments on my PR, will fix the licenses shortly (need to find how the IEEE LRM code is licensed)

tianrui-wei:
Hi @timkpaine, sorry for the delay in response and thanks for fixing the outstanding issues. One reason I kind of gave up on this PR is that it contains source code licensed by Siemens Mentor, and it's not clear to me whether it's compatible with open source licenses that conda forge accepts. Thanks for the effort anyway :)

PR #21086, opened 2022-11-10
Add `openff-models` recipe

mattwthompson:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/openff-models/meta.yaml
+++ b/recipes/openff-models/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "openff-models" %}
+{% set version = "0.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/openforcefield/openff-models/archive/{{ version }}.tar.gz
+  sha256: 38b59e5aefceb3f56df8f63248c0451f6904d310c6195d369678011fe42ae157
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python >=3.8
+    - numpy
+    - pydantic <2.0.0a0
+    - openff-units
+    - openff-utilities
+
+test:
+  requires:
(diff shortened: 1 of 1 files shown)


mattwthompson:

recipes/openff-models/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  run:
+    - python >=3.8
+    - numpy
+    - pydantic <2.0.0a0
+    - openff-units
+    - openff-utilities
+
+test:
+  requires:
+    - setuptools
```
synapticarbors, line 30: Why is `setuptools` listed as a requirement for tests?

recipes/openff-models/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
```
synapticarbors, line 21: Do you also need `openmm`?
https://github.com/openforcefield/openff-models/blob/main/openff/models/types.py#L9

The upstream package doesn't actually list it's `install_requirements` and the `__init__.py` in `openff.models` is basically a no-op, so it seems like other stuff could be missing.

recipes/openff-models/meta.yaml
mattwthompson: Not sure; probably not needed

recipes/openff-models/meta.yaml
mattwthompson: It should just be OpenMM; we track dependencies in our test environments (and subtract out those only needed for tests) since with feedstocks dependencies declared in the main repo is duplicate information and easy to fall out of date

install_requirements

mattwthompson: And thanks for catching the empty `__Init__.py` - we use a cookiecutter but need to make some modifications because we use namespace packages, and somehow the logic for setting `__version__` was not copied over correctly.

mattwthompson:
Thanks!

PR #21087, opened 2022-11-10
Add docbook-xsl-1.79.2

hattne:
Checklist
- [X] Build number is 0.


hattne:
@conda-forge/staged-recipes: ready for review.

hattne:
@conda-forge/staged-recipes: ping (ready for review).

hmaarrfk (changes requested):
The installation and uninstallation seems to be getting done in the postlink and preunlink scripts. can this be avoided?

hattne:
> Even given your explanation, There seems to be references to a webpage: `"http://cdn.docbook.org/release/xsl/"` in the post scripts. I do not think we should be referencing external websites during post scripts unless we are absolutely required to.

This is how the stylesheets are identified: the XSLT processor (_e.g._ `xsltproc`) is asked to apply a stylesheet (_e.g._ `http://cdn.docbook.org/release/xsl/current/html/docbook.xsl`) to some XML files. The processor then uses the catalog to map the URI to the locally installed stylesheets; there will be no need for the processor to actually visit the external site. Other stylsheets will be idintified by different URIs; they cannot be arbitrarily changed or removed.

hattne:
> To me, it is unclear that the desired recipe will have the intended effect. I say this because the man page references an environment variable: `XML_CATALOG_FILES` which seems to be undefined on my conda installation. I have libxml

Yes, the compile-time default (`/etc/xml/catalog`) will not work with conda.  In my testing, I was working around that by setting `XML_CATALOG_FILES` while building; not pretty, I admit.

> is there a way to list the entries in the catalog? Perhaps through `xmlcatalog dump` ? we could add that as a test.

Excellent idea! I'll try to fix that.

> It may be that you should add activate/deactivate scripts to `libxml` to add and remove the XML_CATALOG_FILE respectively, and likely to create the default catalog file for you.

I had not thought of that!  Much nicer than setting it up each time it is needed. I'll submit a PR to `libxslt` tomorrow.

I think catalog-editing should stay in prelink and postunlink, because the catalog only changes when _this_ package (or similar packages providing other stylesheets) is installed/removed.

hmaarrfk:
> I think catalog-editing should stay in prelink and postunlink, because the catalog only changes when this package (or similar packages providing other stylesheets) is installed/removed.

I understand. It might be that you can add "multiple catalogs" but i think that with a comment that repeats:
https://github.com/conda-forge/staged-recipes/pull/21087#issuecomment-1358745916

in the post/pre scripts it is more acceptable.

hattne:
OK, done.

* Submitted [PR against `libxslt`](https://github.com/conda-forge/libxslt-feedstock/pull/36).  If and when that PR is merged, I'll revise this recipe to avoid setting the `XML_CATALOG_FILES` variable.
* Commented on the URI hardcoding
* Checking that the catalog has the relevant entry in the test.  This required moving the Windows tests to `run_tests.bat`; I moved the Unix tests for consistency.

hmaarrfk:
Ocefpaf is quite active. It will be good to get his opinion on this. Let's give him 7 days?

hattne:
> Ocefpaf is quite active. It will be good to get his opinion on this. Let's give him 7 days?

Certainly, no worries!

hattne:
https://github.com/conda-forge/libxml2-feedstock/pull/91 was merged and it looks like it has not caused any trouble for a week. There is no need to set environment variables in the tests any longer—that is now handled by (de)activate scripts in `libxml2`.

@hmaarrfk: unless you have other concerns, I think this is finally good to go.

recipes/docbook-xsl/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+requirements:
+  run:
+    # xmlcatalog is required by post-link.{bat,sh} and
+    # pre-unlink.{bat,sh}.
+    - libxml2
+
+test:
+  requires:
+    # Really need 2.10.3 build 6 or later.
+    - libxml2
```
hmaarrfk, line 26: hmm this isn't really required here, but lets go with it.

hattne: The idea really was to remember to depend on 2.10.4 or some such whenever that becomes available.

hmaarrfk: you will want to make it a run constraint to be honest. Otherwise things like the migrator will have trouble.

hattne: I see; I'll make a note fix this on the next update.

PR #21103, opened 2022-11-10
Adding "inflation" package

ecboghiu:
Python implementations of the Inflation Technique for causal inference. Associated paper: https://arxiv.org/abs/2211.04483

I'm the proposed maintainer of this package. I think it is ready for review.

### DIFF
--- a/recipes/inflation/meta.yaml
+++ b/recipes/inflation/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - scipy
     - numba
     - tqdm
+    - mosek
 
 test:
   imports:
--- a/recipes/inflation/meta.yaml
+++ b/recipes/inflation/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "inflation" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ecboghiu/inflation/archive/v{{ version }}.tar.gz
+  sha256: fcc471f8477a732f40e2034faa200f8a70675880d563aa9af9af06e317c353f0
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - numpy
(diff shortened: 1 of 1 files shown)


ecboghiu:
@conda-forge-admin, please ping team @conda-forge/help-python

ecboghiu:
Could I get some help with the failed checks?

Locally I can install the package with no problems:

`pip install git+https://github.com/ecboghiu/inflation@main`

To create the recipe I followed the steps detailed [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html):

1. install grayskull
2. generate recipe with grayskull pypi https://github.com/ecboghiu/inflation, and modify it slightly to add a longer description
3. fork the staged-recipes on repository
4. create a branch and add the recipe to it inside a folder called after my package inside `staged-recipes/recipes`
5. create the pull request

The modification that I did, that is different w.r.t the GitHub repository for "inflation", is that I removed "mosek" as a dependency, as it is not available on conda-forge and I thought it would be safer this way.

ecboghiu:
The problem is that when importing the package, it is reporting that "mosek" is missing. Mosek can be installed from the "mosek" channel via 

`conda install -c mosek mosek`

I will try to check if one can add as a requirement a package from a different channel.

BastianZim:
You cannot add a package from a different channel, but you can add the package as well. Either in a different PR or, the faster way, in this PR.

BastianZim:
> Would simply removing it from the required_installs list in the repository be enough for this problem?

Yes. You can just add it as an optional dependency there and list it here not under `run` but `run_constrained`: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained

> It is proprietary, but available on Anaconda

If it is proprietary and the license doesn't allow redistribution, then no, it is impossible to add it here. We can only accept packages that are actually built here. There are exceptions to this rule but they all require a license that allows redistribution. 

What I would recommend is:

1. Add the package under `run_constrained`
2. Inform your users to install the package like this: `conda install inflation mosek --channel conda-forge --channel mosek` 

It is not full best practice, because we recommend against mixing channels but it seems like this is the only way you can accomplish this. By using `run_constrained`, the solver will make sure to only install the versions of mosek that you specify.

PR #21112, opened 2022-11-11
CI: set exit_code=1 in create_feedstocks to notice failures

mbargull:
Checklist
- [ ] Build number is 0.

We recently had a couple of feedstocks (`r-adiv`, `throttle`) for which empty feedstocks where created.
Only after it was reported (core chat, collaboration chat) it was corrected (empty feedstock repos deleted to reinit the process).

Would it create too much noise to mark the jobs as failed? In gh-13095 it was explicitly turned off by @beckermr, presumably to reduce noise.
If yes, would/should we need to add some logic to only indicate a failure after a certain number of runs/time (e.g., we could check if the feedstock has existed for a couple of hours/days).

cc @conda-forge/core, @johanneskoester

### DIFF
--- a/.github/workflows/scripts/create_feedstocks.py
+++ b/.github/workflows/scripts/create_feedstocks.py
@@ -348,7 +348,7 @@ if __name__ == '__main__':
                 subprocess.check_call(
                     ['conda', 'smithy', 'rerender'], cwd=feedstock_dir)
             except subprocess.CalledProcessError:
-                exit_code = 0
+                exit_code = 1
                 traceback.print_exception(*sys.exc_info())
                 continue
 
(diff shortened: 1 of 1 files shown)


beckermr (changes requested):
Travis fails all the time and usually ends up succeeding on retries. Unless someone (you?) is going to actively monitor the failing status and do something about it, setting this to failing is more noise than helpful IMHO. Right now, users notice the failures at the feedstock level and find us as needed which is fine.

beckermr:
A better Pr would be one that deletes the feedstock if it is empty before trying to make it again.

mbargull:
> Unless someone (you?) is going to actively monitor the failing status and do something about it, setting this to failing is more noise than helpful IMHO.

Hehe, yeah, I (intentionally) don't get notified for failures here.
But I also don't know who gets notifications, hence why marking this as draft and asking about potential noise.
Thanks for your assessment!

----
> A better Pr would be one that deletes the feedstock if it is empty before trying to make it again.

Makes total sense. There's really no need to force a manual intervention for something we could automate.
Maybe I'll look into that later; I'll leave this draft open for now as a reminder.

PR #21154, opened 2022-11-14
Add kmedoids python package

kno10:
@conda-forge-admin, please ping conda-forge/help-python, ready for review!

A Python package (with Rust backend), already on https://pypi.org/project/kmedoids/
Documentation: https://python-kmedoids.readthedocs.io/

Conda packaging was requested in kno10/python-kmedoids#19

### DIFF
--- a/recipes/kmedoids/meta.yaml
+++ b/recipes/kmedoids/meta.yaml
@@ -1,0 +1,64 @@
+{% set version = "0.4.1" %}
+
+package:
+  name: kmedoids
+  version: {{ version }}
+
+source:
+  url: https://github.com/kno10/python-kmedoids/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d334848ecc0d4b17c0e041b975f29b1963cd64e7c3743d9b9ec157ae5780e2c2
+
+build:
+  number: 0
+  script:
+    - export CARGO_NET_GIT_FETCH_WITH_CLI=true
+    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+    - {{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv
+
+requirements:
+  build:
+    - {{ compiler('c') }} # used by Rust
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  host:
+    - python
+    - pip
+    - maturin
+    - numpy
+  run:
+    - python
+    - numpy
+
+test:
+  imports:
+    - kmedoids
+    - kmedoids.kmedoids
+  requires:
+    - pip
+    - pytest
+  source_files:
+    - tests
+  commands:
+    - pip check
+    - cd tests && python -m unittest discover
+
+about:
+  home: https://github.com/kno10/python-kmedoids
+  summary: 'k-Medoids Clustering in Python with FasterPAM'
+  description:
+    This python package implements k-medoids clustering with PAM and
+    variants of clustering by direct optimization of the (Medoid) Silhouette.
+    It can be used with arbitrary dissimilarites, as it only requires
+    a dissimilarity matrix as input. A sklearn compatible API is included.
+  license: GPL-3.0-or-later
+  license_family: GPL
+  license_file:
+    - THIRDPARTY.yml
+    - LICENSE
+  doc_url: https://python-kmedoids.readthedocs.io/
+  dev_url: https://github.com/kno10/python-kmedoids
+
+extra:
+  recipe-maintainers:
+    - kno10
+    - lale1009


lale1009:
Hi!

Looks good! Glad to maintain the package.

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+test:
+  imports:
+    - kmedoids
+    - kmedoids.kmedoids
+  requires:
+    - python
+    - pip
+    - pytest
+    - numpy
```
hmaarrfk, line 42:
```suggestion
    - pip
    - pytest
```

numpy and python are dependencies through the `run` section, no need to relist them,

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - export CARGO_NET_GIT_FETCH_WITH_CLI=true
+    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+    - {{ PYTHON }} -m pip install . --no-deps --ignore-installed -vv
+
+requirements:
+  build:
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+    - numpy
+  host:
```
hmaarrfk, line 25:
```suggestion
  host:
    - numpy
```

Typically numpy is a host dependency, not a build dependency. If this doesn't work, you should really explain why it is critical to have numpy in the build section requirement (sometimes this is required for cross compilation).

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "kmedoids" %}
+{% set version = "0.4.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/kno10/python-kmedoids/archive/refs/tags/v{{ version }}.tar.gz
+  #  `openssl sha256 <file name>`.
```
hmaarrfk, line 10:
```suggestion
```

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "kmedoids" %}
+{% set version = "0.4.1" %}
+
+package:
+  name: {{ name|lower }}
```
hmaarrfk, line 5:
```suggestion
{% set version = "0.4.1" %}

package:
  name: kmedoids
```
This variable really adds a lot of indirection. Its mostly there to help pull the package more easily from pypi, but honestly, i like to just hardcode the package name most places since the "pattern" is really not that useful to read.

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - cargo-bundle-licenses
+    - numpy
+  host:
+    - python
+    - pip
+    - maturin
+  run:
+    - python
+    - numpy
+    - libgcc-ng  # [linux]
```
hmaarrfk, line 32: it is really strange that you need this particular line. Are you missing a dependency on the c compiler?

recipes/kmedoids/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  source_files:
+    - tests
+  commands:
+    - pip check
+    - cd tests && python -m unittest discover
+
+about:
+  home: https://github.com/kno10/python-kmedoids
+  summary: 'k-Medoids Clustering in Python with FasterPAM'
+  description: 'This python package implements k-medoids clustering with PAM and variants of clustering by direct optimization of the (Medoid) Silhouette. It can be used with arbitrary dissimilarites, as it requires a dissimilarity matrix as input.'
```
hmaarrfk, line 52: Can you follow the example and wrap the lines in a more readable fashion?

recipes/kmedoids/meta.yaml
kno10: It failed to build in the docker container without this.
IMHO this should be inferred from the rust compiler dependency.
I have just pushed without this line, so you can see the build logs and maybe you have a better solution.

kno10: You can see the error at
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=615973&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823

kno10:
Pushed with 'libgcc-ng' back again, because it would not build otherwise. The other changes are incorporated.

recipes/kmedoids/meta.yaml
hmaarrfk: cc: @conda-forge/help-rust <-- hmm this doesn't seem to exit.

cc: @conda-forge/help-c-cpp any ideas?

xhochy: This (the rust build process) probably calls at one stage to a C compiler. As it is transitive, it builds but we need to specify the dependency explicitly for the `run_exports` to work.

kno10: So is it fine as it is, or am I supposed to pull in the C compiler/dependency using
```
- {{ compiler('c') }}
- {{ compiler('rust') }}
```
which seems more portable?

hmaarrfk: I would opt for:
```
- {{ compiler('c') }}
- {{ compiler('rust') }}
```
since it will keep the correct dependency exported.

Without it, the pinning may be to an inconsistent version.

kno10: The version with `- {{ compiler('c') }}` seems to build, and now no longer needs the `#  [linux]` extra dependency. I flag this as resolved.

kno10:
Pushed with `- {{ compiler('c') }}` instead of the linux-only `libgcc-ng` dependency. All requests resolved - ready to go?

xhochy (approved):

ThomasElliot74:
Hi,

On the package page it seems to support Linux and macOS only:

image

The `pip` version works out of the box on Windows.

kno10:
@ThomasElliot74 the windows build daemon just hadn't finished yet, I guess. It shows windows for me now, too.
https://anaconda.org/conda-forge/kmedoids

PR #21208, opened 2022-11-19
Add uutils coreutils

sshockwave:
Checklist
- [ x License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/5eddbd7fc9d1502169089da06c3688d9759be978/recipes/example/meta.yaml#L64-L73) for an example).
- [x] Build number is 0.

### DIFF
--- a/recipes/uutils/meta.yaml
+++ b/recipes/uutils/meta.yaml
@@ -15,8 +15,8 @@ source:
 
 build:
   number: 0
-  ignore_run_exports:
-    - libcxx-ng
+  ignore_run_exports_from:
+    - {{ compiler('cxx') }}
 
 requirements:
   build:
--- a/recipes/uutils/meta.yaml
+++ b/recipes/uutils/meta.yaml
@@ -26,6 +26,9 @@ requirements:
     - {{ compiler('rust') }}
     - make
     - cargo-bundle-licenses
+  run_constrained:
+    - coreutils ==999999999
+    - gnu-coreutils ==999999999
 
 test:
   commands:
--- a/recipes/uutils/meta.yaml
+++ b/recipes/uutils/meta.yaml
@@ -17,13 +17,11 @@ source:
 
 build:
   script:
-    - echo "${PREFIX}" | sed -e 's|/|=|g'
     - CARGO_HOME="${BUILD_PREFIX}"/cargo  # [not win]
     - set CARGO_HOME="%BUILD_PREFIX%"\cargo  # [win]
     - cargo build --release --features {{ feature_set }}
     - cargo install --path . --root "${PREFIX}" --features {{ feature_set }}
     - make PROFILE=Release PREFIX="${PREFIX}" MULTICALL=y install
-    - ls -l "${PREFIX}"/bin
     - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
   number: 0
 
--- a/recipes/uutils/meta.yaml
+++ b/recipes/uutils/meta.yaml
@@ -1,10 +1,6 @@
 {% set name = "uutils" %}
 {% set version = "0.0.16" %}
 
-{% set feature_set = "macos" %}  # [osx]
-{% set feature_set = "unix" %}  # [unix and not osx]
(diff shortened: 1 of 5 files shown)


sshockwave:
Ready for review!

recipes/uutils/meta.yaml
```diff
@@ -0,0 +1,56 @@
+{% set name = "uutils" %}
```
carterbox, line 1: Shouldn't the name of this package be `uutils-coreutils`? We already have a `GNU-coreutils` package. This is a replacement for that package, right? If the have the same API/named outputs, then you also need to make these packages mutually exclusive using `run_constrained` so they don't squash each other's artifacts.

https://github.com/conda-forge/coreutils-feedstock
https://conda-forge.org/feedstock-outputs/

recipes/uutils/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - make
+    - cargo-bundle-licenses
+
+test:
+  commands:
+    - coreutils --help
+    - test -f "${PREFIX}"/bin/coreutils  # [not win]
+    - test -f "${PREFIX}"/bin/cat  # [not win]
+    - if not exist %LIBRARY_BIN%\coreutils.exe exit 1  # [win]
+    - if not exist %LIBRARY_BIN%\cat.exe exit 1  # [win]
```
carterbox, line 34:
```suggestion
    - coreutils --help
    - cat --help
```
This should be sufficient.

recipes/uutils/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  sha256: 1bdc2838e34b6ca8275cb8e3bac5074ca59158ea077e3195fb91ec0cc6b594f6
+  patches:
+    - 0001-FIX_BUILDDIR_IN_MAKE.patch  # [not win]
+    # Fixes long command and quotes problems
+    - 0002-SHORTEN_SH_COMMANDS.patch
+
+build:
+  number: 0
+  ignore_run_exports:
+    - libcxx-ng
```
carterbox, line 19:
```suggestion
  ignore_run_exports_from:
    - {{ compiler('cxx') }}
```
The cxx standard library is also not used.

carterbox (changes requested):

carterbox:
Thanks for gettting the licensing right!

recipes/uutils/meta.yaml
sshockwave: Adding the `-coreutils` suffix would be a nice hint of clobbering artifacts. Thanks for the careful review!

sshockwave:
@carterbox I might need some help here. The package needs cxx compiler to compile. It also depends on `libgcc`, but not `libcxx`. Is it ok to leave `libcxx` as an unused dependency?

carterbox:
No. Then add more packages to `ignore_run_exports` appropriately to avoid dependencies on the unused libs.

sshockwave:
@conda-forge/help-c-cpp ready for review!

recipes/uutils/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - libstdcxx-ng  # [linux]
+  ignore_run_exports_from:
+    - {{ compiler('cxx') }}  # [not linux]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('rust') }}
+    - make
+    - cargo-bundle-licenses
```
carterbox, line 28:
```suggestion
requirements:
  build:
    - {{ compiler('cxx') }}
    - {{ compiler('rust') }}
    - make
    - cargo-bundle-licenses
  run_constrained:
    - coreutils ==999999999
    - gnu-coreutils ==999999999
```
I assume that your package clobbers files from these packages, so they should never be installed into the same environment.

carterbox (changes requested):

recipes/uutils/meta.yaml
sshockwave: Thank you! I didn't realize that `run_constrained`needs to be added in this `meta.yaml`.

PR #21215, opened 2022-11-20
Add zimg

sshockwave:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/zimg/meta.yaml
+++ b/recipes/zimg/meta.yaml
@@ -10,7 +10,6 @@ source:
   sha256: 219d1bc6b7fde1355d72c9b406ebd730a4aed9c21da779660f0a4c851243e32f
 
 build:
-  skip: True  # [win]
   number: 0
   run_exports:
     - {{ pin_subpackage('zimg', min_pin='x.x', max_pin='x.x') }}
@@ -19,8 +18,10 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
-    - automake
-    - libtool
+    - automake  # [unix]
+    - m2-automake1.15  # [win]
+    - libtool  # [unix]
+    - m2-libtool  # [win]
     - make
 
 test:
--- a/recipes/zimg/meta.yaml
+++ b/recipes/zimg/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "zimg" %}
+{% set version = "3.0.4" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/sekrit-twc/zimg/archive/release-{{ version }}.tar.gz
+  sha256: 219d1bc6b7fde1355d72c9b406ebd730a4aed9c21da779660f0a4c851243e32f
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('zimg', min_pin='x.x', max_pin='x.x') }}
+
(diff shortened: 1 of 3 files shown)


sshockwave:
Ready to go!

recipes/zimg/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - make
+
+test:
+  commands:
+    - test -f ${PREFIX}/lib/libzimg${SHLIB_EXT}  # [not win]
+    - if not exist %PREFIX%\bin\zimg.dll exit 1  # [win]
+
+    # CFEP 18
+    - test ! -f ${PREFIX}/lib/libzimg.a  # [not win]
+    - if exist %PREFIX%\lib\zimg.lib exit 1  # [win]
```
hmaarrfk, line 33:
```suggestion
    - if exist %PREFIX%\Library\lib\zimg.lib exit 1  # [win]
```

You might want to change 
```
--prefix=%PREFIX% 
```
to
```
--prefix=%LIBRARY_PREFIX% 
```
in your build script to get this to pass the tests.

sshockwave: Thanks for pointing that out. I didn't notice the difference between `%LIBRARY_PREFIX%` and `%PREFIX`.

hmaarrfk: its a common annoyance of conda. I think the original designers agreed that this discrepancy was a mistake.

recipes/zimg/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/sekrit-twc/zimg/archive/release-{{ version }}.tar.gz
+  sha256: 219d1bc6b7fde1355d72c9b406ebd730a4aed9c21da779660f0a4c851243e32f
+
+build:
+  skip: True  # [win]
```
hmaarrfk, line 13: are you actually meaning to skip windows? if so, maybe just remove everything that is windows related.

sshockwave: I just found `m2-automake` which can run on win64. Gonna add it back to the builds now.

hmaarrfk: Actually, we would prefer that you do not use `m2-` since the `m2` tools generate binaries that are incomaptible with the rest of the windows stuff.

See
https://github.com/conda-forge/ffmpeg-feedstock/blob/main/recipe/meta.yaml#L34

autotools_clang_conda may help

hmaarrfk: All said, I don't really understand the difference between the windows tools too much.

If you do decide to use `m2-` we would have to wait for somebody other than me to review sadly.

sshockwave: Really appreciate the help! I was looking for a tool to simply get the job done and `m2-` popped up on the search. The toolchains on Windows look like magic to me. They sometimes depend on each other, while sometimes they are incompatible.

hmaarrfk: I think `m2-` is actually ok. if you can get it to work. i know it is tough.

It is "ok" to just leave the windows as "future work". I would just ask that you delete all the build files.

sshockwave: I see. Interestingly, `autotools_clang_conda` is just a wrapper around `m2-` tools, so it seems better to let it do the tough part.

PR #21267, opened 2022-11-24
Add awkward-cpp

henryiii:
Adding awkward-cpp, which is the C++ component of AwkwardArray, preparing for the 2.0 release.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/awkward-cpp/meta.yaml
+++ b/recipes/awkward-cpp/meta.yaml
@@ -11,8 +11,7 @@ source:
 
 build:
   script:
-    - set CMAKE_GENERATOR_PLATFORM=  # [win]
-    - set CMAKE_GENERATOR_TOOLSET=  # [win]
+    - set CMAKE_GEN  # [win]
     - {{ PYTHON }} -m pip install . -vv --config-settings=logging.level=DEBUG
   number: 0
 
--- a/recipes/awkward-cpp/meta.yaml
+++ b/recipes/awkward-cpp/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   script:
     - set CMAKE_GENERATOR_PLATFORM=  # [win]
+    - set CMAKE_GENERATOR_TOOLSET=  # [win]
     - {{ PYTHON }} -m pip install . -vv --config-settings=logging.level=DEBUG
   number: 0
 
--- a/recipes/awkward-cpp/meta.yaml
+++ b/recipes/awkward-cpp/meta.yaml
@@ -11,9 +11,6 @@ source:
 
 build:
   script:
-    - set CMAKE_GEN  # [win]
-    - set CMAKE_GENERATOR  # [win]
-    - set CMAKE_GENERATOR=%CMAKE_GEN%  # [win]
     - {{ PYTHON }} -m pip install . -vv
   number: 0
 
--- a/recipes/awkward-cpp/meta.yaml
+++ b/recipes/awkward-cpp/meta.yaml
@@ -11,8 +11,7 @@ source:
 
 build:
   script:
-    - set CMAKE_GEN  # [win]
-    - {{ PYTHON }} -m pip install . -vv --config-settings=logging.level=DEBUG
+    - {{ PYTHON }} -m pip install . -vv
   number: 0
 
 requirements:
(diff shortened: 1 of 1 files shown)


mbargull:
Copied message from @mbargull at https://gitter.im/conda-forge/conda-forge.github.io?at=6384afcfabfb5860bb37e2ae :
> scikit-build really isn't the place to hack around it.
> I'm also not happy the way `CMAKE_GENERATOR` and similar are set and/or hardcoded in different places. There are at least 3 places that set them:
> - `conda-forge-pinning` sets the NMake default for Windows
> - `conda-build` sets the Unix Makesfiles default for non-Win
> - `vs*_win-64` (from the vc-feedstock) set it to VS if unset
> 
> The `CMAKE_GEN` is essentially a (temporary?) variable in the `vs*` activation scripts that I'm not aware of if used/should be used anywhere else.
> The behavior you encountered (N.B.: it would've been good to link to https://github.com/conda-forge/staged-recipes/pull/21267 in your messages ;) ) is just due to some unfortunate interaction within staged-recipes that is different from what you'd encounter in actual feedstock builds: `conda-build` adds all entries (for the selected platform) from `conda_build_config.yaml` as environment variables during the build. At staged-recipes the `conda-forge-pinning` package is installed for the build and as such you get all those variables set in the environment -- including `CMAKE_GENERATOR=NMake Makefiles` on Windows. Since it gets set by `conda-build` that way, the `vs*` activation script does not `set CMAKE_GENERATOR=%CMAKE_GEN%`. For a build in an actual feedstock repository, only a subset of `conda-forge-pinning`'s `conda_build_config.yaml` is generated by `conda-smithy rerender` which won't set `CMAKE_GENERATOR` by default (i.e., unless read from the recipe YAML/script files).
> If you add `CMAKE_GENERATOR: ''` in the recipe's `conda_build_config.yaml` it probably behaves like you'd want.

mbargull:
> I'm also not happy the way `CMAKE_GENERATOR` and similar are set and/or hardcoded in different places. There are at least 3 places that set them:
> 
> * `conda-forge-pinning` sets the NMake default for Windows
> * `conda-build` sets the Unix Makesfiles default for non-Win
> * `vs*_win-64` (from the vc-feedstock) set it to VS if unset

I wouldn't mind if you'd open issues in the repositories mentioned above so we can come up with a cleaner and more consistent solution to this problem. (and ping me and -- if he's available -- `@isuruf` because he probably has a better overview on how/where `CMAKE_GENERATOR` is used in other feedstocks etc.).

henryiii:
Is `CMAKE_GENERATOR` really needed? Conda-build shouldn't be touching it, since "Unix Makefiles" is already the CMake default, and it breaks the ability of scikit-build or scikit-build-core to depend on `ninja` and use that if no default is set (since a default is set, just incorrectly). And it's ridiculous to set it to "Unix Makefiles" if `make` isn't even in your environment (and `ninja` is, for example)!

For `conda-forge-pinning`, the "NMake Makefiles" is broken (since `CMAKE_GENERATOR_*` are set by `vs*_win-64`, and setting it breaks `vs*_win-64`'s activation. And it sounds like staged-recipes could be broken but the created recipe would work.

`vs*_win-64` setting it seems quite reasonable. Also, on Windows, the default is tricky in CMake - it checks the registry for VS, and if it can't find it, the default is NMake Makefiles (and the bittiness in `CMAKE_GENERATOR_PLATFORM` is useful too).

I am setting temporary workaround in scikit-build-core, since it's very important for me that it work out of the box correctly. I think, from the above, we should try to get it removed in `conda-forge-pinning` and `conda-build`?

henryiii:
Will reopen once scikit-build-core rebuilds)

henryiii:
I believe I can't actually patch out the `CMAKE_GENERATOR` setting from conda-build, forcing the use of `make` over `ninja` if it's not fixed upstream. :(

henryiii:
@conda-forge/help-python this ready for review! @agoose77 If you add yourself to the regular awkward feedstock you'll get access to this one too.

henryiii:
(just restarting to make sure this is fine with scikit-build-core 0.1.2)

henryiii:
Anything needed here? Awkward 2 is planning to ship early Dec.

henryiii:
@conda-forge/help-python https://github.com/conda-forge/awkward-feedstock/pull/89 is blocked on this.

beckermr:
Happy new year!

beckermr:
It reruns itself. We'll have to wait.

PR #21285, opened 2022-11-25
Add sphobjinv recipe to conda-forge

anjos:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/sphobjinv/meta.yaml
+++ b/recipes/sphobjinv/meta.yaml
@@ -11,7 +11,6 @@ source:
 
 build:
   noarch: python
-  skip: True  # [win]
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
--- a/recipes/sphobjinv/meta.yaml
+++ b/recipes/sphobjinv/meta.yaml
@@ -11,6 +11,7 @@ source:
 
 build:
   noarch: python
+  skip: True  # [win]
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
--- a/recipes/sphobjinv/meta.yaml
+++ b/recipes/sphobjinv/meta.yaml
@@ -1,0 +1,63 @@
+{% set name = "sphobjinv" %}
+{% set version = "2.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/bskinn/sphobjinv/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 29c2287afe54682bdc6ff66337c36af7dda63b1198166bae7aa0185012e46dde
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+    - attrs >=19.2
+    - certifi
+    - jsonschema >=3.0
+
+test:
+  requires:
+    - pip
+    - pytest
+    - pytest_check >=1.0,<1.1
+    - pytest-ordering
+    - pytest-timeout
+    - sphinx
+    - dictdiffer
+    - stdio-mgr
+  imports:
+    - sphobjinv
+  source_files:
+    - README.rst
+    - tox.ini
+    - requirements-dev.txt
+    - conftest.py
+    - tests/
+  commands:
+    - pip check
+    - pytest -k 'not readme'
+
+about:
+  home: https://github.com/bskinn/sphobjinv
+  summary: 'Toolkit for manipulation and inspection of Sphinx objects.inv files'
+  license: MIT
+  license_family: MIT
+  license_file:
+    - LICENSE.txt
+    - src/sphobjinv/_vendored/fuzzywuzzy/LICENSE.txt
+  doc_url: https://sphobjinv.readthedocs.io/en/latest/
+  dev_url: https://github.com/bskinn/sphobjinv
+
+extra:
+  recipe-maintainers:
+    - bskinn
+    - anjos
--- a/recipes/sphobjinv/meta.yaml
+++ b/recipes/sphobjinv/meta.yaml
@@ -46,7 +46,7 @@ test:
     - tests/
   commands:
     - pip check
-    - pytest -k 'not readme'
+    - pytest -k "not readme"
 
 about:
   home: https://github.com/bskinn/sphobjinv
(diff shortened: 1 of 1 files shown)


anjos:
@bskinn: please confirm your willingness to become a co-maintainer of this package as per conda-forge rules.

anjos:
@ocefpaf: this PR concludes the work started on #21238, so we'd appreciate your advice.  This is a noarch package, but it fails on Windows.  I do not understand why it is tested on windows in the first place, nor why it fails.  @bskinn - any ideas?

anjos:
@conda-forge/help-python I'm tagging this for review.

recipes/sphobjinv/meta.yaml
```diff
@@ -0,0 +1,63 @@
+{% set name = "sphobjinv" %}
+{% set version = "2.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/bskinn/sphobjinv/archive/refs/tags/v{{ version }}.tar.gz
```
ocefpaf, line 9: I'm guessing they are not publishing the sdist on PyPI, right? :weary:

recipes/sphobjinv/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  sha256: 29c2287afe54682bdc6ff66337c36af7dda63b1198166bae7aa0185012e46dde
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
```
ocefpaf, line 15: You are missing the entry_point here. This should fix the Windows build.

```suggestion
  noarch: python
  script: {{ PYTHON }} -m pip install . -vv
  number: 0
  entry_points:
    - sphobjinv = sphobjinv.cli.core:main
```

recipes/sphobjinv/meta.yaml
anjos: Right, not yet - may change in a future release.  I'm working with @bskinn on this item.

recipes/sphobjinv/meta.yaml
anjos: Right, thx for catching this.  One of the issues of not using grayskull, clearly!

ocefpaf: Yes and no. Grayskull does a great job at finding these in setup.py. I'm not sure if setup.cfg is supported and I know pyporject.toml is not. We are working on it though. BTW, you can use grayskull on GitHub releases.

recipes/sphobjinv/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - sphobjinv
+  source_files:
+    - README.rst
+    - tox.ini
+    - requirements-dev.txt
+    - conftest.py
+    - tests/
+  commands:
+    - pip check
+    - pytest -k 'not readme'
```
ocefpaf, line 49: This test is failing on Windows now, not sure why. But, if that is not a deal breaker we can merge this b/c the package is noarch.

bskinn: I've ID-ed the problem and am narrowing down on a fix. Will be getting a new sdist to @anjos soon, which will hopefully provide an all-green test board.

bskinn: (There's been a rabbit hole of Win vs Unix EOL silliness to dive down.)

bskinn:
@anjos:

> @ocefpaf: this PR concludes the work started on #21238, so we'd appreciate your advice. This is a noarch package, but it fails on Windows. I do not understand why it is tested on windows in the first place, nor why it fails. @bskinn - any ideas?

It's tested on Windows because, if used on Windows, it needs to correctly handle EOLs when working with decompressed inventories. (This is less *strictly* true than it used to be, with Windows tools getting better at tolerating Unix EOLs, but is still necessary IMO.)

I'm not sure exactly which rev you're testing now, so I can't speak precisely to the Win test failures without more info. I think I've got everything chased down, the Azure Pipelines are looking good in this most recent CI run on sphobjinv#268. Not sure why the GH Actions job failed, now, though

bskinn:
Ah. Newly broken link in my Sphinx docs. Yay.

Definitely not an issue for the conda-forge packaging. 

@anjos , new sdist coming shortly

anjos:
> It's tested on Windows because, if used on Windows, it needs to correctly handle EOLs when working with decompressed inventories. (This is less _strictly_ true than it used to be, with Windows tools getting better at tolerating Unix EOLs, but is still necessary IMO.)

@bskinn: I meant the conda package itself not your upstream.  The so-called `noarch` conda-forge packages only build/test on linux.  I'm not sure if there is any conda-forge provision to generate a single noarch package, but still test it across various architectures.  @ocefpaf may be able to confirm this.

bskinn:
> bskinn: I meant the conda package itself, not your upstream. 

Aha, that's what I get for trying to diagnose on my phone.

Change the pytest invocation to use double-quotes around `not readme`. Single quotes apparently don't mean "treat this as a single argument" to Windows... sigh.

bskinn:
Full disclosure: `sphobjinv` ***does*** currently vendor an old version of `fuzzywuzzy`, tweaked so that the pieces that `sphobjinv` needs are Python 3 compatible. 

This old version is required due to a licensing conflict with newer `fuzzywuzzy`. Simply pinning to this license-compatible old version is not feasible because of the lack of Python 3 compatibility available at that version. Vendoring and adapting seemed the least-worst option. The vendored code will not be updated with new fuzzywuzzy code over time, and my goal is to eventually be able to remove the vendored code, as part of bskinn/sphobjinv#178 and following work.

recipes/sphobjinv/meta.yaml
bskinn: Aha, nevermind -- this is a separate problem.

This should fix:

```suggestion
    - pytest -k "not readme"
```

bskinn: (depending on whether YAML needs special escaping for double-quotes...can't recall......)

ocefpaf:
> but still test it across various OSes. @ocefpaf may be able to confirm this.

Nope. It would be nice but the whole idea of noarch is that the test is equivalent on the OSes and platforms. What you can do is to try the "brand new" some-arch packages. Those are packages that have platform specific stuff, build and test on each platform but not all all Pythons.

bskinn:
I test `sphobjinv` aggressively cross-platform (Linux/MacOS/Win), cross-Python-version, and cross-dependency-version on all my releases, so I'm pretty comfortable with the conda-forge recipe only testing on POSIX.  The odds of a platform-specific bug emerging between my tests and the conda-forge build seem awfully small.

ocefpaf:
> so I'm pretty comfortable with the conda-forge recipe only testing on POSIX

That is what I needed to ~~hear~~ (read).

PR #21339, opened 2022-11-28
Adding foxes

SchmJo:


SchmJo:
I will re-submit after dropping tests, not sure how to upload test data

PR #21346, opened 2022-11-28
Add pyessv recipe

Zeitsperre:
Checklist
- [x] Build number is 0.


recipes/pyessv/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 043949af39a00c0c01ba69ad27709b9ca03086166bd5aa4e279264b9ed495b57
+
+build:
+  skip: True  # [win]
```
BastianZim, line 13: What is the reason for disabling windows? Seems like this package could be noarch.

Zeitsperre: Addressed. Thanks for the review! Feel free to merge when ready!

BastianZim: If you want to restrict this to unix only, you can just add the __unix virtual package. Or what was the reason for including that in the first place?

recipes/pyessv/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - pytest
+    - tox
```
BastianZim, line 24:
```suggestion
    - tox
    - __unix
```
This would be the way to restrict it to unix.

recipes/pyessv/meta.yaml
Zeitsperre: I think the reason is that I spend a lot of my time in Linux. Nothing stopping pyessv from working for Windows users!

recipes/pyessv/meta.yaml
Zeitsperre: Thank you! But no need here, it's OS independent.

PR #21350, opened 2022-11-28
Adding cuda_nvcc recipe

adibbley:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/cuda_nvcc/meta.yaml
+++ b/recipes/cuda_nvcc/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "cuda_nvcc" %}
+{% set version = "11.8.89" %}
+{% set platform = "linux-x86_64" %} # [linux64]
+{% set platform = "linux-ppc64le" %} # [ppc64le]
+{% set platform = "linux-sbsa" %} # [aarch64]
+{% set platform = "windows-x86_64" %} # [win]
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/{{ name }}/{{ platform }}/{{ name }}-{{ platform }}-{{ version }}-archive.tar.xz
+  sha256: 7ee8450dbcc16e9fe5d2a7b567d6dec220c5894a94ac6640459e06231e3b39a5 # [linux64]
+  sha256: 16fcfac1ef89584a36bf725b1706c51ecf9754acc712600f5f3e70f6ba119c8c # [ppc64le]
+  sha256: 17d30d924a1d6ff5748a0b6a4e7c25fd39644250cef9184fba520362f70ff798 # [aarch64]
+  sha256: 4cdd7555f31186e5af0b14ab761838bbc8b5e6441589f5bb326930c7a502dcd3 # [win]
+  patches:
+    - nvcc.profile.patch
+
+build:
+  # uncomment the following line if the package is pure python and the recipe is exactly the same for all platforms.
+  # it is okay if the dependencies are not built for all platforms/versions, although selectors are still not allowed.
+  # see https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python for more details.
+  # noarch: python
+  # if the installation is complex, or different between unix and windows, use separate bld.bat and build.sh files instead of this key.
+  # by default, the package will be built for the python versions supported by conda-forge and for all major oss.
+  # add the line "skip: true  # [py<35]" (for example) to limit to python 3.5 and newer, or "skip: true  # [not win]" to limit to windows.
+  # more info about selectors can be found in the conda-build docs: 
+  # https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors
+  #script: {{ python }} -m pip install . -vv
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - /lib64/libpthread.so.0
+    - /lib64/libdl.so.2
+    - /lib64/libm.so.6
+    - /lib64/libgcc_s.so.1
+    - /lib64/libc.so.6
+    - /lib64/ld-linux-x86-64.so.2
+    - /lib64/librt.so.1
+    - /lib64/libstdc++.so.6
+
+outputs:
+  - name: cuda-nvcc
+    files:
+      - bin
+      - include
+      - lib
+      - nvvm
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  description: |
+    Compiler for CUDA applications.
+  doc_url: https://docs.nvidia.com/cuda/index.html
+
+extra:
+  recipe-maintainers:
+    - adibbley
--- a/recipes/cuda_nvcc/bld.bat
+++ b/recipes/cuda_nvcc/bld.bat
@@ -1,0 +1,17 @@
+if not exist %PREFIX% mkdir %PREFIX%
+
+rem Directories...
+for /D %%i in (.\*) do (
+    robocopy /E %%i %PREFIX%\%%i
+)
+
+rem Files...
+for %%i in (.\*) do (
+    if not %%~ni==build_env_setup (
+    if not %%~ni==conda_build (
+    if not %%~ni==metadata_conda_debug (
+        xcopy %%i %PREFIX%
+    )
+    )
+    )
+)
--- a/recipes/cuda_nvcc/build.sh
+++ b/recipes/cuda_nvcc/build.sh
@@ -1,0 +1,11 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+
+for i in `ls`; do
+	[[ $i == "build_env_setup.sh" ]] && continue
+	[[ $i == "conda_build.sh" ]] && continue
+	[[ $i == "metadata_conda_debug.yaml" ]] && continue
+	cp -r $i ${PREFIX}
+done
--- a/recipes/cuda_nvcc/nvcc.profile.patch
+++ b/recipes/cuda_nvcc/nvcc.profile.patch
@@ -1,0 +1,14 @@
+--- bin/nvcc.profile        2021-07-06 11:22:41.714913066 -0700
++++ bin/nvcc.profile.new    2021-07-06 11:22:33.467108854 -0700
+@@ -6,9 +6,9 @@
+ LD_LIBRARY_PATH += $(TOP)/lib:
+ PATH            += $(TOP)/$(_NVVM_BRANCH_)/bin:$(_HERE_):
+
+-INCLUDES        +=  "-I$(TOP)/$(_TARGET_DIR_)/include" $(_SPACE_)
++INCLUDES        +=  "-I$(TOP)/include" $(_SPACE_)
+
+-LIBRARIES        =+ $(_SPACE_) "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)/stubs" "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)"
++LIBRARIES        =+ $(_SPACE_) "-L$(TOP)/lib/stubs" "-L$(TOP)/lib"
+
+ CUDAFE_FLAGS    +=
+ PTXAS_FLAGS     +=


recipes/cuda_nvcc/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - name: cuda-nvcc
+    files:
+      - bin
+      - include
+      - lib
+      - nvvm
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE
```
mariusvniekerk, line 62:
```suggestion
  license_file: LICENSE
  license: LicenseRef-NVIDIA-End-User-License-Agreement
```

recipes/cuda_nvcc/meta.yaml
```diff
@@ -0,0 +1,71 @@
+{% set name = "cuda_nvcc" %}
```
beckermr, line 1: So I would strongly argue for `cuda-nvcc` over the underscore version. This is a general statement on other `cuda-*` packages.

adibbley: The output packages are in the format of `cuda-*`. This is what I'm seeing in the log: `/home/conda/staged-recipes/build_artifacts/linux-64/cuda-nvcc-11.8.89-ha770c72_0.conda` 
The underscore just comes from how components are named internally, so if that is still and issue I can change that here.

recipes/cuda_nvcc/meta.yaml
jakirkham: For reference this comes from [here]( https://github.com/conda-forge/cudatoolkit-feedstock/blob/531e4594992258568fe187bc5c4e40d8c9c57b27/recipe/meta.yaml#L600 ). This was done to add a `license` metadata element, which fits with SPDX (expected by the linter)

recipes/cuda_nvcc/meta.yaml
beckermr: Yes please change it. It will keep the feedstock names with dashes too.

recipes/cuda_nvcc/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - test -f "${PREFIX}/bin/nvcc"  # [linux]
+    - if not exist %LIBRARY_BIN%\\bin\\nvcc.exe exit 1  # [win]
+
+outputs:
+  - name: cuda-nvcc
+    files:
+      - bin
+      - include
+      - lib
+      - nvvm
```
beckermr, line 58: This build should not use outputs since it has only one.

recipes/cuda_nvcc/meta.yaml
jakirkham: Think the issue is we can't have the packages in `outputs` and the feedstock `name` match

adibbley: This is renamed to cuda-nvcc now. Should the recipe folder also be renamed?

recipes/cuda_nvcc/meta.yaml
adibbley: Removed this block now.

recipes/cuda_nvcc/meta.yaml
jakirkham: Thanks Alex! 🙏

> Should the recipe folder also be renamed?

For consistency yes. Though I don't think we do anything with the folder name

recipes/cuda_nvcc/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  # uncomment the following line if the package is pure python and the recipe is exactly the same for all platforms.
+  # it is okay if the dependencies are not built for all platforms/versions, although selectors are still not allowed.
+  # see https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python for more details.
+  # noarch: python
+  # if the installation is complex, or different between unix and windows, use separate bld.bat and build.sh files instead of this key.
+  # by default, the package will be built for the python versions supported by conda-forge and for all major oss.
+  # add the line "skip: true  # [py<35]" (for example) to limit to python 3.5 and newer, or "skip: true  # [not win]" to limit to windows.
+  # more info about selectors can be found in the conda-build docs: 
+  # https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors
+  #script: {{ python }} -m pip install . -vv
```
jakirkham, line 34: We can drop these comments that come from the example recipe. They are mainly meant to be instructive

```suggestion
```

recipes/cuda-nvcc/nvcc.profile.patch
```diff
@@ -0,0 +1,14 @@
...
++++ bin/nvcc.profile.new    2021-07-06 11:22:33.467108854 -0700
+@@ -6,9 +6,9 @@
+ LD_LIBRARY_PATH += $(TOP)/lib:
+ PATH            += $(TOP)/$(_NVVM_BRANCH_)/bin:$(_HERE_):
+
+-INCLUDES        +=  "-I$(TOP)/$(_TARGET_DIR_)/include" $(_SPACE_)
++INCLUDES        +=  "-I$(TOP)/include" $(_SPACE_)
+
+-LIBRARIES        =+ $(_SPACE_) "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)/stubs" "-L$(TOP)/$(_TARGET_DIR_)/lib$(_TARGET_SIZE_)"
++LIBRARIES        =+ $(_SPACE_) "-L$(TOP)/lib/stubs" "-L$(TOP)/lib"
```
isuruf, line 11: Why do this? This would make it hard to enable cross compiling

recipes/cuda_nvcc/meta.yaml
adibbley: Sounds good, moved the recipe to cuda-nvcc directory now.

recipes/cuda_nvcc/meta.yaml
adibbley: Removed these comments, thanks!

recipes/cuda-nvcc/nvcc.profile.patch
adibbley: Our packages have been installing directly into `lib` and `include` directories, likely to follow what `cudatoolkit` did. It doesn't appear there was a plan to add cross packages at this point.

jakirkham: Yeah it would be good to add cross-compiling if we can. This has recently come up a few times as a user request. Here's a few references:

* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/209
* https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/210

adibbley: Is there any recommendation on where to move the libraries to? Other installer formats just use something like `targets/<multiarch>/lib` and create a `lib` symlink to the host target, which only works when cuda doesn't share the `lib` directory.

recipes/cuda_nvcc/meta.yaml
jaimergp: Extreme nitpick, but: To avoid confusion down the line, we should also rename the title of this PR accordingly. Thanks <3

jakirkham: Done

recipes/cuda-nvcc/nvcc.profile.patch
jakirkham: Maybe it would make sense to put these in `${CONDA_BUILD_SYSROOT}/lib`. Though thoughts from others welcome here

Somewhat related question, do we include some sort of `libcuda.so` stub library as well? We've made [use of the stub library before during builds]( https://github.com/conda-forge/nvcc-feedstock/blob/9c8e43393d9c798c52a9d8e56785920ef767cabb/recipe/linux/activate.sh#L97-L110 ), which has been pretty helpful.

m3vaz:
Doesn't nvcc have a maximum supported compiler version? Is that limit monitored here?

recipes/cuda-nvcc/nvcc.profile.patch
isuruf: We can add them to `lib/targets/<multiarch>/lib` and do symlinks for each file in all the packages.

adibbley: Removed this patch now. Files should now be under `targets/<multiarch>/lib` and `targets/<multiarch>/include` with symlinks to each file

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - /lib64/libpthread.so.0
+    - /lib64/libdl.so.2
+    - /lib64/libm.so.6
+    - /lib64/libgcc_s.so.1
+    - /lib64/libc.so.6
+    - /lib64/ld-linux-x86-64.so.2
+    - /lib64/librt.so.1
+    - /lib64/libstdc++.so.6
+
+requirements:
```
jakirkham, line 35: Would it make sense to add the C & C++ compilers as `build` dependencies?

This would alleviate the need for the `missing_dso_whitelist` as well as letting the compiler runtime libraries be dropped from `host` and `run`

```suggestion
requirements:
  build:
    - {{ compiler("c") }}
    - {{ compiler("cxx") }}
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_nvcc/{{ platform }}/cuda_nvcc-{{ platform }}-{{ version }}-archive.{{ extension }}
+
+  sha256: 160ca9f8828f53daa4b2066c5361aacb45fa2575885f70c223cda1d11df53d6f  # [linux64]
+  sha256: dc5bd7eac57b32a9645c0c23c6a1ba8a9533e846eeec31414e1aa4230afee2f8  # [ppc64le]
+  sha256: 85c8c643ea2cfa398e1b5ffcd597f5bb3b738526c9e3fb8e39dd909c55345f36  # [aarch64]
+  sha256: f67406f44263f0e4a110a8cf0200bff1d3d0e32917db8a7c9e417a87eb7ca372  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
```
jakirkham, line 24:
```suggestion
  skip: true  # [osx]
  run_exports:
    strong:
      - __glibc >=2.17        # [linux]
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - /lib64/libdl.so.2
+    - /lib64/libm.so.6
+    - /lib64/libgcc_s.so.1
+    - /lib64/libc.so.6
+    - /lib64/ld-linux-x86-64.so.2
+    - /lib64/librt.so.1
+    - /lib64/libstdc++.so.6
+
+requirements:
+  host:
```
jakirkham, line 36:
```suggestion
  host:
    - sysroot_{{ target_platform }} 2.17   # [linux]
    - __glibc >=2.17                       # [linux]
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - libstdcxx-ng  # [linux]
+    - libgcc-ng  # [linux]
+  run:
```
jakirkham, line 39:
```suggestion
  run:
    - sysroot_{{ target_platform }} 2.17   # [linux]
    - __glibc >=2.17                       # [linux]
```

jakirkham (commented):
Given this will require a newer GLIBC, think we need these changes. Though there may be things I've missed here. For example how `aarch64`/`ppc64le` are handled

recipes/cuda-nvcc/meta.yaml
jakirkham: Are there any compiler runtime libraries that should be added as dependencies here?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+
+test:
+  commands:
+    - nvcc --version
```
kkraus14, line 42: It would maybe be nice to have a small test file that gets compiled and runs (even if it doesn't have any device code) just to make sure nothing is breaking in the usage path.

We can absolutely do this in follow up PRs though.

kkraus14 (approved):

recipes/cuda-nvcc/meta.yaml
jakirkham: Was thinking about that too. We can borrow [this test]( https://github.com/conda-forge/nvcc-feedstock/blob/9c8e43393d9c798c52a9d8e56785920ef767cabb/recipe/meta.yaml#L68-L74 )

hmaarrfk:
> Doesn't nvcc have a maximum supported compiler version? Is that limit monitored here?

I've somewhat been routinely ignoring this for various reasons (sometimes C++ versions require it). Not sure what the latest best practice is.

h-vetinari:
> > Doesn't nvcc have a maximum supported compiler version? Is that limit monitored here?
> 
> I've somewhat been routinely ignoring this for various reasons (sometimes C++ versions require it). Not sure what the latest best practice is.

Not sure how you'd achieve that, nvcc hard-errors when a too-new gcc is found (at least whenever I tried). Context I'm aware of w.r.t. this: [here](https://gist.github.com/ax3l/9489132) (xref also [here](https://github.com/conda-forge/nvcc-feedstock/issues/51)),

jakirkham:
In terms of `nvcc` & compiler version compatibility, currently we constrain that at [the global level]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/96948c8e97deb6f75ceb32bbc036c1382ace0732/recipe/conda_build_config.yaml#L158-L165 ). It likely makes sense to keep doing this going forward. Though happy to hear thoughts if others have any

kkraus14:
It would be nice to have proper constraints in the recipe for users who pull this package into their environment alongside the host compiler packages. I.E. I imagine `cuda-nvcc 12.0` shouldn't allow `gcc 13.0` (once its released).

jakirkham:
We could add `run_constrained` to the recipe as well

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
```
jakirkham, line 38: Maybe we could use something like this for the Linux compiler constraint (assuming I've understood [the system requirements]( https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#system-requirements ) correctly)

Not sure if we need something similar for Windows

```suggestion
    - __glibc >=2.17                       # [linux]
 run_constrained:
    - {{ compiler("c") }} >=6,12.1         # [linux]
```

beckermr: what does this constraint mean? It appears to say >=6 and ==12.1. is that right?

jakirkham: Good eye. That should have been an upper bound. Fixed

recipes/cuda-nvcc/meta.yaml
jakirkham: Proposed something in PR ( https://github.com/adibbley/staged-recipes/pull/1 )

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  run_constrained:
+    - {{ compiler("c") }} >=6,<12.2        # [linux]
```
isuruf, line 40:
```suggestion
    - gcc_impl_{{ target_platform }} >=6,<12.2        # [linux]
```

recipes/cuda-nvcc/meta.yaml
jakirkham: As part of creating this test, we realized we need to package `cuda-cudart`. Alex put together a draft PR with that recipe ( https://github.com/conda-forge/staged-recipes/pull/21723 )

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - __glibc >=2.17        # [linux]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run:
```
robertmaynard, line 36: Since you need a host compiler ( Specifically `gcc` on PATH when no `-ccbin` provided ), don't we need a gcc / `{{ compiler("cxx") }}` as a run requirement?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run_constrained:
+    - gcc_impl_{{ target_platform }} >=6,<13        # [linux]
+
+test:
+  requires:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - cuda-cudart-static
```
jakirkham, line 46: Would it be possible to use the dev package here?

```suggestion
    - cuda-cudart-dev
```

robertmaynard: No it needs `cuda-cudart-static` since by default nvcc will link against the cudart_static library:

```
> nvcc -v t.cu
#$ _NVVM_BRANCH_=nvvm
#$ _SPACE_= 
#$ _CUDART_=cudart
#$ _HERE_=/usr/local/cuda/bin
...
#$ g++ -D__CUDA_ARCH_LIST__=520 -m64 -Wl,--start-group "/tmp/tmpxft_000fce61_00000000-13_a_dlink.o" "/tmp/tmpxft_000fce61_00000000-11_t.o"   "-L/usr/local/cuda/bin/../targets/x86_64-linux/lib/stubs" "-L/usr/local/cuda/bin/../targets/x86_64-linux/lib"  -lcudadevrt  -lcudart_static  -lrt -lpthread  -ldl  -Wl,--end-group -o "a.out"
```

robertmaynard: I just brought up that this requirement seems weird here: https://github.com/conda-forge/staged-recipes/pull/21723/files#r1084611245

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  files:
+    - test.cu
+  commands:
+    - nvcc --version
+    - nvcc -ccbin "${CXX}" test.cu
```
jakirkham, line 51:
```suggestion
    - nvcc -ccbin --cudart shared "${CXX}" test.cu
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+test:
+  requires:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - cuda-cudart-dev
+  files:
+    - test.cu
+  commands:
+    - nvcc --version
```
robertmaynard, line 50: we should add `nvcc --ccbin "${CXX} -arch=native test.cu` to validate that the `__nvcc_device_query` executable that nvcc calls is also packaged

jakirkham: How should we handle cross-compiling use cases?

recipes/cuda-nvcc/meta.yaml
jakirkham: Sticking to default (static) linking

recipes/cuda-nvcc/meta.yaml
jakirkham: The requirement order was fixed and the dependency was updated

recipes/cuda-nvcc/meta.yaml
jakirkham: A test has been added. Depends on `cuda-cudart`, which is being packaged separately

recipes/cuda-nvcc/meta.yaml
robertmaynard: I think what I am really asking for is a check that the expected internal executables exist in the package ( `fatbinary`, `nvlink`, `cudafe++`, `__nvcc_device_query`, etc ). 

That would avoid any execution of `nvcc`

jakirkham: Will this still work even when there is no GPU?

robertmaynard: > Will this still work even when there is no GPU?

Yes. The compiler will generate a warning ( `nvcc warning : Cannot find valid GPU for '-arch=native', default arch is used` ) and continue to execute.

recipes/cuda-nvcc/build.sh
```diff
@@ -0,0 +1,24 @@
...
+		mkdir -p ${PREFIX}/${targetsDir}
+		mkdir -p ${PREFIX}/$i
+		cp -r $i ${PREFIX}/${targetsDir}
+		for j in `ls $i`; do
+			ln -s ../${targetsDir}/$i/$j ${PREFIX}/$i/$j
+		done
+	else
+		cp -r $i ${PREFIX}
+	fi
+done
```
jakirkham, line 24:
```suggestion
done

echo '' >> "${PREFIX}/bin/nvcc.profile"
echo "NVCC_PREPEND_FLAGS += -ccbin "\""${CXX}"\""" >> "${PREFIX}/bin/nvcc.profile"
```

recipes/cuda-nvcc/bld.bat
```diff
@@ -0,0 +1,17 @@
+if not exist %PREFIX% mkdir %PREFIX%
```
jakirkham, line 1:
```suggestion
if not exist %PREFIX% mkdir %PREFIX%
if not exist %LIBRARY_PREFIX% mkdir %LIBRARY_PREFIX%
```

recipes/cuda-nvcc/bld.bat
```diff
@@ -0,0 +1,17 @@
...
+
+rem Directories...
+for /D %%i in (.\*) do (
+    robocopy /E %%i %PREFIX%\%%i
```
jakirkham, line 5:
```suggestion
    robocopy /E %%i %LIBRARY_PREFIX%\%%i
```

recipes/cuda-nvcc/bld.bat
```diff
@@ -0,0 +1,17 @@
...
+)
+
+rem Files...
+for %%i in (.\*) do (
+    if not %%~ni==build_env_setup (
+    if not %%~ni==conda_build (
+    if not %%~ni==metadata_conda_debug (
+        xcopy %%i %PREFIX%
```
jakirkham, line 13:
```suggestion
        xcopy %%i %LIBRARY_PREFIX%
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - nvcc -ccbin "${CXX}" test.cu
```
jakirkham, line 51:
```suggestion
    - nvcc test.cu
```

jakirkham (commented):
Think we have a better option for setting the C++ compiler used. Have proposed that change below

Also noticed on Windows files are being put in `%PREFIX%`. Though believe we want these in [`%LIBRARY_PREFIX%`]( https://docs.conda.io/projects/conda-build/en/latest/user-guide/environment-variables.html?highlight=LIBRARY_PREFIX#environment-variables-set-during-the-build-process )

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run_constrained:
+    - gcc_impl_{{ target_platform }} >=6,<13        # [linux]
+
+test:
+  requires:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - cuda-cudart-dev
```
jakirkham, line 46: Should we move this to `requirements/run`?

recipes/cuda-nvcc/build.sh
```diff
@@ -0,0 +1,27 @@
...
+
+echo '' >> "${PREFIX}/bin/nvcc.profile"
+echo "NVCC_PREPEND_FLAGS += -ccbin "\""${CXX}"\""" >> "${PREFIX}/bin/nvcc.profile"
```
jakirkham, line 27: Looks like `$BUILD_PREFIX` is contained in `$CXX`, but is not being expanded. Maybe using [indirect expansion]( https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html ) helps?

```suggestion
echo "NVCC_PREPEND_FLAGS += -ccbin "\""${!CXX}"\""" >> "${PREFIX}/bin/nvcc.profile"
```

jakirkham: Seems this results in an empty string. Maybe we can just rely on `BUILD_PREFIX` and simply make replacement more straightforward. Perhaps like so

```suggestion
echo "NVCC_PREPEND_FLAGS += -ccbin "\""${CXX/BUILD_PREFIX/(BUILD_PREFIX)}"\""" >> "${PREFIX}/bin/nvcc.profile"
```

jakirkham: After looking at this more, think we need to move this to [activation/deactivation scripts]( https://docs.conda.io/projects/conda-build/en/latest/resources/activate-scripts.html ). The scripts themselves would follow these patterns to [activate]( https://github.com/conda-forge/nvcc-feedstock/blob/9c8e43393d9c798c52a9d8e56785920ef767cabb/recipe/linux/activate.sh#L26-L29 ) and [deactivate]( https://github.com/conda-forge/nvcc-feedstock/blob/9c8e43393d9c798c52a9d8e56785920ef767cabb/recipe/linux/deactivate.sh#L30-L34 ). In particular activate would run

```bash
export NVCC_PREPEND_FLAGS = "${NVCC_PREPEND_FLAGS} -ccbin ${CXX}"
```

isuruf: Why do you need activation scripts? What @adibbley had here was correct.

jakirkham: No it didn't actually work. We tried to set this in `nvcc.profile` (also experimented with this a number of different ways locally) and it didn't work. Setting the environment variable in the activation script does work.

Should add personally would have been happier not to use activation scripts (and use `nvcc.profile` or similar), but that doesn't seem to be option.

isuruf: Can you be more specific about what didn't work?

jakirkham: It kept trying to use `gcc`, which it could not find instead of the compiler specified here

isuruf: Can you upload the packages to some anaconda.org channel so that I can have a look?

jakirkham: Sure. Think it's [this one]( https://anaconda.org/jakirkham/cuda-nvcc/files?version=12.0.76&channel=nvcc_tst )

isuruf: Have you got an x86_64 build?

isuruf: Try 
```bash
echo "NVCC_PREPEND_FLAGS += -ccbin "\"${PREFIX}/bin/$(basename ${CXX})\""">> "${PREFIX}/bin/nvcc.profile"
```

jakirkham: No sorry. Only the ARM build

Yeah we tried a bunch of stuff like this without luck. AFAICT `NVCC_PREPEND_FLAGS` isn't respected in `nvcc.profile`

jakirkham: If we soft/hard linked `$PREFIX/bin/gcc` to `$CXX`, that would work (for example the `compilers` package works), but that comes with its own set of tradeoffs

jakirkham: `nvcc.profile` does seem to support `compiler-bindir` (same as `--compiler-bindir` CLI option), but that only specifies what directory to look for `gcc` in (not the name of `gcc`)

isuruf: `--compiler-bindir` and `-ccbin` are aliases right?

jakirkham: Yeah I tried setting `compiler-bindir` to the exact path of `$CXX` in `nvcc.profile`, but got the same error. The only examples I've seen of using this in `nvcc.profile` have been on Windows. It might be ignored on Linux.

isuruf: We can add a symlink in the compiler package like `${PREFIX}/bin/x86_64-conda-linux-gnu-gcc -> ${PREFIX}/x86_64-conda-linux-gnu/bin/gcc`

jakirkham: That seems like an interesting option. Would that be added to the `$PATH` then?

isuruf: No, you can use `${PREFIX}/x86_64-conda-linux-gnu/bin/` as `--compiler-bindir` then.

jakirkham: Ok not following how this would be different here. Given `nvcc.profile` already isn't respecting `compiler-bindir` on Linux, this flag would still be handled via an environment variable in an activation script as it is now. What am I missing?

isuruf: nvcc.profile does support `compiler-bindir`, just not the compiler name. i.e. it supports a custom directory name, but not a custom compiler name. So having `gcc` in `${PREFIX}/x86_64-conda-linux-gnu/bin/` would help.

jakirkham: It supports both. From [the docs]( https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#compiler-bindir-directory-ccbin ):

> Specify the directory in which the default host compiler executable resides.
>
> The host compiler executable name can be also specified to ensure that the correct host compiler is selected.

jakirkham: In any event `nvcc.profile` still seems to be ignoring `compiler-bindir` on Linux

Given we don't seem to be coming up with a better option than the activation script, would suggest we stick with it since we know it works

isuruf: Try the following
```
mamba create -n nvcc cuda-nvcc cuda-cudart-dev cuda-cudart-static compilers -c nvidia
mamba activate nvcc
mv $CONDA_PREFIX/bin/nvcc.profile $CONDA_PREFIX/bin/nvcc.profile.bak
cp $CONDA_PREFIX/bin/nvcc.profile.bak $CONDA_PREFIX/bin/nvcc.profile
echo "compiler-bindir = $CONDA_PREFIX/$HOST/bin" >> $CONDA_PREFIX/bin/nvcc.profile
ln -sf $CONDA_PREFIX/bin/$HOST-gcc $CONDA_PREFIX/$HOST/bin/gcc
ln -sf $CONDA_PREFIX/bin/$HOST-g++ $CONDA_PREFIX/$HOST/bin/g++
echo "int main() {}" > test.cu
nvcc test.cu --verbose
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  sha256: dc5bd7eac57b32a9645c0c23c6a1ba8a9533e846eeec31414e1aa4230afee2f8  # [ppc64le]
+  sha256: 85c8c643ea2cfa398e1b5ffcd597f5bb3b738526c9e3fb8e39dd909c55345f36  # [aarch64]
+  sha256: f67406f44263f0e4a110a8cf0200bff1d3d0e32917db8a7c9e417a87eb7ca372  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+  run_exports:
+    strong:
+      - {{ pin_compatible('cuda-version', max_pin='x.x') }}
```
jakirkham, line 27: Can we relax to `x`?

```suggestion
      - {{ pin_compatible('cuda-version', max_pin='x') }}
```

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - nvcc --verbose test.cu
+
+outputs:
+  - name: cuda-nvcc
+    files:
+      - bin      # [linux]
+      - include  # [linux]
+      - lib      # [linux]
+      - nvvm     # [linux]
+      - Library\nvvm  # [win]
```
kkraus14, line 66: Does it make sense for `nvvm` to be a separate package? I'm thinking about downstream packages like `Numba` who don't use `nvcc` but do use `nvvm`.

Regardless, this could be split out later if desired.

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+      run_constrained:
+        - gcc_impl_{{ target_platform }} >=6,<13  # [linux]
+        - arm-variant * {{ arm_variant_type }}    # [aarch64]
+    test:
+      requires:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - cuda-cudart-dev
+        - cuda-nvcc_{{ target_platform }} =={{ version }}
```
kkraus14, line 67: Is there a reason we need to specify the target platform package here versus just relying on the produced `cuda-nvcc` package?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+        - __glibc >=2.17                       # [linux]
+        - cuda-version {{ cuda_version }}
+      run:
+        - sysroot_{{ target_platform }} 2.17   # [linux]
+        - __glibc >=2.17                       # [linux]
+        - cuda-cudart-dev
+        - cuda-nvcc_{{ target_platform }} =={{ version }}
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+      run_constrained:
+        - gcc_impl_{{ target_platform }} >=6,<13  # [linux]
```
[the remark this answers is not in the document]
jakirkham: Since we only use `clang` on macOS and that is skipped, think we don't need to worry about that.

On Windows we don't distribute compiler packages and instead rely on CI to have them. Not sure what more we can do there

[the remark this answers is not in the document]
kkraus14: > Since we only use `clang` on macOS and that is skipped, think we don't need to worry about that.

A user could presumably install it and want to use it in their environment though?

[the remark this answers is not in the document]
jakirkham: Sorry am not following. How would they install it? This recipe isn't built on macOS

recipes/cuda-nvcc/meta.yaml
adibbley: We can remove the test specific requirements block (done in latest commit).

recipes/cuda-nvcc/meta.yaml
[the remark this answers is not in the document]
kkraus14: There's linux-64 clang builds that someone could install and use in their environment.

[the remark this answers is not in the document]
jakirkham: Ah ok. That makes more sense. Thanks Keith! 🙏

There is a `clang` package. Though the package that encodes `target_platform` would be [in `clang-compiler-activation-feedstock`]( https://github.com/conda-forge/clang-compiler-activation-feedstock/blob/d79fc38aef6f7f267a02623de29cd83f42830bdc/recipe/meta.yaml#L36 ). However this is only configured for macOS targets currently.

That said, suppose one could add Linux targets for `clang`. Maybe we could save this as a follow up issue (for whenever those packages get added)?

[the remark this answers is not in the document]
kkraus14: We can always repodata patch at that time. I'm fine to defer it to a follow up.

kkraus14 (approved):

recipes/cuda-nvcc/nvcc.profile.patch
```diff
@@ -0,0 +1,14 @@
+--- bin/nvcc.profile	2023-03-23 15:56:49.535384866 -0400
++++ bin/nvcc.profile.new	2023-03-23 19:21:08.071151138 -0400
+@@ -5,9 +5,9 @@
+ 
+ PATH            += $(TOP)/$(_NVVM_BRANCH_)/bin;$(_HERE_);$(TOP)/lib;
+ 
+-INCLUDES        +=  "-I$(TOP)/include" $(_SPACE_)
++INCLUDES        +=  "-I$(TOP)/include" $(_SPACE_) "-I$(TOP)/include/targets/$(_WIN_PLATFORM_)" $(_SPACE_)
```
robertmaynard, line 8: Where is `include/targets/` used? I know we have `/targets/arch/include`

adibbley: We put the `cuda-cccl` headers under `include/targets/$(_WIN_PLATFORM_)`. Windows doesn't typically use the `targets` layout, but we needed a safe space to not clobber the GitHub releases.

recipes/cuda-nvcc/bld.bat
vyasr: I'm unfamiliar with Windows, is this something we should be doing in all of our recipes? [We didn't do that in libcublas](https://github.com/conda-forge/libcublas-feedstock/blob/main/recipe/bld.bat), for instance. We're not specifically writing anything to the `%LIBRARY_PREFIX%` directory in this file.

recipes/cuda-nvcc/build.sh
```diff
@@ -0,0 +1,36 @@
...
+        mkdir -p ${PREFIX}/$i
+        cp -rv $i ${PREFIX}/${targetsDir}
+        if [[ $i == "lib" ]]; then
+            for j in "$i"/*.a*; do
+                # Shared and static libraries are symlinked in $PREFIX/lib
+                ln -sv ../${targetsDir}/$j ${PREFIX}/$j
+            done
+        fi
+    else
+        cp -r $i ${PREFIX}
```
vyasr, line 26: Some questions:
- I assume that we're intentionally not including the contents of `bin` into the `targetsDir`?
- We probably want to put the `nvvm` contents into the `targetsDir` right?
- `cicc` is inside `nvvm/bin`, but `nvcc` needs to be able to find it. Does that need special handling to be placed into `${PREFIX}/bin`?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+      - Library\nvvm  # [win]
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - sysroot_{{ target_platform }} 2.17   # [linux]
+        - __glibc >=2.17                       # [linux]
```
vyasr, line 52: Out of scope for fixing right now, but for future note: Is there no relationship between these packages? Should there be?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - cuda-version {{ cuda_version }}
```
vyasr, line 53: What is the benefit of including `cuda-version` in the host environment? Isn't the purpose to be a metapackage ensuring compatibility across dependencies, i.e. something we really only need in `run`? What build requirements are being constrained by including this in `host`?

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+      run:
+        - sysroot_{{ target_platform }} 2.17   # [linux]
+        - __glibc >=2.17                       # [linux]
+        - cuda-cudart-dev
```
vyasr, line 57: Do we need a version constraint here, or is that all managed using `cuda-version`?

vyasr (changes requested):

recipes/cuda-nvcc/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - cuda-nvcc_{{ target_platform }} =={{ version }}
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+      run_constrained:
+        - gcc_impl_{{ target_platform }} >=6,<13  # [linux]
+        - arm-variant * {{ arm_variant_type }}    # [aarch64]
+    test:
```
bdice, line 63: Merge https://github.com/adibbley/staged-recipes/pull/6 to fix the tests. These won't run due to package name conflicts.

recipes/cuda-nvcc/nvcc.profile.patch
robertmaynard: Gotcha, thanks for the explanation.

recipes/cuda-nvcc/build.sh
robertmaynard: The current layout is fine. The `bin` and `nvvm` directories will be copied into `${PREFIX}/bin`/ and `${PREFIX}/nvvm`. Since the relative layot has stayed consistent, `nvcc` doesn't need to be modified to know where the `nvvm` directory is

vyasr: Are we OK writing to `${PREFIX}/nvvm`? My understanding was that we were avoiding writing to such top-level directories as much as possible to avoid clobbering, but perhaps we're OK treating `nvvm` as CUDA-specific in the same way that we are treating `targets` as CUDA-specific.

Do we still want to include some of these contents in `targets` or only put them in the top-level `${PREFIX}/*` directories?

adibbley: We need `nvvm` in `$PREFIX` due to the relative pathing Robert mentioned. `nvcc` expects `bin` `nvvm` and `targets` to all be at the same level. 

Clobbering shouldn't be a concern here as it was with things like `src/example.c`

isuruf: We are already patching `nvcc.profile`, so it's not a big deal to patch a couple more lines.
If we are going with filesystem hierarchy, they should not be in `$PREFIX` and rather in somewhere like `$PREFIX/lib/nvcc/nvvm` and `$PREFIX/lib/nvcc/targets` just like `gcc` does with `$PREFIX/lib/gcc`

Thoughts @mbargull, @conda-forge/core?

jakirkham: `nvcc.profile` is not patched on Linux. It is only patched on Windows to address shipping different CCCL versions as requested in PR ( https://github.com/conda-forge/staged-recipes/pull/21953 ). Preferably it would not be patched at all (even on Windows). Open to a better solution that doesn't require the patch on Windows (if we have ideas). Though this is probably best taken up in the feedstock since we are blocked on Windows builds atm ( https://github.com/conda-forge/staged-recipes/pull/21723#discussion_r1160853681 ).

It's worth noting that there are already cases where we install non-traditional things in `$PREFIX`. For example [`go` adds `$PREFIX/go`]( https://github.com/conda-forge/go-feedstock/blob/865b217fa6b54e87b050e0056d61ad2faf669119/recipe/build-base.sh#L69 ). With `go`, trying to move away from that model breaks the compiler in weird ways.

Typically all CUDA components are placed in the same relative paths where they are then QA’d and deployed. The proposal here suggests dividing them up between different paths that will deviate from the process. This risks introducing new issues unique to this non-standard deployment model and makes it harder to support.

Finally would add Numba is [dependent on this location]( https://github.com/numba/numba/blob/be990a818c377408a1f41c37c7720bc2ba0fb89a/numba/cuda/cuda_paths.py#L144 ). So changing it here means Numba (and maybe other libraries) will need updates.

jakirkham:
Planning on merging EOD Monday if no comments

kkraus14 (approved):

recipes/cuda-nvcc/activate.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+export NVCC_PREPEND_FLAGS_BACKUP="${NVCC_PREPEND_FLAGS}"
+export NVCC_PREPEND_FLAGS="${NVCC_PREPEND_FLAGS} -ccbin ${CXX}"
```
robertmaynard, line 4: Testing the package locally ( `conda install -c ..../staged-recipes/build_artifacts/ cuda-nvcc cuda-cudart` ) in a clean env.

This creates a broken compiler, due to `${CXX}` having no value. We need to either require a CXX compiler or only optionally add this env var when `CXX` exists.

```
(ctk) rmaynard@rmaynard-dt:~/Work/cmake/conda_test_build$ conda list
# packages in environment at /home/rmaynard/miniconda3/envs/ctk:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
cuda-cccl_linux-64        12.0.90              ha770c72_1    conda-forge
cuda-cudart               12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-dev           12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-dev_linux-64  12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-static        12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-static_linux-64 12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-nvcc                 12.0.76              h59595ed_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-version              12.0                 hffde075_0    conda-forge
libgcc-ng                 12.2.0              h65d4601_19    conda-forge
libgomp                   12.2.0              h65d4601_19    conda-forge
libstdcxx-ng              12.2.0              h46fd767_19    conda-forge
````

**Review by robertmaynard — CHANGES_REQUESTED** [review]

**Review by robertmaynard — CHANGES_REQUESTED** [review]
Testing the package locally ( `conda install -c ..../staged-recipes/build_artifacts/ cuda-nvcc cuda-cudart` ) in a clean env.

This creates a broken compiler, due to `${CXX}` having no value. We need to either require a CXX compiler or only optionally add this env var when `CXX` exists.

```
(ctk) rmaynard@rmaynard-dt:~/Work/cmake/conda_test_build$ conda list
# packages in environment at /home/rmaynard/miniconda3/envs/ctk:
#
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                 conda_forge    conda-forge
_openmp_mutex             4.5                       2_gnu    conda-forge
cuda-cccl_linux-64        12.0.90              ha770c72_1    conda-forge
cuda-cudart               12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-dev           12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-dev_linux-64  12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-static        12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-cudart-static_linux-64 12.0.107             hcb278e6_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-nvcc                 12.0.76              h59595ed_0    file:///home/rmaynard/Work/conda-forge/staged-recipes/build_artifacts
cuda-version              12.0                 hffde075_0    conda-forge
libgcc-ng                 12.2.0              h65d4601_19    conda-forge
libgomp                   12.2.0              h65d4601_19    conda-forge
libstdcxx-ng              12.2.0              h46fd767_19    conda-forge
````

jakirkham:
Alex added a clearer error message above. If we want to do more here, let's raise an issue in the feedstock and discuss

jakirkham:
Thanks all! 🙏

Will also include a PR link related to follow ups pertaining to comment ( https://github.com/conda-forge/staged-recipes/pull/21350#issuecomment-1507572414 ) once available

PR #21405, opened 2022-12-02
Added libpcsclite as a package

thewchan:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -17,10 +17,9 @@ build:
     - if test -f Makefile; then make distclean; fi
     - rm -rf *.cache *.m4 config.guess config.log config.status config.sub depcomp ltmain.sh
     - autoreconf --verbose --install --force
-    - ./configure --prefix=$PREFIX --enable-libusb
+    - ./configure --prefix=$PREFIX --sbindir=$PREFIX/bin
     - make install
     - make distcheck
-    - mv $PREFIX/sbin/* $PREFIX/bin/
   number: 0
   run_exports:
     - {{ pin_subpackage('libpcsclite') }}
@@ -39,7 +38,6 @@ requirements:
     - sysroot_linux-64 2.17  # [linux64]
   host:
     - libudev
-    - libusb
   run:
 
 test:
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -14,7 +14,7 @@ build:
   script:
     - ./bootstrap
     - ./configure  # [linux]
-    - ./configure  --disable-libsystemd # [osx]
+    - ./configure --disable-libsystemd  # [osx]
     - make
     # Test is here
     - make distcheck
@@ -29,6 +29,7 @@ requirements:
     - flex
     - libtool
     - libudev  # [linux]
+    - libudev-devel-cos6-x86_64  # [not linux]
     - libusb
     - make
     - pkg-config 
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -6,26 +6,29 @@ package:
   version: {{ version }}
 
 source:
-  url: https://salsa.debian.org/rousseau/PCSC/-/archive/{{ version }}/PCSC-{{ version }}.tar.gz
+  url: https://github.com/LudovicRousseau/PCSC/archive/refs/tags/{{ version }}.tar.gz
   sha256: 856a1daed9a5ae1ac265c75ce9f81743e66a972e5be5a1d3a2196ed23f75ea26
 
 build:
-  skip: True  # [not linux]
+  skip: True  # [win]
   script:
     - ./bootstrap
     - ./configure
     - make
+    # Test is here
+    - make distcheck
   number: 0
 
 requirements:
   build:
     - {{ compiler('c') }}
-    - {{ cdt('systemd-devel') }}
+    - {{ cdt('systemd-devel') }}  # [linux]
     - autoconf-archive
     - automake
     - flex
     - libtool
-    - libudev
+    - libudev  # [linux]
+    - libudev-devel-cos6-x86_64  # [not linux]
     - libusb
     - make
     - pkg-config 
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -12,8 +12,6 @@ source:
 build:
   skip: True  # [not linux]
   script:
-    - ls
-    - cd PCSC
     - ./bootstrap
     - ./configure
     - make
@@ -36,10 +34,12 @@ test:
 
 about:
   home: https://pcsclite.apdu.fr/
-  dev_url: https://pcsclite.apdu.fr/
+  dev_url: https://github.com/LudovicRousseau/PCSC
   summary: Middleware to access a smart card using SCard API (PC/SC).
   license: BSD-3-Clause and GPL-3.0-or-later and BSD-2-Clause and MIT
-  license_file: COPYING
+  license_file:
+    - COPYING
+    - GPL-3.0.txt
 
 extra:
   recipe-maintainers:
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -42,7 +42,7 @@ test:
 
 about:
   home: https://pcsclite.apdu.fr/
-  dev_url: https://github.com/LudovicRousseau/PCSC
+  dev_url: https://salsa.debian.org/rousseau/PCSC
   summary: Middleware to access a smart card using SCard API (PC/SC).
   license: BSD-3-Clause and GPL-3.0-or-later and BSD-2-Clause and MIT
   license_file:
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   skip: True  # [not linux]
   script:
+    - ls
     - cd PCSC
     - ./bootstrap
     - ./configure
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -16,7 +16,6 @@ build:
     - ./configure --prefix=$PREFIX  # [linux]
     - ./configure --prefix=$PREFIX --disable-libsystemd  # [osx]
     - make install
-    # Test is here
     - make distcheck  # [linux]
   number: 0
 
@@ -39,9 +38,7 @@ requirements:
 
 test:
   commands:
-    # Placeholder
-    - ls
-    - ls ..
+    - pcscd --version
 
 about:
   home: https://pcsclite.apdu.fr/
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -13,7 +13,8 @@ build:
   skip: True  # [win]
   script:
     - ./bootstrap
-    - ./configure
+    - ./configure  # [linux]
+    - ./configure  --disable-libsystemd # [osx]
     - make
     # Test is here
     - make distcheck
@@ -22,13 +23,12 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
-    - {{ cdt('systemd-devel') }}
+    - {{ cdt('systemd-devel') }}  # [linux]
     - autoconf-archive
     - automake
     - flex
     - libtool
     - libudev  # [linux]
-    - libudev-devel-cos6-x86_64  # [not linux]
     - libusb
     - make
     - pkg-config 
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -43,6 +43,13 @@ requirements:
 test:
   commands:
     - pcscd --version
+    - pcsc-spy --version
+    - test -f $PREFIX/include/PCSC/pcsclite.h
+    # Test specific SO versions to prevent future package ABI breaks
+    - test -f $PREFIX/lib/libpcsclite.so.1
+    - test -f $PREFIX/lib/libpcscspy.so.0
+    - test -f $PREFIX/lib/pkgconfig/libpcsclite.pc
+
 
 about:
   home: https://pcsclite.apdu.fr/
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -28,13 +28,13 @@ requirements:
     - automake
     - flex
     - libtool
-    - libudev  # [linux]
-    - libudev-devel-cos6-x86_64  # [not linux]
-    - libusb
     - make
     - pkg-config 
     - sysroot_linux-64 2.17  # [linux64]
   host:
+    - libudev  # [linux]
+    - libudev-devel-cos6-x86_64  # [not linux]
+    - libusb
   run:
 
 test:
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -22,7 +22,7 @@ build:
     - make distcheck
   number: 0
   run_exports:
-    - {{ pin_subpackage('libpcsclite') }}
+    - {{ pin_subpackage('pcsclite') }}
 
 requirements:
   build:
@@ -43,7 +43,6 @@ requirements:
 test:
   commands:
     - pcscd --version
-    - pcsc-spy --help
     - test -f $PREFIX/include/PCSC/pcsclite.h
     # Test specific SO versions to prevent future package ABI breaks
     - test -f $PREFIX/lib/libpcsclite.so.1
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -25,6 +25,7 @@ requirements:
     - libsystemd
     - libusb
     - make
+    - pkg-config 
   host:
   run:
 
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -17,6 +17,8 @@ build:
     - make install
     - make distcheck
   number: 0
+  run_exports:
+    - {{ pin_subpackage('libpcsclite') }}
 
 requirements:
   build:
@@ -49,4 +51,5 @@ about:
 
 extra:
   recipe-maintainers:
+    - leofang
     - thewchan
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -1,0 +1,45 @@
+{% set name = "libpcsclite" %}
+{% set version = "1.9.9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://salsa.debian.org/rousseau/PCSC/-/archive/{{ version }}/PCSC-{{ version }}.tar.gz
+  sha256: 856a1daed9a5ae1ac265c75ce9f81743e66a972e5be5a1d3a2196ed23f75ea26
+
+build:
+  skip: True  # [not linux]
+  script:
+    - cd PCSC
+    - ./bootstrap
+    - ./configure
+    - make
+  number: 0
+
+requirements:
+  build:
+    - autoconf-archive
+    - flex
+    - libsystemd
+    - libusb
+    - make
+  host:
+  run:
+
+test:
+  commands:
+    # Placeholder
+    - ls
+
+about:
+  home: https://pcsclite.apdu.fr/
+  dev_url: https://pcsclite.apdu.fr/
+  summary: Middleware to access a smart card using SCard API (PC/SC).
+  license: BSD-3-Clause and GPL-3.0-or-later and BSD-2-Clause and MIT
+  license_file: COPYING
+
+extra:
+  recipe-maintainers:
+    - thewchan
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -42,7 +42,7 @@ requirements:
 
 test:
   commands:
-    - pcscd --version
+    - /usr/sbin/pcscd --version
 
 about:
   home: https://pcsclite.apdu.fr/
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -19,6 +19,7 @@ build:
 
 requirements:
   build:
+    - {{ compiler('c') }}
     - autoconf-archive
     - automake
     - flex
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -32,7 +32,6 @@ requirements:
     - {{ cdt('systemd-devel') }}
     - autoconf-archive
     - automake
-    - file
     - flex
     - libtool
     - make
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -13,7 +13,7 @@ build:
   skip: True  # [win]
   script:
     - ./bootstrap
-    - ./configure --prefix=$PREFIX # [linux]
+    - ./configure --prefix=$PREFIX  # [linux]
     - ./configure --prefix=$PREFIX --disable-libsystemd  # [osx]
     - make install
     # Test is here
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -20,6 +20,7 @@ build:
     - ./configure --prefix=$PREFIX
     - make install
     - make distcheck
+    - mv $PREFIX/sbin/* $PREFIX/bin/
   number: 0
   run_exports:
     - {{ pin_subpackage('libpcsclite') }}
@@ -42,7 +43,7 @@ requirements:
 
 test:
   commands:
-    - /usr/sbin/pcscd --version
+    - pcscd --version
 
 about:
   home: https://pcsclite.apdu.fr/
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -43,7 +43,7 @@ requirements:
 test:
   commands:
     - pcscd --version
-    - pcsc-spy --version
+    - pcsc-spy --help
     - test -f $PREFIX/include/PCSC/pcsclite.h
     # Test specific SO versions to prevent future package ABI breaks
     - test -f $PREFIX/lib/libpcsclite.so.1
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -20,10 +20,11 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
+    - {{ cdt('systemd-devel-cos7-x86_64') }}
     - autoconf-archive
     - automake
     - flex
-    - libsystemd
+    # - libsystemd
     - libtool
     - libudev
     - libusb
--- a/recipes/libpcsclite/meta.yaml
+++ b/recipes/libpcsclite/meta.yaml
@@ -13,8 +13,8 @@ build:
   skip: True  # [win]
   script:
     - ./bootstrap
-    - ./configure  # [linux]
-    - ./configure --disable-libsystemd  # [osx]
+    - ./configure --prefix=$PREFIX # [linux]
+    - ./configure --prefix=$PREFIX --disable-libsystemd  # [osx]
     - make install
     # Test is here
     - make distcheck  # [linux]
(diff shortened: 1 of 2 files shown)


thewchan:
@BastianZim here. I'm particular unsure what I should do for testing

BastianZim:
Often, there is just a make check that's being run or upstream includes tests. 

I would see if you can add anything from here, as that is how upstream builds it: https://salsa.debian.org/rousseau/PCSC/-/blob/master/.github/workflows/build.yml

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  commands:
+    # Placeholder
+    - ls
+
+about:
+  home: https://pcsclite.apdu.fr/
+  dev_url: https://pcsclite.apdu.fr/
+  summary: Middleware to access a smart card using SCard API (PC/SC).
+  license: BSD-3-Clause and GPL-3.0-or-later and BSD-2-Clause and MIT
+  license_file: COPYING
```
BastianZim, line 42:
```suggestion
  license_file: 
    - GPL-3.0.txt
    - COPYING
```
There are some more here that probably need to be linked: https://github.com/LudovicRousseau/PCSC/blob/master/COPYING

thewchan: The text of all the linked license is in COPYING though, right?

BastianZim: Ahh true yes, read to quickly. Just including Copying and updating the license field is enough.

thewchan:
@BastianZim here. I'm particular unsure what I should do for testing

> Often, there is just a make check that's being run or upstream includes tests.
> 
> I would see if you can add anything from here, as that is how upstream builds it: https://salsa.debian.org/rousseau/PCSC/-/blob/master/.github/workflows/build.yml

Got it... looks like I'd need to also figure out what source files to include for test?

BastianZim:
Yes but sometimes the test is included in the build script because of what's needed to execute the tests. I would first focus on getting this built and then look at the tests later.

thewchan:
@BastianZim hmm, looks like the build worked?

thewchan:
@BastianZim @conda-forge/staged-recipes I'm able to get OSX build working, but the test doesn't work on that environment for some reason; something about "not being a git environment". Relatedly, I had to run the test during the build stage because I think otherwise I'll have to take the entire source distribution into the test environment, build again, and run the test there. Any pointers and review welcomed. Thanks!

BastianZim:
That is outside my area of expertise but maybe someone else can help.

thewchan:
@conda-forge/staged-recipes @conda-forge/help-c-cpp review requested

carterbox:
> something about "not being a git environment"

Can't help without the actual error message.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+  run:
+
+test:
+  commands:
+    # Placeholder
+    - ls
+    - ls ..
```
carterbox, line 44: This section should have existence tests for each of the artifact types: config files, libraries, headers, binaries.

Binaries should be tested by calling them with `--help` or `--version`.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - flex
+    - libtool
+    - libudev  # [linux]
+    - libudev-devel-cos6-x86_64  # [not linux]
+    - libusb
+    - make
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+  run:
```
carterbox, line 38:
```suggestion
requirements:
  build:
    - {{ compiler('c') }}
    - {{ cdt('systemd-devel') }}  # [linux]
    - autoconf-archive
    - automake
    - flex
    - libtool
    - make
    - pkg-config 
    - sysroot_linux-64 2.17  # [linux64]
  host:
    - libusb
    - libudev  # [linux]
    - libudev-devel-cos6-x86_64  # [not linux]
```

Probably this package depends on libusb and libudev as libraries (not build tools) so they should go in host.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+build:
+  skip: True  # [win]
+  script:
+    - ./bootstrap
+    - ./configure  # [linux]
+    - ./configure --disable-libsystemd  # [osx]
+    - make
+    # Test is here
+    - make distcheck  # [linux]
```
carterbox, line 20: I don't think you are actually installing anything to $PREFIX. You're just building and testing. Probably you need to call `make install`.

carterbox (changes requested):

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/LudovicRousseau/PCSC/archive/refs/tags/{{ version }}.tar.gz
+  sha256: 856a1daed9a5ae1ac265c75ce9f81743e66a972e5be5a1d3a2196ed23f75ea26
+
+build:
```
carterbox, line 12: There should also be a run_export section which pins this library according to it's ABI compatability.

carterbox (changes requested):

recipes/libpcsclite/meta.yaml
thewchan: I had to run the test suite in the build section because I can't figure out how do to so otherwise, any pointers how I might proceed?

recipes/libpcsclite/meta.yaml
thewchan: let me try this.

thewchan:
@BastianZim @carterbox thanks for your help; specifically @carterbox I will try to resolve your comments one at a time.

recipes/libpcsclite/meta.yaml
thewchan: @carterbox I'll try calling make install instead of make? And see what happens

thewchan: @carterbox I might need to use certain options to make install into $PREFIX, right now it's trying to install into /usr/bin?

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "libpcsclite" %}
+{% set version = "1.9.9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/LudovicRousseau/PCSC/archive/refs/tags/{{ version }}.tar.gz
```
leofang, line 9: Why don't we use the official releases as listed in https://pcsclite.apdu.fr/#download?

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  sha256: 856a1daed9a5ae1ac265c75ce9f81743e66a972e5be5a1d3a2196ed23f75ea26
+
+build:
+  skip: True  # [win]
+  script:
+    - ./bootstrap
+    - ./configure  # [linux]
+    - ./configure --disable-libsystemd  # [osx]
```
leofang, line 17: According to the official website, we should not build on macOS:
> Apple reimplemented PC/SC above their own CryptoTokenKit API.
> You should not need to build pcsc-lite on macOS.

https://pcsclite.apdu.fr/#supported

In fact that's what I see too with `openconnect` (I am the maintainer of openconnect-feedstock): On macOS, the Yubikey support (which would rely on pcsc on Linux) is built and enabled by default, without any effort.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - autoconf-archive
+    - automake
+    - flex
+    - libtool
+    - make
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+    - libudev  # [linux]
+    - libudev-devel-cos6-x86_64  # [not linux]
```
leofang, line 36: ditto, this can be removed.

recipes/libpcsclite/meta.yaml
leofang: just add this
```suggestion
    - pcscd --version
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  dev_url: https://github.com/LudovicRousseau/PCSC
+  summary: Middleware to access a smart card using SCard API (PC/SC).
+  license: BSD-3-Clause and GPL-3.0-or-later and BSD-2-Clause and MIT
+  license_file:
+    - COPYING
+    - GPL-3.0.txt
+
+extra:
+  recipe-maintainers:
+    - thewchan
```
leofang, line 57: Could you add me plz? 🙂
```suggestion
    - thewchan
    - leofang
```

leofang (changes requested):
Appreciate for the effort @thewchan! This PR shows up promptly just when I started looking into it. I left some comments. The main one is macOS builds should be skipped.

recipes/libpcsclite/meta.yaml
carterbox: You can run the test suite during the build phase (or not at all). The test phase of this recipe is for packaging tests (check that the artifacts are packaged correctly) that's why we want existence tests and checks that the CLIs of any binaries work.

recipes/libpcsclite/meta.yaml
carterbox: Yes, you need to install the build artifacts to $PREFIX.

thewchan: Right but... how do I do that?

leofang: I think when you call `./configure` you need to pass `--prefix=$PREFIX`, and then `make install` will work.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+test:
+  commands:
+    # Placeholder
+    - ls
+    - ls ..
+
+about:
+  home: https://pcsclite.apdu.fr/
+  dev_url: https://github.com/LudovicRousseau/PCSC
```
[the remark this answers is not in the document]
thewchan: I can point it back, I think at some point I was using something like that but @BastianZim advised me to switch to GH

recipes/libpcsclite/meta.yaml
thewchan: I can do this as well, and I think again as per @BastianZim I used the GH instead but yeah I can point it here instead

leofang: Ah let's wait for @BastianZim's reply before any change, then. It's unclear to me why we need this. The CF bot can detect new version updates even if the source is not from GH.

recipes/libpcsclite/meta.yaml
thewchan: I'll do the CLI test that @leofang suggested thanks!

thewchan: @carterbox if you have other suggestions please let me know as I'm kinda clueless...

recipes/libpcsclite/meta.yaml
thewchan: It worked! Thanks

recipes/libpcsclite/meta.yaml
thewchan: Do you have more guidance about this part? I'm unfamiliar

recipes/libpcsclite/meta.yaml
thewchan: I'll go ahead and change it for now, easy enough to revert if needed

recipes/libpcsclite/meta.yaml
thewchan: Got it, do you know if Win build is impossible/inappropriate? I sort of assumed based on what was required

thewchan:
@carterbox @leofang this is new, now it's complaining about `libtool` version I think? The version is stuck upstream right now waiting to be merged: https://github.com/conda-forge/libtool-feedstock/pull/29

recipes/libpcsclite/meta.yaml
leofang: According to the ABI lab (which is outdated) https://abi-laboratory.pro/index.php?view=timeline&l=pcsc-lite it's mostly backward compatible, so I guess this is enough:
```suggestion
build:
  run_exports:
    - {{ pin_subpackage('libpcsclite') }}
```
This ensures any packages built against `libpcsclite` v1.x will use v1.x.

recipes/libpcsclite/meta.yaml
leofang: ^^^^^^ @thewchan ?

recipes/libpcsclite/meta.yaml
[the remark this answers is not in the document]
leofang: ditto, not sure why GH is preferred in this case

recipes/libpcsclite/meta.yaml
thewchan: Yes sorry I will add this as well

leofang:
@thewchan the problem is not really in the package source, but in the `bootstrap` script. The script was not sufficient, it should have `--force` added to the last line when calling `autoreconf`:
https://salsa.debian.org/rousseau/PCSC/-/blob/master/bootstrap#L17
It'd be nice if you regenerate the configuration this way, skipping `./bootstrap`, since the green light with the GH artifact may be a false positive.

thewchan:
> @thewchan the problem is not really in the package source, but in the `bootstrap` script. The script was not sufficient, it should have `--force` added to the last line when calling `autoreconf`: https://salsa.debian.org/rousseau/PCSC/-/blob/master/bootstrap#L17 It'd be nice if you regenerate the configuration this way, skipping `./bootstrap`, since the green light with the GH artifact may be a false positive.

Okay, let me try using those lines verbatim instead of `./bootstrap`.

thewchan:
@leofang it doesn't seem like the CLI test is working, either I didn't build the entry points correctly or there's something else amiss: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=632472&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=2492

thewchan:
@leofang I'm checking the logs to make sure it installed properly in `$PREFIX`, I think at least it tried to? https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=632472&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1198

thewchan:
Maybe I need to so something like `mv $PREFIX/sbin/* $PREFIX/bin/`? I see that in the wild. I'll try it

thewchan:
> Maybe I need to so something like `mv $PREFIX/sbin/* $PREFIX/bin/`? I see that in the wild. I'll try it

https://github.com/conda-forge/slurm-feedstock/blob/3ecf889c74e52259b31e24ec58f5f6f518835be4/recipe/build.sh#L3

thewchan:
@carterbox @leofang Okay it runs now! It possibly needs more test, and I also wonder if a Win build is possible. But thanks so much for your help!

recipes/libpcsclite/meta.yaml
leofang: I don't think this works on Windows. Even if it does, we can create a windows build later.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+build:
+  skip: True  # [not linux]
+  script:
+    - set -e
+    - set -x
+    - if test -f Makefile; then make distclean; fi
+    - rm -rf *.cache *.m4 config.guess config.log config.status config.sub depcomp ltmain.sh
+    - autoreconf --verbose --install --force
+    - ./configure --prefix=$PREFIX
```
leofang, line 20: I noticed the build log says libusb support is not built, and I think this is why:
```suggestion
    - ./configure --prefix=$PREFIX --enable-libusb
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - make install
+    - make distcheck
+    - mv $PREFIX/sbin/* $PREFIX/bin/
```
leofang, line 23: This seems fragile, let me see if there's a better way.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ cdt('systemd-devel') }}
+    - autoconf-archive
+    - automake
+    - flex
+    - libtool
+    - make
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
```
leofang, line 38: I noticed `configure` tried to call `/usr/bin/file` and failed:
```suggestion
    - file
    - sysroot_linux-64 2.17  # [linux64]
```

thewchan:
@leofang it seems that we can't build with libudev and libusb at the same time, we have to pick one:
 https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=632569&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1158

leofang:
Yeah... I saw that too... Did a quick search, which to pick doesn't seem important (it is the default of PCSCLite, and both support hot plug), so how about just sticking to what we know works (`libudev`)?

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+build:
+  skip: True  # [not linux]
+  script:
+    - set -e
+    - set -x
+    - if test -f Makefile; then make distclean; fi
+    - rm -rf *.cache *.m4 config.guess config.log config.status config.sub depcomp ltmain.sh
+    - autoreconf --verbose --install --force
+    - ./configure --prefix=$PREFIX --enable-libusb
```
leofang, line 20: Remove `libusb` build and fix the sbin folder:
```suggestion
    - ./configure --prefix=$PREFIX --sbindir=$PREFIX/bin
```

recipes/libpcsclite/meta.yaml
leofang: See above comment, this is no longer needed.
```suggestion
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - automake
+    - file
+    - flex
+    - libtool
+    - make
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+    - libudev
+    - libusb
```
leofang, line 42: ditto
```suggestion
```

leofang (commented):
btw @thewchan A quick tip in case you don't know already: You can just apply suggestions in batch on the GH web interface,  so you can quickly do the work everywhere (cell phone, laptop, ...) without editing in your local git repo 🙂

thewchan:
@carterbox @leofang We're back to passing! Any other tests we should run?

leofang (approved):
LGTM! Thanks @thewchan! 

@carterbox @conda-forge/help-c-cpp This is ready.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - make
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+    - libudev
+  run:
+
+test:
+  commands:
+    - pcscd --version
```
leofang, line 45: Maybe one more test?
```suggestion
    - pcscd --version
    - test -f $PREFIX/include/pcsclite.h
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - set -x
+    - if test -f Makefile; then make distclean; fi
+    - rm -rf *.cache *.m4 config.guess config.log config.status config.sub depcomp ltmain.sh
+    - autoreconf --verbose --install --force
+    - ./configure --prefix=$PREFIX --sbindir=$PREFIX/bin
+    - make install
+    - make distcheck
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('libpcsclite') }}
```
carterbox, line 25: Most of the time ABI compatability does not sync with the API version. Do you have any documentation that they only break the ABI at major API releases?

Ask upstream whether they only do ABI breaks at certain API releases.

```suggestion
    - {{ pin_subpackage('libpcsclite', max_pin='x.x.x') }}
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,66 @@
+{% set name = "libpcsclite" %}
```
carterbox, line 1:
```suggestion
{% set name = "pcsclite" %}
```
This package also contains binaries with CLI. libpcsclite would be the subpackage of this recipe with only the library/headers. You can split the package into multiple outputs now if you want, but it's not necessary.

carterbox (changes requested):

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - pkg-config 
+    - sysroot_linux-64 2.17  # [linux64]
+  host:
+    - libudev
+  run:
+
+test:
+  commands:
+    - pcscd --version
+    - pcsc-spy --version
```
carterbox, line 46:
```suggestion
    - pcsc-spy --help
```

recipes/libpcsclite/meta.yaml
thewchan: @leofang What do  you think?

leofang: @carterbox i think it'd be too strict, see my earlier comment: https://github.com/conda-forge/staged-recipes/pull/21405#discussion_r1054838829 that said I would not mind if you wanna be on the cautious side

leofang: btw

> Most of the time ABI compatability does not sync with the API version.

I am not sure this is true for traditional Linux packages...

recipes/libpcsclite/meta.yaml
leofang: Yes indeed.

recipes/libpcsclite/meta.yaml
carterbox: > @carterbox i think it'd be too strict, see my earlier comment: [#21405 (comment)](https://github.com/conda-forge/staged-recipes/pull/21405#discussion_r1054838829) that said I would not mind if you wanna be on the cautious side

All of the samples on ABI lab are v1.8, so there is not enough evidence to say whether they would break the ABI at a minor API change. I think the breaking ABI changes from 1.8.9 to 1.8.10 were an accident. I still recommend asking upstream, but pinning to `'x.x'` is probably? safe.

leofang:
I feel pcsc-spy should be temporarily removed from the package. It should be made a stand-alone Python noarch package that depends on libpcsclite, but since we have not set it up, it'd take a while to do. I'd prefer to not have it for now.

recipes/libpcsclite/meta.yaml
leofang: @carterbox good catch for adding pcsc-spy, I don't even know it exists...

recipes/libpcsclite/meta.yaml
leofang: Sounds good to me!

thewchan:
Getting this now `/home/conda/staged-recipes/build_artifacts/libpcsclite_1671750348569/test_tmp/run_test.sh: /home/conda/staged-recipes/build_artifacts/libpcsclite_1671750348569/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/bin/pcsc-spy: /usr/bin/python3: bad interpreter: No such file or directory`

leofang:
Yes that's what I meant. We got this error because Python was not installed in the test environment (it's not a dependency as we missed it). I didn't know `pcsc-spy` exists. It's a simple Python script that can be invoked from cmdline. But then it means Python becomes a required run dependency for `pcsclite`, which is not ideal.

recipes/libpcsclite/meta.yaml
leofang:
```suggestion
    - {{ pin_subpackage('libpcsclite', max_pin='x.x') }}
```

thewchan:
@carterbox @leofang okay I think i've made the requested change, name of package and removal of pcsc-spy.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  skip: True  # [not linux]
+  script:
+    - set -e
+    - set -x
+    - if test -f Makefile; then make distclean; fi
+    - rm -rf *.cache *.m4 config.guess config.log config.status config.sub depcomp ltmain.sh
+    - autoreconf --verbose --install --force
+    - ./configure --prefix=$PREFIX --sbindir=$PREFIX/bin
+    - make install
+    - make distcheck
```
leofang, line 22:
```suggestion
    - make distcheck
    - rm $PREFIX/bin/pcsc-spy
    - rm $PREFIX/lib/libpcscspy.so*
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - libudev
+  run:
+
+test:
+  commands:
+    - pcscd --version
+    - test -f $PREFIX/include/PCSC/pcsclite.h
+    # Test specific SO versions to prevent future package ABI breaks
+    - test -f $PREFIX/lib/libpcsclite.so.1
+    - test -f $PREFIX/lib/libpcscspy.so.0
```
leofang, line 49:
```suggestion
```

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - test -f $PREFIX/lib/pkgconfig/libpcsclite.pc
```
leofang, line 50: @carterbox why do we care about this? It has the host prefix hard-coded which is useless.

recipes/libpcsclite/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  run_exports:
+    - {{ pin_subpackage('pcsclite') }}
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ cdt('systemd-devel') }}
+    - autoconf-archive
+    - automake
+    - file
```
leofang, line 33: I struggled to understand why this is still not used, but couldn't figure it out, but it seems no harm.
```suggestion
```

leofang (changes requested):
@thewchan I think removing the test for `pcsc-spy` is not enough. We need to remove it completely.

This is how `pcsc-spy` works: 
- the user do `LD_PRELOAD=libpcscspy.so` which wraps all symbols from `libpcsclite.so` to dump a log to disk
- then, calling `pcsc-spy` separately would launch workers to read the log file

Basically, it is a log monitor that does not provide real functionality. To keep things cleanly separated we should not include the relevant files until we offer a separate subpackage (that depends on Python).

recipes/libpcsclite/meta.yaml
carterbox: Generally checking for CMake/Package config files indicates that they are being shipped without needing to download the package and look. Downstream builds can break if they depend on config files which are here today and removed later.

What do you mean useless? [Paths like `/home/conda/staged-recipes/build_artifacts/pcsclite_1672093677527/_h_env_placehold_placehold_placehold_` are replaced by conda with `$CONDA_PREFIX` at install-time.](https://docs.conda.io/projects/conda-build/en/latest/resources/make-relocatable.html)

thewchan:
@carterbox @leofang I'll let you all duke it out I don't mind one way or another re: the tests. Just let me know.

recipes/libpcsclite/meta.yaml
leofang: You're right, I was being silly about the placeholder prefix. 

Still, I rarely see recipes check for .pc files (plz don't mix/compare them with cmake files; they are of vastly different values) and I can raise many counter examples, so if you think it's the standard practice it'd be better to document it.

leofang (approved):

leofang:
Friendly nudge @carterbox 🙂

recipes/libpcsclite/meta.yaml
carterbox: I'm not commenting on the relative merits of CMake vs Package Config

If upstream developers of this package want to use PC, then we should ship them too. Testing whether they are there indicates to downstream users that we are including all of upstream's artifacts.

carterbox (changes requested):

thewchan:
Wow can't believe this is done! Thank you so much @carterbox and @leofang for your help!

PR #21408, opened 2022-12-02
Add ott-jax recipe

michalk8:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/ott-jax/meta.yaml
+++ b/recipes/ott-jax/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   number: 0
   script: {{ PYTHON }} -m pip install . -vv
-  skip: true  # [py<37]
+  skip: true  # [py<37 or win]
 
 requirements:
   host:
--- a/recipes/ott-jax/meta.yaml
+++ b/recipes/ott-jax/meta.yaml
@@ -11,16 +11,16 @@ source:
 
 build:
   number: 0
+  noarch: python
   script: {{ PYTHON }} -m pip install . -vv
-  skip: true  # [py<37 or win]
 
 requirements:
   host:
-    - python
+    - python >=3.7
     - pip
     - setuptools_scm
   run:
-    - python
+    - python >=3.7
     - jax >=0.1.67
     - jaxlib >=0.1.47
     - numpy >=1.18.4,!=1.23.0
@@ -28,8 +28,8 @@ requirements:
     - flax >=0.5.2
     - optax >=0.1.1
     - scipy >=1.7.0
-    - importlib-metadata >=1.0  # [py<38]
-    - typing-extensions  # [py<38]
+    - importlib-metadata >=1.0
+    - typing-extensions
 
 test:
   # `pip check` is disabled because the release
--- a/recipes/ott-jax/meta.yaml
+++ b/recipes/ott-jax/meta.yaml
@@ -1,0 +1,55 @@
+{% set name = "ott-jax" %}
+{% set version = "0.3.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ott-jax-{{ version }}.tar.gz
+  sha256: adcd3c59ac9223a962b7a9244c4f53cc9857bf2461530583edb1ea8a1fc37ea3
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [py<37]
+
+requirements:
+  host:
+    - python
+    - pip
+    - setuptools_scm
+  run:
+    - python
+    - jax >=0.1.67
+    - jaxlib >=0.1.47
+    - numpy >=1.18.4,!=1.23.0
+    - matplotlib-base >=3.0.0
+    - flax >=0.5.2
+    - optax >=0.1.1
+    - scipy >=1.7.0
+    - importlib-metadata >=1.0  # [py<38]
+    - typing-extensions  # [py<38]
+
+test:
+  imports:
+    - ott
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/ott-jax/ott
+  license: Apache-2.0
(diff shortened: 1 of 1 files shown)


michalk8:
@marcocuturi are you willing to also maintain this recipe?

michalk8:
Removed the `pip check` since previously, we had it in requirements, but was only used when testing. Will re-introduce once we make a new release on PyPI.

marcocuturi:
yes!

michalk8:
@conda-forge/python ready for review!

recipes/ott-jax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ott-jax-{{ version }}.tar.gz
+  sha256: adcd3c59ac9223a962b7a9244c4f53cc9857bf2461530583edb1ea8a1fc37ea3
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [py<37 or win]
```
ocefpaf, line 15: This one can probably be noarch if the Windows failures are only missing packages on that platform.

michalk8: I don't think this will work, I remember that in the past when submitting a recipe for a different package, having `skip: true # [win]`, the following issue was raised: 
```
    noarch packages can't have selectors. If the selectors are necessary, please remove noarch: python.
```

ocefpaf: Making it noarch means removing that skip. Like I said above, if the Windows failures is only a missing package, it is OK b/c the package won't be installed there but it will be ready if/when the missing dependency is available.

recipes/ott-jax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - matplotlib-base >=3.0.0
+    - flax >=0.5.2
+    - optax >=0.1.1
+    - scipy >=1.7.0
+    - importlib-metadata >=1.0
+    - typing-extensions
+
+test:
+  imports:
+    - ott
```
ocefpaf, line 36:
```suggestion
test:
  imports:
    - ott
  commands:
    - pip check
  requires:
    - pip
```

recipes/ott-jax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - python >=3.7
+    - jax >=0.1.67
+    - jaxlib >=0.1.47
+    - numpy >=1.18.4,!=1.23.0
+    - matplotlib-base >=3.0.0
+    - flax >=0.5.2
+    - optax >=0.1.1
+    - scipy >=1.7.0
+    - importlib-metadata >=1.0
+    - typing-extensions
```
ocefpaf, line 32:
```suggestion
    - absl-py >=0.7.0
    - jax >=0.1.67
    - jaxlib >=0.1.47
    - numpy >=1.18.4,!=1.23.0
    - matplotlib-base >=2.0.1
    - flax >=0.3.6
    - optax >=0.1.1
    - scipy >=1.7.0
    - scikit-learn >=1.0
    - importlib-metadata >=1.0
    - typing-extensions
```

michalk8: I removed the `absl-py/scikit-learn`, since they were both used only during testing (unfortunately, they were still present in the previous release on PyPI); this is also what prevents currently from having the `pip check`, as suggested in https://github.com/conda-forge/staged-recipes/pull/21408#discussion_r1038303722.
Imho it's better to not include them here rather than making a new release with minor changes, but don't have strong opinion.

recipes/ott-jax/meta.yaml
michalk8: `jaxlib` seems to be the cause of issues (not supported on Win) and the build fails.
I'd revert back to the selectors unless you have a suggestion as how to fix this.

recipes/ott-jax/meta.yaml
ocefpaf: Please add a comment about that and comment out the pip-check so  we can enable them when there is a new release.

michalk8: Done.

recipes/ott-jax/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - python
+    - jax >=0.1.67
+    - jaxlib >=0.1.47
+    - numpy >=1.18.4,!=1.23.0
+    - matplotlib-base >=3.0.0
+    - flax >=0.5.2
+    - optax >=0.1.1
+    - scipy >=1.7.0
+    - importlib-metadata >=1.0  # [py<38]
+    - typing-extensions  # [py<38]
```
ocefpaf, line 32: It is fine to drop these selectors here for the benefit of keeping it noarch.

```suggestion
    - importlib-metadata >=1.0
    - typing-extensions
```

michalk8: Agree, but even with `noarch: python`, it was the CI for Windows failed.

recipes/ott-jax/meta.yaml
dopplershift: @michalk8 You misunderstand. Go ahead and remove the selectors. We can ignore the CI failure that results because it's only due to a missing dependency on Windows. In practice, noarch packages are only built (usually) on Linux, which will succeed. The package just won't be installable on Windows until (somehow) `jaxlib` packages become available. That's fine.

Keeping this noarch will greatly simplify maintenance and drastically reduce the number of builds.

michalk8: Thanks; I've been trying to figure out what's exactly wrong with `noarch` in our case (since in case for [optax](https://github.com/conda-forge/optax-feedstock/blob/main/recipe/meta.yaml), it seems to cause no problems)
Any help would be much appreciated; my guess are the pins/python version?, but doesn't seem right.

dopplershift: That error is due to there being no `jaxlib` packages available on Windows.

dopplershift (approved):

Issue #4584, opened 2017-12-11
Staged recipes not in `recipes` directory


synapticarbors:
While reviewing a recipe (and subsequently merging it), I failed to notice that the recipe was in the top level directory rather than in `recipes`. Talking with @jakirkham it seems like this is a common problem. 

I suggested that perhaps the linter should be able to automatically catch this common error and wanted to raise the issue for further discussion here. 

@conda-forge/core

jakirkham:
Generally the fix can go in one of three places (potentially in some combination of these). In staged-recipes CIs themselves, in `conda-smithy` (the CLI for the linter lives there), in `conda-forge-webservices` (this runs the linting bot). Would have a slight preference towards solving this in CIs at `staged-recipes` as the issue is localized here. Though could also see value in adding something to the bot (i.e. `conda-forge-webservices`) so that we spell out the problem to the user (instead of having them dig through CI logs). Either way `conda-smithy` itself feels like the wrong place for this as it has no real concept of `staged-recipes`. Other thoughts?

CJ-Wright:
I'd rather not try to fish this out of the CI dump (unless we tried to make it super obvious, eg the error is done before any major output is accumulated).

BastianZim:
Something like this?

https://github.com/conda-forge/staged-recipes/pull/21451

Haven't properly worked on the text yet. It's based on the linter we had in the docs. https://github.com/conda-forge/conda-forge.github.io/pull/1535

cc @beckermr @jakirkham

PR #21451, opened 2022-12-06
Create directory linter

BastianZim:
Context in  #4584

### DIFF
--- a/.github/workflows/correct_directory.yml
+++ b/.github/workflows/correct_directory.yml
@@ -22,9 +22,9 @@ name: directory_linter
                owner: context.repo.owner,
                repo: context.repo.repo,
                body: 'Hi! Thanks for your contribution to conda-forge.' +
-                     '\nUnfortunately, the recipe was added directly in the _recipes_ folder without its own subfolder.\n' +
-                     'Please move the recipe file (_<your_package.yaml>_) into a folder with the name of the package you want to submit.\n\n' +
-                     'For example: if your recipe is currently under `recipes/<your_package>.yaml', '
+                     '\nUnfortunately, the recipe was added directly in the `recipes` folder without its own subfolder.\n' +
+                     'Please move the recipe file (`<your_package.yaml>`) into a folder with the name of the package you want to submit.\n\n' +
+                     'For example: if your recipe is currently under `recipes/<your_package>.yaml`, '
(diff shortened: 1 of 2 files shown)


.github/workflows/correct_directory.yml
```diff
@@ -0,0 +1,28 @@
...
+         with:
+           script: |
+             github.rest.issues.createComment({
+               issue_number: context.issue.number,
+               owner: context.repo.owner,
+               repo: context.repo.repo,
+               body: 'Hi! Thanks for your contribution to conda-forge.' +
+                     '\nUnfortunately, the recipe was added directly in the _recipes_ folder without its own subfolder.\n' +
+                     'Please move the recipe file (_your package.yaml_) into a folder with the name of the package you want to submit.\n' +
+                     'Thanks!'
```
BastianZim, line 27: Bit confusing but not sure how to word it better.

dopplershift:
I'm not sure how to word it better, but I like the change.

BastianZim:
Added a second common case

beckermr:
Do we want this in smithy?

BastianZim:
I followed https://github.com/conda-forge/staged-recipes/issues/4584#issuecomment-350613234 for guidelines on that, and the issue is very localised to here but also happy to close the PR.

.github/workflows/correct_directory.yml
beckermr: Maybe use paths? So you could say "recipe was added at `recipes/<feedstock-name>.yaml` and it should be at `recipes/<feedstock-name>/meta.yaml`?

BastianZim: Good idea. Let me update that.

BastianZim: Do we want `<feedstock-name>` or `your_package`?  `<feedstock-name>` is more consistent but maybe not something everyone knows?

jakirkham:
> Do we want this in smithy?

Wondered about this as well. Though maybe this is less likely to show up in a feedstock since it already has the expected structure unlike staged-recipes where users need to figure it out a bit. So maybe it is ok to keep local to `staged-recipes`. We can always revisit 🙂

jakirkham:
Thanks all! 🙏

Hopefully this helps users that bump their heads on this one

PR #21500, opened 2022-12-10
Openmetadata airflow managed apis

rxm7706:
Checklist
- [x] Build number is 0.


rxm7706:

rxm7706:
@xylar please opine at your convenience. 

OpenMetadata is going to reach 1.0 in a few months. 
- https://docs.open-metadata.org/overview/roadmap 
It would be beneficial to the airflow community to have the integration with OpenMetadata 
- https://github.com/open-metadata/OpenMetadata/blob/main/openmetadata-airflow-apis/README.md
- This is a plugin for Apache Airflow >= 1.10 and Airflow >=2.x that exposes REST APIs to deploy an OpenMetadata workflow definition and manage DAGS and tasks.

Two requests, 
1) Can you review this draft please, the only error is Windows Build Failure with Airflow (Expected)
The reported errors are:
-  nothing provides requested apache-airflow >=2.1.4
2) Would you want to be added as a maintainer for the core packages.
- openmetadata-managed-apis
- openmetadata-ingestion
- openmetadata-airflow-managed-apis

Thank you,

recipes/commonregex/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+{% set version = "1.5.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/madisonmay/CommonRegex
+  git_tag: master
+  git_depth: 1
```
xylar, line 11: This, I'm afraid, is not going to be allowed.  It would also be an absolute nightmare to maintain (which is why it's not allowed).

xylar: Could you ask the repo owner(s) to make tags each time they update the version number?

xylar (changes requested):
I think the main problem is that you absolutely can't get source from `git` and especially not a branch that is a moving target, see below.  That will be a non-starter for any reviewer, I'm quite sure.

> Would you want to be added as a maintainer for the core packages.

I'd prefer not to be a maintainer for now -- I think you'll have these well in hand.  But let me know if you end up needing help and you can add me later.

recipes/commonregex/meta.yaml
rxm7706: Thank you for the help @xylar - let me try.
https://pypi.org/project/commonregex/#history
The last update to this project on pypi was in 2014 ! https://pypi.org/project/commonregex/1.5.4/
So I don't know how much luck I will have with the maintainers, I might have better luck with the OpenMetadata team, if they are interested in supporting conda - this one package might be an issue they might want to solve for on their end. 

One thing I did notice is they do have a packaged distributable on repo
https://github.com/madisonmay/CommonRegex/tree/master/dist
But I could not figure out how to get an static SHA from it, and I was wondering if that was allowed anyway - its not a tag or a release... just packaged and put into a folder.

xylar: Can you get the source from PyPI instead?

xylar: > One thing I did notice is they do have a packaged distributable on repo

No, I don't think that solves the problem.

xylar: I take that back.  I think you could use:
https://raw.githubusercontent.com/madisonmay/CommonRegex/master/dist/commonregex-1.5.5.tar.gz
as the source URL.  It's not great but it's what there seems to be.  As long as you get the sha256 for that file:
```
08e93aec3ce2a3ae29d89cc249af47e70df10218f2161455b3df19fb5cb67d00
```
and it stays the same, I think that should work.

xylar:
```suggestion
  url: https://raw.githubusercontent.com/madisonmay/CommonRegex/master/dist/commonregex-1.5.5.tar.gz
  sha256: 08e93aec3ce2a3ae29d89cc249af47e70df10218f2161455b3df19fb5cb67d00
```

xylar (approved):
Looks good to me.

rxm7706:
@conda-forge/staged-recipes  ready for review

xylar (approved):
Looks okay to me.

xhochy (changes requested):
Please don't put in an upper bound if upstream doesn't do.

rxm7706:
> Please don't put in an upper bound if upstream doesn't do.

@xhochy thank you, I was not aware that the best practice was to leave it unbounded and match upstream, I assumed it was always better to have an upper bound - I will update the recipes. Once again - thank you for reviewing this and the comments.

xhochy:
We only add upper bounds if upstream does it. Otherwise, we expect that it will work with new Python versions.

xhochy (approved):

PR #21505, opened 2022-12-10
Add qt-base and other packages for qt-main

hmaarrfk:
@jschueller my plan is to adapt the build script from qt-webengine, hopefully make it generic enough, so it can be used for all the qt-stuff.

* https://github.com/conda-forge/qt-main-feedstock/issues/78
* Rendered feedstocks: https://github.com/orgs/qtforge/repositories
* Azure CIs: https://dev.azure.com/markharfouche0439/feedstocks/_build

Notable changes
- `krb5` and `glib` are now only dependencies for linux. These were found to provide warnigns like: `in requirements but not used`. 

Libraries to clean up:

- [ ] `qt-base` warnings regarding overlinking to libclang on linux. 
- [ ] `qt-svg` cross compiling on osx-arm64


hmaarrfk:
@jschueller is this effort futile for qt 5.15? Should I just concentrate instead on doing this for qt 6?

jschueller:
> @jschueller is this effort futile for qt 5.15? Should I just concentrate instead on doing this for qt 6?

Though question, if the aarch64 builds can be automated with this, maybe its worthwhile because that was the thing which slowed down the updates.

recipes/xorg-libxxf86vm/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://www.x.org/releases/individual/{{ xorg_category }}/{{ xorg_name }}-{{ version }}.tar.xz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
```
hmaarrfk, line 17:
```suggestion
  skip: true   # [not unix]
  number: 0
```

hmaarrfk:
Generally speaking i'm less interested in this for qt6.

PR #21566, opened 2022-12-17
add pyg-lib

hadim:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pyg-lib/build.sh
+++ b/recipes/pyg-lib/build.sh
@@ -1,0 +1,19 @@
+#!/bin/bash
+set -ex
+
+git submodule sync --recursive
+git submodule update --init --recursive
+
+if [[ ${cuda_compiler_version} != "None" ]]; then
+  export WITH_CUDA=ON
+  export FORCE_CUDA=1
+  export CUDA_ARCH_LIST=75
+else
+  export WITH_CUDA=OFF
+  export FORCE_CUDA=0
+fi
+
+export USE_MKL_BLAS=1
+export Torch_DIR=$(python -c 'import torch; print(torch.utils.cmake_prefix_path)')
+
+pip install . -vvv
--- a/recipes/pyg-lib/build.sh
+++ b/recipes/pyg-lib/build.sh
@@ -7,7 +7,7 @@ else
   export FORCE_CUDA=0
 fi
 
-export USE_MKL_BLAS=1
+# export USE_MKL_BLAS=1  # only used for >0.1.0
 export Torch_DIR=$(python -c 'import torch; print(torch.utils.cmake_prefix_path)')
 
 export FORCE_NINJA=1
--- a/recipes/pyg-lib/build.sh
+++ b/recipes/pyg-lib/build.sh
@@ -11,6 +11,7 @@ fi
 export USE_MKL_BLAS=1
 export Torch_DIR=$(python -c 'import torch; print(torch.utils.cmake_prefix_path)')
 
+export FORCE_NINJA=1
 export EXTERNAL_PHMAP_INCLUDE_DIR="${BUILD_PREFIX}/include/"
 export EXTERNAL_CUTLASS_INCLUDE_DIR="${BUILD_PREFIX}/include/"
 
--- a/recipes/pyg-lib/build.sh
+++ b/recipes/pyg-lib/build.sh
@@ -3,7 +3,6 @@ set -ex
 
 if [[ ${cuda_compiler_version} != "None" ]]; then
   export FORCE_CUDA=1
-  export CUDA_ARCH_LIST=75
 else
   export FORCE_CUDA=0
 fi
--- a/recipes/pyg-lib/build.sh
+++ b/recipes/pyg-lib/build.sh
@@ -1,19 +1,17 @@
 #!/bin/bash
 set -ex
 
-git submodule sync --recursive
-git submodule update --init --recursive
-
 if [[ ${cuda_compiler_version} != "None" ]]; then
-  export WITH_CUDA=ON
   export FORCE_CUDA=1
   export CUDA_ARCH_LIST=75
 else
-  export WITH_CUDA=OFF
   export FORCE_CUDA=0
 fi
 
 export USE_MKL_BLAS=1
 export Torch_DIR=$(python -c 'import torch; print(torch.utils.cmake_prefix_path)')
+
+export EXTERNAL_PHMAP_INCLUDE_DIR="${BUILD_PREFIX}/include/"
+export EXTERNAL_CUTLASS_INCLUDE_DIR="${BUILD_PREFIX}/include/"
 
 pip install . -vvv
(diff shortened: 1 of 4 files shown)


hadim:
windows and osx are expected to fail.

hadim:
@conda-forge/help-python-c: please review

I already asked upstream at https://github.com/pyg-team/pyg-lib/issues/163 to modify the build logic in order to be able to use the conda packages for `cutlass` and `parallel-hashmap`.

hadim:
Blocking for https://github.com/conda-forge/pytorch_geometric-feedstock/pull/29

hadim:
@conda-forge/help-python-c : ready for review.

recipes/pyg-lib/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+  host:
+    - python
+    - pip
+    - pytorch
+    - pytorch =*={{ torch_proc_type }}*
+  run:
+    - python
+    - parallel-hashmap
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
```
xhochy, line 48: These should be added automatically via `run_exports`.
```suggestion
```

recipes/pyg-lib/build.sh
```diff
@@ -0,0 +1,18 @@
...
+set -ex
+
+if [[ ${cuda_compiler_version} != "None" ]]; then
+  export FORCE_CUDA=1
+  export CUDA_ARCH_LIST=75
+else
+  export FORCE_CUDA=0
+fi
+
+export USE_MKL_BLAS=1
```
xhochy, line 11: Does this force linkage against MKL?

hadim: I am not sure how to checl that. Here is the relevant CMakeFiles piece:

```
set(WITH_MKL_BLAS 0)
if(USE_MKL_BLAS AND DEFINED BLAS_INCLUDE_DIR)
  find_file(MKL_INCLUDE_FOUND mkl.h ${BLAS_INCLUDE_DIR} NO_DEFAULT_PATH)
  if(MKL_INCLUDE_FOUND)
    set(WITH_MKL_BLAS 1)
  else()
    if(WITH_COV)
      message(FATAL_ERROR "The mkl.h file was not found - pass the correct directory or set USE_MKL_BLAS=OFF.")
    else()
      message(WARNING "The mkl.h file was not found - building pyg-lib without MKL BLAS support.")
    endif()
  endif()
endif()
```

Then `WITH_MKL_BLAS` is used within the C++ code to use or not MKL at https://github.com/pyg-team/pyg-lib/blob/0a2300ba8db276fc9140a415d4b65c9141135a8f/pyg_lib/csrc/ops/cpu/matmul_kernel.cpp#L15

xhochy: The looks very much as this would introduce a direct dependency on MKL. I'm no MKL expert but that doesn't look like it being compatible with other BLAS implementations. The build works as it seems that some other process is also pulling in MKL. 

There doesn't seem to be any direct linkage to MKL at least reported.

hadim: > but that doesn't look like it being compatible with other BLAS implementations.

That sounds like an issue. Is it?

> There doesn't seem to be any direct linkage to MKL at least reported.

does that mean the current build is fine?

---

Should we ping someone else here?

xhochy: Yes but I'm unsure who.

hadim: @conda-forge/blas: what do you think of the above? It seems unclear to us whether the above MKL integration is ok or not.

h-vetinari: I haven't looked at the recipe here in detail, but first as a general comment: we usually try to keep things compatible with several blas implementations, by ensuring we only compile against reference lapack (which can then be replaced by a given flavour in the final environment).

However, there are a handful of packages I'm aware of that hard-require mkl. I think we should strive for the former, but it's not impossible to do the latter.

hadim: Understood and thanks for the context. According to https://github.com/pyg-team/pyg-lib/blob/fee5fa9cbd6ca559058831e61fec7ee4f32ae459/pyg_lib/csrc/ops/cpu/matmul_kernel.cpp#L16, mkl.h seems hard coded indeed in the source code. If you're fine I'll be to merge here. I can also raise a ticket upstream if you prefer and hold for it.

xhochy: Fine with merging it but then we should also list MKL as an explicit dependency.

hadim: Ok done

hadim:
@xhochy @h-vetinari happy to merge here?

recipes/pyg-lib/build.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+set -ex
+
+if [[ ${cuda_compiler_version} != "None" ]]; then
+  export FORCE_CUDA=1
+  export CUDA_ARCH_LIST=75
```
carterbox, line 6: Please target the same CUDA architectures as pytorch since you are linking/depending on it. Compiling for only 75 would exclude a lot of users.

https://github.com/conda-forge/pytorch-cpu-feedstock/blob/main/recipe/build_pytorch.sh#L94-L111

recipes/pyg-lib/build.sh
carterbox: I see from the code linked above that mkl.h is conditionally included, but are you sure that `WITH_MKL_BLAS()` is evaluating to TRUE with your build? The link inspector on all platforms reports that MKL is unused (not linked dynamically).

carterbox (changes requested):

recipes/pyg-lib/build.sh
hadim: Thanks. I was simply using the upstream recommendation.

Maybe I can simply not provide any `CUDA_ARCH_LIST` variable instead of hard-coding it from your above example.

I will remove it. Please let me know if it's ok or if you prefer me to hard-code those.

recipes/pyg-lib/build.sh
hadim: It's my bad here, I just realized that MKL has only been added for `>0.1.0` (likely for the coming `0.2.0`). I will simply comment the MKL relevant bits for now and uncomment them when 0.2.0 is out in the feedstock.

hadim: Done.

hadim:
@carterbox the builds are all passing.

carterbox:
In the future, consider splitting the compiled and python code into separate packages to reduce the size of the build matrix. PYTHON_VERSIONS x CUDA_VERSIONS x ARCHS is a large matrix.

hadim:
Thanks @carterbox for the review.

> In the future, consider splitting the compiled and python code into separate packages to reduce the size of the build matrix. PYTHON_VERSIONS x CUDA_VERSIONS x ARCHS is a large matrix.

pinging @rusty1s: to consider the above comment eventually in the future so it makes the build matrix smaller and likely reduce the surface for failing builds.

recipes/pyg-lib/build.sh
hadim: Thanks for the context here. It's good to know about this CMake feature.

PR #21572, opened 2022-12-18
WIP: package clif

h-vetinari:
For pykaldi, see [here](https://github.com/conda-forge/kaldi-feedstock/pull/36#issuecomment-1296354677)


h-vetinari:
I should probably try this again now that the dependencies got updated (https://github.com/google/clif/commit/02843bb1d8905d267aceeacd791b710c54a0f054) and the pykaldi maintainer [said](https://github.com/pykaldi/pykaldi/issues/307#issuecomment-1721001605) they won't change to pybind11 (even though that's [recommended](https://github.com/google/clif/issues/54#issuecomment-1414394197) by the clif maintainers).

al-zatv:
Hi! now and then I need pykaldi bc nothing better for python+kaldi was not ever written. it's not very convinient to install pykaldi from source, and current conda package of pykaldi requires ancient python 3.7. So it will be super good for me, if updated pykaldi will once be deployed in conda-forge.

h-vetinari:
It's pykaldi's problem for using clif, but apparently unwilling to change: https://github.com/pykaldi/pykaldi/issues/307

https://github.com/google/clif itself notes

image

You can try to convince pykaldi to switch their approach to building. I'm not going to package clif (though if someone feels up for the task, go ahead and open a PR).

PR #21617, opened 2022-12-27
Added the pvxslibs package

valmar:
Adding the pvxslibs, prerequisite to updating another pacakge (p4p) to the latest version.
Had to use a git url and not a source tar archive because the package uses submodules, but does not provide a tarball which includes the submodules.

Thank you for your help

### DIFF
--- a/recipes/pvxslibs/meta.yaml
+++ b/recipes/pvxslibs/meta.yaml
@@ -6,18 +6,18 @@ package:
   version: {{ version }}
 
 source:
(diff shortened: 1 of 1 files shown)


valmar:
`@conda-forge/help-python, ready for review!`

chrisvam:
It is OK for me to be added as a co-maintainer of this feedstock.

valmar:
Needs more work before merging. Canceling the pull request

PR #21620, opened 2022-12-29
Adding conda-forge recipe for diffpy.pdfgui

dragonyanglong:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -15,16 +15,10 @@ build:
   number: 0
   entry_points:
     - pdfgui = diffpy.pdfgui.applications.pdfgui:main
-  osx_is_app: True
-
-app:
-  entry: pdfgui                                         [not win]
-  entry: pythonw -m diffpy.pdfgui.applications.pdfgui   [win]
-  summary: materials structure analysis from Pair Distribution Function
 
 requirements:
   host:
-    - python >=3.7
+    - python
     - pip
     - setuptools
 
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -29,10 +29,10 @@ requirements:
     - setuptools
     - six
     - matplotlib-base
-    - wxpython
+    - wxpython >=4.0.7
     - menuinst  # [win]
     - diffpy.pdffit2
-    - diffpy.structure
+    - diffpy.structure >=3
     - diffpy.utils
 test:
   imports:
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -20,12 +20,11 @@ build:
 
 requirements:
   host:
-    - python =3.7
+    - python >=3.6
     - pip
-    - setuptools
 
   run:
-    - python =3.7
+    - python >=3.6
     - setuptools
     - six
     - matplotlib-base
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -28,9 +28,9 @@ requirements:
     - python >=3.6
     - setuptools
     - six
-    - matplotlib-base
+   # - matplotlib-base
    # - wxpython >=4.0.7
-    - menuinst  # [win]
+   # - menuinst  # [win]
     - diffpy.pdffit2
     - diffpy.structure >=3
     - diffpy.utils
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -12,6 +12,7 @@ source:
 build:
   # we need egg_dir to avoid overlapping diffpy/__init__.py files.
   preserve_egg_dir: True
+  noarch: python
   number: 0
   script: {{ PYTHON }} -m pip install . --no-deps -vv
   entry_points:
--- a/recipes/diffpy.pdfgui/meta.yaml
+++ b/recipes/diffpy.pdfgui/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     - python
     - setuptools
     - six
-    - matplotlib
+    - matplotlib-base
     - wxpython =4.0.7
     - menuinst                  [win]
     - diffpy.pdffit2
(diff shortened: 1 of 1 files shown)


dragonyanglong:

dragonyanglong:
@conda-forge/python ready for review.

dragonyanglong:
@conda-forge/help-python

dragonyanglong:
@sbillinge , hi simon, please write a comment in the PR that you are good with being included as a maintainer for the package.

sbillinge:
I agree to be included as a maintainer for this project

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  preserve_egg_dir: True
+  number: 0
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  entry_points:
+    - pdfgui = diffpy.pdfgui.applications.pdfgui:main
+  osx_is_app: True
+
+requirements:
+  host:
+    - python =3.7
```
ocefpaf, line 23: Metadata says >=3.6

```suggestion
    - python >=3.6
```

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - pip
+    - setuptools
```
ocefpaf, line 25: Unless you are pinning a version you don't need this.

```suggestion
```

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+  run:
+    - python =3.7
```
ocefpaf, line 28:
```suggestion
    - python >=3.6
```

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - setuptools
```
ocefpaf, line 29: It doesn't like this needs setuptoos or `pkg_resources` at run time.

```suggestion
```

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - six
+    - matplotlib-base
+    - wxpython =4.0.7
```
ocefpaf, line 32: Are these optional? I could not find them in the metadata.

dragonyanglong: These are required dependencies to run the package. Where is the metadata to update this?

dragonyanglong:
Hi @ocefpaf , thanks so much for your time for reviewing this. I updated the yaml, please check.

dragonyanglong:
BTW, @ocefpaf  per your suggestions, I updated the python version from =3.7 into >=3.6, but it shows `Non noarch packages should have python requirement without any version constraints.`. How to fix this?

I tried to add `noarch: python`, but this is not allowed.

recipes/diffpy.pdfgui/meta.yaml
ocefpaf: See https://github.com/diffpy/diffpy.pdfgui/blob/v1.4.0/setup.py#L115-L119

ocefpaf:
> I tried to add noarch: python, but this is not allowed.

Why not? This does look like a `noarch` package. I don't see anything platform specific in https://github.com/diffpy/diffpy.pdfgui/blob/v1.4.0/setup.py

recipes/diffpy.pdfgui/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - python >=3.6
+    - pip
+
+  run:
+    - python >=3.6
+    - setuptools
+    - six
+    - matplotlib-base
+    - wxpython >=4.0.7
+    - menuinst  # [win]
```
ocefpaf, line 33: If this dependency is really necessary we cannot do noarch but it doesn't look like b/c it is not in the metadata.

dragonyanglong:
> 

@ocefpaf  After I add `noarch: python`, the `conda-forge-linter` gives a warning that
```
    noarch packages can't have selectors. If the selectors are necessary, please remove noarch: python.
```
 Though all the checks have been passed successfully.

How to fix it?

recipes/diffpy.pdfgui/meta.yaml
dragonyanglong: I think these dependency are necessary, as we did before for a conda release, see https://github.com/diffpy/diffpy.pdfgui/blob/v1.4.0/conda-recipe/meta.yaml#L39-L41

Is there any way to do noarch and keep these dependencies?

ocefpaf:
> How to fix it?

See my comment about the extra dependencies you added. If they are needed you cannot use noarch but upstream doesn't list them.

dragonyanglong:
> > How to fix it?
> 
> See my comment about the extra dependencies you added. If they are needed you cannot use noarch but upstream doesn't list them.

Dear @ocefpaf , I have updated the upstream setup.py to include these required packages (https://github.com/diffpy/diffpy.pdfgui/blob/v1.4.1/setup.py#L120-L122). But the `noarch` error still persists. Any suggestions?

recipes/diffpy.pdfgui/meta.yaml
ocefpaf: > Is there any way to do noarch and keep these dependencies?

Nope. There is a trick to do a platform noarch but I don't think it is worth in this case. Remove the noarch, add the dependencies and we can merge this.

PS: raise an issue upstream b/c those dependencies are not listed in the package metadata!

recipes/diffpy.pdfgui/meta.yaml
ocefpaf: Is this package really 3.7 only? The metadata says it is `>=3.6`. If that is true you can remove the constraint.

dragonyanglong: It is python>=3.6. But since we are removing `noarch`, the conda-forge-linter warns that  "Non noarch packages should have python requirement without any version constraints.". Shall we ignore this error?

dragonyanglong:
Hi @ocefpaf , thanks for your suggestions so far. I updated the version without `noarch`, please review and merge.

PR #21666, opened 2023-01-06
Add pypdf

MartinThoma:


MartinThoma:
@conda-forge/help-python / @synapticarbors  I think it should be ready now :-)

PR #21691, opened 2023-01-10
add interpret and interpret-core to conda-forge

paulbkoch:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/interpret-core/LICENSE
+++ b/recipes/interpret-core/LICENSE
@@ -1,0 +1,21 @@
+    MIT License
+
+    Copyright (c) Microsoft Corporation. All rights reserved.
+
+    Permission is hereby granted, free of charge, to any person obtaining a copy
+    of this software and associated documentation files (the "Software"), to deal
+    in the Software without restriction, including without limitation the rights
+    to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+    copies of the Software, and to permit persons to whom the Software is
+    furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included in all
+    copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE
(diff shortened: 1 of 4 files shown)


paulbkoch:
@conda-forge/staged-recipes ready for review

paulbkoch:

paulbkoch:

paulbkoch:
@conda-forge/help-python - adding the python label for a review

paulbkoch:
@conda-forge/help-python-c @conda-forge/staged-recipes ready for review

xhochy:
Can you please regenerate the meta.yaml files here using `grayskull pypi`? I see a lot of issues that would be resolved by using grayskull. Before that, it doesn't make sense to give it an in-depth review.

paulbkoch:
Hi @xhochy -- I'm going to take a guess that your primary concern is the large number of packages listed under run_constrained in the interpret-core package, and then duplicated as requirements for the tests and also in the interpret package?  If so, that's an intentional choice and not something that would be understood by grayskull.  Specifically, the "extras_require" is the issue from our setup.py here: 
https://github.com/interpretml/interpret/blob/cf785977ce9fd6a1cdfa6ff1d3f096a099e5bd11/python/interpret-core/setup.py#L216

In pypi we use optional dependency flags to allow a high degree of dependency customization.  You can read more about this in our documentation ( https://interpret.ml/docs/deployment-guide.html ), but I'll summarize here.

In pypi, the interpret-core package is a zero dependency install, but it has many optional dependencies that you can choose from.  The most minimal installation that we would recommend would be:

pip install interpret-core[required,ebm]

And the maximal install is:

pip install interpret-core[required,debug,notebook,plotly,lime,sensitivity,shap,ebm,linear,decisiontree,treeinterpreter,dash,skoperules,testing]

This allows picking and choosing the dependencies that you want, so you could have the combination of shap, linear and ebm, but not the rest like this:

pip install interpret-core[required,shap,ebm,linear]

Of course, most users do not want to understand or care about this, and we don't want to force people to specify all the flags above on each installation, so we created the interpret package that includes all the optional dependencies by default.

pip install interpret -> this is the equivalent of the maximal install above.

Coming back to conda, the issue for us is that conda does not support optional dependencies.  The best equivalent we get is having run_constrained.  Dependency management is better in conda-forge, and I initially thought it might make more sense to have just 1 package that had all the dependencies.  I'm not 100% sold on this decision, but the path I chose was to mirror the format that we use in pypi because:
  1) I felt it would be more confusing to have different packaging methodologies and naming in pypi and conda
  2) conda might add optional dependencies in the future, and in that case we could fully mirror our pypi methodology
  3) I think using run_constrained this way does still have some benefits in that the minimal versions of our dependencies can be managed in our package.  Other packages and/or environments that want subsets of our dependencies then only need to manage the list of requirements instead of both the requirements and version numbers.

paulbkoch:
Hi @xhochy - I ran interpret and interpret-core through grayskull.  It produces the following meta.yaml files, which are incomplete.

https://github.com/paulbkoch/interpret-grayskull

Can you please advise on how we should proceed?

wolfv:
I reviewed this PR and I didn't see any issues. Looked good to me. It will be hard to keep that run_constrained list up-to-date but I hope you'll find a way! 

Another possibility is to mimick optional dependencies by having multiple packages (they can be empty), for example `interpret-notebook` can have the dependencies of `interpret[notebook]` in the PyPI sense etc.

You can also achieve that with multiple outputs. If you need guidance on how to use multiple outputs feel free to let me know.

paulbkoch:
Thanks @wolfv -- Yeah, the run_constrained solution that I went with here is definitely a bit un-ideal.  Your suggestion makes sense and I agree it would be a better long-term solution.  I'll consider moving toward it in the future.  I'd like to get a bit of experience with the conda-forge release process though before undertaking that change as this is my first conda-forge package 👍  Thank you for your offer to help in that process.

PR #21716, opened 2023-01-13
add KaHIP graph partitioner

minrk:
Checklist
- [x] Build number is 0.

kahip vendors argtable3 (included in sources, no library), which is the [recommended way to use argtable3](https://github.com/argtable/argtable3/blob/f46ef208a7ebc941c9f56027112b6894baa4c3f9/README.md?plain=1#L35-L38). I could pull argtable3 out into a separate package, if folks need it to be that way. It hasn't been packaged on conda-forge, yet, though the previous [argtable2](https://github.com/conda-forge/argtable2-feedstock) is.

I patched kahip's build so that the Python packages can be installed separately, so there are:

1. `kahip` (shared libs and binaries)
2. `kahip-python` (Python bindings)

Arguably, kahip could be _3_ packages:

- libkahip (libs only)
- kahip (binaries, depends on libkahip)
- kahip-python (Python bindings, depends on libkahip)

but I don't think there's a huge benefit to that at the moment.

xref https://github.com/conda-forge/fenics-dolfinx-feedstock/issues/7

### DIFF
--- a/recipes/kahip/0001-install-python-component.patch
+++ b/recipes/kahip/0001-install-python-component.patch
@@ -1,0 +1,35 @@
+From f763f968e60e16dac5a216dad891acb61995f0d5 Mon Sep 17 00:00:00 2001
+From: Min RK <benjaminrk@gmail.com>
+Date: Thu, 12 Jan 2023 15:47:20 +0100
+Subject: [PATCH] install python component
+
+---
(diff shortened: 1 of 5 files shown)


minrk:
@conda-forge/help-python-c ready for review, especially the choice to stick with the recommended (by argtable) approach of bundling argtable3 vs building argtable3 as its own package.

PR #21727, opened 2023-01-14
add recipe for morphys

spinoch:


spinoch:
@conda-forge/help-python ready for review :)
I realize that this is not a very clean way of adding the package, but this package has not been updated since 2017 and at least 4 packages which I want to add to conda-forge depend on this :upside_down_face: I'd really appreciate if you can approve this

PR #21759, opened 2023-01-17
Add recipe for pyreaddbc

luabida:


xmnlab:
Thanks for working on that @luabida !
@esloch please confirm you are ok to be listed here as a maintainer pls

luabida:
> * Non noarch packages should have python requirement without any version constraints.

@xmnlab is it ok? Python version is pinned on the package

xmnlab:
@luabida try to use this for the poetry install command: `poetry install --only-root`

xmnlab:
For some reason it seems poetry is using the venv `tmp/tmpurc5w91s/.venv/bin/python`

xmnlab:
Maybe something that you could try would be move back to pip install instead of poetry install .. but adding `--no-deps` to the command

PR #21782, opened 2023-01-18
polar-codes: Add recipe for polar-codes

jdemel:
The `polar-codes` project provides low latency high throughput encoders and decoders for polar codes, a recent type of FEC.

Currently, it is tested on Linux x64. Hopefully, it works on MacOS and Windows too.
Besides, the projects requires a CPU AVX2 support. Thus, it may only be built on x86_64.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/polar-codes/bld.sh
+++ b/recipes/polar-codes/bld.sh
@@ -1,0 +1,22 @@
+setlocal EnableDelayedExpansion
+@echo on
+
+mkdir build
+if errorlevel 1 exit 1
+cd build
+if errorlevel 1 exit 1
+
+
+cmake -G "Ninja" ^
+    -DCMAKE_BUILD_TYPE:STRING=Release ^
+    -DCMAKE_PREFIX_PATH:PATH="%LIBRARY_PREFIX%" ^
+    -DCMAKE_INSTALL_PREFIX:PATH="%LIBRARY_PREFIX%" ^
+    -DPYTHON_EXECUTABLE:PATH="%PYTHON%" ^
+    ..
+if errorlevel 1 exit 1
+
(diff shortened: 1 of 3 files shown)


jdemel:
TCLAP and CppUnit are not available on Windows. Thus, this package should not be available on Windows either. For some reason, the MacOS compiler `x86_64-apple-darwin13.4.0-clang++` throws an error:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=643613&view=logs&j=e35d4f76-8ff2-5536-d795-df91e63eb9f7&t=fa7b4b17-b6ff-5c9c-8cfc-15f888c92310&l=2258
i.e. it does not support `aligned_alloc` (a C11 feature). Thus, this package may not be available on MacOS for now.

Is there a faster way to achieve
```yaml
  skip: True  # [not linux64]
```
than this? The build pipeline is started anyways.

jdemel:
@conda-forge/help-python-c It would be great to get feedback from you.

carterbox:
> Is there a faster way to achieve
> 
> ```yaml
>   skip: True  # [not linux64]
> ```
> 
> than this? The build pipeline is started anyways.

The other builds will be removed when this recipe is issued it's own feedstock.

carterbox (changes requested):
Will every user of your package want to use python? If not, your PR should contain two recipes. One for the CXX library and the other for the Python module. The python package depends on the CXX library.

The python package should be called `pypolar` and the CXX package `libpolarcode`?

carterbox:
Conda also doesn't yet support checking whether the host machine implements a given instruction set, so there is no way to tell if the end-user has an AVX2 compatible processor.

PR #21798, opened 2023-01-19
Add gr-symbolmapping recipe

jdemel:
GNU Radio and Python module for fast standardized symbol (de)mappers.

In digital communication, we rely on symbol mapping such as QPSK, 16QAM, etc. There exists a set of common definitions for these mappings in across a wide variety of standards (WiFi, LTE, 5G NR, ...). Unfortunately, GNU Radio does not implement this standard set. Further, we want fast and efficient demappers that employ efficient approximations for these common mappings.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/gnuradio-symbolmapping/build.sh
+++ b/recipes/gnuradio-symbolmapping/build.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+mkdir build
+cd build
+
+cmake_config_args=(
+    -DCMAKE_BUILD_TYPE=Release
+    -DLIB_SUFFIX=""
+    -DPYTHON_EXECUTABLE=$PYTHON
+    -DENABLE_DOXYGEN=OFF
+)
+
+cmake ${CMAKE_ARGS} -G "Ninja" "${cmake_config_args[@]}" ..
+cmake --build . --config Release -- -j${CPU_COUNT}
+cmake --build . --target install
--- a/recipes/gnuradio-symbolmapping/meta.yaml
+++ b/recipes/gnuradio-symbolmapping/meta.yaml
@@ -1,0 +1,62 @@
+{% set name = "gnuradio-symbolmapping" %}
+{% set version = "1.0.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ant-uni-bremen/gr-symbolmapping/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 4a1002d54ab70dfd0616aa6d3eb1a1a27cf8a60d1445b47b0c9643eb3173cbaf
+
+build:
+  skip: True  # [win]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - cmake >=3.16
+    - git
+    - ninja
+    - pkg-config  # [not win]
+    # cross-compilation requirements
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+    - numpy                               # [build_platform != target_platform]
+    - pybind11                            # [build_platform != target_platform]
+
+  host:
+    - python
+    - numpy >=1.22
+    - fmt
+    - spdlog
+    - pybind11
+    - gmp  # [unix]
+    - fftw
+    - volk >=2.5
+    # the following two entries are for generating builds against specific GR versions
+    - gnuradio-core  # [not gnuradio_extra_pin]
+    - gnuradio-core {{ gnuradio_extra_pin }}.*  # [gnuradio_extra_pin]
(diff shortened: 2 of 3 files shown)


jdemel:
@conda-forge/help-python-c This should be ready for review. I hope this is the correct team. Or should I get in touch with the cpp team?

I added a skip for windows builds for now. The Azure pipeline would just break in every corner before the package itself was even started to be compiled. Maybe, this can be changed at a later date.

recipes/gnuradio-symbolmapping/build.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+mkdir build
+cd build
+
+cmake_config_args=(
+    -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_PREFIX_PATH=$PREFIX
+    -DCMAKE_INSTALL_PREFIX=$PREFIX
```
xhochy, line 9: These are part of CMAKE_ARGS and should not be duplicated.

recipes/gnuradio-symbolmapping/meta.yaml
```diff
@@ -0,0 +1,65 @@
+{% set oot_name = "symbolmapping" %}
+{% set name = "gnuradio-" + oot_name %}
```
xhochy, line 2: Please avoid introducing more Jinja variables than necessary
```suggestion
{% set name = "gnuradio-symbolmapping" %}
```

recipes/gnuradio-symbolmapping/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - git
+    - ninja
+    - pkg-config  # [not win]
+    # cross-compilation requirements
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+    - numpy                               # [build_platform != target_platform]
+    - pybind11                            # [build_platform != target_platform]
+    # Add extra build tool dependencies here
+
```
xhochy, line 31:
```suggestion
```

recipes/gnuradio-symbolmapping/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  host:
+    - python
+    - numpy >=1.22
+    - fmt
+    - spdlog
+    - pybind11
+    - gmp  # [not win]
+    - fftw
+    - volk >=2.5
+    - gmp  # [linux]
```
xhochy, line 41:
```suggestion
    - gmp  # [unix]
    - fftw
    - volk >=2.5
```

recipes/gnuradio-symbolmapping/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    # the following two entries are for generating builds against specific GR versions
+    - gnuradio-core  # [not gnuradio_extra_pin]
+    - gnuradio-core {{ gnuradio_extra_pin }}.*  # [gnuradio_extra_pin]
```
xhochy, line 44: Don't use the `[]` when you actually are not using the functionality.
```suggestion
    - gnuradio-core  # not gnuradio_extra_pin
    - gnuradio-core {{ gnuradio_extra_pin }}.*  # gnuradio_extra_pin
```

xhochy (changes requested):

jdemel:
@xhochy Thanks for your suggestions. I integrated them. Are these suggestions part of a guideline?

recipes/gnuradio-symbolmapping/meta.yaml
jdemel: I suspect either this change, or the `- gmp  # [unix]` change broke the build. At least it seems to work again after i reverted those changes.

jdemel: I've tried all kinds of different combinations. The original version is the only one that works. Also, the original version basically expresses an XOR between the two. This should be the correct behavior.

carterbox: Where does the `gnuradio_extra_pin` variable get defined? You haven't included a 'conda_build_config.yaml', so `gnuradio_extra_pin` should have no value.

`gnuradio-core` is [pinned globally on conda-forge](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/5bac11c21f2a14b9100b9127848310df7721b25f/recipe/conda_build_config.yaml#L347) to the minor version, so ideally we want you to not pin it again.

carterbox (changes requested):

recipes/gnuradio-symbolmapping/meta.yaml
jdemel: I pulled these lines from the GNU Radio OOT template.
If I get this correctly, the way forward would be to just have one line
```yaml
- gnuradio-core
```
without anything extra. Would that be correct?

PR #21818, opened 2023-01-21
Add leocad

SylvainCorlay:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/leocad/build.sh
+++ b/recipes/leocad/build.sh
@@ -16,6 +16,7 @@ if [[ ${HOST} =~ .*linux.* ]]; then
 fi
 
 qmake \
+  DISABLE_UPDATE_CHECK=1 \
   LDRAW_LIBRARY_PATH=${PREFIX}/share/ldraw \
   CONFIG+=release \
   INSTALL_PREFIX=$PREFIX \
--- a/recipes/leocad/build.sh
+++ b/recipes/leocad/build.sh
@@ -1,0 +1,26 @@
+mkdir build
+cd build
+
+if [[ "$ARCH" == "64" ]]; then
+  export PLATFORM=linux-g++-64;
+else
+  export PLATFORM=linux-g++-32;
+fi
+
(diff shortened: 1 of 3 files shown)


SylvainCorlay:
That's weird.

```
[...]/work/conda_build.sh: line 21: [...]/work/leocad.pro: Permission denied
```

SylvainCorlay:
The build appears to work, but I am not having this message upon starting leocad, so the ldraw parts library is not located correctly somehow.

Screenshot from 2023-02-09 10-10-20

SylvainCorlay:
This seems to be due to the encoding of the string:

https://github.com/leozide/leocad/blob/a18f55913bcabe739321e50d84f547bbfc2e7080/qt/qtmain.cpp#L200-L202 

I will patch the sources to use a `const char*`.

SylvainCorlay:
Binary relocation works as expected with the patch, but I am still getting the same error upon starting leocad. Open Issue https://github.com/leozide/leocad/issues/845 upstream.

SylvainCorlay:
I figured it out. The problem comes from the binary relocation done by conda.

- conda allows to install binary packages in any "installation prefix"
- this is made possible by doing the binary relocation of the produced binaries, and then string replacement of the build prefix in text files.
- the way we deal with the case of the prefix path being used in compiled binaries is by using a long (255 chars) installation path at build time, and replace it at installation time by the (zero-padded) installation prefix. A warning is issued to packagers that binary replacement has to be done for relocation.

In the case of leocad, the LDRAW_LIBRARY_PATH is such an instance of the build prefix being present in the compiled binary. I had foreseen that, although I did not realize that the compiler would also put the first characters of this value in the beginning of the small-string-optimization buffer of the QString where it is stored.

The way I could fix it is by preventing this optimization by creating and intemediary `volatile const char*` holding the value of LDRAW_BINARY_PATH to be copied into the QString.

I will post another issue proposing another solution to deal with the LDRAW_LIBRARY_PATH in a way that prevents jumping through these hoops.

PR #21838, opened 2023-01-23
Adding Recipes waves

kbrindley:
Checklist
- [x] (Not Applicable) If static libraries are linked in, the license of the static library is packaged.
- [x] Build number is 0.

### DIFF
--- a/recipes/waves/meta.yaml
+++ b/recipes/waves/meta.yaml
@@ -11,7 +11,15 @@ source:
 
 build:
   noarch: python
-  script: conda install waves-{{ version }}-py_0.tar.bz2
+  script:
+    - echo "waves-{{ version }}-py_0.tar.bz2"
+    - manpath1="$PREFIX/share/man/man1"
+    - manpath2="$PREFIX/man/man1"
+    - mkdir -p $manpath1 $manpath2 $PREFIX/bin $SP_DIR
+    - cp -r man $PREFIX/man
+    - cp -r share $PREFIX/share
+    - cp -r python-scripts/* $PREFIX/bin
+    - cp -r site-packages/* $SP_DIR
 
 requirements:
   build:
@@ -43,7 +51,7 @@ test:
 about:
   home: https://github.com/lanl/waves
   license: BSD-3-Clause
-  license_file: LICENSE.txt
+  license_file: info/licenses/LICENSE.txt
   summary: 'Parameteric study computational engineering workflow tool'
   description: |
     WAVES (LANL code C23004) is a computational engineering workflow tool that integrates parametric studies with
--- a/recipes/waves/meta.yaml
+++ b/recipes/waves/meta.yaml
@@ -23,8 +23,6 @@ build:
 
 requirements:
   build:
-    - python {{ python }}
-    - git
     - scons >=4
     - setuptools >=45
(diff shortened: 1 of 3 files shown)


kbrindley:
This project is already a conda package with a recipe: https://github.com/lanl/waves/blob/main/recipe-github/meta.yaml and a conda package file: https://github.com/lanl/waves/releases/tag/0.6.7.

I'm struggling to find examples of conda-forge recipes that include existing Conda packages. This recipe is a first draft to re-package after ``conda-build`` unpacks the tar archive. I don't think this is the right approach and I will continue to search for a better example. In the meantime, if anyone can point me in the right direction, I think all I need is an example to start from.

Thanks!

Kyle

dopplershift:
The usual way here would be to use the GitHub tarball for the release.

kbrindley:
@dopplershift  As in the source code tarball ``*.tar.gz`` (as opposed to a Conda package ``*.tar.bz2``)?

I'm building and deploying on an internal channel in parallel. I was hoping to re-use the existing Conda package, but if I need to re-build from source for conda-forge, I can. I was afraid that might be the answer, so I'm working on a ``*.tar.gz`` recipe now.

Thanks!

dopplershift:
Since it's pure Python, you're saving no effort whatsoever by trying to reuse the build. And in general, conda-forge works by having everything built by a common toolchain on shared infrastructure. The recipe here needs to be self-contained.

recipes/waves/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+    - {{ PYTHON }} setup.py install
+    - scons man html-github
+    - {{ PYTHON }} package_documentation.py
+  ignore_run_exports:
+    - python_abi
+
+requirements:
+  build:
+    - python {{ python }}
+    - git
```
dopplershift, line 27: Is `git` needed if you're setting `SETUPTOOLS_SCM_PRETEND_VERSION`?

recipes/waves/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  script:
+    - {{ PYTHON }} setup.py install
+    - scons man html-github
+    - {{ PYTHON }} package_documentation.py
+  ignore_run_exports:
+    - python_abi
+
+requirements:
+  build:
+    - python {{ python }}
```
dopplershift, line 26: I don't think this is necessary? It should be enough to have the one under `host` below? 
```suggestion
```
I also don't know what that construct with the template is even trying to do...

recipes/waves/meta.yaml
kbrindley: Probably not. Thanks for the catch.

recipes/waves/meta.yaml
kbrindley: You may be right. I'll admit I'm a bit fuzzy on the ``host`` section. I'll re-read the current documentation and try to trim the requirements list to only those items that are strictly necessary.

The ``{{ python }}`` is a hold-over from an earlier recipe where the Python version was a build variant: https://docs.conda.io/projects/conda-build/en/stable/resources/variants.html#build-variants. It should not be necessary anymore, but I'll test locally to confirm.

dopplershift (approved):

PR #21882, opened 2023-01-27
add dolfinx_mpc

minrk:
separate cpp and python outputs:

- libdolfinx_mpc for cpp
- dolfinx_mpc for python

cc @jorgensd

### DIFF
--- a/recipes/dolfinx_mpc/yum_requirements.txt
+++ b/recipes/dolfinx_mpc/yum_requirements.txt
@@ -1,0 +1,1 @@
+mesa-libGL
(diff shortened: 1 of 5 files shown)


minrk:
needs an 0.6.1 after https://github.com/jorgensd/dolfinx_mpc/pull/45

minrk:
@conda-forge/help-python-c ready for review

recipes/dolfinx_mpc/yum_requirements.txt
```diff
@@ -0,0 +1 @@
+mesa-libGL
```
minrk, line 1: note: this is only needed for `gmsh` in the test env. dolfinx_mpc never uses it directly.

beckermr:
Is the thumbs up by @jorgensd and indication that they are willing to maintain?

jorgensd:
> Is the thumbs up by @jorgensd and indication that they are willing to maintain?
Yes, it is!:)

PR #21901, opened 2023-01-30
Add libcublas recipe

betatim:
Hello all 👋 

this is my first time adding a package to conda-forge. I followed the example `meta.yaml` and this recipe is based on the existing one for `libcublas` in the nvidia conda channel. I'll see what the CI robots say about this first attempt.

@adibbley do you want to continue to be listed as maintainer for this recipe?

cc @jakirkham @adibbley
xref #21382

Checklist
- [x] Build number is 0.


recipes/libcublas/bld.bat
```diff
@@ -0,0 +1,17 @@
...
+rem Files...
+for %%i in (.\*) do (
+    if not %%~ni==build_env_setup (
+    if not %%~ni==conda_build (
+    if not %%~ni==metadata_conda_debug (
+        xcopy %%i %PREFIX%
+    )
+    )
+    )
+)
```
adibbley, line 17:
```suggestion
move lib\x64\* %LIBRARY_LIB%
move bin\* %LIBRARY_BIN%
move include\* %LIBRARY_INC%
move src %PREFIX%
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,75 @@
+{% set name = "libcublas" %}
+{% set version = "11.11.3.6" %}
```
adibbley, line 2:
```suggestion
{% set version = "12.0.1.189" %}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+{% set platform = "linux-x86_64" %}  # [linux64]
+{% set platform = "linux-ppc64le" %}  # [ppc64le]
+{% set platform = "linux-sbsa" %}  # [aarch64]
+{% set platform = "windows-x86_64" %}  # [win]
```
adibbley, line 6:
```suggestion
{% set platform = "windows-x86_64" %}  # [win]
{% set extension = "tar.xz" %}  # [not win]
{% set extension = "zip" %}  # [win]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/{{ name }}/{{ platform }}/{{ name }}-{{ platform }}-{{ version }}-archive.tar.xz
+  sha256: 045e6455c9f8789b1c7ced19957c7904d23c221f4d1d75bb574a2c856aebae98  # [linux64]
+  sha256: 27b07d1fa375404ed0f7ce37573de1c8a5ff8c313b9f388ee7b4ff41d4a8409f  # [ppc64le]
+  sha256: 38fe90cbbc7da3dbdcd8c29e0fcd60f69baf580d9b3f71a4ee102e3c7fc30b3d  # [aarch64]
+  sha256: 67b0934a6359e4ee26fff823c356021589d392c4fd49ca12624f570edc08e2b9  # [win]
```
adibbley, line 17:
```suggestion
  url: https://developer.download.nvidia.com/compute/cuda/redist/{{ name }}/{{ platform }}/{{ name }}-{{ platform }}-{{ version }}-archive.{{ extension }}
  sha256: 38272cc9184b4a57a6250fc2cd914ffc827f9f8cd5a3e1356bc3c31f843caf8f  # [linux64]
  sha256: 2d46e735f8e76e2eb7df1b4f41d56aa8d5d2683f99988bac82d6ae33fb6d344b  # [ppc64le]
  sha256: 96b68f6821684f39f338862639ce74a74c75830bf44dcd17ef372a1cb8c2304d  # [aarch64]
  sha256: 88f29bec81880844da2eee13f1cdc51e1ccd8b4c60af6df633577810b30f80d0  # [win]
```

adibbley (commented):
Looks like the wrong branch was started with, here are the 12.0.0 changes.

betatim:
Thanks for the hint about v12.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+
+  - name: libcublas-static
+    build:
+      skip: True  # [not linux]
+    requirements:
+      run:
+        - libcublas-dev >={{ version }}
+    files:
+      - lib/lib*blas*_static.a
+      - targets/*/lib/lib*blas*.so.*
```
adibbley, line 83: Looks like I sent the wrong line here, sorry about that! This should let you remove the `missing_dso_whitelist` too
```suggestion
      - targets/*/lib/lib*blas*_static.a
```

betatim: 👍  - updated.

recipes/libcublas/post-link.sh
```diff
@@ -0,0 +1,3 @@
+echo "By downloading and using the ${PKG_NAME} conda package, you accept the terms"  >> $PREFIX/.messages.txt
+echo "and conditions of the NVIDIA cuQuantum Software License Agreement:"            >> $PREFIX/.messages.txt
+echo "https://docs.nvidia.com/cuda/cuquantum/license.html"                           >> $PREFIX/.messages.txt
```
jakirkham, line 3: @adibbley, this seems to be referencing cuQuantum (even though this is cuBLAS). Is this meant to be here?

adibbley: I haven't added any such files, I think we can remove this.

leofang: This should be replaced by a pointer to CUDA EULA right?

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas_static.a  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+about:
+  home: https://developer.nvidia.com/cublas
+  # A LicenceRef cannot be used in this case, see:
+  # https://github.com/conda-forge/staged-recipes/pull/12882#discussion_r504086944
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
```
leofang, line 87: I don't understand why this comment is here? We need the `LicenseRef-*` prefix to silent the linter error, so what's wrong with it?

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - include                     # [linux]
+      - targets/*/include/*         # [linux]
+      - lib/pkgconfig               # [linux]
+      - src
+      - Library\include             # [win]
+      - Library\lib                 # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
```
leofang, line 67: I haven't followed the latest discussion on the cudart/nvcc sides. Should we ensure that the headers like `cublas.h` are only packaged in the `*-dev` package?

recipes/libcublas/meta.yaml
jakirkham: It's just referencing the earlier discussion where this came up and this was decided. Seems ok to keep

leofang: Yes it needs to be reworded. It obviously *should* (not "cannot") be used.

recipes/libcublas/post-link.sh
jakirkham: Nope. We are just handling this with `license_url` in `about`

recipes/libcublas/meta.yaml
jakirkham: How about?

```suggestion
  # Uses `LicenceRef-*` per this discussion:
  # https://github.com/conda-forge/staged-recipes/pull/12882#discussion_r504086944
  license: LicenseRef-NVIDIA-End-User-License-Agreement
```

leofang: LGTM

recipes/libcublas/post-link.sh
leofang: Got it, thx

recipes/libcublas/meta.yaml
betatim: Updated, it was copied verbatim from the `cudatoolkit` feedstock, so probably worth updating there as well.

recipes/libcublas/post-link.sh
betatim: Should we keep this file, if yes is there approved/example text for it? If no, is removing it all that needs doing or does something else need adjusting as well?

jakirkham: Gotcha thanks for the context

Yeah we should drop it. Nothing else is needed outside removing the file

betatim: Done

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+      - Library\bin                   # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-dev
+    requirements:
+      run:
+        - libcublas >={{ version }}
```
leofang, line 53: xref: https://github.com/conda-forge/staged-recipes/pull/21924#discussion_r1095297421
```suggestion
        - {{ pin_subpackage("libcublas", max_pin="x") }}
```

recipes/libcublas/meta.yaml
leofang: And we shouldn't test the files that come from dependencies, or we'd have too many to carry all the way down...
```suggestion
        - test -f $PREFIX/include/cublas.h  # [unix]
        - if not exist %LIBRARY_INC%\\cublas.h exit 1  # [win]
```
(I may have missed extra cublas headers, plz double check 🙂)

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+      commands:
+        - test -f $PREFIX/lib/libcublas.so  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-static
+    build:
+      skip: True  # [not linux]
+    requirements:
+      run:
+        - libcublas-dev >={{ version }}
```
leofang, line 74: ditto
```suggestion
        - {{ pin_subpackage("libcublas-dev", max_pin="x") }}
```

recipes/libcublas/meta.yaml
leofang: Sorry Sebastian, I forgot that [cudart](https://github.com/conda-forge/staged-recipes/pull/21723/files#diff-0dd2e318a324e35cf19682e69f69f441a7bbb76fd01f33c3615ac69a66c6f4b3R51) is putting this under `run_exports`, not `run`.

recipes/libcublas/meta.yaml
betatim: Good point. Two questions: why remove the check that the `.so` exists? Does it make sense to check for every single header or go for more of a "sanity check" (just check that `cublas.h` exists).

betatim:
General question: the `libcublas-dev` package depends on `libcublas` but it includes the same files (e.g. the `.so`s). Should it be like that or should the `-dev` package not include the things which are in the "no postfix" package, because to install `-dev` you will also need to install the "no postfix" package?

adibbley:
> General question: the `libcublas-dev` package depends on `libcublas` but it includes the same files (e.g. the `.so`s). Should it be like that or should the `-dev` package not include the things which are in the "no postfix" package, because to install `-dev` you will also need to install the "no postfix" package?

They are different files. The `libcublas-dev` would have `libcublas.so` and `libcublas` would have `libcublas.so.X` and `libcublas.so.X.Y.Z`

recipes/libcublas/meta.yaml
adibbley: We should be testing the `.so` and the `.lib` in the `-dev` packages, they wouldn't be coming from dependencies. 
Testing a single header should be enough. The contents of the source is already tested, so it should be reasonable to assume if `include` is packaged it would contain the expected files.

recipes/libcublas/meta.yaml
betatim: Added this

recipes/libcublas/meta.yaml
betatim: While thinking about your comment I realised that I don't understand why a static library should have any run time dependencies. Isn't the whole point of using a static library that it gets included in what ever you are building (at build time)? This means I removed the `requirements.run` key.

betatim:
> They are different files. The `libcublas-dev` would have `libcublas.so` and `libcublas` would have `libcublas.so.X` and `libcublas.so.X.Y.Z`

Ah yes, I hadn't read the `files` entries closely enough. Surprisingly the test I had added for `libcublas` (`test -f $PREFIX/lib/libcublas.so`) passed. Which shouldn't have happened because as you explained, there shouldn't be a `libcublas.so`

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+
+outputs:
+  - name: libcublas
+    files:
+      - lib/lib*blas*.so.*            # [linux]
+      - targets/*/lib/lib*blas*.so.*  # [linux]
+      - Library\bin                   # [win]
```
adibbley, line 44: We'll have to specify the files here to not pick up compiler files.
```suggestion
      - Library\bin\cublas*.dll                   # [win]
      - Library\bin\nvblas*.dll                   # [win]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - lib/stubs                   # [linux]
+      - include                     # [linux]
+      - targets/*/include/*         # [linux]
+      - lib/pkgconfig               # [linux]
+      - src
+      - Library\include             # [win]
+      - Library\lib                 # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so  # [unix]
```
adibbley, line 66: This should fix the current build failure
```suggestion
        - test -L $PREFIX/lib/libcublas.so  # [unix]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+      - lib/lib*blas*.so.*            # [linux]
+      - targets/*/lib/lib*blas*.so.*  # [linux]
+      - Library\bin                   # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so.*  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-dev
+    requirements:
```
adibbley, line 51: We still need a `run` requirement on `libcublas` here.
```suggestion
    requirements:
      run:
        - libcublas >= {{ version }}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+        - test -f $PREFIX/include/cublas.h  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cublas.h exit 1  # [win]
+
+  - name: libcublas-static
+    build:
+      skip: True  # [not linux]
+    files:
+      - lib/lib*blas*_static.a
+      - targets/*/lib/lib*blas*_static.a
```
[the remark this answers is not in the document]
betatim: Could you explain a bit why it should depend on the `dev` (or any) package? I assumed that the use case for the `static` package is people who want to use it at build time to statically link against `libcublas` so that they don't have any runtime dependence. 

Do you need more than just the `.a` to link to it?

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+      - Library\bin\nvblas*.dll                   # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so.*  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-dev
+    requirements:
+      run:
+        - libcublas >= {{ version }}
```
jakirkham, line 54: Think this should fix the linter warning

```suggestion
        - libcublas >={{ version }}
```

recipes/libcublas/meta.yaml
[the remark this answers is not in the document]
adibbley: I would think linking a static library would still need the headers from the `dev` package to be useful. If that's not the case we could drop the dependency.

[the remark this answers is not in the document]
jakirkham: Yeah think this goes back to how we want to define `-dev` & `-static` that we have had in a couple other places ( https://github.com/conda-forge/staged-recipes/pull/21902#discussion_r1092460458 ) ( https://github.com/conda-forge/staged-recipes/pull/21723#discussion_r1084611245 )

Thus far we've gone with `-dev` should pull in everything except `-static`, which is constrained, but not required. Though an exception to exception is `cuda-cudart`, in which `-dev` pulls in the `-static` portion too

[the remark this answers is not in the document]
betatim: I guess to compile code that relies on things in the static library, you need the headers. But in that case you should be asking for `-dev` no?

In https://github.com/conda-forge/staged-recipes/pull/21723#discussion_r1084611245 the conclusion was that `-dev` should require `-static`. `-static` doesn't require anything else.

The way to use this for a package that wants to link to the static library then is to require the `-dev` package at build time, link to the static library, produce the build output and be done. The finished executable would not have any run time dependency on `-dev` or `-static`.

So I'd add `-static` as a dependency of `-dev`, but leave `-static` as is (unlike the suggestion which suggests adding `-dev` as a requirement for `-static`).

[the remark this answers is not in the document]
adibbley: Yeah I think you have the right of it. Lets follow https://github.com/conda-forge/staged-recipes/pull/21902#discussion_r1092460458 for adding the `run_constrained` in the `-dev` package and leaving `-static` with no dependency.

leofang:
Should we list `cuda-version` under `libcublas`'s `run` or `run_constrained`?

recipes/libcublas/meta.yaml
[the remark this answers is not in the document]
betatim: Done. I'll mark this thread as resolved, so that they don't get super long. We can start a new one if there is a new thing to discuss.

betatim:
> Should we list `cuda-version` under `libcublas`'s `run` or `run_constrained`?

Good question. I was thinking that it should have a runtime dependency on `cuda-cudart`. Does that make sense? And if yes, do we then also need to depend on `cuda-version`? Or is it implicit because of the dependency on `cuda-cudart`?

adibbley:
> Good question. I was thinking that it should have a runtime dependency on `cuda-cudart`. Does that make sense? And if yes, do we then also need to depend on `cuda-version`? Or is it implicit because of the dependency on `cuda-cudart`?

Having `cuda-version` in `libcublas`'s `run_constrained` should be all we need I think. I'd like to avoid a dependency on `cuda-cudart` unless absolutely necessary, as we typically don't enforce that in `libcublas`.

betatim:
Does this mean that I can use `libcublas` without having `cuda-cudart` installed? I assumed that you needed to have the cuda runtime installed in order to use cublas (a bit like, say, you need to have numpy installed in order to use scikit-learn). If you don't need to have it installed, then I guess it shouldn't be a run time dependency.

adibbley:
We do have some existing products, such as cuTENSOR, which choose to depend only on `libcublas`, so there is a use case. 
Ideally these inter-component dependencies are kept to a minimum so we aren't restricting users who want the smallest footprint possible. If someone wants to pull in everything they can use a higher level metapackage.

leofang:
One thing for consideration is that all math libs are statically linked to cudart in order to offer minor ver compatibility. So in this regard cudart, quite counterintuitively, isn't really a run dependency of libcublas and friends.

leofang:
cc: @jakirkham

jakirkham:
Was thinking actually that `libcublas` would depend on `cuda-version` in `requirements/run`

Also [`cuda-version` is in `cuda-nvcc`'s `run_exports/strong`]( https://github.com/conda-forge/staged-recipes/pull/21350/files#diff-df0006df0e9a6f931f97635d6458e58ac98d1b9e555b001383f92742d07e5934R25-R27 ). So anything built with `cuda-nvcc` should have `cuda-version` attached automatically

IOW if a package is using `cuda-nvcc` & `libcublas`, `cuda-nvcc` will handle adding `cuda-version` to the package's `requirements/run`

leofang:
> Was thinking actually that `libcublas` would depend on `cuda-version` in `requirements/run`

So, like this?
```yaml
run:
  - cuda-version >=12.0,<13
```

jakirkham:
Yeah, we could also do...

```yaml
requirements:
  host:
    - cuda-version 12.0
  run:
    - {{ pin_compatible("cuda-version") }}
```

Or even...

```yaml
requirements:
  build:
    - {{ compiler("cuda") }}
```

...once `cuda-nvcc` is plugged in as the `cuda` compiler

betatim:
Implemented the first suggestion from https://github.com/conda-forge/staged-recipes/pull/21901#issuecomment-1447675082.

Marking this as "review for review". Let's see if we can get it merged and then use it as a base for other recipes (instead of having several in flight at the same time).

betatim:
The build failures are because they explicitly test that things work with a particular CUDA version (that isn't 12). What is a good way to deal with these?

leofang:
Shall we merge, then?

betatim:
Waiting to see what the outcome of https://github.com/conda-forge/cuda-version-feedstock/issues/1 is

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - lib/lib*blas*_static.a
+      - targets/*/lib/lib*blas*_static.a
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas_static.a  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+about:
+  home: https://developer.nvidia.com/cublas
+  # Uses `LicenceRef-*` per this discussion:
```
bdice, line 91:
```suggestion
  # Uses `LicenseRef-*` per this discussion:
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+    - cuda-version 12.0
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+    - {{ pin_compatible("cuda-version") }}
```
bdice, line 39: Do some of these host/run pinnings need to be moved to the individual packages, rather than the recipe as a whole? I can't remember the rules for how split packages propagate dependencies, but I think the `run` dependencies are supposed to be defined in the subpackages.

jakirkham: The package artifacts are made available on CI

Though it seems this PR preceded that change. Have merged in the latest changes so we can get a fresh build with those packages uploaded

For example taking [this build]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=673765&view=results ), we can click the "N published" link 

<details>

Screen Shot 2023-03-15 at 2 05 49 PM

</details>

We can then select the build archive to download

<details>

Screen Shot 2023-03-15 at 2 09 22 PM

</details>

The decompressed archive then contains the Conda package within. If one is logged into Azure, the `.conda` package can be downloaded directly from Azure.

<details>

```bash
$ tree conda_pkgs_noarch
conda_pkgs_noarch
├── current_repodata.json
├── current_repodata.json.bz2
├── index.html
├── python-lsp-ruff-1.2.0-pyhd8ed1ab_0.conda
├── repodata.json
├── repodata.json.bz2
├── repodata_from_packages.json
└── repodata_from_packages.json.bz2
```

</details>

The `.conda` package is just a `zip` file (and the extension can be changed). Within it contains `zst` files, which can be explored with any tool that understands that format (like [PeaZip]( https://peazip.github.io ))

leofang: Cool so now stage-recipes by default retains the artifacts?

jakirkham: Yeah though I think they expire at some point

FWICT `libcublas` depends on `cuda-version`, but this doesn't happen for the other packages. Should add this also affects other things (like where compiler runtime libraries are included)

So think we should move `cuda-version` to the dependencies

A separate question is whether we want to do this with `sysroot` & compiler runtime libraries

betatim: > So think we should move `cuda-version` to the dependencies

As in move it to the sub-package? Or what do you mean with "move to the dependencies", for me `requirements.run` is (one of) "the dependencies" :D

bdice: Yes, move `run` dependencies to subpackages. Those don't propagate from the top-level afaik. Examples: 

https://github.com/rapidsai/cudf/blob/8fbfb4ae5f083ac8473023f9e4982e8021b83c6d/conda/recipes/libcudf/meta.yaml#L34
https://github.com/rapidsai/cudf/blob/8fbfb4ae5f083ac8473023f9e4982e8021b83c6d/conda/recipes/libcudf/meta.yaml#L68

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - Library\bin\cublas*.dll                   # [win]
+      - Library\bin\nvblas*.dll                   # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so.*  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-dev
+    requirements:
+      run:
```
jakirkham, line 55: Same `cuda-version` change as discussed before

```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+  run_exports:
+    strong:
+      - __glibc >=2.17
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
```
jakirkham, line 33: Usually `sysroot` is added to `build`. Sometimes (like with compilers) we may put this in `host` & `run`. Though for a library like this it probably makes sense to stick to `build`.

```suggestion
  build:
    - {{ compiler("c") }}
    - {{ compiler("cxx") }}
    - sysroot_{{ target_platform }} 2.17   # [linux]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+  sha256: 2d46e735f8e76e2eb7df1b4f41d56aa8d5d2683f99988bac82d6ae33fb6d344b  # [ppc64le]
+  sha256: 96b68f6821684f39f338862639ce74a74c75830bf44dcd17ef372a1cb8c2304d  # [aarch64]
+  sha256: 88f29bec81880844da2eee13f1cdc51e1ccd8b4c60af6df633577810b30f80d0  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+  run_exports:
+    strong:
+      - __glibc >=2.17
```
jakirkham, line 26: This is added automatically by `sysroot`. So shouldn't need to be explicitly added. If that doesn't work for some reason, we may want to adjust our usage of `sysroot`

```suggestion
```

recipes/libcublas/meta.yaml
jakirkham, line 39: Following the discussion in this thread ( https://github.com/conda-forge/staged-recipes/pull/21901#discussion_r1137751993 ), we should move these into the `outputs`. Would suggest dropping these from the top-level

With `sysroot` & `glibc` specifically, these should be handled by the changes to `build` already. So don't need to be explicitly added.

The rest is about adding `cuda-version` throughout, which is detailed in other comments

```suggestion
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+      - lib/lib*blas*.so.*            # [linux]
+      - targets/*/lib/lib*blas*.so.*  # [linux]
+      - Library\bin\cublas*.dll                   # [win]
+      - Library\bin\nvblas*.dll                   # [win]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas.so.*  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+  - name: libcublas-dev
```
jakirkham, line 53: The `run_exports` section lives under `build` (as opposed to `requirements`). Though can see how this might be confusing. Here are some [docs on this]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements )

```suggestion
  - name: libcublas-dev
    build:
      run_exports:
        - {{ pin_subpackage("libcublas", max_pin="x") }}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    requirements:
+      run:
+        - libcublas >={{ version }}
+      run_exports:
+        - {{ pin_subpackage("libcublas", max_pin="x") }}
```
jakirkham, line 58: Moved to the the `build` section as detailed in the prior comment

```suggestion
```

recipes/libcublas/meta.yaml
jakirkham: Have tried to suggest some changes below to clear this up

jakirkham (commented):
Thanks Tim! 🙏

Have suggested some changes below related to the `cuda-version` changes discussed. Also related to the move of top-level requirements to particular outputs. Lastly included a few around `sysroot`, `run_exports`, and `glibc`, which should hopefully simplify/clarify things a bit.

Please let us know if you have any questions 🙂

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+    - cuda-version 12.0
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+    - {{ pin_compatible("cuda-version") }}
+
+outputs:
+  - name: libcublas
```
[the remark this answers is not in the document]
bdice:
```suggestion
  - name: libcublas
    requirements:
      host:
         - cuda-version {{ cuda_version }}
      run:
         - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    test:
+      commands:
+        - test -L $PREFIX/lib/libcublas.so  # [unix]
+        - test -f $PREFIX/include/cublas.h  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cublas.h exit 1  # [win]
+
+  - name: libcublas-static
+    build:
+      skip: True  # [not linux]
```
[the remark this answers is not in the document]
bdice: Use the newly defined Jinja variable.
```suggestion
  - name: libcublas-static
    build:
      skip: True  # [not linux]
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
+{% set name = "libcublas" %}
+{% set version = "12.0.1.189" %}
```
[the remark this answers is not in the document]
jakirkham: Ok let's prefer Bradley's change. Have hidden mine

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - __glibc >=2.17                       # [linux]
+    - cuda-version 12.0
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+    - {{ pin_compatible("cuda-version") }}
+
+outputs:
+  - name: libcublas
+    files:
```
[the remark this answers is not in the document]
betatim: Super-seeded by https://github.com/conda-forge/staged-recipes/pull/21901/files#r1142611442

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      run_exports:
+        - {{ pin_subpackage("libcublas", max_pin="x") }}
+    requirements:
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - libcublas >={{ version }}
+      run_constrained:
+        - libcublas-static >={{ version }}
```
bdice, line 60: @jakirkham @leofang I don't know how to constrain `libcublas-static` so that it only matches the current major version, but I feel like that might be desired? `libcublas` has a `pin_subpackage("libcublas", max_pin="x")`. Does this need a `pin_compatible`, or is that possible with `run_constrained`?

bdice: Or maybe the `run_exports: - {{ pin_subpackage("libcublas", max_pin="x") }}` above isn't needed at all, due to constraints on `cuda-version`? I feel like there's asymmetry here that we don't want, but I'm unsure of how to improve it.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - targets/*/lib/lib*blas*_static.a
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas_static.a  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+
+about:
+  home: https://developer.nvidia.com/cublas
+  # Uses `LicenseRef-*` per this discussion:
+  # https://github.com/conda-forge/staged-recipes/pull/12882#discussion_r504086944
```
bdice, line 97: Should we include this comment? It's not mentioned on the other CUDA Toolkit staged recipes, so I think it should be safe to remove here.
```suggestion
```

recipes/libcublas/meta.yaml
bdice: For comparison, https://github.com/conda-forge/staged-recipes/pull/21916/ is not defining a `run_exports` on `libcurand-dev`.

leofang: Right, I thought we could do this?
```yaml
run_constrained:
  - {{ pin_subpackage("libcublas", max_pin="x") }}
```
My understanding is that `pin_compatible` is only used if it's a `build` requirement, which `-static` isn't.

jakirkham: We could trying using [`{{ pin_subpackage('libcublas-static', exact=True) }}`]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html?highlight=pin_subpackage%20exact#extra-jinja2-functions )

We should have `run_exports` since using dynamic libraries is the common case

Users of static libraries can add [`ignore_run_exports_from`]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html?highlight=ignore_run_exports_from#export-runtime-requirements ) to drop the runtime dependency

bdice: @jakirkham Can you make suggestions for what you're proposing? Just want to make it concrete so I know I'm understanding correctly.

jakirkham: Since the `-dev` packages are only used in `host` while building packages, am guessing we want exact alignment between headers, libraries, and other build config files (pkg-config, CMake, etc.). Does that seem reasonable or am I missing something?

leofang: OK, so like this?
```yaml
run_constrained:
  - {{ pin_subpackage("libcublas-static", exact=True) }}
```

jakirkham: Do we want to align `libcublas` too?

```yaml
      run:
        ...
        - {{ pin_subpackage("libcublas", exact=True) }}
      run_constrained:
        - {{ pin_subpackage("libcublas-static", exact=True) }}
```

bdice: > Since the -dev packages are only used in host while building other packages

I’m not sure if that’s true. For example, libraft-headers will have runtime dependencies on -dev packages, because it needs to install CUDA headers for any downstream libraries using libraft-headers. But I suppose that transitively, libraft-headers is only used in `host`?

Also I guess if you’re making a developer environment that can build CUDA packages from source, you’ll need to install the dev packages, but forcing the same runtime libraries is okay as long as that run requirement is not exported to recipes using the -dev packages (those should export major-compatible pinnings).

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - name: libcublas-dev
+    build:
+      run_exports:
+        - {{ pin_subpackage("libcublas", max_pin="x") }}
+    requirements:
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - libcublas >={{ version }}
```
leofang, line 58: Based on the discussion https://github.com/conda-forge/cuda-version-feedstock/issues/1#issuecomment-1462998398, we shouldn't need the version constraint anymore:
```suggestion
        - libcublas
```

recipes/libcublas/meta.yaml
leofang: Sorry, let me take one step back. I don't follow either of you anymore... 😅

@jakirkham, why do we have to pin exactly, instead of just letting `cuda-version` to kick in? I think this is Bradley's original question 
```suggestion
        - libcublas-static
```
and based on the discussion https://github.com/conda-forge/cuda-version-feedstock/issues/1#issuecomment-1462998398, it seems to be what we had agreed upon?

@bdice, I don't understand how is `-dev` packages being used at runtime related to `-static` packages here? This is `run_constrained`, not `run`, so it's OK that the `-dev` package is installed to offer the headers at runtime. It won't bring in the `-static` package unless the user explicitly asks for it. Am I missing something?

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    files:
+      - lib/lib*blas*_static.a
+      - targets/*/lib/lib*blas*_static.a
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcublas_static.a  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
```
bdice, line 92: No Windows library should be produced for `libcublas-static`, so we shouldn't have tests for it.
```suggestion
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+
+about:
```
bdice, line 94: This "about" should be copied to each subpackage. See  Discussed offline with @leofang / @jakirkham. https://github.com/conda/infrastructure/discussions/649#discussioncomment-4530683

Example: https://github.com/conda-forge/staged-recipes/pull/21924/commits/fe5604092ee21f8e8a4e8b73fd93667315cfb7ec

recipes/libcublas/meta.yaml
bdice, line 60:
```suggestion
      run_constrained:                      # [linux]
        - libcublas-static >={{ version }}  # [linux]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - libcublas >={{ version }}
+      run_constrained:
+        - libcublas-static >={{ version }}
+    files:
+      - lib/lib*blas*.so            # [linux]
+      - targets/*/lib/lib*blas*.so  # [linux]
+      - lib/stubs                   # [linux]
```
bdice, line 64:
```suggestion
      - lib/stubs/libcublas*.so     # [linux]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - include                     # [linux]
```
bdice, line 65:
```suggestion
      - include/cublas*             # [linux]
      - include/nvblas*             # [linux]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - targets/*/include/*         # [linux]
+      - lib/pkgconfig               # [linux]
+      - src
+      - Library\include             # [win]
```
bdice, line 69:
```suggestion
      - Library\include\cublas*     # [win]
      - Library\include\nvblas*     # [win]
```

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - Library\lib                 # [win]
```
bdice, line 70:
```suggestion
      - Library\lib\cublas*         # [win]
      - Library\lib\nvblas*         # [win]
```

bdice (commented):
I reviewed this offline with @jakirkham and @leofang. I am going to file a PR to @betatim's fork of `staged-recipes` with all the suggested changes applied, so that this is easier to finalize without massive PR comment traffic. We should be able to merge that PR once it's created, and merge this recipe after CI passes.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      run_constrained:
+        - libcublas-static >={{ version }}
+    files:
+      - lib/lib*blas*.so            # [linux]
+      - targets/*/lib/lib*blas*.so  # [linux]
+      - lib/stubs                   # [linux]
+      - include                     # [linux]
+      - targets/*/include/*         # [linux]
+      - lib/pkgconfig               # [linux]
+      - src
```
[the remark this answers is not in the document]
adibbley: They are being copied by the build script, and dropping them isn't an option. What would a more correct location be?

`2023-03-22T09:14:30.6165931Z ‘src/fortran.c’ -> ‘$PREFIX/src/fortran.c’`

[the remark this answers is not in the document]
bdice: I don't think we saw the `src` files being shipped in the artifacts being produced? (I need to look back and confirm that.) I don't know what a more correct location would be. @adibbley Can you explain how these are used, and perhaps what search paths are considered? I'm not familiar with using `$PREFIX/src` in other conda packages and am not sure what precedents to compare against.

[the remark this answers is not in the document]
bdice: Update: I do see the `src/` files in the artifacts being produced. I guess we can just ship those files in `$PREFIX/src`, unless anyone has other ideas.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+    requirements:
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - libcublas >={{ version }}
+      run_constrained:
+        - libcublas-static >={{ version }}
+    files:
+      - lib/lib*blas*.so            # [linux]
```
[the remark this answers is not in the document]
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+    build:
+      skip: True  # [not linux]
+    requirements:
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    files:
+      - lib/lib*blas*_static.a
+      - targets/*/lib/lib*blas*_static.a
```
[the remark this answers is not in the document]
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
bdice: I won't apply this suggestion for now due to: https://github.com/conda-forge/staged-recipes/pull/21916#discussion_r1146880594

recipes/libcublas/meta.yaml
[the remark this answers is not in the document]
leofang: @adibbley The concern during the discussion was that the shipped filenames (ex: `fortran.c`) are hard to tell if they're coming from `libcublas` or another package, and could potentially be clobbering. Also, it's almost impossible for conda-forge users to locate these files, unless the cuBLAS team explicitly documents about `$PREFIX/src` (or wherever they'd be). So it's of very little use. I don't have strong opinions on this, though.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - lib/pkgconfig               # [linux]
+      - src
+      - Library\include             # [win]
+      - Library\lib                 # [win]
+    test:
+      commands:
+        - test -L $PREFIX/lib/libcublas.so  # [unix]
+        - test -f $PREFIX/include/cublas.h  # [unix]
+        - if not exist %LIBRARY_LIB%\\cublas.lib exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cublas.h exit 1  # [win]
```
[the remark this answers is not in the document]
leofang: btw I thought we discussed that for `*.lib` files (defining the linker symbols, not the static library) they should be in `LIBRARY_BIN` not `LIB`?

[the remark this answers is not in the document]
leofang: nvm I was being silly

recipes/libcublas/meta.yaml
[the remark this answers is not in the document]
adibbley: The cublas docs do mention them in `src` specifically: https://docs.nvidia.com/cuda/cublas/#cublas-fortran-bindings
Maybe a `$PREFIX/cuda/src/` would make sense here? We probably have some other files that might be worth putting in a `cuda` directory for clarity. I'm thinking things like `docs` `samples` and the like could make more sense under `$PREFIX/cuda`.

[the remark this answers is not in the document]
bdice: Okay. My primary complaint was that the files are very generically named -- `src/fortran.c` could be literally anything, it has no indication of a relationship to CUDA or cuBLAS.

I kind of like the idea of moving these files into `$PREFIX/cuda/{doc,samples,src}` so that CUDA-related files (besides include/lib) are together.

For comparison: In my primary conda environment (which has a bunch of assorted packages), I see a `doc` directory with a couple things from Qt and CMake, but nothing else. It would feel odd to put extensive documentation from CUDA in that directory, I think. There are also directories like `share` but that doesn't feel quite right either. Since I don't see a better candidate, I'm a weak +1 for `$PREFIX/cuda` for all these files.

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,171 @@
...
+{% set platform = "linux-x86_64" %}  # [linux64]
+{% set platform = "linux-ppc64le" %}  # [ppc64le]
+{% set platform = "linux-sbsa" %}  # [aarch64]
+{% set platform = "windows-x86_64" %}  # [win]
+{% set target_name = "x86_64-linux" %} # [linux64]
+{% set target_name = "ppc64le-linux" %} # [ppc64le]
+{% set target_name = "sbsa-linux" %} # [aarch64]
+{% set target_name = "x64" %} # [win]
+{% set extension = "tar.xz" %}  # [not win]
+{% set extension = "zip" %}  # [win]
```
jakirkham, line 13: Think there may be some `<space>`s needed here to fix the lint

```suggestion
{% set target_name = "x86_64-linux" %}  # [linux64]
{% set target_name = "ppc64le-linux" %}  # [ppc64le]
{% set target_name = "sbsa-linux" %}  # [aarch64]
{% set target_name = "x64" %}  # [win]
{% set extension = "tar.xz" %}  # [not win]
{% set extension = "zip" %}  # [win]
```

jakirkham:
@conda-forge/core @conda-forge/staged-recipes, please take a look when you have a moment

beckermr (commented):
I see no obvious problems with the recipe. I am curious as to why the cuda tests don’t pass and if we expect that.

jakirkham:
Thanks Matt 🙏

Yeah that's expected since they are older CUDA versions. The relevant tests are the vanilla Linux and Windows tests (without CUDA)

beckermr (approved):

recipes/libcublas/meta.yaml
```diff
@@ -0,0 +1,171 @@
...
+      host:
+         - cuda-version {{ cuda_version }}
+      run:
+         - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -L $PREFIX/lib/libcublas.so.{{ version }}                                          # [linux]
+        - test -L $PREFIX/lib/libcublas.so.{{ version.split(".")[0] }}                            # [linux]
```
kkraus14, line 52: I've seen code that uses `dlopen` / `dlsym` as well as `ctypes` code that expects to use just the `lib*.so` as opposed to the versioned shared library at runtime. Not a blocker, but have we considered having the `libcublas.so` symlink in this package as opposed to the `dev` package? I know that the `SONAME` points to `libcublas.so.12`, but there's definitely code out there that will end up requiring `libcublas-dev` at runtime because of this.

Definitely not a blocker for getting this in though!

kkraus14 (approved):

recipes/libcublas/meta.yaml
bdice: We've been reworking RAPIDS code to use the proper SONAME form, like `libcublas.so.12`, anywhere that we've seen that. I could see an argument that it's a bug to rely on `dlopen`ing the unversioned form? That said, I don't have a reference for how we've packaged this in other CUDA Toolkit deployments to know if this approach is "normal" or not. I agree that we shouldn't treat this as a blocker, and I'd be happy to discuss further if users run into specific issues.

bdice: See also:
- https://github.com/rapidsai/cudf/issues/12708
- https://github.com/rapidsai/cudf/issues/12708#issuecomment-1433366212
- https://github.com/rapidsai/cucim/issues/504

jakirkham: Yeah this should be easy to add if we decide. Let's turn this into an issue on the feedstock, which will make it easier to track as well

jakirkham (approved):

jakirkham:
Thanks all! 🙏

Let's get this in. We can iterate on anything else in the feedstock

recipes/libcublas/meta.yaml
jakirkham: Raised an issue ( https://github.com/conda-forge/libcublas-feedstock/issues/1 )

betatim:
Thanks for all the help and work!!

PR #21902, opened 2023-01-30
Add libcufft recipe

seberg:
Attempt at cufft, lets see how far this gets, its my first attempt and I am still a bit confused about the whole setup.

Questions:
* ~What about the static version for windows?  Did I just miss it, or do I need to split it into its own yaml to not create it on windows?  (or can a skip be added just for one of the `outputs`?)~  Nvm, adding the skip seems pretty straight forward.
* ~Was it correct to keep/use the `missing_dso_whitelist`?~

cc @jakirkham, @adibbley
xref gh-21382

EDIT: Arrg, was meant as a draft of course, this probably needs a bit of work still.


recipes/cuda-libcufft/build.sh
```diff
@@ -0,0 +1,11 @@
...
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+
+for i in `ls`; do
+	[[ $i == "build_env_setup.sh" ]] && continue
+	[[ $i == "conda_build.sh" ]] && continue
+	[[ $i == "metadata_conda_debug.yaml" ]] && continue
+	cp -r $i ${PREFIX}
+done
```
adibbley, line 11:
```suggestion
# Install to conda style directories
[[ -d lib64 ]] && mv lib64 lib
[[ -d pkg-config ]] && mv pkg-config ${PREFIX}/lib/pkgconfig
[[ -d "$PREFIX/lib/pkgconfig" ]] && sed -E -i "s|cudaroot=.+|cudaroot=$PREFIX|g" $PREFIX/lib/pkgconfig/cuda*.pc

[[ ${target_platform} == "linux-64" ]] && targetsDir="targets/x86_64-linux"
[[ ${target_platform} == "linux-ppc64le" ]] && targetsDir="targets/ppc64le-linux"
[[ ${target_platform} == "linux-aarch64" ]] && targetsDir="targets/sbsa-linux"

for i in `ls`; do
	[[ $i == "build_env_setup.sh" ]] && continue
	[[ $i == "conda_build.sh" ]] && continue
	[[ $i == "metadata_conda_debug.yaml" ]] && continue
	if [[ $i == "lib" ]] || [[ $i == "include" ]]; then
		mkdir -p ${PREFIX}/${targetsDir}
		mkdir -p ${PREFIX}/$i
		cp -rv $i ${PREFIX}/${targetsDir}
		for j in `ls $i`; do
			ln -s ../${targetsDir}/$i/$j ${PREFIX}/$i/$j
		done
	else
		cp -rv $i ${PREFIX}
	fi
done
```

recipes/cuda-libcufft/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - /lib64/libdl.so.2
+    - /lib64/libm.so.6
+    - /lib64/libpthread.so.0
+    - /lib64/librt.so.1
+    - /lib64/libgcc_s.so.1
+    - /lib64/libc.so.6
+    - /lib64/ld-linux-x86-64.so.2
```
adibbley, line 32:
```suggestion
  run_exports:
    strong:
      - __glibc >=2.17

requirements:
  build:
    - {{ compiler("c") }}
    - {{ compiler("cxx") }}
  host:
    - sysroot_{{ target_platform }} 2.17   # [linux]
    - __glibc >=2.17                       # [linux]
  run:
    - sysroot_{{ target_platform }} 2.17   # [linux]
    - __glibc >=2.17                       # [linux]
```

recipes/cuda-libcufft/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+
+
+outputs:
+  - name: libcufft
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcufft.so.{{ version }}  # [linux]
+        - if not exists %LIBRARY_BIN%\cufft64_11.dll exit 1  # [win]
+    files:
+      - lib/libcufft*.so.*  # [linux]
```
adibbley, line 42:
```suggestion
      - lib/libcufft*.so.*  # [linux]
      - targets/*/lib/*.so.* # [linux]
```

recipes/cuda-libcufft/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    requirements:
+      run:
+        - libcufft >={{ version }}
+        # - cuda-cudart >=12.0
+    files:
+      - lib/stubs/libcufft*.so  # [linux]
+      - libcufft*.so*  # [linux]
+      - include/cufft*  # [linux]
+      - include\cufft*  # [win]
+      - include\cudalibxt*  # [win]
```
adibbley:
```suggestion
      - lib/*.so             # [linux]
      - targets/*/lib/*.so   # [linux]
      - lib/stubs            # [linux]
      - include              # [linux]
      - targets/*/include/*  # [linux]
      - lib/pkgconfig        # [linux]
      - Library\include      # [win]
      - Library\lib          # [win]
```

adibbley (commented):
Here are some missing changes from the 12.0.0 branch

seberg:
> Here are some missing changes from the 12.0.0 branch

Aha, just noticed the windows part of it, but not the rest, thanks!  That also answers my question about the whitelist :)

adibbley:
Please don't rename this to `cuda-libcufft`, it needs to be `libcufft`.

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - lib/pkgconfig        # [linux]
+      - Library\include      # [win]
+      - Library\lib          # [win]
+
+  - name: libcufft-static
+    build:
+       skip: True  # [not linux]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcufft_static.a
```
jakirkham, line 85: @adibbley, it looks like this file test is failing. Is there something we are missing?

seberg: Ah... The targets folder was not included, but the `$PREFIX/lib/libcufft_static.a` is just a symlink to the targets folder.  Pushed a fix for this (I think).

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,110 @@
...
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+
+
+outputs:
+  - name: libcufft
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcufft.so.{{ version }}      # [linux]
+        - if not exist %LIBRARY_BIN%\cufft64_11.dll exit 1  # [win]
```
jakirkham, line 55:
```suggestion
        - test -f $PREFIX/lib/libcufft.so.{{ version }}                              # [linux]
        - if not exist %LIBRARY_BIN%\cufft64_{{ version.split(".")[0] }}.dll exit 1  # [win]
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,110 @@
...
+    strong:             # [linux]
+      - __glibc >=2.17  # [linux]
+  missing_dso_whitelist:
+     - /lib64/libdl.so.2            # [linux]
+     - /lib64/libm.so.6             # [linux]
+     - /lib64/libpthread.so.0       # [linux]
+     - /lib64/librt.so.1            # [linux]
+     - /lib64/libgcc_s.so.1         # [linux]
+     - /lib64/libc.so.6             # [linux]
+     - /lib64/ld-linux-x86-64.so.2  # [linux]
```
jakirkham, line 35: Think we should be able to drop these now that we are requiring the `compiler`s below

```suggestion
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,110 @@
...
+
+  - name: libcufft-dev
+    test:
+      commands:
+        - test -f $PREFIX/include/cufft.h               # [linux]
+        - if not exist %LIBRARY_LIB%\cufft.lib exit 1  # [win]
+    requirements:
+      run:
+        - libcufft >={{ version }}
+        # - cuda-cudart-dev >=12.0
```
jakirkham, line 69: Similar to `cuda-cudart` ( https://github.com/conda-forge/staged-recipes/pull/21723#discussion_r1084611245 ), would depend on the static library here

```suggestion
        - libcufft-static >={{ version }}
```

jakirkham (commented):
Thanks Sebastian! 🙏

This looks pretty good. Commented on a few minor points below

recipes/libcufft/meta.yaml
kkraus14: Can we not make getting the static libraries more explicit? I.E. conda-forge defaults to using shared libraries, why not have the `dev` package depend on the static library and have someone explicitly request the static library if desired?

jakirkham: So the library users of packages would get would be `libcufft`, which just contains the dynamic library. Builds would use `libcufft-dev`, which includes headers, dynamic/static libraries, etc.

That said, if we would rather this be `run_constrained` instead of `run`, that seems fine

recipes/libcufft/meta.yaml
seberg: Hmm, hopefully it works, I had spend quite some time trying to understand why it might be failing (locally mostly) and not making progress.

recipes/libcufft/meta.yaml
seberg: So `run_constrained` would just ensure we don't have an outdated static library around?  That makes sense to me, but I don't have enough experience to have a real opinion :).

recipes/libcufft/meta.yaml
seberg: @jakirkham it is now running into requiring the list and I have no idea how to understand it, it seems to me sysroot should have this and that it gets installed?  On the cublas one it seems to not require it (except for the last build where the library isn't needed to be included).

recipes/libcufft/meta.yaml
jakirkham: Yeah it means if the user installs `libcufft-static`, that requirement is applied. So in this case would be a minimum version

kkraus14: I think a previous comment from @robertmaynard on the `cudart` recipe said that getting a `-dev` package that doesn't supply everything needed for building is funky. I think it should have a `run` dependency on either the dynamic or static library.

I understand `cudart` might be special, but for other libraries shouldn't we default to the shared libraries? Static linking in conda packages is typically the exception where I think the space overhead of getting both the shared and static libraries is a reasonable tradeoff.

jakirkham: Ok have updated the suggestion above to use `run_constrained`

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+  host:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+  run:
+    - sysroot_{{ target_platform }} 2.17   # [linux]
+    - __glibc >=2.17                       # [linux]
+
+
+outputs:
+  - name: libcufft
```
jakirkham, line 40:
```suggestion
  - name: libcufft
    requirements:
      build:
        - {{ compiler("c") }}
        - {{ compiler("cxx") }}
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+      - Library\include      # [win]
+      - Library\lib          # [win]
+
+  - name: libcufft-static
+    build:
+       skip: True  # [not linux]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcufft_static.a
+    requirements:
```
jakirkham, line 76:
```suggestion
    requirements:
      build:
        - {{ compiler("c") }}
        - {{ compiler("cxx") }}
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
jakirkham, line 30:
```suggestion
```

jakirkham (commented):
Maybe we should move the compiler bits to the subpackages. Similar changes may be needed for the other top-level requirements

recipes/libcufft/meta.yaml
seberg: Thanks, lets see. I am completely lost why it is a problem here and not in the other libs...

recipes/libcufft/meta.yaml
seberg: In one place, I feel I squinted a sysroot being resolved to version 2.12 rather than 2.17 (I think).  But even if I didn't misinterpret that not sure if it adds up...

seberg: > A subpackage does not automatically inherit any dependencies from its top-level recipe, so any build or run requirements needed by the subpackage must be explicitly specified.

Does this mean that _all_ requirements on the top level are moot?!  (maybe with the exception of the host requirements?

recipes/libcufft/build.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+mkdir -p ${PREFIX}/lib
```
seberg, line 5: CI was failing with:
```
mv: cannot move ‘pkg-config’ to ‘/home/conda/staged-recipes/build_artifacts/cufft_1675252566075/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/pkgconfig’: No such file or directory
```
I think after moving the `requirements` (although I am not quite sure).

recipes/libcufft/meta.yaml
robertmaynard: Yes that is what I said. To me it is very weird (broken) that a development package doesn't contain the all the components needed for development( headers, static lib, shared lib ).

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,104 @@
+{% set name = "cufft" %}
```
seberg, line 1:
```suggestion
{% set name = "libcufft" %}
```

It now seems to me that this naming must have been the reason for the requirements to behave differently than the other recipes.

seberg: Just changing for now, I suspect this means it is nicer to revert the requirements changes.

recipes/libcufft/meta.yaml
seberg: So that would mean:
- dynamic lib has system runtime dependencies
- static has system runtime dependencies
- dev depends on both dynamic and static (should that actually use `=={{ version.split(',')[0] }}` or exact version matching)?

(I am not clear that the static subpackage can be used without dev, so nobody will ever use them without also having the dynamic libs around.)

Next question would be if:
```
  run_exports:                # [linux]
     strong:                   # [linux]
       - __glibc >=2.17        # [linux]
```
should, at least in principle, be part of the *static* library as:
```
- name: libcufft-static
  run_exports:                # [linux]
     strong:                  # [linux]
       - libcufft-static      # [linux]
```
since arguably if I build depend on `libcufft-dev` and `libcufft-static` (without runtime depend on `libcufft`) I am planning to link in the static lib and require the same runtime exports (sysroot+glibc).

(Sorry, I spend way too much time trying to slowly building a mental model that seems half logical :/)

robertmaynard: > since arguably if I build depend on libcufft-dev and libcufft-static (without runtime depend on libcufft) I am planning to link in the static lib and require the same runtime exports (sysroot+glibc).

I agree with that.

I don't know enough about the conda rule model to state if the above is correctly maps to our shared mental model

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
+{% set name = "libcufft-split" %}
+{% set version = "11.0.0.21" %}
```
bdice, line 2:
```suggestion
{% set version = "11.0.0.21" %}
{% set cuda_version = "12.0" %}
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
...
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+      host:                                    # [linux]
+        - sysroot_{{ target_platform }} 2.17   # [linux]
+        - __glibc >=2.17                       # [linux]
+      run:                                     # [linux]
+        - sysroot_{{ target_platform }} 2.17   # [linux]
+        - __glibc >=2.17                       # [linux]
```
bdice, line 46:
```suggestion
      host:
        - sysroot_{{ target_platform }} 2.17   # [linux]
        - __glibc >=2.17                       # [linux]
        - cuda-version {{ cuda_version }}
      run:
        - sysroot_{{ target_platform }} 2.17   # [linux]
        - __glibc >=2.17                       # [linux]
        - {{ pin_compatible("cuda-version") }}
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
...
+    run_exports:
+      weak:
+        - {{ pin_subpackage("libcufft", max_pin='x') }}
+    requirements:
+      run:
+        - libcufft-dev_{{ target_platform }} =={{ version }}
+        # TODO: Do we need cudart?  But at this point it would fail CI.
+      run_contrained:
+        # Ensure static package is compatible (if installed explicitly)
+        - libcufft-static_{{ target_platform }} >={{ version }}  # [linux]
```
bdice, line 68: Note there's a typo fix in `run_constrained` that might escape the eye.

```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcufft-dev_{{ target_platform }} =={{ version }}
        # TODO: Do we need cudart?  But at this point it would fail CI.
      run_constrained:
        # Ensure static package is compatible (if installed explicitly)
        - libcufft-static_{{ target_platform }} >={{ version }}  # [linux]
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
...
+      - targets/*/lib/stubs  # [linux]
+      - targets/*/include/*  # [linux]
+      - Library\include      # [win]
+      - Library\lib          # [win]
+    run_exports:
+      weak:
+        - {{ pin_subpackage("libcufft", max_pin='x') }}
+    requirements:
+      run_constrained:
+        - libcufft-static_{{ target_platform }} >={{ version }}  # [linux]
```
bdice, line 89:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
      run_constrained:
        - libcufft-static_{{ target_platform }} >={{ version }}  # [linux]
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
...
+  - name: libcufft-static
+    # cuFFT does not ship `*_static.lib` on windows so only relevant on linux
+    build:
+       skip: True  # [not linux]
+    files:
+      - lib/libcufft*_static.a
+      - lib/libcufft_static_nocallback.a
+    requirements:
+      run:
+        - libcufft-static_{{ target_platform }} =={{ version }}
```
bdice, line 105:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcufft-static_{{ target_platform }} =={{ version }}
```

recipes/libcufft/meta.yaml
```diff
@@ -0,0 +1,137 @@
...
+        - test -L $PREFIX/lib/libcufft_static.a
+
+  - name: libcufft-static_{{ target_platform }}
+    build:
+        skip: True  # [not linux]
+        noarch: generic
+    files:
+      - targets/*/lib/libcufft*_static.a
+      - targets/*/lib/libcufft_static_nocallback.a
+    test:
```
bdice, line 117:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
    test:
```

bdice (commented):
Adding `cuda-version` pinnings.

recipes/libcufft/meta.yaml
jakirkham: Would tweak this slightly to move `sysroot` into `requirements/build` (with the compilers) and drop `glibc` & `sysroot` from other sections. Typically `sysroot` is added next to the compilers (this is different for compiler packages like `cuda-nvcc`). Also `glibc` would be added by `sysroot` automatically so shouldn't need hard-coding after this change

seberg: Thanks, will use this as is and then compare with cublas later today.

bdice: Thanks @jakirkham, I wasn't able to concisely edit all the pieces at once. I agree with your suggestions, just wanted to get `cuda-version` fixed first.

seberg: I added a `, max_pin="x.x")` to one of these.  Confirmed that the contents look fine.  The setup warns that the `cxx` requirement may not be needed.

Not sure about:
1. Duplication of about section (I thought it gets inherited, but sounds like its not perfectly)?
2. cudart dependency for the dev versions (we need the cudart headers), either choice makes sense.

bdice:
@seberg I pushed commit f8e8938 with some minor fixes/changes to align with the libcublas recipe.

@jakirkham @adibbley @leofang I think this is ready to go, so I'll approve it.

bdice (approved):
Let's resolve open comment threads that are outdated.

I reviewed the recipe contents and upstream package contents carefully to ensure nothing is being missed. I have not yet validated the output package contents are complete but the tests should be covering us well. (However, I ran into an unexpected situation where tests were passing on libcublas despite having a test that I expected to fail. See https://github.com/conda-forge/libcublas-feedstock/pull/10.) I can verify the package contents on Monday. This looks close enough that I feel comfortable approving it.

robertmaynard (approved):

jakirkham:
Thanks all! 🙏

PR #21924, opened 2023-02-01
Add `libcusolver`, `libcusparse`, and `libnvjitlink` recipes

leofang:
Part of #21382. ~~Merge after #21901.~~

~~Please see the new commits added after commit 93834b6. Discussions are pending.~~

cc: @jakirkham @Adibbley

Checklist
- [ ] Build number is 0.


recipes/libcusolver/build.sh
```diff
@@ -0,0 +1,27 @@
...
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+
+[[ ${target_platform} == "linux-64" ]] && targetsDir="targets/x86_64-linux"
+[[ ${target_platform} == "linux-ppc64le" ]] && targetsDir="targets/ppc64le-linux"
+[[ ${target_platform} == "linux-aarch64" ]] && targetsDir="targets/sbsa-linux"
+
+# cuSOLVER ships a copy of METIS, but this is not OK from the conda-forge perspective
+rm lib/libmetis_static.a
```
leofang, line 11: I added this because on conda-forge this is supposed to be generated from metis-feedstock, not from libcusolver-feedstock which is totally unexpected. While cuSOLVER needs METIS for its reason, this is surprising from conda-forge perspective. We are lucky here even if I don't remove it, though, because metis-feedstock is currently not generating static libs. But it's not guaranteed.

However, there's `libcusolver_metis_static.a` (I didn't remove) and I worry that if this one is removed, `libcusolver_metis_static.a` is rendered useless. Perhaps we should also remove it for the time being?

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  - name: libcusolver
+    files:
+      - lib/libcusolver*.so.*            # [linux]
+      - targets/*/lib/libcusolver*.so.*  # [linux]
+      - Library\bin                      # [win]
+
+  - name: libcusolver-dev
+    requirements:
+      run:
+        - libcublas ={{ major_version }}.*
```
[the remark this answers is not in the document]
leofang: Seems to be documented here: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions

recipes/libcusolver/build.sh
jakirkham: @eriknw, do you have any thoughts on how METIS is handled? No worries if not (just thought this might be of interest)

jakirkham: cc @conda-forge/metis (in case any of you have thoughts here)

leofang:
@jakirkham @adibbley btw as discussed I am adding `libnvjitlink` to the same PR

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+       commands:
+         - test -f $PREFIX/lib/libcusolver.so  # [unix]
+         - test -f $PREFIX/lib/libcusolverMg.so  # [unix]
+         - if not exist %LIBRARY_LIB%\\cusolver.lib exit 1  # [win]
+         - if not exist %LIBRARY_LIB%\\cusolverMg.lib exit 1  # [win]
+
+  - name: libcusolver-dev
+    requirements:
+      run:
+        - libcusolver >={{ version }}
```
leofang, line 52: @Bdice mentioned to me that this is going to have trouble (ex: installing an old header + new & incompatible binary), so I think this is safer:
```suggestion
        - {{ pin_subpackage("libcusolver", max_pin="x") }}
```

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+
+outputs:
+  - name: libcusolver
+    requirements:
+      run:
+        - libcublas =12.*
+        - libnvjitlink =12.*
```
leofang, line 37: nit: 
```suggestion
        - libnvjitlink >=12.0,<13
```
as per [CUDA 12.0 Compiler Support for Runtime LTO Using nvJitLink Library](https://developer.nvidia.com/blog/cuda-12-0-compiler-support-for-runtime-lto-using-nvjitlink-library/):
> Statically or dynamically link the highest version of nvJitLink library from the major release.

recipes/libcusolver/meta.yaml
leofang: It seems [cudart](https://github.com/conda-forge/staged-recipes/pull/21723/files#diff-0dd2e318a324e35cf19682e69f69f441a7bbb76fd01f33c3615ac69a66c6f4b3R51) is putting it under `run_exports`, I'll follow.

recipes/libcusolver/meta.yaml
adibbley: Is there a way we can do this without needing to know the next version? It is likely a safe bet here, but I'd like to remove the guess work on things out of our control if possible.

leofang: We can certainly set up some Jinja variables, but it is a requirement of nvJitLink to be compatible with the CTK major release IIUC. Ultimately, this needs to track back which CTK major release it belongs to. I see no other way based on my current understanding 😞

leofang:
@jakirkham @adibbley I can do cusparse in this PR since it requires nvJitLink, but we better get cublas in first

jakirkham:
Sounds good. Let's go ahead and add `libcusparse` here then

leofang:
Will do tonight or first thing tmr morning.

recipes/libcusolver/meta.yaml
leofang: TODO: Beter to pin `cuda-version` instead (but how?)

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,104 @@
+{% set name = "libcusolver" %}
+{% set version = "11.4.2.57" %}
```
bdice, line 2:
```suggestion
{% set version = "11.4.2.57" %}
{% set cuda_version = "12.0" %}
```

recipes/libcusolver/meta.yaml
bdice, line 37:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcublas =12.*
        - libnvjitlink =12.*
```

recipes/libcusolver/meta.yaml
bdice, line 52:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcusolver >={{ version }}
```

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+        - if not exist %LIBRARY_INC%\\cusolverSp.h exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cusolverSp_LOWLEVEL_PREVIEW.h exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cusolver_common.h exit 1  # [win]
+
+  - name: libcusolver-static
+    build:
+      skip: True  # [not linux]
+    requirements:
+      run:
+        - libcusolver-dev >={{ version }}
```
bdice, line 82:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcusolver-dev >={{ version }}
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,75 @@
+{% set name = "libcusparse" %}
+{% set version = "11.7.5.86" %}
```
bdice, line 2:
```suggestion
{% set version = "11.7.5.86" %}
{% set cuda_version = "12.0" %}
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # uncomment the following line if the package is pure python and the recipe is exactly the same for all platforms.
+  # it is okay if the dependencies are not built for all platforms/versions, although selectors are still not allowed.
+  # see https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-python for more details.
+  # noarch: python
+  # if the installation is complex, or different between unix and windows, use separate bld.bat and build.sh files instead of this key.
+  # by default, the package will be built for the python versions supported by conda-forge and for all major oss.
+  # add the line "skip: true  # [py<35]" (for example) to limit to python 3.5 and newer, or "skip: true  # [not win]" to limit to windows.
+  # more info about selectors can be found in the conda-build docs: 
+  # https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#preprocessing-selectors
+  #script: {{ python }} -m pip install . -vv
```
bdice, line 29: Remove all recipe boilerplate.
```suggestion
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - /lib64/libc.so.6
+    - /lib64/ld-linux-x86-64.so.2
+
+outputs:
+  - name: libcusparse
+    files:
+      - lib/libcusparse.so.*
+    requirements:
+      run:
+        - libnvjitlink =12.*
```
bdice, line 47:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libnvjitlink =12.*
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+  - name: libcusparse-dev
+    requirements:
+      run:
+        - libcusparse >={{ version }}
```
bdice, line 52:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcusparse >={{ version }}
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    files:
+      - lib/libcusparse.so*
+      - lib/stubs/libcusparse.so*
+      - include/cusparse*
+      - src/cusparse_fortran*
+
+  - name: libcusparse-static
+    requirements:
+      run:
+        - libcusparse-dev >={{ version }}
```
bdice, line 62:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libcusparse-dev >={{ version }}
```

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,87 @@
+{% set name = "libnvjitlink" %}
+{% set version = "12.0.76" %}
```
bdice, line 2:
```suggestion
{% set version = "12.0.76" %}
{% set cuda_version = "12.0" %}
```

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+
+outputs:
+  - name: libnvjitlink
```
bdice, line 32:
```suggestion
  - name: libnvjitlink
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    test:
+      commands:
+        - test -f $PREFIX/include/nvJitLink.h               # [linux]
+        - test -f $PREFIX/lib/linvjitlink.so                # [linux]
+        - test -f $PREFIX/lib/linvjitlink.so.{{ version }}  # [linux]
+        - if not exist %LIBRARY_LIB%\nvJitLink.lib exit 1   # [win]
+    requirements:
+      run:
+        - libnvjitlink >={{ version }}
+        - libnvjitlink-static >={{ version }}
```
bdice, line 55:
```suggestion
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libnvjitlink >={{ version }}
        - libnvjitlink-static >={{ version }}
```

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    files:
+      - lib/libnvJitLink*.so            # [linux]
+      - targets/*/lib/libnvJitLink*.so  # [linux]
+      - include/nvJitLink.h             # [linux]
+      - targets/*/include/nvJitLink.h   # [linux]
+      - res/nvJitLink_linux.def
+      - Library\include                 # [win]
+      - Library\lib\nvJitLink.lib       # [win]
+
+  - name: libnvjitlink-static
```
bdice, line 65:
```suggestion
  - name: libnvjitlink-static
    requirements:
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/libcusolver/meta.yaml
leofang: btw I think we don't need the version constraint for `libcublas` and `libnvjitlink` anymore, now that we pin `cuda-version` tightly?

recipes/libcusparse/meta.yaml
leofang: cc: @adibbley for vis (the comments were autogen'd from Alex's script IIRC)

recipes/libcusparse/meta.yaml
leofang: ditto, no need to pin `libnvjitlink` anymore?

recipes/libnvjitlink/meta.yaml
leofang: Q: Do we want to add `-static` as a requirement for `-dev`? We don't do that for other two packages above.

adibbley: We can remove the `-static` requirement here as well. Only `cuda-cudart-dev` is keeping it for nvcc support.

recipes/libcusolver/meta.yaml
adibbley: Agreed we can drop this now that every component should be pinning to `cuda-version` we should be getting this transitively.

recipes/libnvjitlink/meta.yaml
jakirkham: We've been adding to `run_constrained` following this discussion ( https://github.com/conda-forge/staged-recipes/pull/21902#discussion_r1092460458 )

Though following this discussion sounds like we want an exact pin for the static library ( https://github.com/conda-forge/staged-recipes/pull/21901#discussion_r1145407909 )

leofang:
@jakirkham Do you know if these are safe to allowlist?
```
  ERROR (libnvjitlink,Library/bin/vccorlib140.dll): $RPATH/api-ms-win-core-winrt-string-l1-1-0.dll not found in packages, sysroot(s) nor the missing_dso_whitelist.
  ERROR (libnvjitlink,Library/bin/vccorlib140.dll): $RPATH/api-ms-win-core-winrt-error-l1-1-0.dll not found in packages, sysroot(s) nor the missing_dso_whitelist.
  ERROR (libnvjitlink,Library/bin/vccorlib140.dll): $RPATH/api-ms-win-core-winrt-l1-1-0.dll not found in packages, sysroot(s) nor the missing_dso_whitelist.
```

leofang:
I think Dave hit a similar missing dll a while ago: https://github.com/conda-forge/staged-recipes/pull/21913#issuecomment-1429616023, it seems there's no other way but suppress it.

leofang:
OK, now `libcublas` (https://github.com/conda-forge/staged-recipes/pull/21901) is the only blocker.

leofang:
@jakirkham @adibbley @bdice @kkraus14 @robertmaynard This is ready. The package layouts here follow exactly the same as that of `libcublas*` (https://github.com/conda-forge/staged-recipes/pull/21901).

cc: @conda-forge/staged-recipes

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,181 @@
...
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+        - libcublas
+        - libnvjitlink
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - {{ pin_compatible("libcublas", max_pin="x") }}
```
kkraus14, line 48: The `cuda-version` pin here will force the `libcublas` minor versions anyway. Should we pin to the minor version to make thinks more clear?

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,152 @@
...
+      run_exports:
+        - {{ pin_subpackage("libnvjitlink", max_pin="x") }}
+    files:
+      - lib/libnvJitLink*.so                            # [linux]
+      - lib/pkgconfig                                   # [linux]
+      - targets/{{ target_name }}/include/nvJitLink.h   # [linux]
+      - targets/{{ target_name }}/lib/libnvJitLink*.so  # [linux]
+      #- targets/{{ target_name }}/lib/stubs             # [linux]  # bug: currently missing
+      - Library\include                                 # [win]
+      - Library\lib                                     # [win]
```
leofang, line 77:
```suggestion
      - Library\include                                 # [win]
      - Library\lib\nvJitLink.lib                       # [win]
```

recipes/libcusolver/meta.yaml
leofang: Yes, for now this doesn't matter since we are guarded by cuda-version. I slightly prefer to have the pinning on cublas/nvjitlink relaxed, in case in the future we want to allow certain degree of mix-n-match. But I don't have strong opinion here.

recipes/libcusolver/build.sh
leofang: Since we place everything under targets now, this discussion is moot.

recipes/libcusolver/meta.yaml
kkraus14: If the library only requires the major version as opposed to newer than the matching minor version then I'm generally a -1 to having `cuda-version` be a run requirement as opposed to run_constrained.

leofang: Based on the summary in https://github.com/conda-forge/cuda-version-feedstock/issues/1#issuecomment-1462998398 (and the discussion around it) I think we agreed that
- `cuda-version` needs to be a run requirement so that 1. we can use it as an identifier to declare which CUDA Version a package supports, and 2. later we can request the exact components for a certain CTK version can be requested via `cuda-toolkit` & co.
- In theory CTK components from within the same major version could be mix-n-matched (ex: cuSOLVER v11 depends on cuBLAS v12), but for now we're using `cuda-version` as a safe guard to disallow this, and will revisit this decision in the future. 

I'd prefer to stick to the above conclusion for now, and raise an issue in the created feedstocks to track this discussion. I don't think it's a blocker given that mix-n-match is still disallowed. Otherwise, this discussion should be brought back to cuda-version-feedstock, not here.

kkraus14: Then I think we should pin to the minor version of `libcublas` here to reduce the ambiguity of what our dependency constraints are at this time.

leofang: See d6f084e.

kkraus14 (approved):

recipes/libcusolver/build.sh
```diff
@@ -0,0 +1,33 @@
...
+    [[ $i == "build_env_setup.sh" ]] && continue
+    [[ $i == "conda_build.sh" ]] && continue
+    [[ $i == "metadata_conda_debug.yaml" ]] && continue
+    if [[ $i == "lib" ]] || [[ $i == "include" ]]; then
+        # Headers and libraries are installed to targetsDir
+        mkdir -p ${PREFIX}/${targetsDir}
+        mkdir -p ${PREFIX}/$i
+        cp -rv $i ${PREFIX}/${targetsDir}
+        if [[ $i == "lib" ]]; then
+            for j in "$i"/*.so*; do
```
adibbley, line 23:
```suggestion
            for j in "$i"/*.so*; do
            for j in "$i"/{*.so,stubs/*.so}*; do
```
Do we want stubs symlinked from `lib/stubs/*.so` to `targets/{{ target_platform }}/stubs/*.so`?

cudart [tries to](https://github.com/adibbley/staged-recipes/blob/cuda-cudart/recipes/cuda-cudart/meta.yaml#L218)
But libcublas does not

cc @bdice

leofang: cc: @robertmaynard for input

bdice: Let's follow the precedent set by libcublas on this (unless we have a reason to change libcublas as well). The cudart package is somewhat different than the math libraries in structure and use of static linking. We had some discussions about the topic of stubs, but I don't recall a conclusive reason for/against this, from our past conversations. We can always add symlinks in `$PREFIX/lib/stubs` later if needed.

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,181 @@
...
+        - test -L $PREFIX/lib/libcusolver.so.{{ version }}                  # [linux]
+        - test -L $PREFIX/lib/libcusolver.so.{{ version.split(".")[0] }}    # [linux]
+        - test -L $PREFIX/targets/{{ target_name }}/lib/libcusolver.so.{{ version }}                  # [linux]
+        - test -f $PREFIX/targets/{{ target_name }}/lib/libcusolver.so.{{ version.split(".")[0] }}    # [linux]
+        - test -L $PREFIX/lib/libcusolverMg.so.{{ version }}                # [linux]
+        - test -L $PREFIX/lib/libcusolverMg.so.{{ version.split(".")[0] }}  # [linux]
+        - test -L $PREFIX/targets/{{ target_name }}/lib/libcusolverMg.so.{{ version }}                # [linux]
+        - test -f $PREFIX/targets/{{ target_name }}/lib/libcusolverMg.so.{{ version.split(".")[0] }}  # [linux]
+        - if not exist %LIBRARY_BIN%\\cusolver*.dll exit 1                  # [win]
+        - if not exist %LIBRARY_BIN%\\cusolverMg*.dll exit 1                # [win]
```
bdice, line 63: Tiny suggestion: We discussed this on libcublas and it seems that only one slash is needed here. Let's match that style.
```suggestion
        - if not exist %LIBRARY_BIN%\cusolver*.dll exit 1                  # [win]
        - if not exist %LIBRARY_BIN%\cusolverMg*.dll exit 1                # [win]
```

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,181 @@
...
+        - test -f $PREFIX/targets/{{ target_name }}/lib/stubs/libcusolver.so               # [linux]
+        - test -f $PREFIX/targets/{{ target_name }}/lib/stubs/libcusolverMg.so             # [linux]
+        - if not exist %LIBRARY_INC%\\cusolverDn.h exit 1                   # [win]
+        - if not exist %LIBRARY_INC%\\cusolverMg.h exit 1                   # [win]
+        - if not exist %LIBRARY_INC%\\cusolverRf.h exit 1                   # [win]
+        - if not exist %LIBRARY_INC%\\cusolverSp.h exit 1                   # [win]
+        - if not exist %LIBRARY_INC%\\cusolverSp_LOWLEVEL_PREVIEW.h exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\cusolver_common.h exit 1              # [win]
+        - if not exist %LIBRARY_LIB%\\cusolver*.lib exit 1                  # [win]
+        - if not exist %LIBRARY_LIB%\\cusolverMg*.lib exit 1                # [win]
```
bdice, line 124:
```suggestion
        - if not exist %LIBRARY_INC%\cusolverDn.h exit 1                   # [win]
        - if not exist %LIBRARY_INC%\cusolverMg.h exit 1                   # [win]
        - if not exist %LIBRARY_INC%\cusolverRf.h exit 1                   # [win]
        - if not exist %LIBRARY_INC%\cusolverSp.h exit 1                   # [win]
        - if not exist %LIBRARY_INC%\cusolverSp_LOWLEVEL_PREVIEW.h exit 1  # [win]
        - if not exist %LIBRARY_INC%\cusolver_common.h exit 1              # [win]
        - if not exist %LIBRARY_LIB%\cusolver*.lib exit 1                  # [win]
        - if not exist %LIBRARY_LIB%\cusolverMg*.lib exit 1                # [win]
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,156 @@
...
+        - {{ pin_compatible("libnvjitlink", max_pin="x.x") }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -L $PREFIX/lib/libcusparse.so.{{ version }}                                  # [linux]
+        - test -L $PREFIX/lib/libcusparse.so.{{ version.split(".")[0] }}                    # [linux]
+        - test -L $PREFIX/{{ target_name }}/lib/libcusparse.so.{{ version }}                # [linux]
+        - test -f $PREFIX/{{ target_name }}/lib/libcusparse.so.{{ version.split(".")[0] }}  # [linux]
+        - if not exist %LIBRARY_BIN%\\cusparse*.dll exit 1                                  # [win]
```
bdice, line 56:
```suggestion
        - if not exist %LIBRARY_BIN%\cusparse*.dll exit 1                                   # [win]
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,156 @@
...
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```
bdice, line 130: Add arm-variant constraint.
```suggestion
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
      run_constrained:
        - arm-variant * {{ arm_variant_type }} # [aarch64]
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,156 @@
...
+        - {{ pin_subpackage("libcusparse", exact=True) }}
+      run_constrained:
+        - libcusparse-static >={{ version }}  # [linux]
```
bdice, line 93: Add arm-variant constraint.
```suggestion
        - arm-variant * {{ arm_variant_type }} # [aarch64]
        - libcusparse-static >={{ version }}  # [linux]
```

recipes/libnvjitlink/meta.yaml
```diff
@@ -0,0 +1,151 @@
...
+  - name: libnvjitlink-dev
+    build:
+      run_exports:
+        - {{ pin_subpackage("libnvjitlink", max_pin="x") }}
+    files:
+      - lib/libnvJitLink*.so                            # [linux]
+      - lib/pkgconfig                                   # [linux]
+      - targets/{{ target_name }}/include/nvJitLink.h   # [linux]
+      - targets/{{ target_name }}/lib/libnvJitLink*.so  # [linux]
+      #- targets/{{ target_name }}/lib/stubs             # [linux]  # bug: currently missing
```
bdice, line 74: Is there an nvbug filed upstream? Have we confirmed that nvjitlink is _supposed_ to ship stubs?

bdice (commented):
A few minor tweaks are needed for missing `arm-variant` constraints, then I can approve. Otherwise LGTM! I verified that all the upstream contents are making it into the downstream packages.

recipes/libnvjitlink/meta.yaml
leofang: I don't know if a bug is filed (I bet there's), but this is documented:
> For a cross-compile toolkit (such as linux64 host, aarch64 target), we are missing the host-side stub library for libnvJitLink. As a workaround, you can copy the libnvJitlink.so from the target install (for example, /usr/local/cuda-12.1/targets/aarch64-linux/lib/libnvJitLink.so) to the host install (/usr/local/cuda-12.1/targets/aarch64-linux/lib/stubs/libnvJitLink.so). Similarly if you are using the static library version (/usr/local/cuda-12.1/targets/aarch64-linux/lib/libnvJitLink_static.a), can copy it from the target install (i.e. install on the device) to the same path on the host install. For a sbsa cross-compile replace aarch64 with sbsa in the above copies.

https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#general-cuda-known-issues

recipes/libcusparse/meta.yaml
leofang: Technically we don't need `arm-variant` for `-dev` packages, as it'd get this requirement from the `lib` package. But I am fine following the same style.

recipes/libcusparse/meta.yaml
leofang: Good catch, thanks!

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,159 @@
...
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - {{ pin_subpackage("libcusparse", exact=True) }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }} # [aarch64]
```
leofang, line 93:
```suggestion
        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,159 @@
...
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }} # [aarch64]
```
leofang, line 133:
```suggestion
        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```

recipes/libcusparse/meta.yaml
bdice: Agreed -- but let's match the style. There are potentially other factors that would make this matter, rather than being only a transitive dependency/constraint. (At one point we had discussed the possibility of `-dev` packages shipping only headers and stubs, which would eliminate the dependency on the `lib` package.)

recipes/libcusolver/build.sh
adibbley: Thanks, that is what I thought as well. Just wanted to be sure since I couldn't find it explicitly stated anywhere. We can resolve this with no change.

recipes/libnvjitlink/meta.yaml
adibbley: I see libnvJitLink stubs included in the release after 12.1.

leofang: Awesome, thanks Alex. Let's keep this line then, and uncomment it when 12.2 is out. Mark as resolved.

leofang:
> @jakirkham @adibbley @bdice @kkraus14 @robertmaynard This is ready. 

All comments are addressed/resolved. Let's merge?

bdice (approved):

recipes/libcusolver/build.sh
robertmaynard: I agree with @bdice

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,159 @@
...
+      description: |
+        cuSPARSE - Basic Linear Algebra for Sparse Matrices on NVIDIA GPUs
+      doc_url: https://docs.nvidia.com/cuda/cusparse/index.html
+
+  - name: libcusparse-static
+    build:
+      skip: true  # [win]
+    files:
+      - targets/{{ target_name }}/lib/libcusparse_static.a
+    requirements:
```
robertmaynard, line 122: Looks like we are missing a requirement on nvJitLink. The `libcusparse_static.a` ships with undefined nvJitLink symbols

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,159 @@
...
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```
leofang, line 131: @robertmaynard Something like this?
```suggestion
      host:
        - cuda-version {{ cuda_version }}
        - libnvjitlink-static
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - {{ pin_compatible("libnvjitlink-static", max_pin="x.x") }}
```

bdice: Or should it be this? Is it needed in `host`? (I might be unaware of the reason.)
```suggestion
      host:
        - cuda-version {{ cuda_version }}
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - libnvjitlink-static
```

robertmaynard: I think it only needs to be in run, but needs to have the `pin_compatible`

leofang: > I think it only needs to be in run, but needs to have the `pin_compatible`

I don't think this makes sense. `pin_compatible` needs to refer to the version used in the build/host section, see my suggestion above.

bdice: pin_compatible only works if it’s in the host section. That is how it determines compatibility. But shouldn’t compatibility (same CTK minor version) be guaranteed by cuda-version?

leofang: > But shouldn’t compatibility (same CTK minor version) be guaranteed by cuda-version?

@kkraus14 prefers it to be explicit: https://github.com/conda-forge/staged-recipes/pull/21924#discussion_r1155402875 We can revisit this discussion in the feedstocks.

robertmaynard: libcusolver uses `pin_compatbile` in the `run` section:
```
      run:
        - {{ pin_compatible("libcublas", max_pin="x.x") }}
        - {{ pin_compatible("libnvjitlink", max_pin="x.x") }}
```
https://github.com/conda-forge/staged-recipes/blob/5cf9dd7c5648fe5c9eaa0b4c2162ef6a5fd49190/recipes/libcusolver/meta.yaml#L47

leofang: @robertmaynard please read the lines right above line 47 🙂

leofang: If it helps, this is the documentation for `pin_compatible`:
https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html#extra-jinja2-functions

recipes/libcusparse/meta.yaml
leofang: @robertmaynard How did you see the symbols are missing? I just checked and there's no nvJitLink symbols.
```
nm -C /usr/local/cuda-12.0.0/lib64/libcusolver_static.a | grep " U " | sort
```

leofang: nvm looked at the wrong file

leofang:
```
nm -C /usr/local/cuda-12.0.0/lib64/libcusparse_static.a | grep " U " | sort
```

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,181 @@
...
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```
leofang, line 150:
```suggestion
      host:
        - cuda-version {{ cuda_version }}
        - libcublas-static
        - libcusparse-static
      run:
        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
        - {{ pin_compatible("libcublas-static", max_pin="x.x") }}
        - {{ pin_compatible("libcusparse-static", max_pin="x.x") }}
```

recipes/libcusparse/meta.yaml
leofang: Fixed in bc7eec96, great catch @robertmaynard, thanks!

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,181 @@
...
+        cuSOLVER - Direct Linear Solvers on NVIDIA GPUs
+      doc_url: https://docs.nvidia.com/cuda/cusolver/index.html
+
+  - name: libcusolver-static
+    build:
+      skip: true  # [win]
+    files:
+      - targets/{{ target_name }}/lib/libcusolver*_static.a
+      - targets/{{ target_name }}/lib/libmetis_static.a
+    requirements:
```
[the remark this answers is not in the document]
leofang: Fixed in bc7eec96, great catch @robertmaynard, thanks!

robertmaynard (approved):

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,185 @@
...
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+        - libcublas
+        - libnvjitlink
```
kkraus14, line 45: Should these be versioned more explicitly? I get that the `cuda-version` package will constrain them to a specific minor version, but I think we'd be best off explicitly capturing the dependency versions instead of relying purely on the `cuda-version` package.

recipes/libcusolver/meta.yaml
```diff
@@ -0,0 +1,185 @@
...
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+        - libcublas-static
+        - libcusparse-static
```
kkraus14, line 150: +1 here

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - Library\bin\cusparse*.dll                        # [win]
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+        - libnvjitlink
```
kkraus14, line 44: +1

recipes/libcusparse/meta.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - targets/{{ target_name }}/lib/libcusparse_static.a
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - sysroot_{{ target_platform }} 2.17    # [linux]
+      host:
+        - cuda-version {{ cuda_version }}
+        - libnvjitlink-static
```
kkraus14, line 130: +1

recipes/libcusolver/meta.yaml
leofang: It is @adibbley's intention (and also kinda by design) to avoid explicitly spell out the version, and let conda do the work for us. Otherwise, it's too much maintenance overhead, even with automation.

adibbley: Yeah I would much rather let `cuda-version` handle the versioning here. It is common for a CUDA update release to not bump all component versions. 

So we could have a situation where CUDA 12.0.X and CUDA 12.0.Y ship the same version of `libcusolver`, but a new `libcublas`. Leaving these unversioned gives us the `>12.0,<12.1` for free. This constraint would be error prone with every component following their own versioning.

kkraus14 (approved):

jakirkham:
Thanks all! 🙏

Let's follow up on anything else on the feedstocks

recipes/libcusolver/meta.yaml
leofang: Opened two issues in the feedstocks to track this discussion:
- https://github.com/conda-forge/libcusolver-feedstock/issues/5
- https://github.com/conda-forge/libcusparse-feedstock/issues/4

PR #21952, opened 2023-02-03
Adding cuda-compat recipe

adibbley:
Checklist
- [x] Build number is 0.


adibbley:
Here is the recipe for cuda-compat 
cc @jakirkham @kkraus14

kkraus14:
I imagine this would get used similarly to how people use the `cuda-compat` package today.

From the docs https://docs.nvidia.com/deploy/cuda-compatibility/#forward-compatibility-title:
- It seems like it's only officially supported on a subset of GPUs,
> Forward Compatibility is applicable only for systems with NVIDIA Data Center GPUs or select [NGC Server Ready](https://docs.nvidia.com/ngc/ngc-ready-systems/index.html#abstract) SKUs of RTX cards.
- Users are encouraged to explicitly set `LD_LIBRARY_PATH` to opt into using it

I don't think we want to encourage anything explicitly depending on this package given the subset of supported GPUs.

As a starting point, I'd propose we package this similarly to how the current recipe does, not in the default linker search path, which allows a user to set something like `LD_LIBRARY_PATH` themselves. In the future, we could explore using activation scripts to automatically set things if we find there's desired default behaviors.

jakirkham:
Guess the only other question is whether we add a `cuda-version` dependency on this package ( https://github.com/conda-forge/cuda-version-feedstock/issues/1 )

bdice:
> Guess the only other question is whether we add a `cuda-version` dependency on this package ( [conda-forge/cuda-version-feedstock#1](https://github.com/conda-forge/cuda-version-feedstock/issues/1) )

I think this is an exception and does not need a `cuda-version` dependency. It appears to be versioned according to the driver rather than the toolkit:
```
{% set version = "525.60.13" %}
```
However, we might need an explicit `run_constrained: __cuda>=12` pinning on this recipe, if that would normally come from the `cuda-version`.

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+build:
+  number: 0
+  skip: true  # [osx]
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
bdice, line 26: https://github.com/conda-forge/staged-recipes/pull/21952#issuecomment-1468646082
We might need CUDA 12 drivers for this package to be useful.
```suggestion
    - {{ compiler("cxx") }}
  run_constrained:
    - __cuda >=12
```

jakirkham: Do we need a more specific pinning for driver version? Or is a lower bound like this sufficient?

kkraus14:
> I think this is an exception and does not need a `cuda-version` dependency. It appears to be versioned according to the driver rather than the toolkit:
> 
> ```
> {% set version = "525.60.13" %}
> ```
> 
> However, we might need an explicit `run_constrained: __cuda>=12` pinning on this recipe, if that would normally come from the `cuda-version`.

If using `apt`, I can do `apt install cuda-compat-12-0` to get the cuda-compat package for CUDA 12.0. With this versioning scheme there's no easy way for someone to say "Give me the cuda-compat package aligned with CUDA 12.0".

Maybe we could split into `cuda-compat` and `cuda-compat-impl` type packages or something? Where `impl` is versioned based on the driver version, i.e. `525.60.13` and `cuda-compat` is versioned according to the CUDA version, i.e. `12.0`?

adibbley:
> If using `apt`, I can do `apt install cuda-compat-12-0` to get the cuda-compat package for CUDA 12.0. With this versioning scheme there's no easy way for someone to say "Give me the cuda-compat package aligned with CUDA 12.0".
> 
> Maybe we could split into `cuda-compat` and `cuda-compat-impl` type packages or something? Where `impl` is versioned based on the driver version, i.e. `525.60.13` and `cuda-compat` is versioned according to the CUDA version, i.e. `12.0`?

We could add the CUDA version to the package name as we do with `.debs` to match this functionality. The tradeoff is ease of upgrades, but I'm not sure users of this package would want that anyway.

jakirkham:
Interesting would be tempted to use `cuda-version` for this kind of behavior. This seems inline with general `cuda-version` usage as discussed in issue ( https://github.com/conda-forge/cuda-version-feedstock/issues/1 ). Though maybe I'm missing something

kkraus14:
Would you ever want a newer `cuda-compat` package with an older set of cuda runtime libraries? I.E. get the `cuda-compat` for CUDA 12.2 with cuBLAS and cuRAND 12.0?

That feels like a legitimate and reasonable use case that wouldn't work with the `cuda-version` package.

bdice:
Yes, @kkraus14's example is exactly what I had in mind. I also don't really want to bake the version into the package name. I liked Keith's other suggestion of splitting into a `cuda-compat` and `cuda-compat-impl` for double-versioning. Or perhaps a build string would be enough to select the CUDA version "by name" without needing proper version logic (comparators, etc.).

jakirkham:
What about `cuda-compat-version`?

kkraus14:
> What about `cuda-compat-version`?

That doesn't solve the issue if the main package uses `cuda-version` as a run dependency.

kkraus14 (approved):

recipes/cuda-compat/build.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+mkdir -p ${PREFIX}/cuda-compat
+
+cp -v lib/libcuda.so.${DRV_VERSION} ${PREFIX}/cuda-compat
+ln -sv libcuda.so.${DRV_VERSION} ${PREFIX}/cuda-compat/libcuda.so.1
+ln -sv libcuda.so.1 ${PREFIX}/cuda-compat/libcuda.so
+cp -v lib/libnvidia-nvvm.so.${DRV_VERSION} ${PREFIX}/cuda-compat
+ln -sv libnvidia-nvvm.so.${DRV_VERSION} ${PREFIX}/cuda-compat/libnvidia-nvvm.so.4
```
bdice, line 11: How will we know when to change these soversions like ".so.4" and ".so.1"? Are those fixed or do they change?

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,91 @@
+{% set name = "cuda-compat" %}
+{% set version = "525.60.13" %}
+{% set cuda_version = "12.0" %}
```
bdice, line 3: Is this the true cuda-compat versioning scheme / does this need to be a longer version like `12.0.0`? If cuda-compat can release new versions with CUDA updates for the same minor version, we need to be able to distinguish them.

Maybe we just need to use a longer version for the package version, and truncate it to `major.minor` for the `.so.{{ version }}` cases below?

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+    about:
+      home: https://developer.nvidia.com/cuda-toolkit
+      license_file: LICENSE
+      license: LicenseRef-NVIDIA-End-User-License-Agreement
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: CUDA Compatibility Platform
+      description: |
+          Package containing all the necessary CUDA driver files related to forward
+          compatibility. This compatibility enables newer CUDA runtimes to work with
+          older display drivers. Used for TESLA cards only.
```
bdice, line 42:
```suggestion
          older display drivers. Used for Tesla cards only.
```

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
```
bdice, line 74:
```suggestion
          older display drivers. Used for Tesla cards only.
```

bdice (changes requested):

recipes/cuda-compat/build.sh
adibbley: Looking at when the `.deb`s bumped this to `.so.4`, the change was requested to us during testing.

recipes/cuda-compat/meta.yaml
adibbley: I can expand to the full CUDA version, we do bump driver versions in update releases.

This shouldn't need to be truncated for the below `.so`s as those are using the driver version.

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: CUDA Compatibility Platform
+  description: |
+      Package containing all the necessary CUDA driver files related to forward
+      compatibility. This compatibility enables newer CUDA runtimes to work with
+      older display drivers. Used for TESLA cards only.
```
bdice, line 86:
```suggestion
      older display drivers. Used for Tesla cards only.
```

recipes/cuda-compat/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: CUDA Compatibility Platform
+      description: |
+          Package containing all the necessary CUDA driver files related to forward
+          compatibility. This compatibility enables newer CUDA runtimes to work with
+          older display drivers. Used for Tesla cards only.
+      doc_url: https://docs.nvidia.com/cuda/index.html
+
+  - name: cuda-compat-impl
+    version: {{ version }}
```
bdice, line 46: Were we planning to have a "dual version" here, for both CTK and driver versions? I'm not seeing both `version` (driver) and `cuda_version` (CTK version) being used in the impl/non-impl packages. Maybe I'm missing something.

bdice (commented):
I missed fixing one instance of `TESLA` earlier. Also a versioning question.

recipes/cuda-compat/meta.yaml
adibbley: Both versions are being used, `cuda-compat` uses `cuda_version` [here](https://github.com/conda-forge/staged-recipes/pull/21952/files#diff-688d8da022bb4c13e857bc25239b355421f8d5ff0fd0845ba2299ccab64abf45R11)

And here are the artifacts:
```
    /home/conda/staged-recipes/build_artifacts/linux-64/cuda-compat-impl-525.60.13-h59595ed_0.conda \
    /home/conda/staged-recipes/build_artifacts/linux-64/cuda-compat-12.0.0-ha770c72_0.conda
```

jakirkham: Should we rename `version` to `driver_version` for clarity?

adibbley: Sure, renamed it in the latest.

bdice (approved):

jakirkham:
Thanks all! 🙏

Let's get this in. We can follow up on anything else in the feedstock

PR #21953, opened 2023-02-03
Adding cuda-cccl recipe

adibbley:
Checklist
- [x] Build number is 0.


recipes/cuda-cccl/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+
+[[ ${target_platform} == "linux-64" ]] && targetsDir="targets/x86_64-linux"
+[[ ${target_platform} == "linux-ppc64le" ]] && targetsDir="targets/ppc64le-linux"
+[[ ${target_platform} == "linux-aarch64" ]] && targetsDir="targets/sbsa-linux"
+
+for i in `ls`; do
```
robertmaynard, line 10: This script currently causes `lib/cmake` to be a symlink which will break other conda packages that need to install directories into `lib/cmake` ( Or will fail if that exists as a directory ).

Instead we need to symlink the `lib/cmake/[cub,libcudacxx,thrust]`  directories

robertmaynard (changes requested):

recipes/cuda-cccl/build.sh
adibbley: Thanks Robert! This should be fixed now.

robertmaynard: Just verified and the update works, thank you

jakirkham:
@conda-forge/core @conda-forge/staged-recipes, could you please review?

As noted in comment ( https://github.com/conda-forge/staged-recipes/pull/21723#issuecomment-1429835268 ), this is needed for `cuda-cudart`, which is needed for `cuda-nvcc`

jakirkham:
cc @leofang

kkraus14:
Are we not splitting this up into `cub`, `thrust`, `libcu++`, etc. packages?

Also, what is the strategy with regards to this and the public GitHub repos for these libraries?

jakirkham:
cc @robertmaynard (who was thinking about similar questions recently)

leofang:
> Are we not splitting this up into `cub`, `thrust`, `libcu++`, etc. packages?

From what I learned from Jake we should not do this. From the package maintenance viewpoint it's also more work. We should just retire {thrust,cub}-feedstocks going forward.

isuruf:
Preferring closed source builds over open source builds is not something we want to ever encourage.

kkraus14:
> Preferring closed source builds over open source builds is not something we want to ever encourage.

I don't think that's what Leo is saying. Jake = @jrhemstad who leads the cccl team at NVIDIA.

I believe he's saying that we shouldn't package Thrust, Cub, libcudf++, and anything else contained in cccl separately, presumably because there's plans that blur the boundaries between these libraries more than they're already blurred.

That was my first question. The question about how we're handling using the open source repo(s) vs this prepackaged release from NVIDIA is a separate question.

jrhemstad:
> I believe he's saying that we shouldn't package Thrust, Cub, libcudf++, and anything else contained in cccl separately, presumably because there's plans that blur the boundaries between these libraries more than they're already blurred.

Correct. This hasn't been formally announced yet, but the Thrust/CUB/libcu++ repos will be merging into a single CCCL monorepo. Our vision for the future of these libraries is that they are parts of a unified whole and shouldn't be separated.

leofang:
Thanks, Keith/Jake. Indeed, after NVIDIA/CCCL GH repo is set up we can definitely switch to use the open sourced tarballs, but right now we're not there yet, so what Alex has in this PR is just a temporary solution. 

I have made no comment on the choice of close vs open sources. Seems to me like a misinterpretation?

isuruf:
Thanks for the clarification. Until NVIDIA/CCCL repo is setup, I think we should depend on thrust, cub conda packages in this package. Otherwise they will clobber each other.

leofang:
Clobbering is an important concern, but I think we can set up `run_constrained` to avoid this. I can't comment on if it's possible (or how much work is needed) to set up Thrust/CUB/... and make an equivalent cuda-cccl package, though, so I'd leave this to others.

jakirkham:
Would suggest that we not mess with how the compiler has been packaged by pulling in components not tested or used with it. This seems to introduce unnecessary reliability risk in a key component of the `cuda-nvcc` toolchain

kkraus14:
> Would suggest that we not mess with how the compiler has been packaged by pulling in components not tested or used with it. This seems to introduce unnecessary reliability risk in a key component of the `cuda-nvcc` toolchain

The challenge here is that in a normal system install, cccl lands in `/usr/local/cuda` and then if someone was to install Thrust themselves it would typically land in `/usr/local`, so the two wouldn't clobber.

For us in a conda environment, both cccl and Thrust land in `$CONDA_PREFIX` and do clobber. And then historically, Thrust, Cub, and libcu++ have had a release cycle that isn't necessarily aligned to the release cycle of CUDA and projects like RAPIDS have taken advantage of new versions that don't coincide with a CUDA release version. On the other hand, I believe cccl is tied to the CUDA release schedule, at least as of now?

Has anyone inspected the cccl package to see if it's essentially just a metapackage of Thrust, Cub, and libcu++?

jakirkham:
Suppose we could put the compiler in `$PREFIX/cuda` and symlink relevant pieces into `$PREFIX`. IOW follow the analogous structure in Conda packages

isuruf:
> Has anyone inspected the cccl package to see if it's essentially just a metapackage of Thrust, Cub, and libcu++?

Yes, those three plus `include/nv` headers

kkraus14:
> > Has anyone inspected the cccl package to see if it's essentially just a metapackage of Thrust, Cub, and libcu++?
> 
> Yes, those three plus `include/nv` headers

That's part of libcu++: https://github.com/NVIDIA/libcudacxx/tree/main/include%2Fnv

Sounds like we can just take those three repos separately for now and we just need to make it a point as we update to ensure that we get the right versions?

jrhemstad:
> Has anyone inspected the cccl package to see if it's essentially just a metapackage of Thrust, Cub, and libcu++?

I'd hope so, but fwiw, we don't own anything to do with that packaging, so I'm always leery of anything we didn't do ourselves :sweat_smile:.

jakirkham:
Yeah this is essentially my concern. If we start doing custom things to this package, it is going to get a lot harder to get support if we need it. Am pretty reluctant to go down that path

jrhemstad:
@jakirkham are you advocating for only pulling the cccl package from the NVIDIA redist packages? Or for pulling from GH?

One important thing to note is that we will be releasing on GH far more frequently than the CTK. So if we want to enable people to use the latest version of cccl stuff via conda, that would mean you'd want to pull from GH.

jakirkham:
Am preferring using the redist packages for the compiler's needs. It's easier that way to say we are distributing the compiler in the way it was intended (and as it was QA'd) whenever an issue shows up we need help with. Also this will save extra back-and-forth about packaging choices that would deviate from using the redist packages. Plus should keep maintenance of these packages simple (desirable for long term maintainability).

<hr>

That said, I understand folks want to use more recent CCCL versions as part of their builds. For the RAPIDS (or other outside users), continuing to use CMake (or other source retrieval mechanisms) probably makes sense.

For other recipes in conda-forge, think we need to come up with a solution so they don't clobber each other without messing with the compiler. Here are some thoughts:

1. Silo the compiler somehow (maybe `$PREFIX/cuda`?)
2. Have a separate more up-to-date CCCL package for recipes that lives in another location (maybe `$PREFIX/cccl`?)
3. Use [multiple `source`s]( https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#source-from-multiple-sources ) in recipes to fetch source code for CCCL & the package (here's [an example]( https://github.com/conda-forge/apr-feedstock/blob/6af7c12c5cf9dc3ce252937f21036da5f77047b3/recipe/meta.yaml#L10-L25 ))
4. ?

Am open to other suggestions along these lines

isuruf:
I'm fine with doing 3 and keeping the includes only in `$PREFIX/targets/$arch-linux/`. That way they won't clobber with the other conda packages and if they are installed, vendored copies in `cccl` will not be used since `$PREFIX/targets` has lower priority.

kkraus14:
> I'm fine with doing 3 and keeping the includes only in `$PREFIX/targets/$arch-linux/`. That way they won't clobber with the other conda packages and if they are installed, vendored copies in `cccl` will not be used since `$PREFIX/targets` has lower priority.

Is the idea only to enable `nvcc` with these headers? If not, the challenge I see with doing this is that we also can't install things like their CMake / pkgconfig files into a standard location because you risk clobbering them as well and without those in the standard location typical build scripts won't find them as expected. This also will make it really easy for one package to compile against one version in `targets` and another package to compile against a different version from say the `thrust` package and to get into potentially incompatible breaking situations because of it.

I also think requiring downstream recipes to source `cccl` from the esoteric NVIDIA URL that provides the download is asking too much from recipe builders and maintainers.

There's pretty clearly versions of each of the 3 packages shipped that we can align on: https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#cuda-toolkit-major-component-versions. Especially considering these are all header only libraries, we should be able to make them exactly match the CCCL distribution if desired.

Alternatively, maybe we could come up with some kind of build string or selector package or something of the like and use `track_features` to downweigh folks from getting a non-standard configuration by default? Could then ship `CCCL` versions of the 3 packages as well as a `CCCL` metapackage, but give users an escape hatch if they want to explicit request a newer version of some of the packages? Might make the recipe a bit funky where sometimes we'd want to use the `cccl` download from NVIDIA URL, and other times we'd want to use github URLs from `thrust`, `cub`, and `libcudacxx`, but it should still be manageable enough?

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+source:
+  - url: https://github.com/NVIDIA/thrust/archive/refs/tags/{{ version }}.{{ extension }}
+    folder: thrust
+    sha256: 06df373cf487cf3710be6798bc5441a03685b78e5c1338e2f14005f7ccd030e6  # [linux]
+    sha256: 39a9fc88b0b06da118cc3ed28cbfca27b2a7ae192877996cab1295176ceed65e  # [win]
+  - url: https://github.com/NVIDIA/cub/archive/refs/tags/{{ version }}.{{ extension }}
+    folder: cub
+    sha256: beeef0962c87eca994db06cb9e9168a71c63fb369398a5b5f1db923140814701  # [linux]
+    sha256: 22308facf62421d6dd2489b9554ffa4908dca13d3921a3e7f2965c5a6984cd20  # [win]
+  - url: https://github.com/NVIDIA/libcudacxx/archive/refs/tags/1.9.0.{{ extension }}
```
kkraus14, line 23: I think we should use a version tag for this instead of hardcoding `1.9.0`?

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+build:
+  number: 0
+  skip: true  # [osx]
+
+outputs:
+  - name: cuda-cccl
+    requirements:
+      run:
+        - cuda-cccl_{{ target_platform }} =={{ version }}
+        - cuda-cccl-impl >=2.0.1
```
kkraus14, line 30: Do we not want to pin this to an exact version?

recipes/cuda-cccl-impl/meta.yaml
adibbley: Added tags for the 3 components now. This will get cleaned up once the repos are merged and we only have 1 version here.

recipes/cuda-cccl/meta.yaml
adibbley: We want to allow updating to versions of CCCL shipped outside of CTK schedules.

kkraus14: I thought the idea was that if someone installed `cuda-cccl=12.0` they get the exact version shipped with CUDA Toolkit 12.0. Then if someone wanted to deviate from that they would only install the `cuda-cccl-impl` package and not the `cuda-cccl` package.

adibbley: Yeah that makes sense. I was still thinking CUDA would be pulling in `cuda-cccl`, but we would just use `cuda-cccl_{{ target_platform }}` instead here.

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    sha256: 06df373cf487cf3710be6798bc5441a03685b78e5c1338e2f14005f7ccd030e6  # [linux]
+    sha256: 39a9fc88b0b06da118cc3ed28cbfca27b2a7ae192877996cab1295176ceed65e  # [win]
+  - url: https://github.com/NVIDIA/cub/archive/refs/tags/{{ cub_version }}.{{ extension }}
+    folder: cub
+    sha256: beeef0962c87eca994db06cb9e9168a71c63fb369398a5b5f1db923140814701  # [linux]
+    sha256: 22308facf62421d6dd2489b9554ffa4908dca13d3921a3e7f2965c5a6984cd20  # [win]
+  - url: https://github.com/NVIDIA/libcudacxx/archive/refs/tags/{{ libcudacxx_version }}.{{ extension }}
+    folder: libcudacxx
+    sha256: 05b1435ad65f5bdef1bb8d1eb29dc8f0f7df34c01d77bf8686687a4603588bce  # [linux]
+    sha256: d7864460ce8def8e0620cb9dd891e94299ade75eac75aa96b34001af3cdb074c  # [win]
```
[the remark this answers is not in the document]
jakirkham: We were thinking that we could have `cub` and `thrust` depend on this package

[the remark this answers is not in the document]
isuruf: Why? @kkraus14 mentioned that some projects rely on having a newer version of these packages than the cuda release.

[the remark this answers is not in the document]
jakirkham: We eventually want to phase out those packages and were trying to decide on the best migration plan

Right they would be pulling in the latest version of CCCL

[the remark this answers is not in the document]
isuruf: Ah, forgot about that. Then you should do `1`.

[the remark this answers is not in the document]
jakirkham: All good. Lot of moving parts here

Yeah we could add some `<` constraint on CCCL to the existing packages via a repodata patch to ensure they conflict

Does that seem reasonable or there other things we should consider?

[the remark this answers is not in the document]
isuruf: Why add repodata patch on existing packages, when you can add a run_constrained on this new package?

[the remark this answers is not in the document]
jakirkham: Think this goes back to the point above about making new `cub` and `thrust` packages aliases to the CCCL. There are still packages using `cub` and `thrust` and we need a way to gradually move them over to CCCL

leofang:
I downloaded the artifacts and looked at the file contents. No issue found.

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - if not exist %LIBRARY_LIB%\cmake\cub exit 1         # [win]
+    - if not exist %LIBRARY_LIB%\cmake\libcudacxx exit 1  # [win]
+    - if not exist %LIBRARY_INC%\thrust exit 1  # [win]
+    - if not exist %LIBRARY_INC%\cub exit 1     # [win]
+    - if not exist %LIBRARY_INC%\cuda exit 1    # [win]
+    - if not exist %LIBRARY_INC%\nv exit 1      # [win]
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE
```
jakirkham, line 54:
```suggestion
  license_file:
    - cub/LICENSE.TXT
    - thrust/LICENSE
    - libcudacxx/LICENSE.TXT
```

recipes/cuda-cccl-impl/bld.bat
```diff
@@ -0,0 +1,14 @@
...
+move thrust\thrust\cmake %LIBRARY_LIB%\cmake\thrust
+move cub\cub\cmake %LIBRARY_LIB%\cmake\cub
+move libcudacxx\lib\cmake\libcudacxx %LIBRARY_LIB%\cmake
+
+move thrust\thrust %LIBRARY_INC%
+move cub\cub %LIBRARY_INC%
+move libcudacxx\include\cuda %LIBRARY_INC%
+move libcudacxx\include\nv %LIBRARY_INC%
+
+move libcudacxx\LICENSE.TXT LICENSE
```
jakirkham, line 14:
```suggestion
```

recipes/cuda-cccl-impl/build.sh
```diff
@@ -0,0 +1,15 @@
...
+mv thrust/thrust/cmake ${PREFIX}/lib/cmake/thrust
+mv cub/cub/cmake ${PREFIX}/lib/cmake/cub
+cp -r libcudacxx/lib/cmake/libcudacxx ${PREFIX}/lib/cmake
+
+cp -r thrust/thrust ${PREFIX}/include
+cp -r cub/cub ${PREFIX}/include
+cp -r libcudacxx/include/cuda ${PREFIX}/include
+cp -r libcudacxx/include/nv ${PREFIX}/include
+
+cp -r libcudacxx/LICENSE.TXT LICENSE
```
jakirkham, line 15:
```suggestion
```

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
```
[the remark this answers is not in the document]
jakirkham:
```suggestion
  license: BSD-3-Clause AND Apache-2.0
```

jakirkham (commented):
For the licenses in `cuda-cccl-impl`, maybe we can handle them like this?

recipes/cuda-cccl-impl/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+mkdir -p ${PREFIX}/lib/cmake
+mkdir -p ${PREFIX}/include
+
+mv thrust/thrust/cmake ${PREFIX}/lib/cmake/thrust
+mv cub/cub/cmake ${PREFIX}/lib/cmake/cub
+cp -r libcudacxx/lib/cmake/libcudacxx ${PREFIX}/lib/cmake
```
bdice, line 8: Why do we `mv` some and `cp -r` others? I think we should `cp -r` all of them unless I'm missing something. And to align with other libraries' `build.sh` scripts, let's use `-v` for verbosity in the logs.
```suggestion
cp -rv thrust/thrust/cmake ${PREFIX}/lib/cmake/thrust
cp -rv cub/cub/cmake ${PREFIX}/lib/cmake/cub
cp -rv libcudacxx/lib/cmake/libcudacxx ${PREFIX}/lib/cmake
```

recipes/cuda-cccl-impl/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cp -r thrust/thrust ${PREFIX}/include
+cp -r cub/cub ${PREFIX}/include
+cp -r libcudacxx/include/cuda ${PREFIX}/include
+cp -r libcudacxx/include/nv ${PREFIX}/include
```
bdice, line 13:
```suggestion
cp -rv thrust/thrust ${PREFIX}/include
cp -rv cub/cub ${PREFIX}/include
cp -rv libcudacxx/include/cuda ${PREFIX}/include
cp -rv libcudacxx/include/nv ${PREFIX}/include
```

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+test:
+  commands:
+    - test -d $PREFIX/lib/cmake/thrust      # [linux]
+    - test -d $PREFIX/lib/cmake/cub         # [linux]
+    - test -d $PREFIX/lib/cmake/libcudacxx  # [linux]
+    - test -d $PREFIX/include/thrust        # [linux]
+    - test -d $PREFIX/include/cub           # [linux]
+    - test -d $PREFIX/include/cuda          # [linux]
+    - test -d $PREFIX/include/nv           # [linux]
```
bdice, line 43: Fix spacing.
```suggestion
    - test -d $PREFIX/include/nv            # [linux]
```

recipes/cuda-cccl/build.sh
```diff
@@ -0,0 +1,13 @@
...
+[[ -d lib64 ]] && mv lib64 lib
+
+[[ ${target_platform} == "linux-64" ]] && targetsDir="targets/x86_64-linux"
+[[ ${target_platform} == "linux-ppc64le" ]] && targetsDir="targets/ppc64le-linux"
+[[ ${target_platform} == "linux-aarch64" ]] && targetsDir="targets/sbsa-linux"
+
+mkdir -p ${PREFIX}/${targetsDir}
+
+cp -r lib ${PREFIX}/${targetsDir}
+cp -r include ${PREFIX}/${targetsDir}
```
bdice, line 13:
```suggestion
cp -rv lib ${PREFIX}/${targetsDir}
cp -rv include ${PREFIX}/${targetsDir}
```

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+build:
+  number: 0
+  skip: true  # [osx]
+
+outputs:
+  - name: cuda-cccl
+    requirements:
+      run:
+        - cuda-cccl_{{ target_platform }} =={{ version }}
+        - cuda-cccl-impl ==2.0.1
```
kkraus14, line 30: Would maybe be helpful to have this as a jinja variable that can be easily bumped at the top of the recipe

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+outputs:
+  - name: cuda-cccl
+    requirements:
+      run:
+        - cuda-cccl_{{ target_platform }} =={{ version }}
```
kkraus14, line 29: Where is this package being defined? Why do we need platform specific packages for cccl?

recipes/cuda-cccl-impl/build.sh
adibbley: In this case the `cmake` directories get `mv`'d to the lib directory and the `thrust`/`cub` directories are then copied to the `include` directory below. So if we change this to `cp` we also need to add `rm` to maintain the expected layout.
```suggestion
cp -rv thrust/thrust/cmake ${PREFIX}/lib/cmake/thrust
rm -rv thrust/thrust/cmake ${PREFIX}/lib/cmake/thrust
cp -rv cub/cub/cmake ${PREFIX}/lib/cmake/cub
rm -rv cub/cub/cmake ${PREFIX}/lib/cmake/cub
cp -rv libcudacxx/lib/cmake/libcudacxx ${PREFIX}/lib/cmake
```

recipes/cuda-cccl/meta.yaml
adibbley: It is the second output of this recipe. It is needed for the `cuda-cudart-dev` and `cuda-nvcc` target specific packages.

kkraus14: Why couldn't they just depend on `cuda-cccl` instead of `cuda-cccl_{{ target_platform }}`? I don't fully understand why target specific package names are needed but they all seem to be connected to where there's special jinja templates used. Maybe @jakirkham knows more?

recipes/cuda-cccl-impl/build.sh
bdice: Thanks for explaining, I missed that. This suggestion can be resolved with no changes needed.

adibbley: Updated my suggestion so we can still add `-v` and commit this, thanks!

leofang:
I noticed that all `linux_64_cuda_XXX` CIs are green, but they shouldn't be. Did we miss adding `cuda-version` somewhere?

bdice:
> I noticed that all `linux_64_cuda_XXX` CIs are green, but they shouldn't be. Did we miss adding `cuda-version` somewhere?

None of the recipes appear to depend on `cuda-version` at present. It was my expectation that `cuda-cccl` should depend on `cuda-version` because it is derived from the CTK archive, but `cuda-cccl-impl` should not depend on `cuda-version` because it is derived from the thrust/cub/libcudacxx GitHub repositories.

recipes/cuda-cccl/meta.yaml
bdice: This crossed my mind as well, but I think it's probably good practice to align with how the `https://developer.download.nvidia.com/` distributes CCCL: it's provided as a platform-specific archive.

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - cuda-cccl-impl =={{ impl_version }}
```
bdice, line 31: I'm a little confused why `cuda-cccl` needs to depend on `cuda-cccl-impl` at all, because it appears to be fetching the CCCL from the CUDA Toolkit Archive? Won't that mean that both `cuda-cccl` and `cuda-cccl-impl` distribute files that would clobber each other?

leofang: I think there's no clobbering. `cuda-cccl_{{ target_platform }}` installs to `$PREFIX/targets` and `cuda-cccl-impl` installs to $PREFIX`. 

That said, I am sensing people have different interpretations on the proposed resolution here. Even I am no longer sure whether my above question on `cuda-version` makes sense. Perhaps we should clean up the proposal and write up a short, up-to-date summary...

adibbley:
> I noticed that all `linux_64_cuda_XXX` CIs are green, but they shouldn't be. Did we miss adding `cuda-version` somewhere?

Yes. Added now to `cuda-cccl` and `cuda-cccl_{{ target_platform }}` (not `cuda-cccl-impl`)

recipes/cuda-cccl/meta.yaml
adibbley: I thought the idea was if someone wanted the `cuda-cccl-impl` that shipped with CTK then `cuda-cccl` would be the way to do that.

I put a summary of my understanding of the proposal in [an above comment](https://github.com/conda-forge/staged-recipes/pull/21953#issuecomment-1486960066) if that helps.

recipes/cuda-cccl/meta.yaml
kkraus14: Yes, I'm not questioning it being platform-specific, and we could use platform-specific sources, but you don't need separate package names for that. Most packages that have a binary component are platform-specific, but they do so under a certain name. The ones that use the `_{{ target_platform }}` approach are ones that sit behind special templates like `{{ compiler('cxx') }}`. I don't fully understand the reasoning behind this but maybe someone with more understanding could explain.

bdice: Good point. We could use selectors to achieve the platform-specific source. I think I agree that the `_{{ target_platform }}` package is unnecessary. @adibbley @leofang What are your thoughts?

isuruf: > The ones that use the _{{ target_platform }} approach are ones that sit behind special templates like {{ compiler('cxx') }}

That's exactly the reason to use a platform specific name. To allow the cuda compiler to cross compile.

kkraus14: Okay, so if I understand correctly we then require this package to be platformed as well for cross compilation purposes?

isuruf: Yes

bdice: That sounds right. Thanks for the correction @isuruf.

kkraus14: Great, resolving this.

recipes/cuda-cccl/meta.yaml
bdice: Does `cuda-cccl` need to depend on `cuda-cccl-impl =={{ impl_version }}`? It seems it would prevent users from combining compilers or other packages depending on `cuda-version=12.0` with some later version of `cuda-cccl-impl`. The ability to install newer CCCL versions when building libraries with a fixed CUDA Toolkit version is important for downstream applications.

adibbley: I would expect other CTK packages, namely `cuda-cudart-dev_{{ target_platform }}`, to depend on `cuda-cccl_{{ target_platform }}`, not `cuda-cccl`. So a user would only lock `cuda-cccl-impl` to a CTK version if they install `cuda-cccl` explicitly. I [originally ](https://github.com/conda-forge/staged-recipes/pull/21953#discussion_r1150777350) had this as `>=`, but that would it more difficult to lock `cuda-cccl-impl` to a CTK version if that is intended.

bdice: Fantastic. That's a good solution. I apologize, I think you might have said this before and I forgot that.

bdice (approved):
Package contents and dependencies/constraints.

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,102 @@
...
+      home: https://developer.nvidia.com/cuda-toolkit
+      license_file: LICENSE
+      license: LicenseRef-NVIDIA-End-User-License-Agreement
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: 'C++ Core Compute Libraries Documentation and Scripts'
+      description: |
+        C++ Core Compute Libraries Documentation and Scripts
+      doc_url: https://docs.nvidia.com/cuda/index.html
+
+    extra:
```
bdice, line 54: I don't think the `extra` section is needed for each subpackage. As I understood @leofang, only the `about` section needed to be copied to each subpackage. But this is not a blocker imo, we can fix that on the feedstock later.

adibbley: Yup, thanks for the catch! Fixed now.

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+    test:
+      commands:
+        - test -d $PREFIX/targets/{{ target_name }}/lib      # [linux]
+        - test -d $PREFIX/lib/cmake  # [linux]
+    about:
+      home: https://developer.nvidia.com/cuda-toolkit
+      license_file: LICENSE
+      license: LicenseRef-NVIDIA-End-User-License-Agreement
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: 'C++ Core Compute Libraries Documentation and Scripts'
```
bdice, line 49: CCCL stands for CUDA C++ Core Libraries, not C++ Core Compute Libraries ([source](https://github.com/NVIDIA/cccl)). Also this package doesn't contain documentation or scripts. Please fix on all subpackages/recipes/about sections in this PR.
```suggestion
      summary: 'CUDA C++ Core Libraries'
```

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+      description: |
+        C++ Core Compute Libraries Documentation and Scripts
```
bdice, line 51:
```suggestion
        CUDA C++ Core Libraries
```

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+        - if not exist %LIBRARY_LIB%\{{ target_name }}\cmake exit 1    # [win]
+        - if not exist %LIBRARY_INC%\targets\{{ target_name }} exit 1  # [win]
+    about:
+      home: https://developer.nvidia.com/cuda-toolkit
+      license_file: LICENSE
+      license: LicenseRef-NVIDIA-End-User-License-Agreement
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: 'C++ Core Compute Libraries Documentation and Scripts'
+      description: |
+        C++ Core Compute Libraries Documentation and Scripts
```
bdice, line 79:
```suggestion
      summary: 'CUDA C++ Core Libraries'
      description: |
        CUDA C++ Core Libraries
```

recipes/cuda-cccl/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+      doc_url: https://docs.nvidia.com/cuda/index.html
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: 'C++ Core Compute Libraries Documentation and Scripts'
+  description: |
+    C++ Core Compute Libraries Documentation and Scripts
```
bdice, line 89:
```suggestion
  summary: 'CUDA C++ Core Libraries'
  description: |
    CUDA C++ Core Libraries
```

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file:
+    - cub/LICENSE.TXT
+    - thrust/LICENSE
+    - libcudacxx/LICENSE.TXT
+  license: BSD-3-Clause AND Apache-2.0
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: 'C++ Core Compute Libraries Documentation and Scripts'
+  description: |
+    C++ Core Compute Libraries Documentation and Scripts
```
bdice, line 62:
```suggestion
  summary: 'CUDA C++ Core Libraries'
  description: |
    CUDA C++ Core Libraries
```

recipes/cuda-cccl/meta.yaml
leofang: So should we just add "CUDA"?

bdice: Right -- we need the word CUDA somewhere. I'm fine with "CUDA C++ Core Compute Libraries" if that's what the consensus suggests. I also asked the CCCL team to clarify.

kkraus14 (approved):

leofang (approved):

recipes/cuda-cccl/meta.yaml
bdice: Confirmed offline with CCCL team: @wmaxey says "CUDA C++ Core Libraries", and I'd go with that for this PR.

bdice:
@adibbley From what I can tell, we only need to apply the naming fixes and then this should be good to merge.

@kkraus14 Do you have merge power here, after those naming suggestions are applied? @jakirkham is unavailable for a bit, but it'd be wonderful to move this along so we can move to `cuda-cudart` ASAP. Thanks in advance, we greatly appreciate your help.

recipes/cuda-cccl-impl/meta.yaml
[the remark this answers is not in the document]
kkraus14: @isuruf does this need to be addressed before merging?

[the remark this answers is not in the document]
isuruf: Yes, otherwise there'll be clobbering. You need to send a PR to do a repodata patch as @jakirkham said or add a `run_constrained` as I suggested. Your choice.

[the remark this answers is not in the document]
bdice: I’ll add a `run_constrained` suggestion here.

leofang (approved):

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    sha256: 22308facf62421d6dd2489b9554ffa4908dca13d3921a3e7f2965c5a6984cd20  # [win]
+  - url: https://github.com/NVIDIA/libcudacxx/archive/refs/tags/{{ libcudacxx_version }}.{{ extension }}
+    folder: libcudacxx
+    sha256: 05b1435ad65f5bdef1bb8d1eb29dc8f0f7df34c01d77bf8686687a4603588bce  # [linux]
+    sha256: d7864460ce8def8e0620cb9dd891e94299ade75eac75aa96b34001af3cdb074c  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+
```
bdice, line 34: @isuruf @kkraus14 @adibbley I think this is the constraint we are needing.
```suggestion

requirements:
  run_constrained:
    - cub < 0.0.0
    - thrust < 0.0.0

```

recipes/cuda-cccl-impl/meta.yaml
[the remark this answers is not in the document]
bdice: Suggestion for the constraint is here: https://github.com/conda-forge/staged-recipes/pull/21953/files#r1157595912

recipes/cuda-cccl-impl/meta.yaml
kkraus14: It would probably be cleaner to have a run constraint and repodata patch on the deprecated `thrust` and `cub` packages so that this `cuda-cccl` package doesn't need to have this `cub` and `thrust` run constraint in perpetuity, but whatever is decided upon I'm happy either way.

bdice: Agreed, @kkraus14. I would propose adding the constraint on this recipe for now, to have an immediate solution. I agree that the patch would be better to apply on the deprecated packages. However, I don't have the knowledge (or merge-power) to help with implementing that option.

kkraus14: I think we could avoid needing to modify the Thrust / cub recipes where once we merge this we can archive them.

I think we'd just need a pull request to the repodata patches repository: https://github.com/conda-forge/conda-forge-repodata-patches-feedstock which I imagine would look something like:
```
if record_name in {"thrust", "cub"}:
    new_constrains = record.get('constrains', [])
    new_constrains.append("cuda-cccl < 0.0.0a0")
    new_constrains.append("cuda-cccl-impl < 0.0.0a0")
    new_constrains.append(f"cuda-cccl_{subdir} < 0.0.0a0")
    record['constrains'] = new_constrains
```

I have merge-power and am happy to help review + iterate through merging.

kkraus14 (approved):

kkraus14:
This looks good to go from my perspective. @adibbley @bdice @leofang @jakirkham @isuruf anything else that should be addressed before we merge this?

beckermr (commented):
@jakirkham Asked me to take a look here. My only comment is that it'd be nice to include a comment in each recipe explaining how it relates to the other recipes. This doesn't have to be detailed, but it will help us debug things in the future. Otherwise, I don't see anything obviously wrong with the recipes.

recipes/cuda-cccl-impl/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+{% set cub_version = "2.0.1" %}
+{% set libcudacxx_version = "1.9.0" %}
+{% set platform = "linux-x86_64" %}  # [linux64]
+{% set platform = "linux-ppc64le" %}  # [ppc64le]
+{% set platform = "linux-sbsa" %}  # [aarch64]
+{% set platform = "windows-x86_64" %}  # [win]
+{% set extension = "tar.gz" %}  # [not win]
+{% set extension = "zip" %}  # [win]
+
+package:
```
bdice, line 13: Maybe something like this can address @beckermr's request for documenting the relationships between packages. I would appreciate input from @adibbley @jakirkham on this.
```suggestion
# CUDA C++ Core Libraries (CCCL) includes thrust, cub, and libcudacxx.
# The cuda-cccl-impl package ships CCCL headers in the environment's include directory for use in downstream libraries. It follows CCCL upstream versioning.
# The cuda-cccl_{{ target_platform }} package (e.g. cuda-cccl_linux-64) ships CCCL headers in 'targets/.../include' that can be used by nvcc, the CUDA compiler. It follows CUDA Toolkit versioning.
# The cuda-cccl package is a metapackage that depends on cuda-cccl-impl and cuda-cccl_{{ target_platform }} to ensure that versions are consistent. That is, "I want to use the thrust headers that shipped with CUDA Toolkit 12.1."


package:
```

beckermr: Good start. You should cover why there are separate packages per target.

jakirkham: Yeah the separate packages per target is needed due to cross-compilation (as this is a dependency of the compiler, `cuda-nvcc`)

bdice: Edited to mention cross-compilation in the target_platform description.

adibbley: LGTM, thanks!

beckermr (approved):

bdice:
@adibbley Can you copy this package description to the cuda-cccl meta.yaml? https://github.com/conda-forge/staged-recipes/pull/21953#discussion_r1159052062

leofang:
I would suggest we merge as-is, and add the docs to conda-forge.github.io feedstock.

adibbley:
I already copied it. Can't hurt to have it in the docs as well once this is merged.

leofang:
Even better, thanks!

jakirkham:
Planning on merging EOD tomorrow if no comments

kkraus14 (approved):

jakirkham:
Thanks all! 🙏

Let's follow up on anything else on the feedstocks

PR #21974, opened 2023-02-07
Added festival text-to-speech package with a sample voice

peci1:
Checklist
- [ ] Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md). See https://github.com/conda-forge/staged-recipes/pull/21974#issuecomment-1421359869 .
- [x] Build number is 0.

### DIFF
--- a/recipes/festival/build.sh
+++ b/recipes/festival/build.sh
@@ -28,6 +28,9 @@ for i in $(find src/ -type f); do
       -e 's,\"\.\./\(base_class/.*\)\",\<speech_tools/\1\>,g' $i;
 done
 
+# Use shared libs
+echo "GCC_MAKE_SHARED_LIB = gcc -shared -o XXX" >> config/systems/default.mak
+
 make  # cannot build in parallel, the build system does not support it
 
 make install  # only copies files inside the build directory
--- a/recipes/festival/build.sh
+++ b/recipes/festival/build.sh
@@ -1,0 +1,41 @@
+set -exou
+
+# The festival file system hierarchy is insane; here we correct a few preconfigured folders
+
+sed -i "s#^EST=.*\$#EST=${BUILD_PREFIX}/lib/speech_tools#" config/config.in  # Path to libestools
+sed -i "s#^FTLIBDIR =.*\$#FTLIBDIR = ${PREFIX}/lib#" config/project.mak  # Where the installed libraries AND dictionaries/voices can be found
+sed -i "s#^INSTALL_PREFIX=.*\$#INSTALL_PREFIX=${PREFIX}#" config/systems/default.mak
+sed -i "s#\$(FESTIVAL_HOME)/bin/festival#${PREFIX}/bin/festival#" examples/Makefile
+sed -i "s#/projects/festival/lib/dicts#${PREFIX}/share/festival/dicts#" lib/lexicons.scm  # Default path for dictionaires (can be overriden with --libdir).
+
+./configure
+
+# This make call just generates a file on which we need to apply the following sed commands, too
+make modules
+(cd src/modules; make init_modules.cc)
+
+# The include statements do not follow any filesystem hierarchy. This command straightens that up. Let's pray it is correct!
+# Base for this command was taken from Arch linux: https://github.com/archlinux/svntogit-packages/blob/packages/festival/trunk/PKGBUILD
+for i in $(find src/ -type f); do
+  sed -i -e 's,"EST.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/\",speech_tools/,g' -e 's,<\(EST_walloc\.h\)>,\<speech_tools/\1\>,g' \
+      -e 's,"siod.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/\",speech_tools/,g' \
+      -e 's,"instantiate/.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/instantiate/\",speech_tools/instantiate/,g' -e 's,"instantiate,instantiate,g' \
+      -e 's,"ling_class/.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/ling_class/\",speech_tools/ling_class/,g' -e 's,"ling_class,ling_class,g' \
+      -e 's,"rxp/.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/rxp/\",speech_tools/rxp/,g' -e 's,\"rxp.h\",\<speech_tools/rxp/rxp.h\>,' \
+      -e 's,"sigpr/.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/sigpr/\",speech_tools/sigpr/,g' -e 's,"sigpr,sigpr,g' \
+      -e 's,"unix/.*\.h",\<speech_tools/&\>,g' -e 's,speech_tools/unix/\",speech_tools/unix/,g' -e 's,\.h\">,.h\>,g' -e 's,"unix,unix,g' \
+      -e 's,<speech_tools/\(EST_.*Cost.h\)>,\"\1\",g'  -e 's,<speech_tools/\(EST_.*Cache.h\)>,\"\1\",g'  -e 's,<speech_tools/\(EST_.*Coverage.h\)>,\"\1\",g' \
+      -e 's,\"\.\./\(base_class/.*\)\",\<speech_tools/\1\>,g' $i;
+done
+
+make  # cannot build in parallel, the build system does not support it
+
+make install  # only copies files inside the build directory
+
+# Now also impose hierarchy on the festival-own includes (cannot do that before build, it would fail)
+for i in $(find src/ -type f); do
+  sed -i -e 's,"festival\.h",\<festival/festival.h\>,g' \
+      -e 's,"ModuleDescription\.h",\<festival/ModuleDescription.h\>,g' \
+      -e 's,"Phone\.h",\<festival/Phone.h\>,g' $i
+done
+
--- a/recipes/festival/install.sh
+++ b/recipes/festival/install.sh
@@ -1,0 +1,35 @@
+set -exou
+
+# All files have been built by build.sh, this just installs them into ${PREFIX}
+
+mkdir -p "${PREFIX}/bin"
+mkdir -p "${PREFIX}/lib"
+mkdir -p "${PREFIX}/lib/festival"
+mkdir -p "${PREFIX}/share/festival"
+
+# Prepare the dicts and voices directories. Dependent packages will put their dicts and voices there.
+mkdir -p "${PREFIX}/share/festival/dicts"
+mkdir -p "${PREFIX}/share/festival/voices"
+# Conda would delete the directories if they were left empty, so we use a placeholder file.
+echo "keep" > "${PREFIX}/share/festival/dicts/.keep"
+echo "keep" > "${PREFIX}/share/festival/voices/.keep"
+
+cp src/main/festival "${PREFIX}/bin/"
+cp src/main/festival_client "${PREFIX}/bin/"
+cp bin/text2wave "${PREFIX}/bin/"
+
+# Yes, we're copying lib files into share, as there it makes sense. In a later part of this script, we'll make symlinks from lib.
+cp -a lib/* "${PREFIX}/share/festival/"
+rm -fv $(find "${PREFIX}/share/festival" -name Makefile)
+
+# Crazy...
+ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/lib/festival/audsp"
+ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/bin/audsp"
+
+# text2wave contains a hardcoded path to the festival binary, so fix it
+sed -i "s:$(pwd):${PREFIX}:g" "${PREFIX}/bin/text2wave"
+
+# Crazy... All files from share/festival have to also be accessible from lib/
+for f in "${PREFIX}"/share/festival/*; do
+  ln -s "$f" "${PREFIX}/lib/$(basename "$f")"
+done
--- a/recipes/festival/install.sh
+++ b/recipes/festival/install.sh
@@ -10,26 +10,25 @@ mkdir -p "${PREFIX}/share/festival"
 # Prepare the dicts and voices directories. Dependent packages will put their dicts and voices there.
 mkdir -p "${PREFIX}/share/festival/dicts"
 mkdir -p "${PREFIX}/share/festival/voices"
-# Conda would delete the directories if they were left empty, so we use a placeholder file.
-echo "keep" > "${PREFIX}/share/festival/dicts/.keep"
-echo "keep" > "${PREFIX}/share/festival/voices/.keep"
 
 cp src/main/festival "${PREFIX}/bin/"
 cp src/main/festival_client "${PREFIX}/bin/"
 cp bin/text2wave "${PREFIX}/bin/"
+cp lib/etc/unknown_Linux/audsp "${PREFIX}/bin/audsp"
 
-# Yes, we're copying lib files into share, as there it makes sense. In a later part of this script, we'll make symlinks from lib.
+# The upstream package doesn't have separate libdir and datadir - it puts everything into lib, even data files like
+# voices. Debian has more proper patches for this, however, we work it around by making the files and directories
+# accessible both from lib and share/festival. This allows us to put data files into share/festival without the
+# need for maintaining the more proper patch introducing separate datadir.
+# https://salsa.debian.org/tts-team/festival/-/blob/master/debian/patches/20-debian-filesystem-standard.diff
 cp -a lib/* "${PREFIX}/share/festival/"
 rm -fv $(find "${PREFIX}/share/festival" -name Makefile)
-
-# Crazy...
-ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/lib/festival/audsp"
-ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/bin/audsp"
-
-# text2wave contains a hardcoded path to the festival binary, so fix it
-sed -i "s:$(pwd):${PREFIX}:g" "${PREFIX}/bin/text2wave"
-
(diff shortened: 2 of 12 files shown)


peci1:
This feedstock depends on https://github.com/conda-forge/libestools-feedstock/pull/1 . I will also need second eyes on that PR as I am unsure with a few things.

One thing I don't understand is how transitive dependencies are handled. I build locally this PR and got the 4 packages. All festvox* and festlex* packages depend on festival and this dependency is correctly rendered. However, festvox-kallpc16k also depends on festlex-cmu and festlex-poslex, but these dependencies do not seem to appear in the list of dependencies of festvox-kallpc16k. I'm sorry for posting a screenshot instead of text, but mouse selection doesn't work in the console where I tried it... I created a clean mamba env and requested just the install of festvox-kallpc16k (in hope it would install all 4 packages).

Snímek obrazovky pořízený 2023-02-07 20-46-10

I would expect the two festlex* packages to be also automatically installed. But they're not.

peci1:
This package also ships a static library libFestival.a. The upstream build system is not capable of producing shared libs. Debian has an extensive patch allowing the build of shared libs, but wouldn't want to maintain that patch here. So, either the static lib is accepted, or I can remove the -dev package altogether as there is veery low probability anyone would actually want to link to this library. I mainly provided festival to expose its CLI tools.

peci1:

peci1:
https://github.com/conda-forge/libestools-feedstock/pull/1 is now merged. I adapted this PR with the changes.

I locally tested this PR and the built binaries work.

To test, install festvox-kallpc16k and try:

    echo "test" | text2wave -

It should output a binary stream in WAV format (starting with text `RIFF$WAVEfmt`).

I needed to add dependency on `libestools-static` which provides static libraries from libestools. I think the static libs are not used anywhere, but the super-silly build system requires them to be present (because of the weird make rule `libestools.so : libestools.a`). For the same reason, I also provided package `festival-static` in case anyone needs to build against `libFestival.so`.

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - libestools-static  # [linux]  This is needed to satisfy Make, but not used
+
+  host:
+    - libestools  # [linux]
+  run:
+    - libestools  # [linux]
+
+outputs:
+  - name: festival
+    script: install.sh  # [linux]
```
carterbox, line 34: Just FYI, conda also accepts a list of files with glob pattern matching instead of the name of an install script for separating outputs. It's less commonly used, but sometimes it's easier to implement especially if the artifacts are grouped in folders.

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    - {{ compiler('cxx') }}
+    - make  # [linux]
+    - libgomp  # [linux]
+    - libestools-dev  # [linux]
+    - libestools-static  # [linux]  This is needed to satisfy Make, but not used
+
+  host:
+    - libestools  # [linux]
+  run:
+    - libestools  # [linux]
```
carterbox, line 30: build is for build tools. host is for libraries which will be linked against and need to match the target platform. run is for dependencies that do not need to be around during build but are needed at runtime.

Should be something like this.

```suggestion
requirements:
  build:
    - {{ compiler('c') }}
    - {{ compiler('cxx') }}
    - make  # [linux]
    - libgomp  # [linux]
  host:
    - libestools-dev  # [linux]
    - libestools-static  # [linux]  This is needed to satisfy Make, but not used
```

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - libgomp  # [linux]
+        - libestools-dev  # [linux]
+      host:
+        - libestools  # [linux]
+      run:
+        - libestools  # [linux]
```
carterbox, line 44:
```suggestion
      build:
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        - libgomp  # [linux]
      host:
        - libestools-dev  # [linux]
```

recipes/festlex-cmu/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+
+build:
+  skip: True  # [not linux]
+  number: 0
+
+requirements:
+  host:
+    - festival
+  run:
+    - festival
```
carterbox, line 20: My understanding is that these festlex packages are just data packages and are thus the same on any platform. In that case, they should all be `noarch: generic` packages, and this section should be something like:

```suggestion
build:
  skip: True  # [not linux]
  number: 0
  noarch: generic

requirements:
  run:
  # add version constraints to festival package here
    - festival
```

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    requirements:
+      host:
+        - {{ pin_subpackage("festival", exact=True) }}
+      run:
+        - {{ pin_subpackage("festival", exact=True) }}
+    test:
+      commands:
+        - test -f $PREFIX/include/festival/festival.h  # [linux]
+
+  - name: festival-static
```
carterbox, line 66:
```suggestion
  - name: libfestival-static
```

recipes/festival/install.sh
```diff
@@ -0,0 +1,35 @@
...
+ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/lib/festival/audsp"
+ln -s "${PREFIX}/share/festival/etc/unknown_Linux/audsp" "${PREFIX}/bin/audsp"
+
+# text2wave contains a hardcoded path to the festival binary, so fix it
+sed -i "s:$(pwd):${PREFIX}:g" "${PREFIX}/bin/text2wave"
+
+# Crazy... All files from share/festival have to also be accessible from lib/
+for f in "${PREFIX}"/share/festival/*; do
+  ln -s "$f" "${PREFIX}/lib/$(basename "$f")"
+done
```
carterbox, line 35: For what reason? This may be a conda portability related thing. Are there hardcoded paths somewhere that need to be replaced?

carterbox (changes requested):

recipes/festival/meta.yaml
peci1: Should the -dev package really be in host section? I understand "host" as "build"+"run", but the -dev version is only required during build.

recipes/festival/install.sh
peci1: No, the library build and configuration system is really crazy. I don't have the time to write a proper build system for the library, so this seems like the solution that is quick and not that harmful.

These symlinks are also needed when building locally on Ubuntu.

recipes/festival/meta.yaml
peci1: The docs are not really verbose on how to specify where to put the files. I.e. I need to pick folder `src/include/festival` and put it into `$PREFIX/include/festival`. How do I specify that with the file lists?

recipes/festlex-cmu/meta.yaml
peci1: Done, thanks.

recipes/festival/meta.yaml
peci1: Done.

peci1:
@carterbox I think this PR is now waiting for a few answers from you in the inline comments. Could you please have a look?

carterbox:
If you need additional review, please tag your PR by mentioning the appropriate team @conda-forge/help-c-cpp

peci1:
I wanted to get a response to my questions regarding your review, @carterbox : https://github.com/conda-forge/staged-recipes/pull/21974#discussion_r1293607389 and https://github.com/conda-forge/staged-recipes/pull/21974#discussion_r1293628321 .

recipes/festival/meta.yaml
carterbox: The paths are relative to $PREFIX. So if you want to include `$PREFIX/lib/libfestival.so` in the output, you would list `lib/libfestival*` or `lib/libfestival.so  # [linux]`

recipes/festival/meta.yaml
carterbox: Host is not build "build + run". It is the list of packages which must match the target platform when cross-compiling. I'm not sure why it is called 'host' instead of 'target'.

https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#host

> It represents packages that need to be specific to the target platform when the target platform is not necessarily the same as the native build platform.

recipes/festival/install.sh
carterbox: These links are needed by a downstream package? Not this package?

carterbox:
Reviewers can read through every notification of activity that they get from every PR that they have ever interacted with. That's why we have the labelling system, so that we know that you actually need review and aren't just adding a few commits or writing notes in a comment.

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    printed manual, info files and HTML.
+  license: BSD-4-Clause
+  license_family: BSD
+  license_file: COPYING
+  doc_url: https://www.cstr.ed.ac.uk/projects/festival
+  dev_url: https://github.com/festvox/festival
+
+extra:
+  recipe-maintainers:
+    - peci1
```
carterbox, line 103:
```suggestion
extra:
  feedstock-name: festival
  recipe-maintainers:
    - peci1
```
Otherwise, the feedstock will be named festival-split

recipes/festival/meta.yaml
carterbox, line 30: Why are all these dependencies selected for linux if you are skipping `not linux`?

recipes/festival/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+  name: {{ name|lower }}-split
+  version: {{ version }}
+
+source:
+  url: https://github.com/festvox/festival/archive/5db6ee949f4e4e9da25c87bd03b90e69f9011393.tar.gz
+  sha256: 9f0d0e4f0e91fca4430ee4a6ca4d3d24bac444f36b4f4e9653c6119aadfae13e
+
+build:
+  run_exports:
+    - {{ pin_subpackage('festival', max_pin='x.x') }}
```
carterbox, line 14: Because you have multiple outputs, I believe that you need to move this run_export into both of the output sections. Or at least add it to the libfestival-dev package. Please check the detemplated recipes in your built packages metadata.

recipes/festival/build.sh
```diff
@@ -0,0 +1,44 @@
...
+sed -i "s#\$(FESTIVAL_HOME)/bin/festival#${PREFIX}/bin/festival#" examples/Makefile
+sed -i "s#/projects/festival/lib/dicts#${PREFIX}/share/festival/dicts#" lib/lexicons.scm  # Default path for dictionaires (can be overriden with --libdir).
+
+./configure
+
+# This make call just generates a file on which we need to apply the following sed commands, too
+make modules
+(cd src/modules; make init_modules.cc)
+
+# The include statements do not follow any filesystem hierarchy. This command straightens that up. Let's pray it is correct!
```
carterbox, line 17: You can check your built package to see if it is correct. Download the build artifacts on the Azure Pipelines page.

recipes/festival/install.sh
```diff
@@ -0,0 +1,35 @@
...
+mkdir -p "${PREFIX}/lib"
+mkdir -p "${PREFIX}/lib/festival"
+mkdir -p "${PREFIX}/share/festival"
+
+# Prepare the dicts and voices directories. Dependent packages will put their dicts and voices there.
+mkdir -p "${PREFIX}/share/festival/dicts"
+mkdir -p "${PREFIX}/share/festival/voices"
+# Conda would delete the directories if they were left empty, so we use a placeholder file.
+echo "keep" > "${PREFIX}/share/festival/dicts/.keep"
+echo "keep" > "${PREFIX}/share/festival/voices/.keep"
```
carterbox, line 15: Does the package require these directories to exist even if they are empty?

recipes/festlex-cmu/meta.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+build:
+  number: 0
+  noarch: generic
+
+test:
+  commands:
+    - test -f $PREFIX/share/festival/dicts/cmu/cmulex.scm  # [linux]
+    - test -f $PREFIX/share/festival/dicts/cmu/cmudict-0.4.out  # [linux]
+    - test -f $PREFIX/share/festival/dicts/cmu/cmu_lts_rules.scm  # [linux]
```
carterbox, line 20:
```suggestion
build:
  number: 0
  noarch: generic

requirements:
  run:
    - __unix  # [unix]

test:
  commands:
    - test -f $PREFIX/share/festival/dicts/cmu/cmulex.scm  # [unix]
    - test -f $PREFIX/share/festival/dicts/cmu/cmudict-0.4.out  # [unix]
    - test -f $PREFIX/share/festival/dicts/cmu/cmu_lts_rules.scm  # [unix]
```
You probably want to structure your noarch packages like this because the UNIX prefix and Windows prefix are different. This way, you will be able to add support for MacOS without rebuilding these data packages.



https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-packages-with-os-specific-dependencies

recipes/festvox-kallpc16k/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: http://festvox.org/packed/festival/2.5/voices/festvox_kallpc16k.tar.gz
+  sha256: 809c4ab5ed9e4df4a658b58d5c56fe35055723f00d81a238168f5a1ebdaed08c
+
+build:
+  skip: True  # [not linux]
+  number: 0
```
carterbox, line 14: Why is festvox-kallpc16k platform specifc instead of generic?

carterbox (changes requested):

peci1:

recipes/festival/meta.yaml
peci1: @carterbox I don't understand your request here. The run_exports are both in the global scope and in each individual package. Where else should I add it?

recipes/festlex-cmu/meta.yaml
peci1: @carterbox How did you tell the unix and windows prefixes are different? The path to the data files is configured in the conda-forge build files. In the unlikely event that somebody adds windows support to festival feedstock, he could easily make sure these paths are the same on Windows.

recipes/festvox-kallpc16k/meta.yaml
peci1: @carterbox I tried making it generic, but in that case it depends on festival on all platforms, however only linux64 has the package. So what is the correct way of specifying this? It doesn't make sense to me to add the festival dependency only for linux and make the other platforms without any.

recipes/festival/meta.yaml
peci1: The upstream package does provide a `make install` target. However, it's not very useful. It doesn't follow any standard FHS rules. Instead of patching the installer, I decided for the bunch of custom scripts which do the same work, but aren't compatible with the YAML install file lists.

recipes/festival/build.sh
peci1: The result looks okay. These seds were taken from Arch linux: https://gitlab.archlinux.org/archlinux/packaging/packages/festival/-/blob/main/PKGBUILD?ref_type=heads#L127 .

recipes/festival/meta.yaml
peci1: This was probably some preparation for easier adding of other OS builds. But as that is improbably, I'm dropping these selectors.

recipes/festival/install.sh
peci1: Yes, further in this script, I need to create the symlinks from share/festival to lib, and they are essential for the packages with voices. If these directories would be deleted, the symlinks would not work.

recipes/festival/install.sh
peci1: Not exactly downstream packages - it is where this package searches e.g. for installed voices. This package doesn't have a libdir and datadir - instead, it puts everything into lib. Ubuntu has a more proper patch that adds datadir, and then these symlinks are not needed. My approach is to create the datadir (share/festival) and symlink it to lib, so that all code can still refer to the files via libdir, while the files (like voices) can be more properly put into share/festival, where it makes more sense.

I know this is not a proper solution, but it works and doesn't hurt anything. And it doesn't require to maintain a difficult patch.

peci1:

peci1:
@conda-forge/help-c-cpp This PR is ready for another round of review. I answered most questions from the previous review, added a few explanations and also a few more questions.

recipes/festvox-kallpc16k/meta.yaml
peci1: The reason why I made this package depend on the festival binary is because this kind of dependency is present in both Ubuntu and Arch Linux.

recipes/festival/meta.yaml
carterbox: Yep. Looks good. Maybe I was looking at some old commits.

recipes/festlex-cmu/meta.yaml
carterbox: Windows packages are installed to `%LIBRARY_PREFIX%` i.e. `%PREFIX%\Library`. I don't know why. It just is that way.

https://docs.conda.io/projects/conda-build/en/stable/user-guide/environment-variables.html#environment-variables-set-during-the-build-process
https://docs.conda.io/projects/conda-build/en/stable/resources/use-shared-libraries.html#shared-libraries-in-windows

recipes/festvox-kallpc16k/meta.yaml
carterbox: > however only linux64 has the package.

That's OK. Noarch is a description of the contents of the package, not of its dependencies.

carterbox: This package will just be uninstallable by win/osx until the festival package is build for those platforms. That's fine. If it's not fine, then this package shouldn't depend on festival. i.e. You can remove those depenencies if someday someone wants to use these files with another program.

carterbox:
These failed builds due to missing dependencies are fine/expected. The noarch: generic packages should be built only on linux runners by default, so you don't see them in your feedstock.

PR #21976, opened 2023-02-07
Add flite text-to-speech recipe

peci1:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/flite/build.sh
+++ b/recipes/flite/build.sh
@@ -1,0 +1,7 @@
+set -exou
+
+./configure --with-vox=cmu_us_kal16 --enable-shared --prefix="$(pwd)/build"
+
+make
+
+make install
--- a/recipes/flite/install.sh
+++ b/recipes/flite/install.sh
@@ -5,5 +5,6 @@ mkdir -p "${PREFIX}/lib"
 if [[ "$target_platform" == linux* ]]; then
   cp -v -P build/lib/lib*.so.* "${PREFIX}/lib/"
 elif [[ "$target_platform" == osx* ]]; then
+  ls -la build/lib/
   cp -v -P build/lib/lib*.*.dylib "${PREFIX}/lib/"
 fi
--- a/recipes/flite/install.sh
+++ b/recipes/flite/install.sh
@@ -1,4 +1,9 @@
 set -exou
 
 mkdir -p "${PREFIX}/lib"
-cp -P build/lib/lib*.so.* "${PREFIX}/lib/"
+
+if [[ "$target_platform" == linux* ]]; then
+  cp -v -P build/lib/lib*.so.* "${PREFIX}/lib/"
+elif [[ "$target_platform" == osx* ]]; then
+  cp -v -P build/lib/lib*.*.dylib "${PREFIX}/lib/"
+fi
--- a/recipes/flite/install.sh
+++ b/recipes/flite/install.sh
@@ -2,9 +2,4 @@ set -exou
 
 mkdir -p "${PREFIX}/lib"
 
-if [[ "$target_platform" == linux* ]]; then
-  cp -v -P build/lib/lib*.so.* "${PREFIX}/lib/"
-elif [[ "$target_platform" == osx* ]]; then
-  ls -la build/lib/
-  cp -v -P build/lib/lib*.*.dylib "${PREFIX}/lib/"
-fi
+cp -P build/lib/lib*.so.* "${PREFIX}/lib/"
--- a/recipes/flite/install.sh
+++ b/recipes/flite/install.sh
@@ -1,0 +1,4 @@
+set -exou
+
+mkdir -p "${PREFIX}/lib"
+cp -P build/lib/lib*.so.* "${PREFIX}/lib/"
--- a/recipes/flite/install_devel.sh
+++ b/recipes/flite/install_devel.sh
@@ -1,0 +1,7 @@
+set -exou
+
+mkdir -p "${PREFIX}/include/"
+mkdir -p "${PREFIX}/lib"
+
+cp -a build/include/flite "${PREFIX}/include/"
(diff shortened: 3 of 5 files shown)


peci1:

ocefpaf:
Let's close and open this one to toggle the CIs. The logs are long gone here.

recipes/flite/build.sh
```diff
@@ -0,0 +1,7 @@
+set -exou
+
+./configure --with-vox=cmu_us_kal16 --enable-shared --prefix="$(pwd)/build"
+
+make
+
```
ocefpaf, line 6: Is there a `make check` or `ctest`?

recipes/flite/install.sh
```diff
@@ -0,0 +1,5 @@
+set -exou
+
+mkdir -p "${PREFIX}/lib"
+
+cp -P build/lib/lib*.so.* "${PREFIX}/lib/"
```
ocefpaf, line 5: Should we use `SHLIB_EXT` instead of `.so` to catch the macOS `.dylib` too?

recipes/flite/install_devel.sh
```diff
@@ -0,0 +1,7 @@
+set -exou
+
+mkdir -p "${PREFIX}/include/"
+mkdir -p "${PREFIX}/lib"
+
+cp -a build/include/flite "${PREFIX}/include/"
+cp -P build/lib/lib*.so "${PREFIX}/lib/"
```
ocefpaf, line 7: Same as above.

recipes/flite/meta.yaml
```diff
@@ -0,0 +1,86 @@
...
+      run_exports:  # [linux]
+        - {{ pin_subpackage('libflite', max_pin='x.x') }}  # [linux]
+    requirements:  # [linux]
+      build:  # [linux]
+        - {{ compiler('c') }}  # [linux]
+        - make  # [linux]
+    test:  # [linux]
+      commands:  # [linux]
+        - test -f $PREFIX/lib/libflite.so.2.3  # [linux]
+        - test -f $PREFIX/lib/libflite_usenglish.so.2.3  # [linux]
```
ocefpaf, line 34: If you use `SHLIB_EXT` would these pass macOS?

recipes/flite/meta.yaml
```diff
@@ -0,0 +1,86 @@
...
+    requirements:  # [linux]
+      host:  # [linux]
+        - {{ pin_subpackage("libflite", exact=True) }}  # [linux]
+      run:  # [linux]
+        - {{ pin_subpackage("libflite", exact=True) }}  # [linux]
+    test:  # [linux]
+      commands:  # [linux]
+        - test -f $PREFIX/include/flite/flite.h  # [linux]
+        - test -f $PREFIX/lib/libflite.so  # [linux]
+        - test -f $PREFIX/lib/libflite_usenglish.so  # [linux]
```
ocefpaf, line 49: Same as above.

recipes/flite/build.sh
peci1: No, there isn't. There is an unmerged PR from gentoo: https://github.com/festvox/flite/pull/90 . I don't feel like maintaining an unmerged PR.

But the test suite is actually there, so it can be run manually. But it can't be run with a simple `make something`. It has to make the test and the run it from CLI. But I don't think this is worth it. The test already present in this recipe tests that the main binary is working. I think that's enough, isn't it?

ocefpaf: > The test already present in this recipe tests that the main binary is working. I think that's enough, isn't it?

The more tests the better but this is probably good enough to merge and get the first package out. Just add an issue in the feedstock to try adding more tests in the future.

ocefpaf:
macOS is failing with:

```
cp -P 'build/lib/lib*.dylib.*' /Users/runner/mambaforge/conda-bld/flite-tts_1689110067507/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/
cp: build/lib/lib*.dylib.*: No such file or directory
```

You can try a local build to see the right name of this file but I'm guessing that maybe a change from ` build/lib/lib*.dylib.*` to  `build/lib/*.dylib*` may be enough.

recipes/flite/install.sh
```diff
@@ -0,0 +1,5 @@
+set -exou
+
+mkdir -p "${PREFIX}/lib"
+
+cp -P build/lib/lib*${SHLIB_EXT}.* "${PREFIX}/lib/"
```
ocefpaf, line 5: Maybe this fixes the macOS build.

```suggestion
cp -P build/lib/*${SHLIB_EXT}* "${PREFIX}/lib/"
```

peci1:
This is the list of all files on OS X build:

```
+ ls build/lib/
libflite.a
libflite_cmu_grapheme_lang.a
libflite_cmu_grapheme_lex.a
libflite_cmu_indic_lang.a
libflite_cmu_indic_lex.a
libflite_cmu_time_awb.a
libflite_cmu_us_awb.a
libflite_cmu_us_kal.a
libflite_cmu_us_kal16.a
libflite_cmu_us_rms.a
libflite_cmu_us_slt.a
libflite_cmulex.a
libflite_usenglish.a
```

The package just doesn't build shared libs on OS X. That's what the referenced PR would solve. I can certainly turn it into an issue in the feedstock once it is created.

carterbox (commented):
The dev packages are skipped for osx because upstream does not support building non-static libs for osx yet. However, the flite binary is still built for osx.

carterbox:
This package is strange because `libflite.so.1` links to `libflite.so.2.3`. As opposed to `libflite.so.2` linking to `libflite.so.2.3`.

PR #22006, opened 2023-02-10
Adding kt-legacy for `keras-tuner`

Anselmoo:
Checklist
- [x] Build number is 0.


recipes/kt-legacy/meta.yaml
```diff
@@ -22,8 +22,6 @@ requirements:
     - python >=3.7
 
 test:
-  imports:
-    - kt_legacy
```
Anselmoo: Remove import because it will require `keras-tuner` and this is probably a cyclic dependency. 

See also: https://github.com/haifeng-jin/kt-legacy

dopplershift: Unless I'm missing something, shouldn't that import instead be `kerastuner`? That's the only package name/directory I see in the upstream repo.

Anselmoo: Exactly, it is for importing `keras_tuner` in the old fashion.

I used `greyskull`, and the local build ran fine. So, I am wondering, what happens, if I test `import kerastuner`? Should I?

dopplershift: If at all possible, we want an import test so we can validate that created conda package isn't broken. So try adding:
```yaml
    imports:
      - kerastuner
```
and see what happens.

Anselmoo: It is not possible because of

```shell
from keras_tuner import *
ModuleNotFoundError: No module named 'keras_tuner'
import: 'kerastuner'
```

Or is there a possibility for a test import?

dopplershift (approved):
Ignoring noarch Windows build failure due to missing dependency.

This looks good to me. Are you done with this @Anselmoo?

Anselmoo:
> Ignoring noarch Windows build failure due to missing dependency.
> 
> This looks good to me. Are you done with this @Anselmoo?

I am good

PR #22027, opened 2023-02-14
add libefp

loriab:
libefp is a computational chemistry library. This packages the C library itself and pybind11 bindings to it; these projects live in separate repositories but can be built in one cmake command. Please let me know of any questions. 

This is the c-f-ification and extension to windows of longstanding packages (i.e., https://anaconda.org/psi4/libefp and https://anaconda.org/psi4/pylibefp). I expect the greatest concerns will be (1) that this is pulling from two repos for the core C and python bindings and (2) the core C lib is a couple releases behind the upstream repo. The upstream maintainer (also listed as maintainer here) and I have a strategy to resolve those concerns. But in the meantime, the current setup has been stable for several years and the current version is what any downstream already seeking the libefp project on conda (namely psi4) is needing.

Checklist
- [x] Build number is 0.


loriab:

carterbox (changes requested):
Great job with this recipe. It seems to have all the correct bits (tests, run_exports, etc).

> I expect the greatest concerns will be (1) that this is pulling from two repos for the core C and python bindings

Yes. Because these are two projects versioned separately, they should have separate recipes. Please split this recipe into two recipes instead of building once and spliting the outputs. Keep this single PR, just add an additional folder for `pylibefp`. The bot will be able to figure out that one needs to be built before the other. In they end they will be provisioned separate feedstocks.

> (2) the core C lib is a couple releases behind the upstream repo.

This is fine. The first release on conda-forge can be an old release.

> The upstream maintainer (also listed as maintainer here) and I have a strategy to resolve those concerns.

The packages should be separated now because if we were to try to separately them later it requires manual intervention to overcome our package hijacking checks.

loriab:
Thank you for the review, @carterbox. On the single/multiple recipe issue, perhaps I should have mentioned that the next step I've discussed with the upstream maintainer is to move the python bindings code into the upstream C lib codebase, retire the pylibefp repository, and adopt identical versioning for both subpackages. So I think the single recipe / multiple outputs strategy is more sensible longer term than it appears now. I'd prefer to retain the current single feedstock setup, but if you definitely prefer a split, please let me know.

loriab:
Thanks for the merge!

PR #22029, opened 2023-02-14
Add recipe for goss

finsberg:
Checklist
- [x] Build number is 0.


finsberg:

recipes/goss/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "goss" %}
```
carterbox, line 1: Why use a different name on PYPI and conda?

recipes/goss/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - {{ compiler('cxx') }}
+  host:
+    - pip
+    - python
+    - pybind11
+    - ninja
+    - scikit-build
+    - cmake
+    - libgomp    # [linux]
+    - llvm-openmp     # [osx]
```
carterbox, line 28: Please read our documentation regarding openmp. OpenMP dependencies should only appear in the requirements/build section.

recipes/goss/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+source:
+  url: "https://pypi.io/packages/source/p/pygoss/pygoss-{{ version }}.tar.gz"
+  sha256: bb63b5f135b11f3366b92c3bc24c735f57c503ff6d04bb7de25ebdf053bd5d55
+
+build:
+  number: 1
+  script: "{{ PYTHON }} -m pip install . -vv --no-deps"
+  skip: true  # [win]
+
```
carterbox, line 16:
```suggestion
  skip:  # [py<37]
```

carterbox (changes requested):
It seems like you intend for downstream users to link directly to the libgoss C-API because your package contains a dynamically linked libgoss with CMake config and header files.

While pip is a python-only package manager, and conda is a general package manager, so please install libgoss into the standard location for shared-objects and headers.

You will do this by creating separate recipes for "pygoss" and "libgoss". Both recipes can be merged in this same PR, but they should have their own folders. Ultimately they will be provisioned their own feedstocks.

"pygoss" should depend on "libgoss" as it would any other cxx library. The solver should recognize that libgoss needs to be built first. Multiple "pygoss" builds will share one "libgoss" build. Please re-mention the team if you have questions.

recipes/goss/meta.yaml
finsberg: Thanks for pointing that out. This should now be fixed in https://github.com/conda-forge/staged-recipes/pull/22029/commits/0963ef5f9636fa5befd203cfee4715d3ff8618d8

recipes/goss/meta.yaml
carterbox: Consistency is important because requirements.txt so that requirements.txt files can be used to create a conda environment without updating the package names.

finsberg: OK, I have now changed the same so that it matches the one on pypi, see https://github.com/conda-forge/staged-recipes/pull/22029/commits/40707725c680aba8c9ee75f7f211d85520a47161

finsberg:
> It seems like you intend for downstream users to link directly to the libgoss C-API because your package contains a dynamically linked libgoss with CMake config and header files.
> 
> While pip is a python-only package manager, and conda is a general package manager, so please install libgoss into the standard location for shared-objects and headers.
> 
> You will do this by creating separate recipes for "pygoss" and "libgoss". Both recipes can be merged in this same PR, but they should have their own folders. Ultimately they will be provisioned their own feedstocks.
> 
> "pygoss" should depend on "libgoss" as it would any other cxx library. The solver should recognize that libgoss needs to be built first. Multiple "pygoss" builds will share one "libgoss" build. Please re-mention the team if you have questions.
> 
> NOTE: It is possible to have one recipe/feedstock that produces multiple outputs, but IMO, it's easier to reason about separate feedstocks. It's a trade off between bevity and complexity.

Thanks for the explanation. I will try to make two separate packages and ping you when it is done.

finsberg:
@carterbox I think this should be OK now.

recipes/goss/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+      test:
+        imports:
+          - goss
+
+        requires:
+          - pip
+
+        commands:
+          - pip check
```
carterbox, line 75:
```suggestion
          - pip check
          - goss --help
```
Test that your CLI starts correctly.

carterbox (changes requested):

carterbox:
I believe that the OSX build is failing due to a problem with conda-build version 3.24, hopefully that's fixed soon?

PR #22053, opened 2023-02-16
Add cuda-version recipe

adibbley:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/cuda-version/meta.yaml
+++ b/recipes/cuda-version/meta.yaml
@@ -1,0 +1,29 @@
+{% set name = "cuda-version" %}
+{% set version = "12.0.0" %}
+{% set major_version = version.split('.')[0]|int %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  run_constrained:
+    - __cuda >={{ major_version }}
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  #license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: A meta-package for pinning to a CUDA release version
+  description: |
+    A meta-package for pinning to a CUDA release version
+  doc_url: https://docs.nvidia.com/cuda/index.html
+
+extra:
+  recipe-maintainers:
+    - adibbley
--- a/recipes/cuda-version/meta.yaml
+++ b/recipes/cuda-version/meta.yaml
@@ -14,6 +14,10 @@ requirements:
   run_constrained:
     - __cuda >={{ major_version }}
 
+test:
+  commands:
+    - exit 0
+
 about:
   home: https://developer.nvidia.com/cuda-toolkit
   #license_file: LICENSE
--- a/recipes/cuda-version/meta.yaml
+++ b/recipes/cuda-version/meta.yaml
@@ -8,6 +8,7 @@ package:
 
 source:
   url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_documentation/LICENSE.txt
+  sha256: 5db25d4fd138013b563f9a3d1d87f7de7df1dac014fd4cccdfbb3435a5cff761
 
 build:
   number: 0
--- a/recipes/cuda-version/meta.yaml
+++ b/recipes/cuda-version/meta.yaml
@@ -6,6 +6,9 @@ package:
   name: {{ name|lower }}
   version: {{ version }}
 
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_documentation/LICENSE.txt
+
 build:
   number: 0
   noarch: generic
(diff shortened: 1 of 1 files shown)


recipes/cuda-version/meta.yaml
```diff
@@ -0,0 +1,29 @@
...
+{% set version = "12.0.0" %}
+{% set major_version = version.split('.')[0]|int %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+  skip: true  # [osx]
```
jakirkham, line 11: Since this is package is architecture independent (nothing different between Windows and Linux), we can just make it `noarch`

```suggestion
  noarch: generic
```

recipes/cuda-version/meta.yaml
```diff
@@ -0,0 +1,33 @@
+{% set name = "cuda-version" %}
+{% set version = "12.0.0" %}
+{% set major_version = version.split('.')[0]|int %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
```
jakirkham, line 8: Maybe we can include the license file from here? Though open to other ideas

```suggestion

source:
  - url: https://gitlab.com/nvidia/headers/cuda/-/raw/master/LICENSE
```

recipes/cuda-version/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  run_constrained:
+    - __cuda >={{ major_version }}
+
+test:
+  commands:
+    - exit 0
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  #license_file: LICENSE
```
jakirkham, line 23: Since it would be downloaded in `source`, it should be possible to include it here

```suggestion
  license_file: LICENSE
```

recipes/cuda-version/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  run_constrained:
+    - __cuda >={{ major_version }}
```
jakirkham, line 15: Curious whether we want the major version constraint here or if we want this more specified

jakirkham (commented):
Thanks Alex! 🙏

Had a couple comments below. Namely about how we might package the license file, making this package `noarch` (since it is relatively simple), and how we want to constrain the `__cuda` version

jakirkham:
Just as a note, the CUDA CI job failures are expected. As those jobs don't have CUDA 12 support, the install fails there

For context here's a snippet from [this CI job]( https://github.com/conda-forge/staged-recipes/pull/22053/checks?check_run_id=11404733203 ):

```
Mamba failed to solve:
 - cuda-version 12.0.0.* ha770c72_0

with channels:
 - local
 - conda-forge
 - conda-forge

The reported errors are:
-   - package cuda-version-12.0.0-ha770c72_0 has constraint __cuda >=12 conflicting with __cuda-11.2-0
- 
failed to get install actions, retrying.  exception was: Unsatisfiable dependencies for platform linux-64: {MatchSpec("__cuda==11.2=0"), MatchSpec("cuda-version==12.0.0=ha770c72_0")}
```

That said, we don't need CUDA support to build this package

jakirkham:
This is a metapackage for allowing users to configure packages installed by CUDA version that was used during the build. IOW this would replace how `cudatoolkit` does this now. It could be added via `run_exports` by `cuda-nvcc`. We could also hot-fix existing packages to also be constrained by this package for a more consistent user experience. Please let us know your thoughts on this approach

cc @conda-forge/core @conda-forge/staged-recipes

recipes/cuda-version/meta.yaml
adibbley: This would match what `cudatoolkit` was doing for the versions we would support here:
https://github.com/conda-forge/cudatoolkit-feedstock/blob/main/recipe/meta.yaml#L576-L580
But open to changing as necessary

jakirkham: Yeah starting out with matching `cudatoolkit` seems reasonable

Let's see if others have thoughts here

kkraus14: This seems reasonable to me. If someone is using newer driver features or something of the like they can always add an explicit constraint directly on `__cuda` if needed.

recipes/cuda-version/meta.yaml
```diff
@@ -0,0 +1,33 @@
...
+  number: 0
+  noarch: generic
+
+requirements:
+  run_constrained:
+    - __cuda >={{ major_version }}
+
+test:
+  commands:
+    - exit 0
```
kkraus14, line 19: Is this test just to ensure the package installs as expected?

jakirkham: It's just there to keep the linter happy

recipes/cuda-version/meta.yaml
jakirkham: Done below

recipes/cuda-version/meta.yaml
jakirkham: Done below

jakirkham:
Thanks Alex! Also thanks Keith for the review! 🙏

Let's get this one out. If we find something else needed here, please raise an issue on the feedstock and we can follow up there 🙂

PR #22085, opened 2023-02-20
add ogdf

hadim:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/ogdf/build.sh
+++ b/recipes/ogdf/build.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+set -ex
+
+mkdir build/
+cd build/
+
+if [ "$(uname)" == "Darwin" ]; then
+    EXTRA_CMAKE_ARGS="${EXTRA_CMAKE_ARGS} -DCMAKE_OSX_SYSROOT=${CONDA_BUILD_SYSROOT}"
+fi
+
+cmake .. \
+    ${CMAKE_ARGS} \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DBUILD_SHARED_LIBS=ON \
+    -DWITH_EXTERNAL_COIN=True \
+    -DWITH_EXTERNAL_BACKWARD=True \
+    -DWITH_EXTERNAL_PUGIXML=True \
+    ${EXTRA_CMAKE_ARGS}
+
+make -j ${CPU_COUNT}
+make install
--- a/recipes/ogdf/meta.yaml
+++ b/recipes/ogdf/meta.yaml
@@ -1,0 +1,67 @@
+{% set name = "ogdf" %}
+{% set version = "2022.02" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ogdf/ogdf/archive/refs/tags/dogwood-202202.tar.gz
+  sha256: 308cc2749c6a63520f7979bac86a04066dfea2fb9d3a5e89831318db404bfbf5
+
+  patches:
+    - cmake.patch
+
+build:
+  number: 0
+  skip: True  # [win]
+  run_exports:
+    - {{ pin_subpackage('ogdf', max_pin='2022.02') }}
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - backward-cpp                           # [build_platform != target_platform]
+    - pugixml                                # [build_platform != target_platform]
+    - coincbc                                # [build_platform != target_platform]
+    - libdwarf-dev                           # [build_platform != target_platform]
+  host:
+    - cmake >=3.1
+    - make               # [not win]
+    - ninja              # [win]
+    - backward-cpp
+    - pugixml
+    - coincbc
+    - libdwarf-dev
+  run:
+    # To prevent DSO linking errors
+    - coin-or-clp
+    - coin-or-utils
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libOGDF.so                 # [linux]
+    - test -f $PREFIX/lib/libOGDF.dylib              # [osx]
+    - if not exist %LIBRARY_LIB%\\OGDF.lib exit 1    # [win]
+
+about:
+  home: http://www.ogdf.net/
+  license: GPL-2.0-only OR GPL-3.0-only
+  summary: OGDF is a self-contained C++ class library for the automatic layout of diagrams
+
+  description: |
+    OGDF is a self-contained C++ class library for the automatic layout of diagrams.
+  doc_url: http://www.ogdf.net/doku.php
+  dev_url: https://github.com/ogdf/ogdf
+
+  license_file:
+    - LICENSE.txt
+    - LICENSE_GPL_v2.txt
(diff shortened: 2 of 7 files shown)


hadim:
Context and history about that package at https://github.com/conda-forge/staged-recipes/pull/12498#issuecomment-1002208390

hadim:
@conda-forge/help-c-cpp, please review. The OSX error seems to be due to lower/uppercase mismatch in CMake (everything works fine on Linux). I tried a few things without success. If nobody has an idea, I'll simply disable osx before merging.

recipes/ogdf/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - LICENSE.txt
+    - LICENSE_GPL_v2.txt
+    - LICENSE_GPL_v3.txt
+    - LICENSE_abacus.txt
+    - LICENSE_minisat.txt
+
+extra:
+  recipe-maintainers:
+    - hadim
+    - N-Coder
```
hadim, line 67: @N-Coder: are you interested in being a maintainer?

recipes/ogdf/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/ogdf/ogdf/archive/refs/tags/dogwood-202202.tar.gz
+  sha256: 308cc2749c6a63520f7979bac86a04066dfea2fb9d3a5e89831318db404bfbf5
+
+  patches:
+    - cmake.patch
```
N-Coder, line 13: Maybe it would make the patch file easier to maintain (GitHub doesn't even want to display it online due to its size) if you only put actual changes to files' contents into the diff and then delete the respective folders in the script?

N-Coder:
Hey, here are some notes from my side...
- OGDF grew a [PyPi package](https://github.com/N-Coder/ogdf-wheel) in the meantime (which is also about to become part of the official distribution), so it would probably be a good idea to keep the conda version tightly in sync with that, otherwise I see people getting problems because they installed the inofficial conda version that is somehow different than the official PyPi version.
- The PyPi version has a few difference in how OGDF is built, see [here](https://github.com/N-Coder/ogdf-wheel/blob/3a34892cd7c411ae507388ab714c1e13ab06fa47/hatch_build.py#L120) for the cmake flags and the fix for generating portable builds.
- Furthermore, the PyPi version will probably switch to a debug build (possibly also including things like asan) soon because that makes building things with its python interface a lot easier. This probably conflicts with the efficiency oriented use-case of tmap (in which I still would recommend static linkage with the OGDF, but that seems to be a problem with conda -- have you thought about publishing a PyPi package to a wider audience instead of being restricted to the conda-forge ecosystem? On PyPi, you could simply distribute a statically linked version of the OGDF within tmap and everything would be fine and even faster), so we'd probably need to find a way how to support both usages next to each other.
- Did you try running this build of the OGDF? IIRC the packaged version of COIN was somewhat old because there were unannounced breaking changes at  runtime in a COIN release along the way, which led to the OGDF vendoring a working version. So it might be that the COIN part is actually not working with this build (same possibly with pugixml). Maybe running these [smoke tests](https://github.com/N-Coder/ogdf-wheel/blob/main/test_layouts.py) on some graph loaded from an (xml-based) graphml file would provide some first insights into this (although also no guarantee that everything actually works).
- I'm not super happy with having a redundant coda-forge version of the OGDF wheel as that is also available on PyPi to a superset of conda-forge's userbase. So you can keep me in the loop if you insist on pushing this to conda-forge, but I don't see me or other OGDF devs investing further resources into conda-forge that could be put to better use by working on the PyPi release or the library itself.

recipes/ogdf/meta.yaml
carterbox: I agree. Please only include the changes to the CMake files in this diff, and do the file deletions in the build script. Otherwise, the patch seems to be 99% file deletion.

recipes/ogdf/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  run:
+    # To prevent DSO linking errors
+    - coin-or-clp
+    - coin-or-utils
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libOGDF.so                 # [linux]
+    - test -f $PREFIX/lib/libOGDF.dylib              # [osx]
+    - if not exist %LIBRARY_LIB%\\OGDF.lib exit 1    # [win]
```
carterbox, line 45: Please add an existence check for at least one file of each of the following categories: Cmake Config, Header,

recipes/ogdf/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - coincbc                                # [build_platform != target_platform]
+    - libdwarf-dev                           # [build_platform != target_platform]
+  host:
+    - cmake >=3.1
+    - make               # [not win]
+    - ninja              # [win]
+    - backward-cpp
+    - pugixml
+    - coincbc
+    - libdwarf-dev
```
carterbox, line 35: Please address static/underlinking warnings. We do not allow static linking.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=661341&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=3106

carterbox, line 35: Looks like the errors on CMake for OSX are some combination of needing to provide CMake some better search paths, ogdf find_library implementations not working correctly for OSX, and/or the upstream packages not exporting the same CMake targets on all platforms.

recipes/ogdf/build.sh
```diff
@@ -0,0 +1,23 @@
...
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DCMAKE_INSTALL_LIBDIR=lib \
+    -DBUILD_SHARED_LIBS=ON \
+    -DWITH_EXTERNAL_COIN=True \
+    -DWITH_EXTERNAL_BACKWARD=True \
+    -DWITH_EXTERNAL_PUGIXML=True \
+    ${EXTRA_CMAKE_ARGS}
+
+make -j ${CPU_COUNT}
```
carterbox, line 22: Add the --target option here and only build OGDF (the tests and example targets are unused).

carterbox (changes requested):

hadim:
I don't think I'll have time to maintain that repo. Someone should take the lead.

N-Coder:
If you need any help bringing a pre-built tmap package (i.e. wheel) to PyPI feel free to contact me. Doing that for the OGDF was pretty easy and anything that is available via pip is also available to users of conda (but not the other way around), so I don't really see a big need to invest too much work into abiding to the rules that are imposed on conda-forge.

hadim:
> If you need any help bringing a pre-built tmap package (i.e. wheel) to PyPI feel free to contact me. Doing that for the OGDF was pretty easy and anything that is available via pip is also available to users of conda (but not the other way around), so I don't really see a big need to invest too much work into abiding to the rules that are imposed on conda-forge.

Conda forge has a lot of advantages compared to built binaries on PyPI. I just don't have the time to work on it, but ultimately, OGDF should be packaged on conda-forge.

Thank you for your proposal on the TMAP wheel, but we already have an internal TMAP conda package that suits our needs for the moment.

PR #22125, opened 2023-02-23
Adding cuda-profiler-api recipes

adibbley:
xref: https://github.com/conda-forge/staged-recipes/issues/21382

Checklist
- [x] Build number is 0.


adibbley:
cc @jakirkham

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,42 @@
+{% set name = "cuda-profiler-api" %}
+{% set version = "12.0.76" %}
```
bdice, line 2:
```suggestion
{% set version = "12.0.76" %}
{% set cuda_version = "12.0" %}
```

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_profiler_api/{{ platform }}/cuda_profiler_api-{{ platform }}-{{ version }}-archive.{{ extension }}
+  sha256: 0fb470a3065ad39666dd231c12f8c8a9a94f34c4ac8494e5a399ea8e6ad69346  # [linux64]
+  sha256: 211513a8a194557a2b838a46700a820af6f38932fac5a82fa229c162779d9a25  # [ppc64le]
+  sha256: b93d4e99077c31e7034eeaea09068b43abc0dd2868a01d489ebcd1e51ceab816  # [aarch64]
+  sha256: e8bccbfda35abc7076ba6623578d11d05966a9c7fed817cbfe7d4dea9627057e  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
```
bdice, line 23:
```suggestion
build:
  number: 0
  skip: true  # [osx]

requirements:
  host:
    - cuda-version {{ cuda_version }}
  run:
    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

bdice (commented):
Recipe looks fine aside. I haven't checked the package contents yet.

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+requirements:
+  build:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
robertmaynard, line 32: We need a build dependency on `cuda-cudart-dev` since `cuda_profiler_api.h ` brings in headers provided by that package.

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "cuda-profiler-api" %}
```
robertmaynard, line 1: Shouldn't this be `cuda-profiler-api-dev`? this is just shipping header files which based on our pattern with `nvml` would make this a dev package.

robertmaynard (changes requested):

recipes/cuda-profiler-api/build.sh
```diff
@@ -0,0 +1,25 @@
...
+for i in `ls`; do
+    [[ $i == "build_env_setup.sh" ]] && continue
+    [[ $i == "conda_build.sh" ]] && continue
+    [[ $i == "metadata_conda_debug.yaml" ]] && continue
+    if [[ $i == "lib" ]] || [[ $i == "include" ]]; then
+        # Headers and libraries are installed to targetsDir
+        mkdir -p ${PREFIX}/${targetsDir}
+        mkdir -p ${PREFIX}/$i
+        cp -rv $i ${PREFIX}/${targetsDir}
+        # Header only component
```
vyasr, line 19: If we're removing the lib-specific logic since this is header-only we could also consider removing the lib part of the conditional as well as the else clause here. I don't have strong opinions either way but seems strange to remove parts of it but not the rest.

recipes/cuda-profiler-api/meta.yaml
vyasr: Agree with this.

recipes/cuda-profiler-api/meta.yaml
adibbley: Shouldn't that be a run dependency? Or both?

vyasr (commented):
Couple of open questions, nothing blocking from me but Rob's questions look like they do need to be addressed before we can merge.

recipes/cuda-profiler-api/meta.yaml
adibbley: We have other cases where `-dev` naming conventions aren't followed. Mostly devtools, but also `cuda-cccl`. I'd rather stay consistent with how this component has shipped previously.

recipes/cuda-profiler-api/meta.yaml
vyasr: run dependency sounds right since it's not required to build this package, but rather to build other packages that use this one. Terminology is confusing since this is a header-only package that is only used to build downstream packages, so it's build-time for the downstream packages (not runtime) but that is still a "run" requirement for this one (i.e. must be installed when this one is installed).

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/include/cuda_profiler_api.h  # [linux]
+    - if not exist %LIBRARY_INC%\cuda_profiler_api.h exit 1  # [win]
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
```
vyasr, line 47: Don't know if there's a better URL to be used here.

recipes/cuda-profiler-api/build.sh
adibbley: Sure, removed the lib statement. I originally removed the other block as it doesn't skip gracefully when no `.so` is found.

recipes/cuda-profiler-api/meta.yaml
leofang: But don't we also do `cuda-nvml-dev`?

adibbley: That also matches the given component name and historical naming, which would not be the case here.

recipes/cuda-profiler-api/build.sh
```diff
@@ -0,0 +1,25 @@
...
+for i in `ls`; do
+    [[ $i == "build_env_setup.sh" ]] && continue
+    [[ $i == "conda_build.sh" ]] && continue
+    [[ $i == "metadata_conda_debug.yaml" ]] && continue
+    if [[ $i == "include" ]]; then
+        # Headers and libraries are installed to targetsDir
+        mkdir -p ${PREFIX}/${targetsDir}
+        mkdir -p ${PREFIX}/$i
+        cp -rv $i ${PREFIX}/${targetsDir}
+        # Header only component
```
bdice, line 19: Clean up a bit.
```suggestion
        # Headers are installed to targetsDir (header only)
        mkdir -p ${PREFIX}/${targetsDir}
        cp -rv $i ${PREFIX}/${targetsDir}
```

recipes/cuda-profiler-api/build.sh
```diff
@@ -0,0 +1,25 @@
...
+    else
+        # Put all other files in targetsDir
```
bdice, line 21: Like cuda-nvml-dev, I would consider skipping this section. There's nothing to do here, right? See also: https://github.com/conda-forge/staged-recipes/pull/22127#discussion_r1167256968

recipes/cuda-profiler-api/meta.yaml
bdice: I'd also like to keep consistency with the existing package names.

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    - cuda-cudart-dev
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/include/cuda_profiler_api.h  # [linux]
+    - if not exist %LIBRARY_INC%\cuda_profiler_api.h exit 1  # [win]
```
bdice, line 44: For good measure, let's check for the headers with the driver API too:
```suggestion
    - test -f $PREFIX/targets/{{ target_name }}/include/cuda_profiler_api.h  # [linux]
    - test -f $PREFIX/targets/{{ target_name }}/include/cudaProfiler.h       # [linux]
    - if not exist %LIBRARY_INC%\cuda_profiler_api.h exit 1  # [win]
    - if not exist %LIBRARY_INC%\cudaProfiler.h exit 1       # [win]
```

bdice (commented):
Minor updates suggested.

bdice (approved):
All looks good. 😄

leofang (approved):

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    - cuda-cudart-dev
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/include/cuda_profiler_api.h  # [linux]
+    - test -f $PREFIX/targets/{{ target_name }}/include/cudaProfiler.h       # [linux]
```
kkraus14, line 44: Should these exist in `$PREFIX/include` as well?

kkraus14: See comment here: https://github.com/conda-forge/staged-recipes/pull/22127#discussion_r1168746777

kkraus14 (approved):

recipes/cuda-profiler-api/meta.yaml
robertmaynard: I am fine with keeping the current names so we have consistency.

robertmaynard (approved):

recipes/cuda-profiler-api/meta.yaml
jakirkham: Asked about this offline. There's not a great user facing page. We could improve the docs link though. Have added a suggestion related to that

recipes/cuda-profiler-api/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: CUDA Profiler API Headers.
+  description: |
+    CUDA Profiler API Headers.
+  doc_url: https://docs.nvidia.com/cuda/index.html
```
jakirkham, line 56:
```suggestion
  doc_url: https://docs.nvidia.com/nsight-compute/ProfilingGuide/index.html#range-replay-define-range
```

jakirkham (commented):
Small suggestion to improve the URLs here based on feedback

jakirkham:
Thanks all! 🙏

PR #22127, opened 2023-02-23
Add cuda-nvml-dev recipe

adibbley:
Checklist
- [x] Build number is 0.


adibbley:
cc @jakirkham

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,47 @@
+{% set name = "cuda-nvml-dev" %}
+{% set version = "12.0.76" %}
```
bdice, line 2:
```suggestion
{% set version = "12.0.76" %}
{% set cuda_version = "12.0" %}
```

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  sha256: abc366280d0aa1ca047323a3e316270b80ffabb649021184647cd16b44c60a71  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
bdice, line 28:
```suggestion
requirements:
  build:
    - {{ compiler("c") }}
    - {{ compiler("cxx") }}
  host:
    - cuda-version {{ cuda_version }}
  run:
    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/lib/stubs/libnvidia-ml.so  # [linux]
+    - test -f $PREFIX/targets/{{ target_name }}/include/nvml.h             # [linux]
+    - if not exist %LIBRARY_LIB%\nvml.lib exit 1  # [win]
+    - if not exist %LIBRARY_INC%\nvml.h exit 1    # [win]
+    - if not exist %PREFIX%\nvml\example\example.c exit 1    # [win]
```
jakirkham, line 48: Should we place this in a doc package?

adibbley: We could. Another question would be where we should install it on Windows? On Linux it will be here: `${PREFIX}/${targetsDir}/${PKG_NAME}`, but I don't think we decided on a similar place for Windows.

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  host:
+    - cuda-version {{ cuda_version }}
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/lib/stubs/libnvidia-ml.so  # [linux]
```
bdice, line 44: Where is the non-stub library shipped?

recipes/cuda-nvml-dev/meta.yaml
bdice: We decided not to ship Fortran bindings for libcublas on Windows partly for this reason, I think. I don't have an opinion here. I think it's acceptable to ship the example somewhere in this package without splitting since it's only about 20 kB (not like the 10 MB of PDF files in cuda-cupti).

recipes/cuda-nvml-dev/meta.yaml
bdice: Sounds good. Maybe a comment could help explain that? I wasn't aware.

bdice (approved):
LGTM aside from the unresolved question about where to ship `nvml/example/`.

robertmaynard (approved):
Approving presuming outstanding issues are resovled.

recipes/cuda-nvml-dev/meta.yaml
robertmaynard: I vote for not shipping this

vyasr: I also vote to not ship, at least for now.

vyasr (approved):

vyasr:
Once the PR is merged can we open an issue on the feedstock to track the packaging of the examples?

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
bdice, line 34: Most of the other packages have listed a sysroot here. Should we copy that? Is it needed?

bdice (approved):
Looks good, aside from one small question about sysroot.

recipes/cuda-nvml-dev/build.sh
```diff
@@ -0,0 +1,28 @@
...
+        # Headers and libraries are installed to targetsDir
+        mkdir -p ${PREFIX}/${targetsDir}
+        mkdir -p ${PREFIX}/$i
+        cp -rv $i ${PREFIX}/${targetsDir}
+        # Nothing to be symlinked in $PREFIX/lib
+    else
+        # Skip all other files (only samples here)
+        # mkdir -p ${PREFIX}/${targetsDir}/${PKG_NAME}
+        # cp -rv $i ${PREFIX}/${targetsDir}/${PKG_NAME}
+    fi
```
bdice, line 27: CI is failing here. I think you need a no-op to make the "else...fi" happy. https://unix.stackexchange.com/questions/133972/bash-syntax-error-when-else-follows-an-empty-then-clause
```suggestion
    else
        # Skip all other files (only samples here)
        :
        # mkdir -p ${PREFIX}/${targetsDir}/${PKG_NAME}
        # cp -rv $i ${PREFIX}/${targetsDir}/${PKG_NAME}  
    fi
```

recipes/cuda-nvml-dev/meta.yaml
jakirkham: The default is to build on CentOS 6 with GLIBC 2.12. If that works, then we don't need `sysroot`

If this requires CentOS 7 with GLIBC 2.17, then we need to add `sysroot` (as done elsewhere)

bdice: Okay. So if CI passes without it, we don't need it. Sounds good.

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - cuda-version {{ cuda_version }}
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/targets/{{ target_name }}/lib/stubs/libnvidia-ml.so  # [linux]
+    - test -f $PREFIX/targets/{{ target_name }}/include/nvml.h             # [linux]
```
kkraus14, line 45: Should this land in `$PREFIX/include` as well?

recipes/cuda-nvml-dev/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - if not exist %LIBRARY_LIB%\nvml.lib exit 1  # [win]
+    - if not exist %LIBRARY_INC%\nvml.h exit 1    # [win]
```
kkraus14, line 47: Should we test that the pkgconfig files are landing as expected?

adibbley: Yes, added that test now

recipes/cuda-nvml-dev/meta.yaml
adibbley: We held off on putting symlinks in `$PREFIX/include` for all packages. IIRC there were concerns with the generic naming of headers, such as `math_functions.h` and others. Using `nvcc` will include the `targets` directory by default.

kkraus14: Fair enough. I think we should prioritize a strategy around this because most of these are host functions where `nvcc` isn't required and generally shouldn't be used if there's no kernels / device code.

jakirkham:
Thanks all! 🙏

PR #22171, opened 2023-02-28
Add HiPACE++

ax3l:
Add HiPACE++: Highly efficient Plasma Accelerator Emulation, quasistatic particle-in-cell code.

Checklist
- [ ] Package does not vendor other packages. [AMReX](https://github.com/AMReX-Codes/amrex) is pulled on the fly, under the same license. Many binary variants make separating them very hard for now. They use the same license and the same LBNL/UC copyright holder.
- [x] Build number is 0.

### DIFF
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -17,7 +17,8 @@ cmake ^
     -DCMAKE_NM=llvm-nm                    ^
     -DCMAKE_VERBOSE_MAKEFILE=ON           ^
     -DHiPACE_amrex_branch=23.02          ^
-    -DHiPACE_COMPUTE=NOACC ^
+    -DHiPACE_openpmd_internal=OFF  ^
+    -DHiPACE_COMPUTE=NOACC  ^
     -DHiPACE_MPI=OFF
 if errorlevel 1 exit 1
 
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -17,7 +17,7 @@ cmake ^
     -DCMAKE_LINKER=lld-link        ^
     -DCMAKE_NM=llvm-nm             ^
     -DCMAKE_VERBOSE_MAKEFILE=ON    ^
-    -DHiPACE_amrex_branch=23.05    ^
+    -DHiPACE_amrex_branch=23.07    ^
     -DHiPACE_openpmd_internal=OFF  ^
     -DHiPACE_COMPUTE=NOACC         ^
     -DHiPACE_MPI=OFF
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -1,0 +1,34 @@
+@echo on
+
+:: simple install prep
+::   copy all impactx*.exe and impactx*.dll files
+if not exist %LIBRARY_PREFIX%\bin md %LIBRARY_PREFIX%\bin
+if errorlevel 1 exit 1
+
+:: configure
+cmake ^
+    -S %SRC_DIR% -B build                 ^
+    %CMAKE_ARGS%                          ^
+    -G "Ninja"                            ^
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo     ^
+    -DCMAKE_C_COMPILER=clang-cl           ^
+    -DCMAKE_CXX_COMPILER=clang-cl         ^
+    -DCMAKE_LINKER=lld-link               ^
+    -DCMAKE_NM=llvm-nm                    ^
+    -DCMAKE_VERBOSE_MAKEFILE=ON           ^
+    -DHiPACE_amrex_branch=23.02          ^
+    -DHiPACE_COMPUTE=NOACC ^
+    -DHiPACE_MPI=OFF
+if errorlevel 1 exit 1
+
+:: build
+cmake --build build --config RelWithDebInfo --parallel 2
+if errorlevel 1 exit 1
+
+:: test
+ctest --test-dir build --build-config RelWithDebInfo --output-on-failure
+if errorlevel 1 exit 1
+
+:: install
+cmake --build build --config RelWithDebInfo --target install
+if errorlevel 1 exit 1
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -16,7 +16,7 @@ cmake ^
     -DCMAKE_LINKER=lld-link               ^
     -DCMAKE_NM=llvm-nm                    ^
     -DCMAKE_VERBOSE_MAKEFILE=ON           ^
-    -DHiPACE_amrex_branch=23.02          ^
+    -DHiPACE_amrex_branch=23.03          ^
     -DHiPACE_openpmd_internal=OFF  ^
     -DHiPACE_COMPUTE=NOACC  ^
     -DHiPACE_MPI=OFF
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -1,7 +1,7 @@
 @echo on
 
 :: simple install prep
-::   copy all hipace*.exe and hipace*.dll files
+::   copy all hipace*.exe files
 if not exist %LIBRARY_PREFIX%\bin md %LIBRARY_PREFIX%\bin
 if errorlevel 1 exit 1
 
@@ -36,11 +36,5 @@ for /r "build\bin" %%f in (*.exe) do (
     echo %%~nf
     dir
     copy build\bin\%%~nf.exe %LIBRARY_PREFIX%\bin\
-    if errorlevel 1 exit 1
-)
-for /r "build\lib" %%f in (*.dll) do (
-    echo %%~nf
-    dir
-    copy build\lib\%%~nf.dll %LIBRARY_PREFIX%\lib\
     if errorlevel 1 exit 1
 )
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -7,29 +7,29 @@ if errorlevel 1 exit 1
 
 :: configure
 cmake ^
-    -S %SRC_DIR% -B build                 ^
-    %CMAKE_ARGS%                          ^
-    -G "Ninja"                            ^
-    -DCMAKE_BUILD_TYPE=RelWithDebInfo     ^
-    -DCMAKE_C_COMPILER=clang-cl           ^
-    -DCMAKE_CXX_COMPILER=clang-cl         ^
-    -DCMAKE_LINKER=lld-link               ^
-    -DCMAKE_NM=llvm-nm                    ^
-    -DCMAKE_VERBOSE_MAKEFILE=ON           ^
-    -DHiPACE_amrex_branch=23.03          ^
+    -S %SRC_DIR% -B build          ^
+    %CMAKE_ARGS%                   ^
+    -G "Ninja"                     ^
+    -DCMAKE_BUILD_TYPE=Release     ^
+    -DCMAKE_C_COMPILER=clang-cl    ^
+    -DCMAKE_CXX_COMPILER=clang-cl  ^
+    -DCMAKE_LINKER=lld-link        ^
+    -DCMAKE_NM=llvm-nm             ^
+    -DCMAKE_VERBOSE_MAKEFILE=ON    ^
+    -DHiPACE_amrex_branch=23.03    ^
     -DHiPACE_openpmd_internal=OFF  ^
-    -DHiPACE_COMPUTE=NOACC  ^
+    -DHiPACE_COMPUTE=NOACC         ^
     -DHiPACE_MPI=OFF
 if errorlevel 1 exit 1
 
 :: build
-cmake --build build --config RelWithDebInfo --parallel 2
+cmake --build build --config Release --parallel 2
 if errorlevel 1 exit 1
 
 :: test -> deferred to test.bat
 
 :: install
-::cmake --build build --config RelWithDebInfo --target install
+::cmake --build build --config Release --target install
 ::if errorlevel 1 exit 1
 for /r "build\bin" %%f in (*.exe) do (
     echo %%~nf
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -26,9 +26,7 @@ if errorlevel 1 exit 1
 cmake --build build --config RelWithDebInfo --parallel 2
 if errorlevel 1 exit 1
 
-:: test
-ctest --test-dir build --build-config RelWithDebInfo --output-on-failure
-if errorlevel 1 exit 1
+:: test -> deferred to test.bat
 
 :: install
 cmake --build build --config RelWithDebInfo --target install
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -29,5 +29,17 @@ if errorlevel 1 exit 1
 :: test -> deferred to test.bat
 
 :: install
-cmake --build build --config RelWithDebInfo --target install
-if errorlevel 1 exit 1
+::cmake --build build --config RelWithDebInfo --target install
+::if errorlevel 1 exit 1
+for /r "build\bin" %%f in (*.exe) do (
+    echo %%~nf
+    dir
+    copy build\bin\%%~nf.exe %LIBRARY_PREFIX%\bin\
+    if errorlevel 1 exit 1
+)
+for /r "build\lib" %%f in (*.dll) do (
+    echo %%~nf
+    dir
+    copy build\lib\%%~nf.dll %LIBRARY_PREFIX%\lib\
+    if errorlevel 1 exit 1
+)
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -1,7 +1,7 @@
 @echo on
 
 :: simple install prep
-::   copy all impactx*.exe and impactx*.dll files
+::   copy all hipace*.exe and hipace*.dll files
 if not exist %LIBRARY_PREFIX%\bin md %LIBRARY_PREFIX%\bin
 if errorlevel 1 exit 1
 
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -10,6 +10,7 @@ cmake ^
     -S %SRC_DIR% -B build          ^
     %CMAKE_ARGS%                   ^
     -G "Ninja"                     ^
+    -DAMReX_INSTALL=OFF            ^
     -DCMAKE_BUILD_TYPE=Release     ^
     -DCMAKE_C_COMPILER=clang-cl    ^
     -DCMAKE_CXX_COMPILER=clang-cl  ^
--- a/recipes/hipace/bld.bat
+++ b/recipes/hipace/bld.bat
@@ -17,7 +17,7 @@ cmake ^
     -DCMAKE_LINKER=lld-link        ^
     -DCMAKE_NM=llvm-nm             ^
     -DCMAKE_VERBOSE_MAKEFILE=ON    ^
-    -DHiPACE_amrex_branch=23.03    ^
+    -DHiPACE_amrex_branch=23.05    ^
     -DHiPACE_openpmd_internal=OFF  ^
     -DHiPACE_COMPUTE=NOACC         ^
     -DHiPACE_MPI=OFF
--- a/recipes/hipace/build.sh
+++ b/recipes/hipace/build.sh
@@ -28,6 +28,7 @@ cmake \
     -DHiPACE_COMPUTE=NOACC               \
     -DHiPACE_IPO=${HiPACE_IPO}           \
     -DHiPACE_amrex_branch=23.02          \
+    -DHiPACE_openpmd_internal=OFF        \
     -DHiPACE_MPI=OFF
 
 # build
--- a/recipes/hipace/build.sh
+++ b/recipes/hipace/build.sh
@@ -21,7 +21,7 @@ cmake \
     -DCMAKE_INSTALL_LIBDIR=lib        \
     -DCMAKE_INSTALL_PREFIX=${PREFIX}  \
     -DHiPACE_COMPUTE=NOACC            \
-    -DHiPACE_amrex_branch=23.05       \
+    -DHiPACE_amrex_branch=23.07       \
     -DHiPACE_openpmd_internal=OFF     \
     -DHiPACE_MPI=OFF
 
--- a/recipes/hipace/build.sh
+++ b/recipes/hipace/build.sh
@@ -1,0 +1,42 @@
+#!/usr/bin/env bash
+
+# avoid side-injection of -std=c++14 flag in some toolchains
+if [[ ${CXXFLAGS} == *"-std=c++14"* ]]; then
+    echo "14 -> 17"
+    export CXXFLAGS="${CXXFLAGS} -std=c++17"
+fi
+# Darwin modern C++
+#   https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
+if [[ ${target_platform} =~ osx.* ]]; then
+    export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+# IPO/LTO does only work with certain toolchains
+HiPACE_IPO=ON
+if [[ ${target_platform} =~ osx.* ]]; then
+    HiPACE_IPO=OFF
+fi
+
+# configure
+cmake \
+    -S ${SRC_DIR} -B build                \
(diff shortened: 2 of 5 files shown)


recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+    - cmake >=3.20.0
+    - lld           # [win]
+    # TODO: add libxml2 dep in lld-feedstock
+    - libxml2       # [win]
+    - make          # [unix]
+    - ninja         # [win]
+    - libgomp       # [linux]
+    - llvm-openmp   # [osx or win]
+  host:
+    - pkgconfig
```
ax3l, line 37:
```suggestion
    - fftw
    - pkgconfig
```

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - ninja         # [win]
+    - libgomp       # [linux]
+    - llvm-openmp   # [osx or win]
+  host:
+    - fftw
+    - pkgconfig
+    # ctest
+    - matplotlib-base  # [build_platform == target_platform]
+    - numpy            # [build_platform == target_platform]
+    - pandas           # [build_platform == target_platform]
```
ax3l, line 42:
```suggestion
    - openpmd-viewer  # [build_platform == target_platform]
    - pandas           # [build_platform == target_platform]
```

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+  source_files:
+    - examples/beam_in_vacuum/inputs_SI
+  files:
+    - test.sh   # [unix]
+    - test.bat  # [win]
+  imports:
+    - hipace
+  commands:
+    - test -f ${PREFIX}/bin/hipace.NOMPI.NOACC.DP  # [unix]
+
```
ax3l, line 66:
```suggestion
```

recipes/hipace/bld.bat
```diff
@@ -0,0 +1,34 @@
...
+    %CMAKE_ARGS%                          ^
+    -G "Ninja"                            ^
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo     ^
+    -DCMAKE_C_COMPILER=clang-cl           ^
+    -DCMAKE_CXX_COMPILER=clang-cl         ^
+    -DCMAKE_LINKER=lld-link               ^
+    -DCMAKE_NM=llvm-nm                    ^
+    -DCMAKE_VERBOSE_MAKEFILE=ON           ^
+    -DHiPACE_amrex_branch=23.02          ^
+    -DHiPACE_COMPUTE=NOACC ^
```
ax3l, line 20:
```suggestion
    -DHiPACE_openpmd_internal=OFF  ^
    -DHiPACE_COMPUTE=NOACC  ^
```

recipes/hipace/build.sh
```diff
@@ -0,0 +1,42 @@
...
+cmake \
+    -S ${SRC_DIR} -B build                \
+    ${CMAKE_ARGS}                         \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo     \
+    -DCMAKE_VERBOSE_MAKEFILE=ON           \
+    -DCMAKE_INSTALL_LIBDIR=lib            \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX}      \
+    -DHiPACE_COMPUTE=NOACC               \
+    -DHiPACE_IPO=${HiPACE_IPO}           \
+    -DHiPACE_amrex_branch=23.02          \
```
ax3l, line 30:
```suggestion
    -DHiPACE_amrex_branch=23.02          \
    -DHiPACE_openpmd_internal=OFF        \
```

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - lld           # [win]
+    # TODO: add libxml2 dep in lld-feedstock
+    - libxml2       # [win]
+    - make          # [unix]
+    - ninja         # [win]
+    - libgomp       # [linux]
+    - llvm-openmp   # [osx or win]
+  host:
+    - fftw
+    - pkgconfig
```
ax3l, line 38:
```suggestion
    - openpmd-api
    - pkgconfig
```

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - lld           # [win]
+    # TODO: add libxml2 dep in lld-feedstock
+    - libxml2       # [win]
+    - make          # [unix]
+    - ninja         # [win]
+    - libgomp       # [linux]
+    - llvm-openmp   # [osx or win]
+  host:
+    - fftw
+    - openpmd-api
```
ax3l, line 38:
```suggestion
    - openpmd-api >=0.14.5
```

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
+{% set name = "hipace" %}
+{% set version = "23.02" %}
+{% set build = 0 %}
+{% set sha256 = "f4f5958f0600dd59cc1bb02218d5b7d1e6e0da34461c504e4d0367c98d5d6113" %}
```
ax3l, line 4: Update to 23.03 to unbreak Windows build.

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - lld           # [win]
+    # TODO: add libxml2 dep in lld-feedstock
+    - libxml2       # [win]
+    - make          # [unix]
+    - ninja         # [win]
+    - libgomp       # [linux]
+    - llvm-openmp   # [osx or win]
+  host:
+    - fftw
+    - openpmd-api >=0.14.5
```
ax3l, line 38: This triggers the need for a newer numpy, that for some reason does not exist in the `staged-recipes` repo...?

MaxThevenet: HiPACE++ now requires `openpmd-api >=0.15.0`

ax3l: Thanks!

To be precise, now as of 23.04+, currently working on packaging 23.03+.

ax3l: Let's just take the latest:
```suggestion
    - openpmd-api
```

ax3l: Urgh, does not do the trick and tries to pull in 0.14.4 o.0

ax3l: @mfansler do/can we use `libmamba` or `mamba` here? The dependency resolver does not find the latest versions (0.15.1).

ax3l:
@conda-forge/help-c-cpp I am a bit puzzled what I can do to overcome this numpy version resolution issue here.

recipes/hipace/meta.yaml
mfansler: Sorry, just saw this. Conda Forge already uses `mamba` as the solver for all CI ops. The situation looks like `numpy=1.24` is getting used in recent `openpmd-api` builds. But currently using plain `numpy` here gets replaced with `numpy=1.21`, hence only older `openpmd-api` versions get resolved.

In local solving tests (not building) I was able to get `openpmd-api >=0.14.5` to work with setting `numpy=1.23`. You can either try:

 - setting a stricter bound like this on `numpy` in this recipe; or 
 - loosen the `openpmd-api` builds to let it use the `numpy` global pinning. 

I think the latter could be achieved by removing [the explicit bound on `numpy` (`>=1.15.0`)](https://github.com/conda-forge/openpmd-api-feedstock/blob/d4bb38a4b402c533b11287a128a582f3c0cf95e9/recipe/meta.yaml#L73), which I suspect is what is interfering with the default global pinning (`numpy=1.21`).

Maybe someone from @conda-forge/help-python could weigh in on this. I don't want to recommend rebuilding that feedstock unnecessarily, since it has so many variants. 😬 But I believe getting the `openpmd-api` builds compatible with global pinning would be preferable.

carterbox: Please see https://github.com/conda-forge/openpmd-api-feedstock/pull/102#issuecomment-1539217537 for a concrete example about what our docs say regarding pinning numpy versions. Consider rebuilding necessary if it prevents unsolvable environments.

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+    - openpmd-api
+    - pkgconfig
+    # ctest
+    - matplotlib-base  # [build_platform == target_platform]
+    - numpy            # [build_platform == target_platform]
+    - openpmd-viewer   # [build_platform == target_platform]
+    - pandas           # [build_platform == target_platform]
+    - pytest           # [build_platform == target_platform]
+    - python           # [build_platform == target_platform]
+    - scipy            # [build_platform == target_platform]
```
carterbox, line 47:
```suggestion
    - pkgconfig
  host:
    - fftw
    - openpmd-api
    - numpy
    - python
```
My suggestion to get a solveable environment is to try removing all of these dependencies that are only used for testing.

You can either skip running unit tests (unit tests are the upstream project's responsibility) or delay them to the test phase instead of doing them during the build phase.

Also, pkgconfig is a build dependency not a host dependency?

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,93 @@
...
+
+source:
+  url: https://github.com/Hi-PACE/hipace/archive/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: {{ build }}
+  run_exports:
+    # releases are not (yet) compatible
+    - hipace {{ version }}
```
carterbox, line 18:
```suggestion
    - {{ pin_compatible('hipace', max_pin='x.x.x') }}
```

recipes/hipace/meta.yaml
carterbox: Also, it's not uncommon to need to rebuild packages with newer versions of their dependencies.

ax3l: Thanks a lot for the guidance - fix coming in https://github.com/conda-forge/openpmd-api-feedstock/pull/103

recipes/hipace/meta.yaml
ax3l: Thanks! The reason so many of the test dependencies are in the build phase is that they are used during the `ctest` phase, which is run in the build directory.

recipes/hipace/meta.yaml
ax3l: Oh, this [complains below](https://github.com/conda-forge/staged-recipes/pull/22171#issuecomment-1541176100):
```
    pin_subpackage should be used instead of pin_compatible for hipace because it is one of the known outputs of this recipe: ['hipace'].
```

Will try `pin_subpackage`

recipes/hipace/meta.yaml
ax3l: I try to add `build/` to the `source_files:` section to defer the `ctest` run to the test phase... let's see if that can work.

recipes/hipace/meta.yaml
carterbox: Yes. Listen to the linter. 😄

ax3l:
Thanks! `numba` is pulled by `openpmd-viewer` to make some operations faster (optional). Let's see if we can add it explicitly in the tests.

ax3l:
@mfansler ranom thought: we cannot use numba from the mainline `conda` channel, do we? Because there is already a Python 3.11 build there: https://anaconda.org/numba/numba/files

recipes/hipace/test.bat
```diff
@@ -0,0 +1,12 @@
...
+set "OMP_NUM_THREADS=2"
+set "TEST_DIR=example\beam_in_vacuum"
+
+:: executable
+impactx.NOMPI.NOACC.DP.exe %TEST_DIR%\inputs_SI
+if errorlevel 1 exit 1
+
+:: ctest
+ctest --test-dir build --build-config RelWithDebInfo --output-on-failure
+if errorlevel 1 exit 1
```
ax3l, line 12: I will skip this because @isuruf told me on Gitter:
> adding build/ would work, but keep in mind that whatever is in source_files would end up in the final conda package in order to facilitate testing in user machines. (They won't get extracted to the user environment, but will be in the .conda file

mfansler:
> @mfansler random thought: we cannot use numba from the mainline `conda` channel, do we? Because there is already a Python 3.11 build there: https://anaconda.org/numba/numba/files

Unfortunately, no. Use of outside channels gets flagged and won't build.

recipes/hipace/test.bat
ax3l: Reworked - using a subset of the test scripts now to validate correctness in the test phase (w/o `ctest`).

mfansler:
Looks like Windows build is installing into system folder instead of the prefix. So Conda finds nothing has been added to the prefix to bundle up. Is the install directory be specified correctly?

MaxThevenet:
I confirm to be listed as co-maintainer

ax3l:
@mfansler thanks, fixed!

SeverinDiederichs:
I confirm to be listed as co-maintainer

ax3l:
@conda-forge/help-c-cpp ready for merge :)

recipes/hipace/bld.bat
```diff
@@ -0,0 +1,45 @@
...
+    -DCMAKE_NM=llvm-nm                    ^
+    -DCMAKE_VERBOSE_MAKEFILE=ON           ^
+    -DHiPACE_amrex_branch=23.03          ^
+    -DHiPACE_openpmd_internal=OFF  ^
+    -DHiPACE_COMPUTE=NOACC  ^
+    -DHiPACE_MPI=OFF
+if errorlevel 1 exit 1
+
+:: build
+cmake --build build --config RelWithDebInfo --parallel 2
```
carterbox, line 26: Please switch to Release without Debugging info. You can create a dev label if you want to publish pre-release / dev builds.

recipes/hipace/build.sh
```diff
@@ -0,0 +1,40 @@
...
+HiPACE_IPO=ON
+if [[ ${target_platform} =~ osx.* ]]; then
+    HiPACE_IPO=OFF
+fi
+
+# configure
+cmake \
+    -S ${SRC_DIR} -B build                \
+    ${CMAKE_ARGS}                         \
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo     \
```
carterbox, line 24: See comment above about shipping debugging info.

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,84 @@
...
+build:
+  number: 0
+  run_exports:
+    # releases are not (yet) compatible
+    - {{ pin_subpackage('hipace', max_pin='x.x') }}
+
+requirements:
+  build:
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
```
carterbox, line 21: Python in the build and not host requirements implies that you are just using python as build tool and that your package does not link to or have API requirements of the python version.

If this is the case, then you don't need to build against multiple python versions. You should probably be ignoring run_exports from python.

If this is not the case, then python should also be listed in the host section.

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - tests/
+  files:
+    - test.sh   # [unix]
+    - test.bat  # [win]
+  commands:
+    - test -f ${PREFIX}/bin/hipace.NOMPI.NOACC.DP  # [unix]
+    - if not exist %LIBRARY_PREFIX%\bin\hipace.NOMPI.NOACC.DP.exe exit 1  # [win]
+    # smoke tests: execute
+    - ./test.sh      # [unix and build_platform == target_platform]
+    - call test.bat  # [win]
```
carterbox, line 61: For each platform, please add an existence test for a header file, and a cmake file. Also test that the library is shared and not static. This will hint to readers of the recipe that these files are expected to exist in the package without them needing to download it.

recipes/hipace/bld.bat
```diff
@@ -0,0 +1,45 @@
+@echo on
+
+:: simple install prep
+::   copy all hipace*.exe and hipace*.dll files
+if not exist %LIBRARY_PREFIX%\bin md %LIBRARY_PREFIX%\bin
+if errorlevel 1 exit 1
+
+:: configure
+cmake ^
```
carterbox, line 9: Please add the BUILD_SHARED_LIBS CMake variable. We do not allow shipping static libraries.

carterbox, line 9: Please add the BUILD_SHARED_LIBS CMake variable. We do not allow shipping static libraries.

carterbox, line 9: Please add the BUILD_SHARED_LIBS CMake variable. We do not allow shipping static libraries.

recipes/hipace/build.sh
```diff
@@ -0,0 +1,40 @@
...
+    -DHiPACE_openpmd_internal=OFF        \
+    -DHiPACE_MPI=OFF
+
+# build
+cmake --build build --parallel ${CPU_COUNT}
+
+# test -> deferred to test.sh
+
+# install
+cmake --build build --target install
```
carterbox, line 40: What is being installed in "$PREFIX/Tools" ? Maybe remove it.

carterbox (changes requested):

recipes/hipace/bld.bat
carterbox: The static library being shipped is AMReX, which is not this package. Definitely, you should not be shipping the headers / libraries of another package with this package. If you need to link your binary to AMReX statically because you are using a non-release / patched version that's probably fine, but if you are linking against one of their [stable releases](https://github.com/AMReX-Codes/amrex/releases), we would like you to package AMReX separately and link to it dynamically.

recipes/hipace/meta.yaml
ax3l: This was due to the `ctest` requiring Python scripts. Since we do not run ctest anymore in the build phase, we can absolutely simplify the build combinations now :tada:

recipes/hipace/bld.bat
ax3l: Thx. Disabling the lib via `AMReX_INSTALL=OFF`

ax3l: Thx. Disabling the lib via `AMReX_INSTALL=OFF`

recipes/hipace/build.sh
ax3l: This should be fixed with `AMReX_INSTALL=OFF`, too :)

recipes/hipace/meta.yaml
ax3l: The goal is only to install a binary, which we check three lines above.

I disabled install of all other artifacts, thanks for catching!

ax3l: Update: now a clean
```
-- Install configuration: "Release"
-- Installing: $PREFIX/bin/hipace.NOMPI.NOACC.DP
```

Thanks for catching this!

recipes/hipace/bld.bat
ax3l: Update: now a clean
```
Install the project...
$BUILD_PREFIX/bin/cmake -P cmake_install.cmake
-- Install configuration: "Release"
-- Installing: $PREFIX/bin/hipace.NOMPI.NOACC.DP
```

Thanks for catching this!

ax3l:
@carterbox thank you for the review and issues caught! :pray:
PR updated to address them :)

ax3l:
@conda-forge/help-c-cpp ready for merge :)

ax3l:
@carterbox updated & ready :)

recipes/hipace/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Hi-PACE/hipace/archive/v{{ version }}.tar.gz
+  sha256: f32104af1e7ca33b7ef8a3fd7964ee62db13a68a00af305d4ace5d0e86e2efb2
+
+build:
+  number: 0
```
carterbox, line 13:
```suggestion
build:
  # Prevent conda-build from deciding to build for each python version when hipace doesn't depend on python
  skip: True  # [py != 310]
  number: 0
```
I'm pretty sure from our discussions that hipace does NOT depend on Python, but I think some bug in conda-build is causing it to decide to build your recipe three times. Use skip here to prevent building for multiple python versions since it is not needed.

carterbox (changes requested):

carterbox:
LGTM once my patches build successfully.

carterbox (changes requested):
What's going on with static links to libadios2? We have this available on our channel, so it should be listed in the host section if you're using it.

From the build logs:
```
-- Found ADIOS: $PREFIX/lib/libadios_nompi.a;$PREFIX/lib/libz.so;$PREFIX/lib/libbz2.so;$PREFIX/lib/libblosc.so;$BUILD_PREFIX/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so;$BUILD_PREFIX/x86_64-conda-linux-gnu/sysroot/usr/lib/libm.so;$BUILD_PREFIX/x86_64-conda-linux-gnu/sysroot/usr/lib/librt.so (found version "1.13.1") 
-- Found ADIOS2: $PREFIX/lib/cmake/adios2/adios2-config.cmake (found version "2.8.3") found components: C CXX 
-- openPMD-api: Found version '0.15.1'
```

```

WARNING (hipace): run-exports library package conda-forge::adios2-2.8.3-nompi_py311hc31d2c9_105 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
```

ax3l:
Oh, I was not aware of the ADIOS1 static lib build. Let me look into that in https://github.com/conda-forge/adios-feedstock

ax3l:
Hm, that is wild because I pull it via `openPMD-api`, which is in `host`. Do I need to list transitive dependencies from other libraries in `host`?

recipes/hipace/build.sh
```diff
@@ -0,0 +1,41 @@
...
+if [[ ${target_platform} =~ osx.* ]]; then
+    export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+# IPO/LTO does only work with certain toolchains
+HiPACE_IPO=ON
+if [[ ${target_platform} =~ osx.* ]]; then
+    HiPACE_IPO=OFF
+fi
+
```
ax3l, line 19: Not yet implemented:
```suggestion
```

recipes/hipace/build.sh
```diff
@@ -0,0 +1,41 @@
...
+cmake \
+    -S ${SRC_DIR} -B build            \
+    ${CMAKE_ARGS}                     \
+    -DAMReX_INSTALL=OFF               \
+    -DCMAKE_BUILD_TYPE=Release        \
+    -DCMAKE_VERBOSE_MAKEFILE=ON       \
+    -DCMAKE_INSTALL_LIBDIR=lib        \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX}  \
+    -DHiPACE_COMPUTE=NOACC            \
+    -DHiPACE_IPO=${HiPACE_IPO}        \
```
ax3l, line 30:
```suggestion
```

carterbox:
> Thank you for the guidance! Oh, that is interesting because I pull it via `openPMD-api`, which is in `host`. Do I need to list transitive dependencies from other libraries in `host`?

If hipace doesn't link directly to adios2, then it should not be listed in your recipe. Why do you[ export a dependency on adios2 from the openpmd pacakage](https://github.com/conda-forge/openpmd-api-feedstock/blob/03a1b7cb812f07de235b8462fa417f953d14c7ef/recipe/meta.yaml#L51)? It is very unusual for openpmd to export a dependency on a package which is not built by the openpmd-feedstock.

ax3l:
Thank you for catching that. Fixed via https://github.com/conda-forge/openpmd-api-feedstock/pull/104

ax3l:
@conda-forge/help-c-cpp ready for merge :)

carterbox:
There is a package collision between adios and adios2. Open an issue on one of those feedstocks and we can discuss the details. i.e. Should aido and adios2 be mutually exclusive? If so how to accomplish that, etc.

ax3l:
They are complementary and can be used at the same time (e.g., in openPMD-api).

There seems to be a collision in created cli tools, where in case of conflict the tools of ADIOS2 should be selected.

ax3l:
> Open an issue on one of those feedstocks and we can discuss the details.

Thank you :pray: 
https://github.com/conda-forge/adios2-feedstock/issues/72

ax3l:
@conda-forge/help-c-cpp ready for merge :)

ax3l:
@carterbox thank you for your help, all green now :)

PR #22206, opened 2023-03-04
Add `rb-connection_pool`

svengiegerich:
Checklist
- [x] Build number is 0.


recipes/rb-connection_pool/LICENSE.txt
```diff
@@ -0,0 +1,20 @@
+Copyright (c) 2011 Mike Perham
```
janjagusch, line 1: @svengiegerich, you should be able to get rid of the license file by following the same steps we discussed here: https://github.com/conda-forge/staged-recipes/pull/22205

recipes/rb-connection_pool/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE.txt
+  summary: Generic connection pool for Ruby
+  doc_url: https://www.rubydoc.info/gems/{{ name }}
+  dev_url: https://github.com/mperham/{{ name }}
+
+extra:
+  recipe-maintainers:
+    - svengiegerich
```
janjagusch, line 40: happy to co-maintain. :)

```suggestion
    - svengiegerich
    - janjagusch
```

janjagusch:
> @janjagusch how can I add two packages within the same PR (as one depends on the other)? I've tried in [e52addb](https://github.com/conda-forge/staged-recipes/commit/e52addb71afaab1c42f7756e1a5d66527f7c72cb) but failed with the msg that the dependency cannot be found... 😶‍🌫️ (I'd like to do that so that we don't need another iteration cycle for approvals...)

Just check `rb-net-http-persistent/meta.yaml` into this branch, and then CI should build both packages.

recipes/rb-net-http-persistent/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+requirements:
+  host:
+    - ruby <2.6, >=2.3
+  run:
+    - ruby <2.6, >=2.3
+
+test:
+  commands:
+    - ruby -r {{ name }} -e 'exit 0'
+    - rb-connection_pool >=2.2, <3
```
janjagusch, line 28: You probably added this by accident, right?

```suggestion
```

recipes/rb-net-http-persistent/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  host:
+    - ruby <2.6, >=2.3
+  run:
+    - ruby <2.6, >=2.3
+
+test:
+  commands:
+    - ruby -r {{ name }} -e 'exit 0'
```
janjagusch, line 27: I think the import for this package is different, see [here](https://github.com/drbrain/net-http-persistent#synopsis-).

```suggestion
    - ruby -r 'net/http/persistent' -e 'exit 0'
```

recipes/rb-net-http-persistent/LICENSE.txt
```diff
@@ -0,0 +1,9 @@
+(The MIT License)
```
janjagusch, line 1: @svengiegerich, you should be able to get rid of the license file. :)

janjagusch (approved):

recipes/rb-connection_pool/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - gem install -N -l -V --norc --ignore-dependencies {{ name }}-{{ version }}.gem
+    - gem unpack {{ name }}-{{ version }}.gem
+    - cp {{ name }}-*/LICENSE .
+  skip: true  # [win]
+
+requirements:
+  host:
+    - ruby <2.6
+  run:
+    - ruby <2.6
```
xhochy, line 24: For now, let's pin ruby more explicit.
```suggestion
    - ruby >=2.5,<2.6
  run:
    - ruby >=2.5,<2.6
```

xhochy:
I would prefer if this could be ported to a newer Ruby version but it seems that our general Ruby stack is way behind and thus this is probably a reasonable compromise for now.

PR #22230, opened 2023-03-07
add rsatoolbox

ilogue:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/rsatoolbox/meta.yaml
+++ b/recipes/rsatoolbox/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "rsatoolbox" %}
+{% set version = "0.1.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/rsatoolbox-{{ version }}.tar.gz
+  sha256: fee6e0134c345f4d7e7f2d7bdde20b13724e9d0d32f89e8b22b7fc17ed21ef9a
+
+build:
+  skip: true  # [py<37 or py>310 or py2k]
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - cython # >=3.0.0a11
+    - numpy >=1.21.2
+    - scipy
+    - setuptools
+    - setuptools-scm
+    - pip
+  run:
+    - python
+    - {{ pin_compatible('numpy') }}
+    - {{ pin_compatible('scipy') }}
+    - pandas
+    - scikit-learn
+    - scikit-image
+    - matplotlib-base
+    - h5py
+    - joblib
+    - tqdm
+    - coverage
+
+
+test:
+  imports:
+    - rsatoolbox
+  commands:
+    - pip check
+    - python -m unittest -v rsatoolbox.test
+  requires:
+    - pip
+
+about:
+  home: https://github.com/rsagroup/rsatoolbox
+  summary: Representational Similarity Analysis (RSA) in Python
+  license: MIT
+  license_file: LICENSE
+  doc_url: https://rsatoolbox.readthedocs.io/
+  dev_url: https://github.com/rsagroup/rsatoolbox
+
+extra:
+  recipe-maintainers:
+    - ilogue


ilogue:
hi @conda-forge/help-python this is ready for review :)

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - cython # >=3.0.0a11
+    - numpy >=1.21.2
```
beckermr, line 23: You don't want this constraint here or if you do, you need to specify the minimum version. Otherwise, conda will install the latest version of numpy at build time (e.g., 1.24 or w/e) and you won't be able to run against say 1.23.

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - scipy
+    - setuptools
+    - setuptools-scm
+    - pip
+  run:
+    - python
+    - {{ pin_compatible('numpy') }}
+    - {{ pin_compatible('scipy') }}
```
beckermr, line 31: I don't follow this constraint. I didn't think scipy had any ABI dependence?

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - cython # >=3.0.0a11
+    - numpy >=1.21.2
+    - scipy
```
beckermr, line 24: Does this package link to scipy? I didn't think that was possible.

recipes/rsatoolbox/meta.yaml
beckermr, line 23: If this package doesn't link to numpy by compiling code, then you don't need numpy in host. The same is true for scipy.

beckermr (changes requested):
I have some questions about host vs run and the constraints.

recipes/rsatoolbox/meta.yaml
h-vetinari: There's a small C-API in scipy, but I kinda doubt that this is using it.

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/rsatoolbox-{{ version }}.tar.gz
+  sha256: 2d091cbaa33373bf9da4df5ca8d127f0e427431a3db726076090ab2d54fe1213
+
+build:
+  skip: true  # [py<37 or py>310 or py2k]
```
h-vetinari, line 13:
```suggestion
```

None of these are necessary: python 2 & <3.7 aren't built in conda-forge anymore, and 3.11 will only be added by the migrator (which can then be deferred until there's a rsatoolbox version that supports it).

recipes/rsatoolbox/meta.yaml
beckermr: Ahh cool! I did not know that!

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - cython # >=3.0.0a11
```
h-vetinari, line 22: If you absolutely want/need cython 3, you can add a `conda_build_config.yaml` file like [this one](https://github.com/conda-forge/aimrocks-feedstock/blob/main/recipe/conda_build_config.yaml).

recipes/rsatoolbox/meta.yaml
h-vetinari: In other words, assuming you're compiling against numpy's C-API
```suggestion
    - numpy
```

If not, remove

recipes/rsatoolbox/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - {{ pin_compatible('numpy') }}
+    - {{ pin_compatible('scipy') }}
+    - pandas
+    - scikit-learn
+    - scikit-image
+    - matplotlib-base
+    - h5py
+    - joblib
+    - tqdm
+    - coverage
```
h-vetinari, line 39: Coverage is a very test-specific dependence, I'd be surprised that this is necessary for correct operation of the package?

h-vetinari (commented):
Stumbled over this since https://github.com/rsagroup/rsatoolbox/pull/302 links to one of my PRs. 🙃

recipes/rsatoolbox/meta.yaml
ilogue: Good catch! removed.

recipes/rsatoolbox/meta.yaml
ilogue: Turned out this was just because we wanted to design against Cython3, but it also builds on the latest Cython 2.9

recipes/rsatoolbox/meta.yaml
ilogue: Got it. Removed the skip entry.

recipes/rsatoolbox/meta.yaml
ilogue: I've removed the `pin_compatible` restraint. I thought there could be cases when the user installs `rsatoolbox` in an existing environment (an old one e.g. python 3.7) and it wouldn't update numpy/scipy to the version rsatoolbox was built against.

recipes/rsatoolbox/meta.yaml
ilogue: I'm not sure I understand fully, I thought `>=1.21.2` would mean that the minimum is `1.21.2`? Either way I've let go of the version constraints as this seems to be the typical way of doing things in conda, and it seems unlikely that people would now have something old like `1.19`.

ilogue: Thank you @beckermr and @h-vetinari !! I don't have much experience with conda but these all seem like good suggestions and I've implemented them.

Having a closer look at our Cython code it turns out we don't currently have a build-time dependency on `numpy` (it returns regular memory views which are then converted to numpy arrays in the calling code). We do depend on scipy (`blas.dgemv`):

https://github.com/rsagroup/rsatoolbox/blob/bef1775960773e39ed6e9487047e298ad0b25542/src/rsatoolbox/cengine/similarity.pyx#L8

So I've switched out the host requirements but got rid of the `pin_compatible()` restraint. 

Thanks again and let me know if there's anything else to fix

h-vetinari (commented):
I haven't verified the run-deps against the upstream requirements (and it would be nicer if everything but `python` was alphabetized), but basically this LGTM. :)

beckermr:
Do we not inherit Abi constraints from scipy?

ilogue:
> Do we not inherit Abi constraints from scipy?

Sorry I wasn't sure if you meant to ask me or @h-vetinari .
If you mean, are there old versions of scipy that rsatoolbox would not work / build on, the answer is yes. But I understood from the comments above that it's not required to specify those constraints as they do not occur within the restrictions that conda applies. But if I misunderstood I'm happy to add `>=1.5` to the scipy entry?

ilogue:

beckermr:
If you compile code or use cython against scipy in host, then you would want the pin_compatible for scipy in run. This is what I meant by an ABI constraint.

PR #22236, opened 2023-03-07
NEW: Add ABI compliance checker

carterbox:
These software tools analyze the symbols of shared libraries to see whether they are ABI compatible.

The tools themselves are actually self-contained perl scripts, which is why make/perl are the only build/host dependencies. The compilers are listed in the run section because they are used at runtime to analyze header files.

I'm a bit unsure about whether I got the runtime dependencies right because the documentation is inconsistent, so I just made a best guess.

Checklist
- [x] Build number is 0.


carterbox:
Not available for OSX because of missing dependencies.
```
The reported errors are:
-   - nothing provides requested vtable-dumper >=1.1
-   - nothing provides requested elfutils
```

carterbox:

recipes/abi-compliance-checker/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  url: https://github.com/lvc/abi-compliance-checker/archive/refs/tags/{{ version }}.tar.gz
+  sha256: b1e32a484211ec05d7f265ab4d2c1c52dcdb610708cb3f74d8aaeb7fe9685d64
+
+build:
+  number: 0
+  skip: true  # [not linux]
+  script:
+    - sed -i "s|#!/usr/bin/perl|#!$PREFIX/bin/perl|" abi-compliance-checker.pl
```
beckermr, line 16: What does conda do this line at runtime? I am wondering if it leaves the string $PREFIX or actually replaces it? You may need jinja2 here (i.e. `${{ PREFIX }}`) so that conda will rewrite the prefix at runtime?

I have the same question below as well.

beckermr (changes requested):
One question on the #! line. I might be wrong but I don't understand.

recipes/abi-compliance-checker/meta.yaml
carterbox: If you download the build artifacts, and open this file with a text editor, $PREFIX has been replaced with /home/conda/staged-recipes/build_artifacts/abi-compliance-checker_1678293645399/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol.

I believe $PREFIX it is evaluated by `sed` in the build script when it does the replacement. Not by bash when the script itself is run. Later, at install time, conda replaces the placeholder with the end users CONDA_PREFIX.

carterbox:

carterbox:
Thanks, @beckermr !

PR #22280, opened 2023-03-13
add SystemC

timsnyder:
Checklist
   - It does vendor a subset of boost header only libs into the `include/sysc/packages/boost` namespace. License file included
   - Vendors QuickThreads headers into `include/sysc/packages/qt` and it's `copyright.h` is also included as a license file.
   - I don't see any static libraries being linked. (Visually inspected linux build)
   -  ~~static library is split into `libsystemc-static` subpackage~~ Actually, after I switched to using `cmake` on unix, the static library is no longer built and I'm just going to not ship it for now.  If someone needs it in the future, I'll add a separate `libsystemc-static` subpackage.
   - On Windows, the static library is shipped because `The compilation of SystemC as a DLL on Windows is currently not supported!`
- [x] Build number is 0.


timsnyder:
I unchecked the "this package doesn't vendor other packages" checkbox because I just noticed the `src/sysc/packages` subdirectory... I guess I didn't look hard for vendored stuff.

recipes/systemc/build.sh
```diff
@@ -0,0 +1,33 @@
...
+# Just running configure with 2.3.4, it complains about not finding Makefile.in so
+# rerun autotools to generate the automake outputs
+autoreconf --install --force
+
+mkdir build
+cd build
+../configure \
+    --prefix="$PREFIX" \
+    --with-unix-layout \
+    --with-pthreads \
```
timsnyder, line 16: `--with-pthreads` doesn't make sense on OS-X.  Need to make this linux-specific

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/accellera-official/systemc/archive/refs/tags/{{ version }}.tar.gz
+  sha256: bfb309485a8ad35a08ee78827d1647a451ec5455767b25136e74522a6f41e0ea
+  patches:
+    - patches/0001-examples-sysc-2.3-simple_async-needs-PTHREADS.patch  # https://github.com/accellera-official/systemc/pull/44
```
timsnyder, line 12: This patch needs to be linux-only

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,117 @@
...
+    the intellectual property (IP), electronic design automation (EDA), semiconductor, electronic
+    systems, and embedded software industries currently use SystemC for architectural exploration,
+    to deliver high-performance hardware blocks at various levels of abstraction and to develop
+    virtual platforms for hardware/software co-design. SystemC has been standardized by the Open
+    SystemC Initiative (OSCI) and Accellera Systems Initiative and ratified as IEEE Std. 1666(TM)-2011.
+  license: Apache-2.0
+  license_family: Apache
+  license_file:
+    - LICENSE
+    - NOTICE
```
timsnyder, line 111: I think https://github.com/accellera-official/systemc/blob/39740075f47fedbce242808d357fcfb6d3d33957/src/sysc/packages/boost/LICENSE_1_0.txt#L13-L15 means that we don't _have_ to include the boost license since they are only using some of the header-only libraries.  But, I'm not a lawyer and I'm going to err on the side of including more than less:
```suggestion
    - NOTICE
    - src/sysc/packages/boost/LICENSE_1_0.txt
```

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,118 @@
...
+          - include/systemc*                                     # [not win]
+          - include/tlm*                                         # [not win]
+          - include/tlm_core                                     # [not win]
+          - include/tlm_utils                                    # [not win]
+          - etc/conda/*activate.d/*activate-systemc.sh           # [not win]
+      test:                                                      # [not win]
+        requires:                                                # [not win]
+          - {{ compiler('cxx') }}                                # [not win]
+          - {{ pin_subpackage('systemc-doc', exact=True) }}      # [not win]
+          - make                                                 # [not win]
```
timsnyder, line 69:
```suggestion
          - make                                                 # [not win]
          - sed                                                    # [not win]
```
also install cf GNU sed because the sed on os-x is wonk

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+          - sed                                                  # [not win]
+        commands:                                                # [not win]
+          - test "$SYSTEMC_HOME" == "$PREFIX"                    # [not win]
+          - cp -r "$PREFIX"/share/doc/systemc/examples .         # [not win]
+          # patch Makefile.rules because we want to use the installed libsystemc  
+          # AND we want to use the dynamic lib which the linker should be able to find
+          - "sed -i 's|$(EXE): $(OBJS) $(SYSTEMC_LIB_DIR)/libsystemc.a|$(EXE): $(OBJS)|' examples/build-unix/Makefile.rules"  # [not win]
+          - make -C examples/build-unix -j $CPU_COUNT run        # [not win]
+          - make -C examples/build-unix -j $CPU_COUNT check      # [not win]
```
timsnyder, line 78:
```suggestion
          - make -C examples/build-unix -j $CPU_COUNT run        # [linux]
          - make -C examples/build-unix -j $CPU_COUNT check      # [linux]
          - make -C examples/build-unix -j $CPU_COUNT run LDFLAG_RPATH=-Wl,-rpath,        # [osx]
          - make -C examples/build-unix -j $CPU_COUNT check LDFLAG_RPATH=-Wl,-rpath,     # [osx]
```

`examples/build-unix/Makefile.rules` uses GNU syntax for the rpath option that doesn't work with the macos linker.  Need to use a comma instead of `=`.

timsnyder:

timsnyder:
@conda-forge/help-c-cpp @conda-forge/staged-recipes, can I please get a review of this recipe?

timsnyder:
@timkpaine do you need SystemC for your work or are you pure verilog/SV?

timkpaine:
im pure sv but happy to be a maintainer here if you want someone else on the feedstock

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,145 @@
...
+    - LICENSE
+    - NOTICE
+    - src/sysc/packages/boost/LICENSE_1_0.txt
+    - src/sysc/packages/qt/copyright.h
+  doc_url: https://ieeexplore.ieee.org/document/6134619
+  dev_url: https://github.com/accellera-official/systemc
+
+extra:
+  recipe-maintainers:
+    - timsnyder
```
timsnyder, line 145:
```suggestion
    - timkpaine
    - timsnyder
```

recipes/systemc/meta.yaml
carterbox: Use a selector?

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,146 @@
...
+    - nmake /nologo CONFIG=Release PLATFORM=x64 VCBUILD=vcbuild check  # [win64]
+    - nmake /nologo CONFIG=Release PLATFORM=Win32 VCBUILD=vcbuild check  # [win32]
+
+# split unix packages into:
+# libsystemc dynamic-library
+# systemc-dev package depends on libsystemc and adds header and unversioned .so symlink
+# systemc-static just the static library
+# systemc-doc  documentation including the installed examples
+
+outputs:                                                         # [not win]
```
carterbox, line 46: Thanks for your patience. Why not split the Windows package too?

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,146 @@
...
+          - cp -r "$PREFIX"/share/doc/systemc/examples .         # [not win]
+          # patch Makefile.rules because we want to use the installed libsystemc  
+          # AND we want to use the dynamic lib which the linker should be able to find
+          - "sed -i 's|$(EXE): $(OBJS) $(SYSTEMC_LIB_DIR)/libsystemc.a|$(EXE): $(OBJS)|' examples/build-unix/Makefile.rules"  # [not win]
+          - make -C examples/build-unix -j $CPU_COUNT run        # [linux]
+          - make -C examples/build-unix -j $CPU_COUNT check      # [linux]
+          # osx ld64 doesn't support '-rpath=' must be '-rpath,'
+          - make -C examples/build-unix -j $CPU_COUNT run LDFLAG_RPATH=-Wl,-rpath,    # [osx]
+          - make -C examples/build-unix -j $CPU_COUNT check LDFLAG_RPATH=-Wl,-rpath,  # [osx]
+    - name: systemc                                              # [not win]
```
carterbox, line 112: Each output that could possibly be a direct build dependency for downstream packages needs to have a run export for a compatible version of libsystemc. This includes, systemc, systemc-dev, and libsystemc.

carterbox (changes requested):

recipes/systemc/meta.yaml
timsnyder: System-C doesn't support shared-library builds on Windows and also doesn't include the documentation (need to double check that).  I put a comment about the shared library support not working in https://github.com/conda-forge/staged-recipes/pull/22280/files#diff-b8d2b920295573b173d5ed931eae74bca220fad3f892fc99d644f20d84debd9aR9-R11 Perhaps I should also explain that here.

Without shared library or docs, splitting the package didn't make sense to me (although I agree that it is ugly).

carterbox: Ideally we want consistency so that downstream packages don't need to use platform selectors to build correctly against this package. Maybe I'm wrong, but it seems like unix packages would build against "systemc-dev", but windows would build against "systemc"?

timsnyder: That's a very good point.  I think in this case, it may make sense for the windows package to ship `systemc-dev` that downstreams would build against and possibly a `systemc` metapackage that on windows only depends on `systemc-dev`?

carterbox: I think the outputs for unix and windows should be according to this chart

systemc        both
systemc-dev    both
systemc-doc    both
libsystemc     unix-only because static links for windows

timsnyder:
This fell off my radar while I have been switching jobs this year.  I'm getting back to it and will make the updates requested by Daniel soon.

timsnyder:
I've just hit the first reason I need this at work and I should be able to allocate some higher BW priority to wrapping it up.

timsnyder:

timsnyder:
I want to download the windows package so I can update the split-package to also handle windows.

timsnyder:
@carterbox , this is ready for a re-review.  I just need to get the tests passing again on Windows (see below for more details on that).

I've split the windows package similarly to how it was split on Unix.  As you suggested, I've got the following packages:
systemc both
systemc-dev both
systemc-doc both
libsystemc unix-only because static links for windows


I also added run-exports for libsystemc from systemc-dev, libsystemc and systemc metapackage on Unix.

I have the examples included in systemc-doc package and I require that package to test systemc-dev.  This works fine on Unix but for some reason in the Windows Variant systemc-dev packages before systemc-doc and the testing fails because conda can't find the systemc-doc package.  I don't see why conda-build isn't packaging the packages in the order they are in the meta.yaml on Windows.  I'm probably missing something really dumb.

I may go back to cramming the examples into the test files of systemc-dev on Windows like I was before, just to get past this issue.  I'm going to ask about it in the gitter channel first though.

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,280 @@
...
+          - LICENSE
+          - NOTICE
+          - src/sysc/packages/boost/LICENSE_1_0.txt
+          - src/sysc/packages/qt/copyright.h
+        doc_url: https://ieeexplore.ieee.org/document/6134619
+        dev_url: https://github.com/accellera-official/systemc
+    - name: systemc-dev
+      build:                                                     # [not win]
+          run_exports:                                           # [not win]
+              - {{ pin_subpackage('libsystemc', exact=True) }}   # [not win]
```
carterbox, line 128: A max_pin of 'x.x.x' or 'x.x' should be sufficient, no? There aren't build variants, and the ABI shouldn't be changing between build numbers of this package.

```suggestion
              - {{ pin_subpackage('libsystemc', max_pin='x.x.x') }}   # [not win]
```

carterbox (changes requested):
See my comment about export pinnings. Otherwise, looks good!

recipes/systemc/meta.yaml
timsnyder: Did this in [4b65a1c](https://github.com/conda-forge/staged-recipes/pull/22280/commits/4b65a1cfad89ff47dc9544fff4008167d3dbeda2)

carterbox:
> This works fine on Unix but for some reason in the Windows Variant systemc-dev packages before systemc-doc and the testing fails because conda can't find the systemc-doc package. I don't see why conda-build isn't packaging the packages in the order they are in the meta.yaml on Windows. I'm probably missing something really dumb.

I think the builder only considers dependency relations when determining the build order, so since the dev package doesn't depend on the doc package. It isn't built first. You could add the doc package to the host requirements of the dev package which would force them to be build in order without affecting the run requirements.

timsnyder:
Ahh.  Ok.  I'll try adding it to the host requirements to make the dependency honored.

recipes/systemc/meta.yaml
timsnyder: Yes.  I'll fix this too.

recipes/systemc/meta.yaml
```diff
@@ -0,0 +1,285 @@
...
+      requirements:
+          build:
+              - {{ compiler('cxx') }}
+          run:                                                   # [not win]
+              - {{ pin_subpackage('libsystemc', exact=True) }}   # [not win]
```
carterbox, line 133:
```suggestion
      requirements:
          build:
              - {{ compiler('cxx') }}
          host:
              - {{ pin_subpackage('systemc-doc', exact=True) }}
          run:                                                   # [not win]
              - {{ pin_subpackage('libsystemc', exact=True) }}   # [not win]
```
Add dependency here, so systemc-doc MUST be built before systemc-dev in order to systemc-doc to be available for tests.

timsnyder: Thanks.  I did this in [83245d6](https://github.com/conda-forge/staged-recipes/pull/22280/commits/83245d65a2b88e3297808a22c9a1d44daa176e66)

recipes/systemc/meta.yaml
timsnyder: Done in 
a896e1d

timsnyder:
@carterbox, I finally have this straightened out with the tests passing again.  The details of CMD scripting are not my forte.

carterbox:
> The details of CMD scripting are not my forte.

Same. 😐

PR #22313, opened 2023-03-17
Added grep

calebzulawski:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/grep/meta.yaml
+++ b/recipes/grep/meta.yaml
@@ -27,7 +27,7 @@ test:
 
 about:
   home: https://www.gnu.org/software/grep/
-  license: GPL-3.0
+  license: GPL-3.0-or-later
   license_file: COPYING
   summary: |
     Grep searches one or more input files for lines containing a match to a
--- a/recipes/grep/meta.yaml
+++ b/recipes/grep/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "grep" %}
+{% set version = "3.9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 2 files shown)


recipes/grep/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  build:
+    - {{ compiler('c') }}
+    - make
+  host:
+    - pcre2
+
+test:
+  commands:
+    - ${CONDA_PREFIX}/bin/grep --version
+    - echo hello world | ${CONDA_PREFIX}/bin/grep hello
```
hmaarrfk, line 26:
```suggestion
    - test -x ${PREFIX}/bin/grep
    - if [ $(which grep) != ${PREFIX}/bin/grep ]; then echo 1; fi
    - grep --version
    - echo hello world | grep hello
```

one would use `${PREFIX}` in recipes, and `${CONDA_PREFIX}` during other usage.

hmaarrfk:
@conda-forge/core do we want this package at all? seems like it is more updated than the version i get from ubuntu 22.10 (3.7)

recipes/grep/meta.yaml
calebzulawski: thanks, makes sense!

calebzulawski:
In my particular use case, I am managing the entire environment with conda and am not using system packages (including macOS, where the alternative to acquire GNU grep would be something like homebrew).  I just picked 3.9 because it's the latest--ubuntu 23.04 seems to have 3.8, so I expect 3.9 to be available somewhat soon.  If this version is considered too new, I can use an older version.

hmaarrfk:
thanks Isuru!

PR #22322, opened 2023-03-18
add amaranth / nmigen

timkpaine:
Checklist
- [x] Build number is 0.


timkpaine:
package is published to pypi but only wheel, so using GH link

timkpaine:
@conda-forge/help-python all set for review, note that due to historical reasons, amaranth is basically 2 packages in one (`amaranth` and `nmigen`), and (at least at the present) `amaranth==0.3` is basically the successor to `nmigen==0.2`, so these do properly belong as a single feedstock rather than as 2 separate feedstocks (at least at the time of writing this).

recipes/amaranth/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - nmigen-rpc = nmigen.rpc:main
+requirements:
+  host:
+    - python >=3.7
+    - pip
+    - setuptools
+    - setuptools_scm
+  run:
+    - python >=3.7
+    - importlib_resources
```
dopplershift, line 33: Upstream also has:
```suggestion
    - importlib_metadata
```

timkpaine: I don't think this is still needed 3.8+?

dopplershift: But the recipe says `python >=3.7` 😉

timkpaine: conda-forge is only building 3.8+ these days though, im cool with either, I always forget I can't use selectors: https://github.com/conda-forge/staged-recipes/pull/22322/commits/27286641b46c6fe1ce84eb92020668b93eacf820

dopplershift: Well, it's noarch so it will only build one package that relies on the version ranges to exclude it from environments. Right now, this recipe is installable on Python 3.7, but would need `importlib_metadata` to avoid an error..

timkpaine: ok I can add

dopplershift (approved):

PR #22323, opened 2023-03-18
add cocotb

timkpaine:
Checklist
- [x] Build number is 0.

—-

Closes https://github.com/conda-forge/staged-recipes/pull/21056

### DIFF
--- a/recipes/cocotb/meta.yaml
+++ b/recipes/cocotb/meta.yaml
@@ -47,6 +47,7 @@ test:
 
 about:
   home: https://www.cocotb.org
+  dev_url: https://github.com/cocotb/cocotb
   summary: cocotb is a coroutine based cosimulation library for writing VHDL and Verilog testbenches in Python.
   license: BSD-3-Clause
   license_family: BSD 
--- a/recipes/cocotb/meta.yaml
+++ b/recipes/cocotb/meta.yaml
@@ -19,7 +19,7 @@ build:
   run_exports:
     - {{ pin_subpackage("cocotb", max_pin="x") }}
   script:
-    - {{ PYTHON }} -m pip install -U --no-deps . -vvv
+    - {{ PYTHON }} -m pip install --no-deps . -vvv
 
 requirements:
   build:
--- a/recipes/cocotb/meta.yaml
+++ b/recipes/cocotb/meta.yaml
@@ -1,0 +1,64 @@
+{% set name = "cocotb" %}
+{% set version = "1.7.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cocotb-{{ version }}.tar.gz
+  sha256: 959892eb94bd0b3ff40e0fca51d33a3936416deb853e2bac4f7f766b40002650 
+  patches:
+    - patches/0001-remove-static-stdc.patch
+
+build:
+  number: 0
+  skip: true  # [win or py<310]
+  entry_points:
+    - cocotb-config = cocotb.config:main
+  run_exports:
+    - {{ pin_subpackage("cocotb", max_pin="x") }}
+  script:
+    - {{ PYTHON }} -m pip install -U --no-deps . -vvv
+
+requirements:
+  build:
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - {{ compiler('cxx') }}
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - find-libpython
+  run_constrained:
+    - iverilog >=10.3
+    - verilator >=5.004
+
+test:
+  imports:
+    - cocotb
+  commands:
+    - pip check
+    - cocotb-config --help
+  requires:
+    - pip
+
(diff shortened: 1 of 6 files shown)


timkpaine:
need to pull the 3 missing licenses of vendored files, 2 of which come from IEEE language reference manuals and 1 is apache 2.0 (but i need to get the license text).

timkpaine:
@conda-forge/help-python-c need some help diagnosing behavior, it looks like the first build works fine but then subsequent versions end up with errors like:

```
ERROR (cocotb,lib/python3.9/site-packages/cocotb/libs/<all the libs>): lib/python3.8/site-packages/cocotb/libs/libgpi.so not found in any packages
```

e.g. during the build for `3.9`, its looking for libs from `3.8`. 

on OS X its similar, except it builds `3.10` before `3.9`

```
  ERROR (cocotb,lib/python3.9/site-packages/cocotb/libs/libpygpilog.so): lib/python3.10/site-packages/cocotb/libs/libgpilog.so not found in any packages
```

recipes/cocotb/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+pip install -U --no-deps . -vvv --no-build-isolation
```
dhirschfeld, line 3:
```suggestion
$PYTHON -m pip install -U --no-deps . -vvv --no-build-isolation
```

timkpaine: I can try but it didn't do anything before

timsnyder:
I started trying to look at this because it's really weird.  It has been a really long time since I looked at the overlinking checks code but I'm starting to think this is definitely an issue with how conda-forge's `build_all.py` is building multiple versions and state in the overlinking checks is bleeding from one build to the next.  I don't know if I'll find time to look at this any more but there may be something really broken here in the infrastructure.  I think restricting it to a single python until you're in the feedstock and the builds are isolated is a good idea.

timkpaine:
@conda-forge/staged-recipes @conda-forge/core fyi

timkpaine:
@isuruf / @conda-forge/staged-recipes i opened https://github.com/conda/conda-build/pull/4821 but seems it will be a wait, since this issue only effects in `staged-recipes` can we merge this and get the feedstock rolled out?

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+build:
+  number: 0
+  skip: true  # [win or py<310]
+  entry_points:
+    - cocotb-config = cocotb.config:main
+  run_exports:
+    - {{ pin_subpackage("cocotb", max_pin="x") }}
+  script:
+    - {{ PYTHON }} -m pip install -U --no-deps . -vvv
```
carterbox, line 22:
```suggestion
    - {{ PYTHON }} -m pip install --no-deps . -vvv
```
Why do you specify `--upgrade`?

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  patches:
+    - patches/0001-remove-static-stdc.patch
+
+build:
+  number: 0
+  skip: true  # [win or py<310]
+  entry_points:
+    - cocotb-config = cocotb.config:main
+  run_exports:
+    - {{ pin_subpackage("cocotb", max_pin="x") }}
```
carterbox, line 20: This run_export tells me that your shared libraries are not python version specific because you haven't included a python version in the pinning.

Will your package still run if you move the shared libs from site-packages/cocotb/libs to $PREFIX/lib? This would allow you to build the shared libraries once per platform and use a `noarch: python` package for the python bits (which would reduce the build matrix to one build variant and allow users of your C API to avoid installing python).

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  run:
+    - python
+    - find-libpython
+  run_constrained:
+    - iverilog >=10.3
+    - verilator >=5.004
+
+test:
+  imports:
+    - cocotb
```
carterbox, line 41: You are also installing a python module called "pygpi", but that module is not tested here. Is the missing test a mistake or is the install a mistake?

recipes/cocotb/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  commands:
+    - pip check
+    - cocotb-config --help
+  requires:
+    - pip
+
+about:
+  home: https://www.cocotb.org
```
carterbox, line 49:
```suggestion
  home: https://www.cocotb.org
  dev_url: https://github.com/cocotb/cocotb
```

carterbox (changes requested):

recipes/cocotb/meta.yaml
timkpaine: they are, but it wont. This package is a bit complicated due to the way it is used
https://github.com/conda-forge/staged-recipes/pull/21056#issuecomment-1309329721
https://github.com/cocotb/cocotb/pull/2268
https://github.com/ktbarrett/find_libpython

recipes/cocotb/meta.yaml
timkpaine: it comes with `cocotb`, I can add a simple import test?

PR #22345, opened 2023-03-20
Add Kokkos-kernels recipe.

vincentmr:
Checklist
- [x] Build number is 0.


vincentmr:

.azure-pipelines/azure-pipelines-linux.yml
```diff
@@ -30,115 +30,3 @@ jobs:
 
   - publish: build_artifacts/noarch/
     artifact: conda_pkgs_noarch
-
```
dopplershift, line 33: You should not be making changes to this file.

vincentmr: Hi @dopplershift , if I don't want to build any CUDA backend, should I revert the changes to this file and set some condition in my recipe?

dopplershift: I'm not an expert on the compiled side here, but IIRC the CUDA builds are opt-in. The CI jobs that ran look like the standard ones to me. @conda-forge/help-c-cpp anything to add?

vincentmr: I had the impression it was opt-out, but I think you're right. I've just reverted the changes.

BastianZim:
That looks good to me, thank you!

@leofang Would you mind having a look as well? It's a dependency for Pennylane.

recipes/kokkos-kernels/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+test:
+  commands:
+  # Lots more CMake files; just checking for one
+    - test -f $PREFIX/lib/cmake/KokkosKernels/KokkosKernelsConfig.cmake  # [unix]
+  # Testing explicitly for all shared libs, but not links
+    - test -f $PREFIX/lib/libkokkoskernels.so  # [linux]
+    - test -f $PREFIX/lib/libkokkoskernels.dylib  # [osx]
+  # There are lots more headers; only test for one.
+    - test -f $PREFIX/include/KokkosKernels_Handle.hpp  # [unix]
+  # There are no binary
```
leofang, line 39: It seems to me one can compile some dummy code against `kokkos-kernels`. How hard is it to come up with one? If we can do a simple compilation test here, it'd be great. You'd be surprised how many issues a naive test like this could catch 🙂

leofang (commented):
Sorry I missed the ping. LGTM overall. Left a few perhaps naive questions, since I don't know Kokkos well enough.

recipes/kokkos-kernels/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+source:
+  url: https://github.com/kokkos/kokkos-kernels/archive/refs/tags/{{ version }}.tar.gz
+  sha256: b2060f5894bdaf7f7d4793b90444fac260460cfa80595afcbcb955518864b446
+
+build:
+  skip: true  # [win]
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('kokkos-kernels') }}
```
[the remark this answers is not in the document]
vincentmr: I'm not 100% sure, but I would say yes.

recipes/kokkos-kernels/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+requirements:
+  build:
+    - cmake >=3.21.1,<4
+    - {{ compiler('cxx') }}
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
+    - ninja
+  host:
+    - kokkos
```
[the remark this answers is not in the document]
vincentmr: I guess an analogy might be:

- CUDA => Kokkos
- cuBLAS, cuSPARSE => Kokkos-kernels

There are other Kokkos extensions, like simd-math. As far as I know, they are all tightly coupled and for some the goal is to eventually merge them into the main Kokkos library.

recipes/kokkos-kernels/meta.yaml
vincentmr: I guess we could try building the examples and tests, would that work?

vincentmr: There is an error when compiling the tests, which I have reported [here](https://github.com/kokkos/kokkos-kernels/issues/1791).

leofang (approved):

recipes/kokkos-kernels/meta.yaml
vincentmr: Hi @leofang , what's the best way to get that file if the file is in `kokkos-kernels/example/batched_solve`. Should I copy from `$SRC_DIR/../work`?

vincentmr: Found my answer to the previous questions, but my test fails because the c++ compiler is not found in the test phase. Where does one find g++ for a test?

recipes/kokkos-kernels/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  run:
+    - kokkos
+
+test:
+  source_files:
+    - example/batched_solve/team_GMRES.cpp
+  commands:
+    - gxx team_GMRES.cpp -o team_GMRES -I$PREFIX/include -L$PREFIX/lib -lkokkoscontainers -lkokkoscore -lkokkoskernels -fopenmp
```
leofang, line 34:
```suggestion
    - g++ team_GMRES.cpp -o team_GMRES -I$PREFIX/include -L$PREFIX/lib -lkokkoscontainers -lkokkoscore -lkokkoskernels -fopenmp
```

recipes/kokkos-kernels/meta.yaml
leofang: Sorry for late reply, Vincent, have a long day... I think we just need to switch to `g++`.

recipes/kokkos-kernels/meta.yaml
leofang: better:
```suggestion
    - $CXX team_GMRES.cpp -o team_GMRES -I$PREFIX/include -L$PREFIX/lib -lkokkoscontainers -lkokkoscore -lkokkoskernels -fopenmp
```

recipes/kokkos-kernels/meta.yaml
vincentmr: For sure, no worries. I think I've tried that, but let me try again.

leofang: Yes you did it in [153418a](https://github.com/conda-forge/staged-recipes/pull/22345/commits/153418a048e8bdfb997ae3bc67e81472a521d8fc), but you didn't add the compiler until [1a59fdc](https://github.com/conda-forge/staged-recipes/pull/22345/commits/1a59fdc5b276be5c9f9b2a3fd2bf73260704bb5d) 🙂

leofang: linux-64 passed, great work Vincent!

vincentmr: Thanks for your feedback ;)

recipes/kokkos-kernels/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  host:
+    - kokkos
+  run:
+    - kokkos
+
+test:
+  source_files:
+    - example/batched_solve
+  commands:
+    - $CXX example/batched_solve/team_GMRES.cpp -o team_GMRES -I$PREFIX/include -L$PREFIX/lib -lkokkoscontainers -lkokkoscore -lkokkoskernels -fopenmp
```
leofang, line 34: We may need to use C++17. It fails on macOS.
```suggestion
    - $CXX example/batched_solve/team_GMRES.cpp -o team_GMRES -std=c++17 -I$PREFIX/include -L$PREFIX/lib -lkokkoscontainers -lkokkoscore -lkokkoskernels -fopenmp
```

recipes/kokkos-kernels/meta.yaml
leofang: All passed!

leofang (approved):

leofang:
@BastianZim This is ready.

BastianZim:
Thanks everyone!

recipes/kokkos-kernels/meta.yaml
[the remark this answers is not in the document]
vincentmr: Hi Leo, I think I misunderstood you question here. I had in mind to match the versions of Kokkos and Kokkos-kernels, but I realize this has nothing to do with it. So I guess we might remove that, but I'm also wondering what's the best way to lock these together. Just 
```
  host:
    - kokkos =={{ version }}
```
?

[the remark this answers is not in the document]
leofang: @vincentmr I see. My question was "for those who depend on Kokkos Kernels, would they pin kokkos-kernels tightly that they use the same version at run time as at build time?" which is what this `run_exports` is about.

Your concern was that "Kokkos Kernels depends on Kokkos, will they be version-matched?" The answer is currently no. The [`run_exports` of Kokkos](https://github.com/conda-forge/kokkos-feedstock/blob/ba2a1102b1b9e99adc5afb1a62e4b93789822d81/recipe/meta.yaml#L19-L20) only guarantees that the run time and build time `kokkos` is the same version, but it's not necessary the same as the `kokkos-kernels` version. We should fix this in the feedstock.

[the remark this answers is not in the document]
leofang: In particular, note that kokkos v4.0.0 just came out. It seems `kokkos-kernels` v3.7.01 should be rebuilt, and the build number 0 should be marked as broken (since they depend on `kokkos` v4.0.0).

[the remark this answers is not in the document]
leofang: > It seems `kokkos-kernels` v3.7.01 should be rebuilt, 

Ah, I just saw that you already did it: https://github.com/conda-forge/kokkos-kernels-feedstock/pull/1, thanks!

> and the build number 0 should be marked as broken (since they depend on `kokkos` v4.0.0).

This needs to be done.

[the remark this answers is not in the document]
vincentmr: > In particular, note that kokkos v4.0.0 just came out a few weeks ago. It seems `kokkos-kernels` v3.7.01 should be rebuilt, and the build number 0 should be marked as broken (since they depend on `kokkos` v4.0.0).

I thought iterating the build number would forcefully deprecate lower build numbers. Ok, so let me try to do that. And thx for the feedback.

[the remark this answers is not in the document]
leofang: > I thought iterating the build number would forcefully deprecate lower build numbers.

Usually it's the case. But I've seen some complicated situations where the solver thinks it's better to pick the ones with a lower build number. Plus, in this case it's literally broken; we're not supposed to use Kokkos 4 to build Kokkos Kernels 3, unless there's strong compatibility guarantee from the Kokkos team.

PR #22368, opened 2023-03-23
Add chemnlp to conda.

knc6:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/chemnlp/chemnlp/meta.yaml
+++ b/recipes/chemnlp/chemnlp/meta.yaml
@@ -1,0 +1,55 @@
+{% set name = "chemnlp" %}
+{% set version = "2023.3.4" %}
+
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/chemnlp-{{ version }}.tar.gz
+  sha256: b2709f8eb6c0d167aef1ad9c6dc345b73fe3023797e1fc71174de01a5c3a2398 
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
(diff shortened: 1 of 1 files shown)


recipes/alignn/meta.yaml
dopplershift, line 1: Please don't modify other recipe files, this can cause problems.

knc6: Hi, [chemnlp ](https://github.com/usnistgov/chemnlp) and [alignn ](https://github.com/usnistgov/alignn) are two different repos.
I want to add only chemnlp to conda-froge and not alignn at the moment.
I thought we need to keep example recipe only according to docs.

dopplershift: It's confusing because the diff in this PR shows the file being deleted, which should indicate removal of a file that's present in the upstream repository (i.e. conda-forge/staged-recipes). Looks like that file was only present (and removed) in #16794 (by you).

Regardless, it's not present now so I guess we can ignore the diff here.

recipes/chemnlp/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  host:
+    - pip
+    - python >=3.7
+  run:
+    - numpy >=1.22.0
+    - scipy >=1.6.3
+    - jarvis-tools >=2021.07.19
+    - chemdataextractor >=1.3.0
+    - matplotlib-base >=3.4.1
+    - pandas ==1.2.4
```
dopplershift, line 30: Is this tight pinning of pandas necessary? I don't see it in upstream's `setup.py`?

knc6: Just modified the file.

dopplershift (approved):

PR #12093, opened 2020-07-08
Adding Kaleido

jonmmease:
Checklist
- [x] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml"
- [x] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/master/recipes/example/meta.yaml#L57-L66) for an example)
- [x] Source is from official source
- [x] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged)
- [x] If static libraries are linked in, the license of the static library is packaged.
- [x] Build number is 0
- [x] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details)
- [x] GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there

PR #22418, opened 2023-03-29
Add R package chromote

jdblischak:
CRAN package chromote https://cran.r-project.org/package=chromote

It requires a chromium-based browser to function. I couldn't find an obvious option from conda-forge, and given the complexity of building a headless chrome driver (eg #12093), I'm inclined to just let the user decide how to best install a web browser

> SystemRequirements: Google Chrome or other Chromium-based browser. chromium: chromium (rpm) or chromium-browser (deb)

---

Checklist
- [x] Build number is 0.


jdblischak:
@conda-forge/help-r Ready for review

jdblischak:
> Recipe looks good - thanks for adding!

@mfansler Thanks for the quick review!

> I don't object to the omission, though do prefer when the Conda Forge build brings the CRAN-declared SystemRequirements, since that's what makes Conda awesome 😉

I 100% agree that conda is awesome. I mostly agree about installing system requirements. For system dependencies like libxml2 that are required during build time, I agree that conda should always install them (well, er, has to). On the other hand, I go back and forth on the decision to install runtime-only dependencies (eg pandoc for rmarkdown). This is convenient but is also less flexible for power users (eg what if they want to use the system pandoc or a self-installed version). The Debian/Ubuntu binaries only install buildtime dependencies like libxml2, and it's trivial enough to simply add `pandoc` to the list of software to install. I wish there was a way to specify these preferences during install, kind of like how you can do with pip, eg hypothetically `pip install pkg[minimal]` vs `pip install pkg[full]`. I guess we could create a minimal output in the recipe that contains everything but pandoc, and power users could then `conda install rmarkdown-minimal`.

But until someone complains loudly enough, it's not really a priority.

> If it stays omitted, perhaps it would make sense to at least have a test for compatibility with a system level chromium? E.g., since it's noarch, add [a yum_requirements.txt](https://conda-forge.org/docs/maintainer/knowledge_base.html#yum-deps) that installs chromium in the Linux system then add some sort of integration test to check that the build can interop with that. Does that sound reasonable?

Yes, I think this is a great idea. Though I'm not sure how to install chromium. Both the `SystemRequirements` field and this [help article](https://wpcademy.com/how-to-install-chromium-on-centos-7/) I consulted say that the rpm package name is "chromium". But the build fails to find any software with this name:

```sh
No package chromium available.
Error: Nothing to do
```

jdblischak:
I confirmed that the package "chromium" can't be installed in the Docker container:

```sh
docker run -it --rm condaforge/linux-anvil-cos7-x86_64

[conda@d2146f3acf89 ~]$ /usr/bin/sudo -n yum install -y chromium

Loaded plugins: fastestmirror, ovl
Determining fastest mirrors
 * base: mirror.cogentco.com
 * extras: mirrors.tripadvisor.com
 * updates: nyc.mirrors.clouvider.net
base                                                                                             | 3.6 kB  00:00:00
http://mirrors.tripadvisor.com/centos/7.9.2009/extras/x86_64/repodata/repomd.xml: [Errno 12] Timeout on http://mirrors.tripadvisor.com/centos/7.9.2009/extras/x86_64/repodata/repomd.xml: (28, 'Operation too slow. Less than 1000 bytes/sec transferred the last 30 seconds')
Trying other mirror.
extras                                                                                           | 2.9 kB  00:00:00
updates                                                                                          | 2.9 kB  00:00:00
(1/4): base/7/x86_64/group_gz                                                                    | 153 kB  00:00:00
(2/4): extras/7/x86_64/primary_db                                                                | 249 kB  00:00:00
(3/4): base/7/x86_64/primary_db                                                                  | 6.1 MB  00:00:02
(4/4): updates/7/x86_64/primary_db                                                               |  20 MB  00:00:05
No package chromium available.
Error: Nothing to do
```

mfansler:
Going on your example, my impression is the idiom for Conda is more to have a `pandoc=*=external` version that Conda can use as an indicator that the user has an *external* version to satisfy that requirement. We don't need an alternative `r-rmarkdown`, but rather an empty version of the dependency.

I.e., if we had a functional `chromium-headless`, we could also have an "external" build that let's the power user BYOC.

Anyway, seems like the testing idea is a no go - I'll re-review once the CI passes.

mfansler:
> > my impression is the idiom for Conda is more to have a pandoc=*=external version that Conda can use as an indicator that the user has an external version to satisfy that requirement. We don't need an alternative r-rmarkdown, but rather an empty version of the dependency.
> 
> 
> 
> Are you saying this is a hypothetical workaround or do you know of examples of conda recipes that use this approach? I've never seen this before, and I'm having trouble wrapping my head around how this would work in practice.


It's not in widespread use, but [Conda Forge uses this idiom](https://conda-forge.org/docs/user/tipsandtricks.html#using-external-message-passing-interface-mpi-libraries) for the `mpich` package. [I've tested it and recommended it](https://stackoverflow.com/a/71302552/570918) for users wanting to use an external `clang`, and seemed to work. But still kind of anecdotal.

jdblischak:
@mfansler This is ready for a final review/merge. I left the test script there for future troubleshooting. Please let me know if you'd prefer that I remove it.

Also, I tested it locally on Ubuntu to confirm that it runs correctly when a suitable chromium browser is available:

```sh
# Ubuntu 22.04
sudo apt install chromium-browser

Rscript recipes/r-chromote/test-chromium.R
## Path to browser executable: /usr/bin/chromium-browser
## Version info:
## List of 5
##  $ protocolVersion: chr "1.3"
##  $ product        : chr "HeadlessChrome/111.0.5563.110"
##  $ revision       : chr "@3da375c9e9e5ad9b556af5df708a6e6c01d8cfb1"
##  $ userAgent      : chr "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) HeadlessChrome/111.0.5563.110 Safari/537.36"
##  $ jsVersion      : chr "11.1.277.17"

dbast (approved):

PR #22456, opened 2023-03-31
Adding conda-project

AlbertDeFusco:
Checklist
- [x] Build number is 0.


recipes/conda-project/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    dependencies, specifications for how to run your code, and any other files that it may need.
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE
+  doc_url: https://conda-incubator.github.io/conda-project/user_guide.html
+  dev_url: https://github.com/conda-incubator/conda-project
+
+extra:
+  recipe-maintainers:
+    - AlbertDeFusco
```
mattkram, line 49:
```suggestion
    - AlbertDeFusco
    - mattkram
```

mattkram (commented):
Would appreciate an add as secondary maintainer

recipes/conda-project/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
```
mattkram, line 50:
```suggestion
```

AlbertDeFusco:
I seem to be having trouble with the automated builds. Is there something wrong with the recipe?

rxm7706:
> 



> Forgot to @conda-forge/help-python
> 
> The recipe worked once but I'm getting pip errors about `extras_require`
> 
> If you have a moment is there something I'm doing wrong?

@AlbertDeFusco @mattkram 
Thank you for creating conda-project ; looking forward to using it. 
I think https://github.com/conda-incubator/conda-project/pull/91 should fix the issue. 

I've tested a bit locally and also made a few edits to this conda recipe that you can see at
https://github.com/rxm7706/staged-recipes/blob/conda-project/recipes/conda-project/meta.yaml

recipes/conda-project/LICENSE
dbast, line 1: Please remove, as this file is part of the release tarball, which is a better warranty that it is correct.

dbast:
The error from `python setup.py egg_info` is:
```error in conda-project setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.```

You don't see this in the CI jobs in the upstream repo as defaults have an older setuptools version. You can build with an older version here `setuptools <=65.0.0`, which will make CI green.

If you decide to update the setup.py regarding extras_require to be compatible with newer setuptools, then please also update https://github.com/conda-incubator/conda-project/blob/main/versioneer.py as that is potentially not compatible with recent Python >3.6 versions.

**Comment by mattkram** [comment]
Thank you both for the fixes and explanations. That all makes complete sense.

The root issue is the error in the `setup.py` upstream for the sphinx dependency specifier, but I like the approach of pinning the setuptools version for now so we don't have to re-tag the release. We can resolve both in the next release. 

If @AlbertDeFusco is good with that, I'll work towards that this morning.

**Comment by rxm7706** [comment]
I think this should fix the setuptools issue
https://github.com/AlbertDeFusco/staged-recipes/pull/1
Also closed the test PR https://github.com/conda-forge/staged-recipes/pull/22461

**Comment by mattkram** [comment]
I've addressed the outstanding comments, and updated the branch to track `main`. CI should pass now (it passed before I updated the branch).

**Review by dbast — DISMISSED** [review]

**Comment by mattkram** [comment]
Hi @conda-forge/help-python, should be ready for review now :)

**Comment by rxm7706** [comment]
@AlbertDeFusco 
https://github.com/conda-incubator/conda-project/blob/0172b893635a17dd1e10665fc549560b72f66210/setup.py#L46
For noarch on conda forge, shouldn't python in meta.yaml and setup.py/pyproject.toml in source driving pip/pypi, support the same versions of python. py >3.7 ;

**Comment by dbast** [comment]
Seems like 3.8-3.10 are tested via CI and the setup.py needs an update https://github.com/conda-incubator/conda-project/blob/main/.github/workflows/main.yaml#L51 ... the meta.yaml already requires >3.8 (defaults and conda-forge anyways stopped to support Python <=3.7)

**Comment by dbast** [comment]
@rxm7706 Seems to be addressed in here https://github.com/conda-incubator/conda-project/pull/95

**Comment by mattkram** [comment]
Yes, we are going to update the source package to remove 3.7. Unless there's any issue, we were going to plan to re-tag the release, since it hasn't been published yet.

Alternately, we can release a fast-follow release after this initial one.

**Comment by mattkram** [comment]
We have merged conda-incubator/conda-project#95, which resolves the source package metadata. We have re-tagged the release as well, and updated the sha256 hash.

We have also pinned the upper-bound python version to `<3.12` since, although not all of our 3.11 tests complete because of dependency issues during installation of test packages, we are confident that the package will work if installed (it does pass [on macOS](https://github.com/conda-incubator/conda-project/actions/runs/4603979870/jobs/8134495161)).

Finally, I removed the upper pin on `setuptools`, since the source metadata for the package spec should also have been fixed in the same PR as referenced above.

With that, I believe we should be good to go if CI passes.

CC: @AlbertDeFusco

**Comment by mattkram** [comment]
We have merged conda-incubator/conda-project#95, which resolves the source package metadata. We have re-tagged the release as well, and updated the sha256 hash.

We have also pinned the upper-bound python version to `<3.12` since, although not all of our 3.11 tests complete because of dependency issues during installation of test packages, we are confident that the package will work if installed (it does pass [on macOS](https://github.com/conda-incubator/conda-project/actions/runs/4603979870/jobs/8134495161)).

Finally, I removed the upper pin on `setuptools`, since the source metadata for the package spec should also have been fixed in the same PR as referenced above.

With that, I believe we should be good to go if CI passes.

CC: @AlbertDeFusco

**Comment by mattkram** [comment]
I've diagnosed the above failures and I'm working on another fix. This one is due to the specification of search path for package files in `setuptools`.

I will reply back once I get that fixed and fix confirmed.

**Comment by mattkram** [comment]
I believe we have finally resolved the outstanding issues in the source repo now. The CI/CD passes, and the branch is updated now!

File: recipes/conda-project/meta.yaml [hunk 3]
```diff
@@ -0,0 +1,56 @@
...
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  noarch: python
+  entry_points:
+    - conda-project = conda_project.cli.main:main
+
+requirements:
+  host:
+    - python >=3.8, <3.12
```
dbast, line 21:
```suggestion
    - python >=3.8,<3.12
```

recipes/conda-project/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - pip
+    - setuptools
+  run:
+    - python >=3.8, <3.12
```
dbast, line 25:
```suggestion
    - python >=3.8,<3.12
```

recipes/conda-project/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  number: 0
+  noarch: python
+  entry_points:
+    - conda-project = conda_project.cli.main:main
+
+requirements:
+  host:
+    - python >=3.8, <3.12
+    - pip
+    - setuptools
```
dbast, line 23: According to https://github.com/conda-incubator/conda-project/blob/0.1.0/pyproject.toml#L2 
```suggestion
    - setuptools
    - setuptools-scm
```

mattkram: 👍

Out of curiosity, why do we have to list these explicitly if `pip` will grab them when it processes the install backend?

AlbertDeFusco:
We fixed an issue where the package was not getting the right version. We've switched to pypi sdist to correctly capture the version string in the package.

recipes/conda-project/meta.yaml
dbast: pip is not allowed to install anything from pypi during processing, as that would be a supply chain risk. Everything to be defined via conda available dependencies and sources with sha256.

You strictly don't need setuptools-scm anymore as the switch to pypi is packaging an sdist with burned in version... that afaik makes setuptools-scm an optional requirement.

You need setuptools-scm + such a file https://github.com/conda/conda/blob/main/.git_archival.txt to be able to package github release tarballs (it allows setuptools-scm to calculate the right version from the archive during build).

mattkram: Got it. Thanks for the detailed explanation!

I tried the sit archive support in setuptools-scm but didn't have any luck getting it working. The PyPI solution works well though, so I gave up on that yak shave 🙂. Maybe I'll come back to it another time.

PR #22522, opened 2023-04-10
nvim: add recipe and dependencies

danielnachun:
Checklist
- [x] Build number is 0.


danielnachun:
@conda-forge-admin, please ping team conda-forge/staged-recipes.

This PR should resolve some of the issues that previous failed attempts to submit `nvim` have encountered:

- Use `luajit-openresty`, a maintained branch of LuaJIT that still produces tagged releases.  For some reason the main LuaJIT upstream no longer tags releases which is problematic, particularly because the last tagged release is incompatible with Apple Silicon among other issues.
- Devendor all bundled dependencies into their own packages, including two Lua modules and a header only C library.  For now I think it makes sense to just target `luajit-openresty` for the Lua modules, but eventually these could also be built for Lua 5.1/5.2/5.3.  `lua-compat-5.3` also includes a Lua module which is not used by `nvim`, hence why I've only put the C library in `lua-compat-5.3-c`
- Build only the shared C library for `tree-sitter`.  There is a separate `tree-sitter` CLI written in Rust that fails to build on older macOS and should be packaged separately.  `tree-sitter` also provides an optional WebAssembly binding that is not used by `neovim` and should be packaged separately.
- For now I have skipped the build on Windows since it will be more complex.  I did borrow the `CmakeLists.txt` provided by `neovim` for `libtermkey`, `libvterm` and `tree-sitter` so that these packages now use `cmake` with `ninja` instead of `autotools`.  However the CMake installation used for `luajit` by `neovim` is unsuitable for `conda`.  I hope to eventually make a `CMakeLists.txt` for `luajit-openresty` so that we don't have to use `autotools` at all and that should make the Windows build easier, but I'd prefer to at least get `nvim` packaged on Linux and macOS.

danielnachun:
@conda-forge-admin please ping team conda-forge/staged-recipes
@conda-forge-admin please ping team conda-forge/help-c-cpp

I wanted to ping again about a review on this.   Thanks!

danielnachun:
@conda-forge-admin please ping team conda-forge/help-c-cpp

Also pinging here as this may be more specific to this recipe.

ocefpaf:
@danielnachun I guess that the amount of packages here make it hard to find a reviewer. I understand they it is better to submit then together to check consistency though.

At  a first glance I see that most cmake call do not have a make check/test or ctest. If there are any can you add them?

danielnachun:
@ocefpaf thank you for reviewing this.  I realize this was a fairly ambitious PR with so many recipes, but as you said it seemed to make the most sense to test these together since `nvim` has many dependencies.

Regarding the use of `make test`, I did add tests where I could, for `libvterm` and `unibilium`.  I realized that the custom `CMakelists.txt` I was adding had no way to run the `make test` target.  Since Windows support is not a goal of this PR, I changed the build recipes for all the recipes that normally use `make` to use it instead of `cmake` and `ninja`.  I would eventually love to see all these recipes use `cmake` and `ninja` but that will require more work on the `CMakelists.txt` for those recipes than I think is worth doing right now.

For the other C recipes I was unable to use `make test` for various reasons:

- `libtermkey` does have a `make test` target, but it fails in the last test on both macOS and Linux.  Oddly if I run the `conda_build.sh` script in `SRC_DIR` afterward, the test succeeds, so I don't think this failure is worth trying to track down right now.
- `luv` has a `make test` target, but it assumes that the build is happening in a Git repo.  I tried to directly invoke the Lua tests that it runs with `luajit tests/run.lua` and this fails because some of the tests want network access which is quite flaky to get right.  No network functionality is used in `luv` by `nvim` so I think this isn't critical either.
- `luajit-openresty` has a Perl testing suite.  I have gotten this to work locally, but it requires 3 Perl modules that are not in `conda-forge` (they are available in  `bioconda`).  I have a long term interest in bringing over Perl modules from `bioconda` and I am happy to do this at some point in a follow up, but this PR is big enough as is.
- `nvim` has a `make test` target for running tests with `busted`, a Lua testing framework.  This is not packaged by `conda-forge` either and I think it would be better to add this in a separate PR.
- `libtree-sitter` doesn't have a `make test` target.

PR #22531, opened 2023-04-11
Packaging TkinterWeb for Conda Forge

Andereoo:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/tkinterweb/meta.yaml
+++ b/recipes/tkinterweb/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "tkinterweb" %}
-{% set version = "3.17.5" %}
+{% set version = "3.17.7" %}
 
 package:
   name: {{ name|lower }}
--- a/recipes/tkinterweb/meta.yaml
+++ b/recipes/tkinterweb/meta.yaml
@@ -19,7 +19,7 @@ requirements:
     - python >=2.7
     - pip
   run:
-    - python >=32.7
+    - python >=2.7
     - pillow >=7.1
 
 test:
--- a/recipes/tkinterweb/meta.yaml
+++ b/recipes/tkinterweb/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "tkinterweb" %}
-{% set version = "3.17.4" %}
+{% set version = "3.17.5" %}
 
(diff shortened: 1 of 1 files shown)


Andereoo:
@conda-forge-admin I'm looking for some help with this pull request. Would anyone here happen to know why this is failing on OSX and Linux? Thanks!

dopplershift:
It looks like this package is shipping its own pre-compiled binary DLLs and .so files from the upstream repo? That's problematic for so many reasons, so you really need to build those here. Where does tkhtml come from?

Andereoo:
I'm not sure I completely understand the problem? The precompiled binaries are modified from Tkhtml and are specific to this project. They are packaged with the PyPi project.

dopplershift:
Is there a repo with the code for those modifications somewhere that could be built?

Except in extraordinary circumstances, we don't repackage binaries in our conda-forge recipes as it creates support problems in maintaining compatibility across our full collection of packages. We also strongly favor having builds that are open and inspectable.

Andereoo:
Yes, the modifications can be found at https://github.com/Andereoo/TkinterWeb-Tkhtml. How would I go about using this code instead of the precompiled binaries?

dopplershift:
Just FYI, I can't access that URL, I assume it's currently a private repo? You'd need to add an additional recipe that packages that repo as a separate package, then depend on that for TkinterWeb and avoid having it install its pre-compiled version.

Andereoo:
Hey! Sorry about the delay. I made the repository public.

The instructions for compiling it vary across platforms but generally follow this order of operations:
```
TCL=/pathtotcl
export PATH=$TCL/bin:$PATH
cd htmlwidget
tclsh src/cssprop.tcl && tclsh src/tokenlist.txt && tclsh src/mkdefaultstyle.tcl > htmldefaultstyle.c
mv *.c src && mv *.h src
mkdir build && cd build
../configure CC="gcc -static-libgcc" --with-tcl=$TCL/lib --with-tk=$TCL/lib --with-tclinclude=$TCL/include --with-tkinclude=$TCL/include
make binaries
```
Would you happen to have any insights on how to get conda-forge to compile this itself?

PR #22578, opened 2023-04-15
adding `libevalhyd` `evalhyd-python` `evalhyd-cli` `r-evalhyd`

ThibHlln:
Checklist
- [x] Build number is 0.


ThibHlln:
Hi @conda-forge/help-python-c

I don't understand why I get "No module named pip" for my Python recipe (relying on the C++ library in my C++ recipe), would you be able to tell me what I am doing wrong please?

ThibHlln:
Hi https://github.com/orgs/conda-forge/teams/help-python-c,

I am getting the following compiler error message:
> error: 'value' is unavailable: introduced in macOS 10.13 - see https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
          ? diagnostics.value()

I tried to set the environment variable `CXXFLAGS=${CXXFLAGS} -D _LIBCPP_DISABLE_AVAILABILITY"` in my build / script_env section in the recipe:
https://github.com/ThibHlln/staged-recipes/blob/b9ee4c838866d6e359be85ea7a31cf222af9c73e/recipes/evalhyd-python/meta.yaml#L19

But it does not seem to do the trick. Any idea how I can fix this please?

ThibHlln:
Hi `@conda-forge/help-r`,

I am trying to use a *configure*/*configure.win* file, but it says:
> ERROR: 'configure' exists but is not executable -- see the 'R Installation and Administration Manual'

So I think the TMPDIR that R is trying to use may be mounted as 'noexec' (https://cran.r-project.org/doc/manuals/R-admin.html#Running-R), any idea how I can fix this please?

ThibHlln:
The headers could not be found on Windows, probably because `xtensor` R lib was not in *Imports* and in *LinkingTo* sections of *DESCRIPTION* (like Rcpp was), this should fix the Windows CI now.

ThibHlln:
Hi @conda-forge/r-help,

Any idea why my C++ headers cannot be found for my Windows build please?

ThibHlln:
The MinGW GCC version is rather old (5.3.0) which resulted in some strange compilation error, never encountered before neither locally nor on our own GitLab CI. The solution was to use a different -more recent- compiler, Clang, using the conda package `r_clang_win-64`.

But then, *R.dll* could not be found, so we also needed to add it to the recipe's build section "missing_dso_whitelist" for it to work (see https://github.com/conda/conda-build/pull/4786).

ThibHlln:

Thank you.

recipes/evalhyd-cli/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+{% set version = "0.1.0.0" %}
+{% set sha256 = "38ebcc4915542b0887d4178724643539355b043a7aa4a68bdeacce4629bd1f11" %}
+
+package:
+  name: {{ project_name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.irstea.fr/HYCAR-Hydro/evalhyd/{{ project_name }}/-/archive/v{{ version }}/{{ project_name }}-v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
```
h-vetinari, line 14: Since all packages seem to be built from the same sources, it would make sense IMO to build everything in one multi-output recipe.

That way the dependence of (e.g.) the python bindings on the C++ lib is also much clearer and less fragile.

If you don't know how to do this, there are many examples (e.g. [arrow](https://github.com/conda-forge/arrow-cpp-feedstock/blob/main/recipe/meta.yaml)), or I can try to help.

recipes/evalhyd-cpp/bld.bat
```diff
@@ -0,0 +1,8 @@
+cmake -G "NMake Makefiles" -B build/ -D EVALHYD_BUILD_TEST=OFF -D CMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% -D CMAKE_INSTALL_LIBDIR=lib -D CMAKE_BUILD_TYPE=Release
```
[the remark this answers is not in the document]
h-vetinari: Also you're specifying ninja as a build dep in the meta.yaml, but using a different generator here?

recipes/evalhyd-cpp/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+cmake ${CMAKE_ARGS}                     \
+      -B build/                         \
+      -D EVALHYD_BUILD_TEST=OFF         \
+      -D CMAKE_INSTALL_PREFIX="$PREFIX" \
+      -D CMAKE_INSTALL_LIBDIR=lib       \
+      -D CMAKE_BUILD_TYPE=Release
```
h-vetinari, line 8:
```suggestion
      -D CMAKE_BUILD_TYPE=Release \
      -D BUILD_SHARED_LIBS=ON
```
also on windows

recipes/evalhyd-cpp/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - {{ compiler('cxx') }}
+    - cmake >=3.15
+    - make                    # [unix]
+    - ninja                   # [win]
+  host:
+    - xtl >=0.7.5
+    - xtensor >=0.24.6
+
+test:
+  commands:
```
h-vetinari, line 31: Please add tests for the presence of shared libs, absence of static ones.

recipes/evalhyd-cpp/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+package:
+  name: {{ project_name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.irstea.fr/HYCAR-Hydro/evalhyd/{{ project_name }}/-/archive/v{{ version }}/{{ project_name }}-v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
```
h-vetinari, line 17: Probably should have a run-export
```
  run_exports:
    - {{ pin_subpackage("libevalhyd", max_pin="x.x") }}
```
`x.x.x`  or `x.x` depending on whether you expect the ABI to break in patch releases or not

recipes/evalhyd-python/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  script_env:
+    - CPPFLAGS=-D_LIBCPP_DISABLE_AVAILABILITY -std=c++17    # [osx]
+    - EVALHYD_PYTHON_VENDOR_XTL=FALSE
+    - EVALHYD_PYTHON_VENDOR_XTENSOR=FALSE
+    - EVALHYD_PYTHON_VENDOR_XTENSOR_PYTHON=FALSE
+    - EVALHYD_PYTHON_VENDOR_EVALHYD_CPP=FALSE
```
h-vetinari, line 23: Better to set these in the build scripts directly and avoid `script_env`

recipes/evalhyd-python/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+
+test:
+  imports:
+    - {{ package_name }}
+  commands:
+    - python -c "import {{ package_name }}; assert {{ package_name }}.__version__ == '{{ version }}'"
+    - python -c "import {{ package_name }}; print({{ package_name }}.__doc__)"
+    - python -c "import {{ package_name }}; print({{ package_name }}.evald.__doc__)"
+    - python -c "import {{ package_name }}; print({{ package_name }}.evalp.__doc__)"
+    - python -c "import {{ package_name }}; import numpy; assert {{ package_name }}.evald(numpy.array([1, 2, 3]), numpy.array([1, 2, 3]), ['NSE'])[0][0, 0, 0] == 1"
```
h-vetinari, line 58: Please remove all the useless templating here and use the names that are expected (goes for everywhere else as well, but that will need to change anyway for a multi-output recipe).

A variable is useful when a value can change. The name in such a recipe cannot change, hence it just obstructs rather than helps.

h-vetinari (commented):
You're already off to a good start! But it's a pretty involved set of packages, so I think we should still iterate some more.

recipes/evalhyd-cli/meta.yaml
ThibHlln: Thank you very much for your review @h-vetinari 

I was not sure which way to go. And I tried with separate recipes, because I used different versions for the different packages (e.g. _0.1.0_ for the C++ core lib, and _0.1.0_.0 for the Python/R bindings), i.e. I keep an additional number for small changes to the bindings not affecting the C++ core lib.

If I had a multi-output recipe, would that prevent me from having separate versions?

h-vetinari: > If I had a multi-output recipe, would that prevent me from having separate versions?

If each source release is consistent, the versioning of the individual components is secondary. And yes, it's possible, see [scipopt](https://github.com/conda-forge/scipoptsuite-feedstock/blob/main/recipe/meta.yaml) for an extreme example (all components have a separate version). BTW: that's a case where variables are warranted. ;-)

recipes/evalhyd-cli/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+build:
+  number: 0
+  skip: true                  # [win and vc<14]
```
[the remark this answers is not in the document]
ThibHlln: removed in all recipes

recipes/evalhyd-cpp/bld.bat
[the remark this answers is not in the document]
ThibHlln: added line breaks and removed ninja as dependency where applicable

recipes/evalhyd-cpp/bld.bat
```diff
@@ -0,0 +1,8 @@
...
+if errorlevel 1 exit 1
```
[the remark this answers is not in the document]
ThibHlln: replaced in all batch file

recipes/evalhyd-cpp/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
ThibHlln: added in all bash script

recipes/evalhyd-cpp/build.sh
ThibHlln: added shared lib flag in batch and bash

recipes/evalhyd-cpp/meta.yaml
ThibHlln: Could you help me here please? I am not sure how to do this.

recipes/evalhyd-cpp/meta.yaml
ThibHlln: added with `x.x` only, patches not expected to break the ABI

recipes/evalhyd-cpp/bld.bat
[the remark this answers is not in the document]
h-vetinari: Not that you have to change it, but for completeness, I would have changed it in the other direction (i.e. use `ninja`). It's more transparent, better tested, and I also like to use it on unix, then I have a consistent toolset (CMake + ninja) across platforms.

recipes/evalhyd-python/meta.yaml
ThibHlln: tidied these up, in the R recipe also

recipes/evalhyd-cpp/meta.yaml
```diff
@@ -0,0 +1,56 @@
+# name for conda package (i.e. `conda install -c conda-forge <project_name>`)
+{% set project_name = "evalhyd-cpp" %}
```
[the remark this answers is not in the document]
ThibHlln: Ok, I have changed for `libevalhyd`, best to follow the best practice/convention

recipes/evalhyd-cli/meta.yaml
ThibHlln: OK, thank you.

And when you say that my packages are built fron the same sources, they come from the same GitLab project, but they are in different repositories. Is this a problem?

recipes/evalhyd-cpp/bld.bat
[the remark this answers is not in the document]
ThibHlln: This sounds like a noble approach, happy to give it a go. :-) Thank you for the advice.

recipes/evalhyd-cli/meta.yaml
h-vetinari: Ah, actually, I overlooked that with the templated variables, apologies. In that case, the multiple recipes are fine actually.

Please just use the strings directly, rather than a fake variable.

recipes/evalhyd-python/meta.yaml
ThibHlln: This has the advantage of not having to duplicate the information in two separate build files?

recipes/evalhyd-cli/meta.yaml
ThibHlln: Ok, understood. I have removed jinja templating where possible.

recipes/evalhyd-python/meta.yaml
h-vetinari: Yes

recipes/evalhyd-cpp/meta.yaml
h-vetinari: The link I gave above for the arrow-cpp feedstock has a nicely worked-out example of this (for several libraries; if there's only one you can drop the for loop)

ThibHlln: Thank you. I have added similar tests, but they seem to fail. So I must not be building the shared library properly?

ThibHlln: But that's because I create an interface library, isn't it?

h-vetinari: Sorry, I haven't looked at the library specifically, I was talking in general principles. If it's header only and there are no library artefacts, then that's fine and no further tests are needed

ThibHlln:
Hi @h-vetinari 

Thank you very much for your help. 

Do you think this is ready to ship now?

recipes/r-evalhyd/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+  script_env:
+    - EVALHYD_R_VENDOR_XTL=FALSE
+    - EVALHYD_R_VENDOR_XTENSOR=FALSE
+    - EVALHYD_R_VENDOR_XTENSOR_R=FALSE
+    - EVALHYD_R_VENDOR_EVALHYD_CPP=FALSE
+  rpaths:
+    - lib/R/lib/
+    - lib/
+  missing_dso_whitelist:
+    - '*/R.dll'                   # [win]
```
h-vetinari, line 26: This needs a comment IMO; what's the reason we need this exception?

recipes/evalhyd-cli/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - libevalhyd ==0.1.0
+
+test:
+  commands:
+    - |
+      evalhyd --help
+    - |
+      evalhyd evald --help
+    - |
+      evalhyd evalp --help
```
h-vetinari, line 32: You shouldn't need the `- |` synatx, just regular `-` should suffice.

h-vetinari (commented):
I don't have a commit bit here, and I'm not knowledgeable with R recipes. That said, the library and python parts look OK to me.

@conda-forge/help-python-c
@conda-forge/help-r

recipes/evalhyd-cli/meta.yaml
ThibHlln: I had issues (on Windows I think) because of the "dash" denoting the YAML sequence, and the "dashes" denoting the help option, but quoting the whole line would not work either, that is why I had it this way. https://github.com/ThibHlln/staged-recipes/commit/4185e1b2e3c0c8c55265fe2a102d7eb019f93199

ThibHlln: Ok, I have tried replacing `--help` with `-h` to keep it on one line. In my other recipes, I never use double dashes, but single dashes do not seem to trigger problems, so let's see what the CI says.

recipes/r-evalhyd/meta.yaml
ThibHlln: Fair point. I have added a comment, referring to a PR on conda-build, explanation why the missing R.dll message needs to be ignored.

ThibHlln:
> I don't have a commit bit here, and I'm not knowledgeable with R recipes. That said, the library and python parts look OK to me.
> 
> @conda-forge/help-python-c @conda-forge/help-r

Thank you @h-vetinari, I will be waiting for someone with merging rights to further review this PR then.

recipes/evalhyd-cli/bld.bat
```diff
@@ -0,0 +1,12 @@
+cmake -G "Ninja" ^
+      -B build/ ^
+      -D CMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% ^
+      -D CMAKE_INSTALL_LIBDIR=lib ^
+      -D CMAKE_BUILD_TYPE=Release
+if %ERRORLEVEL% neq 0 exit 1
+
+cmake --build build/ --parallel %CPU_COUNT%
```
ocefpaf, line 8: Is there a `make test` or `ctest` we can run in these cmake builds?

recipes/evalhyd-python/build.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+set -ex
+
+${PYTHON} -m pip install . -vvv
```
ocefpaf, line 4: Why not use the `script` entry in the meta.yaml and remove both build and bld files?

recipes/evalhyd-python/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - numpy
+
+test:
+  imports:
+    - evalhyd
+  commands:
+    - python -c "import evalhyd; assert evalhyd.__version__ == '{{ version }}'"
+    - python -c "import evalhyd; print(evalhyd.__doc__)"
+    - python -c "import evalhyd; print(evalhyd.evald.__doc__)"
+    - python -c "import evalhyd; print(evalhyd.evalp.__doc__)"
```
ocefpaf, line 51: Probably not needed.

recipes/evalhyd-cli/bld.bat
ThibHlln: No, the unittests for this CLI are actually written in Python (calling the executable). Should these be used in the "test" section of the recipe?

recipes/evalhyd-python/build.sh
ThibHlln: Yes, you are right, that's an overkill. I will fix this.

recipes/evalhyd-python/meta.yaml
ThibHlln: Yes, now there are some more meaningful tests afterwards, these are not very useful anymore. I will drop them.

ThibHlln:
Thank you very much @ocefpaf

PR #22613, opened 2023-04-19
Add geo-espresso

jwhhh:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/geo-espresso/meta.yaml
+++ b/recipes/geo-espresso/meta.yaml
@@ -10,7 +10,6 @@ source:
   sha256: 504c0a6341ecc7d385a3632fbf3628e83d6d77ffd97bb1fd68761197c120c0f9
 
 build:
-  skip: True  # [py<37]
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
@@ -62,4 +61,3 @@ about:
 extra:
   recipe-maintainers:
     - jwhhh
-
--- a/recipes/geo-espresso/meta.yaml
+++ b/recipes/geo-espresso/meta.yaml
@@ -1,0 +1,60 @@
+{% set name = "geo-espresso" %}
+{% set version = "0.2.2.dev1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/geo-espresso-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/geo-espresso/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set name = "geo-espresso" %}
+{% set version = "0.2.2.dev1" %}
```
dopplershift, line 2: We generally want to package release builds for conda-forge. Does this package have any regular releases?

jwhhh: Thanks for the comment. We haven't had a release yet, but maybe that's due to us being too conservative about versioning. The API is relatively stable already, but currently the user base is small so I'm not entirely sure we are ready in terms of having as fewer bugs as possible.

If it's really necessary, I can make a release for going to conda-forge. Otherwise I think it should be fine and I'm pretty sure the development is ongoing (not going to be abandoned or something)..

jwhhh:
Dear maintainers, could I please get some help on this PR? More specifically, is it allowed to disable `win64` build if you agree that the below issue is not worth it?

## Detailed explanation

This Python package is built with `scikit-build` and hence `cmake`. From my understanding, it fails on `win64` because the `gfortran` and the generators used by `scikit-build` are somehow incompatible ([related article](https://www.kitware.com//fortran-for-cc-developers-made-easier-with-cmake/))

If we can specify the generator on Windows to be `MinGW Makefiles` then *maybe* this could be solved (like in this [PCMSolver example](https://github.com/conda-forge/staged-recipes/pull/22286/files#diff-cc705a338a3541d60482ba07d04218fddfd2dd93ca67cdaa417a8faa2d5a4ea1R8)). But from the [scikit-build documentation page](https://scikit-build.readthedocs.io/en/latest/generators.html#build-system-generator) I read that this generator is not supported by them.

Another option I've seen is using `flang` instead of `m2w64_fortran` (inspired from [fastscapelib-fortran-feedstock](https://github.com/conda-forge/fastscapelib-f2py-feedstock/blob/main/recipe/meta.yaml#L25)), but adding it didn't help this project as you may see in the most recent build. And I'm not an expert in Windows so thought maybe we'll skip `win64` for now and will come back later when we do have the expertise.

recipes/geo-espresso/meta.yaml
jwhhh: Hi, we've released v0.3.0 on PyPI and I'll update the version number and hash after [the conversation below](https://github.com/conda-forge/staged-recipes/pull/22613#issuecomment-1519484088) has been reviewed. Thanks and cheers!

PR #22651, opened 2023-04-21
Add recipe for `semimap`

jjerphan:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/semimap/meta.yaml
+++ b/recipes/semimap/meta.yaml
@@ -27,10 +27,6 @@ requirements:
     - make  # [unix]
 
 test:
-  requires:
-    - {{ compiler('cxx') }}
-  source_files:
-    - test.cpp
   commands:
     - bash {{ RECIPE_DIR }}/run_test.sh  # [not win]
     - {{ RECIPE_DIR }}/run_test.bat  # [win]
--- a/recipes/semimap/meta.yaml
+++ b/recipes/semimap/meta.yaml
@@ -1,13 +1,7 @@
 {% set name = "semimap" %}
-# NOTE: Semimap does not have a release or a version.
-#
-# We rely on the last commit on upstream at the time of writing
-# and simply use it as the version identifier.
-#
-# Q: Is there a better alternative?
-{% set version = "9a106db4a8eb38a5b6e7c94a2a7addf8de7f6dd2" %}
+{% set version = "1.0.0" %}
 {% set organization = "QuantStack" %}
-{% set sha256 = "4d24673f4b83d8de1e573ef7cd06060e3e1a765e272f15a8d7d9a470f2a80b0e" %}
+{% set sha256 = "77534da5e5c8c582945210a5665c9f711e323e17853bc92fa45601c410f42611" %}
 
 package:
   name: {{ name|lower }}
--- a/recipes/semimap/meta.yaml
+++ b/recipes/semimap/meta.yaml
@@ -1,6 +1,10 @@
 {% set name = "semimap" %}
-# Semimap does not have any release nor version.
-# We rely on the last commit on upstream at the time of writing.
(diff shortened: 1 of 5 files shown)


jjerphan:
I added support for the test suite.

I think we need to discuss how we handle the lack of a version identifier.

JohanMabille:
We usually avoid to copy header files. A better practice is to do it via `make install`, which also installs a cmake package file. This way, you can explicitely state the dependency in the CMakeLists.txt of the downstream package:

```cmake
find_package(semimap REQUIRED)
```

This requires a projectConfig.cmake file that is usually configured in the CMakeLists.txt in the install section. You can take inspiration from [this config file](https://github.com/xtensor-stack/xtl/blob/master/xtlConfig.cmake.in) and the [corresponding section](https://github.com/xtensor-stack/xtl/blob/master/CMakeLists.txt#L98) in the CMakeLists.txt.

This repo does not have such a config file; when packaging these repos for conda-forge, we usually contribute upsteam and then we have two options:
- Wait for the contribution to be accepted and for a new release including it. This is fine when the upstream maintainers are very responsive and can tag a new release quickly
- patch the repo in the recipes and release the package; when a new release is published, with the contribution, the patch can be removed.

In the case of `semimap`, since the main repo does not appear to be matained, I woud suggest to contribute upstream, but also to the fork in the QuantStack org (since we use it for having tags).

JohanMabille (changes requested):
We usually avoid to direcly copy header files. A better practice is to run a `make install` command, that will copy the header and a cmake package file. Thi way, we can explicitely state the dependency in the downstream CMakeLists.txt:

```cmake
find_package(semimap REQUIRED)
```

This package file is usually the result of a config file used in the install process. Here is an example of such a [config file](https://github.com/xtensor-stack/xtl/blob/master/xtlConfig.cmake.in) and the [section of CMakeLists](https://github.com/xtensor-stack/xtl/blob/master/CMakeLists.txt#L98) using it.

In this case, the source repository does not contain this package file. When packaging such a repo for conda-forge, we usually contribute upstream, and then we have two options:
- wait for a new release with the contribution included
- patch the repo in the feedstock, release the package. The patches will be droped when a new release with the contribution is out.

Since the original `semimap` repo doe snot seem actively maintain, and since we have a fork in the QuantStack org for the release tags, I suggest to update the fork directly and not patch the repo in the recipes. (A PR should be opened to the original repo too).

recipes/semimap/meta.yaml
```diff
@@ -0,0 +1,50 @@
+{% set name = "semimap" %}
+# NOTE: Semimap does not have a release or a version.
+#
+# We rely on the last commit on upstream at the time of writing
+# and simply use it as the version identifier.
+#
+# Q: Is there a better alternative?
+{% set version = "9a106db4a8eb38a5b6e7c94a2a7addf8de7f6dd2" %}
+{% set organization = "QuantStack" %}
+{% set sha256 = "4d24673f4b83d8de1e573ef7cd06060e3e1a765e272f15a8d7d9a470f2a80b0e" %}
```
jjerphan, line 10: To me, the last remaining piece of complexity to address is to have a proper version identifier for semimap.

https://github.com/QuantStack/semimap/commit/3e2175dbaf3f17004f91e0a0f26de39665607241 somewhat has introduced a v.1.0.0 version; after the first review of this recipe, the repository was restructured to use cmake.

Should we bump the version number, use a tag and create a release before this recipe gets approved? What do you think, @JohanMabille?

JohanMabille: Yes, we should tag a release on the current commit (since the version is already set) and use it for this PR.

jjerphan: What would be an appropriate version identifier? v1.1.0?

PR #22664, opened 2023-04-23
Add openvdb

versatran01:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/openvdb/meta.yaml
+++ b/recipes/openvdb/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "openvdb" %}
+{% set version = "10.0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/AcademySoftwareFoundation/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 887a3391fbd96b20c77914f4fb3ab4b33d26e5fc479aa036d395def5523c622f
+
+
+build:
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
+    - make  # [unix]
+    - boost-cpp
+    - blosc
+    - zlib
+    - jemalloc
+
+test:
+  commands:
+    - test -f ${PREFIX}/include/openvdb.hpp  # [unix]
+
+about:
+  home: https://github.com/AcademySoftwareFoundation/openvdb
+  summary: 'OpenVDB - Sparse volume data structure and tools'
+  description: |
+    OpenVDB is an open source C++ library comprising a novel hierarchical data structure and a large suite of tools for the efficient storage and manipulation of sparse volumetric data discretized on three-dimensional grids. It was developed by DreamWorks Animation for use in volumetric applications typically encountered in feature film production.
+  license: Mozilla Public License 2.0
+  license_family: MPL
+  license_file: LICENSE
+  dev_url: https://github.com/AcademySoftwareFoundation/openvdb
+
+extra:
+  recipe-maintainers:
+    - versatran01
--- a/recipes/openvdb/build.sh
+++ b/recipes/openvdb/build.sh
@@ -1,0 +1,11 @@
+mkdir -p build
+cd build
+cmake ${CMAKE_ARGS} \
+      -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+      -DCMAKE_PREFIX_PATH=${PREFIX} \
+      -DOPENVDB_CORE_SHARED=ON \
+      -DOPENVDB_CORE_STATIC=OFF \
+      -DUSE_EXPLICIT_INSTANTIATION=OFF \
+      ..
+make -j${CPU_COUNT}
+make install


versatran01:

versatran01:

versatran01:

versatran01:
@tetov 
Yeah, I'm open to adding python/numpy support.

tetov:
@versatran01: Ok, I'll work on that (prob over the weekend). I don't think it should hold this PR back though if it's ready to be merged.

versatran01:
I can add you as a maintainer to this recipe if you are interested.

versatran01:
Can this be merged?

tetov:
I'll work on the recipe when I find time, life happened, sorry!

I think it should be merged in the meantime.

versatran01:
@tetov, sure no problem. I'm don't think I can merge this myself, it has to be some one from conda?

recipes/openvdb/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  run:
+    - boost-cpp
+    - blosc
+    - zlib
+    - tbb-devel
+    - jemalloc
+
+test:
+  commands:
+    - test -f ${PREFIX}/include/openvdb/openvdb.h  # [unix]
```
carterbox, line 39: This doesn't seem to be a header only library, so please test for the other artifacts that you are installing to $PREFIX including a cmake file, a shared library. Also, check that you binary runs in the test phase by calling `vdp_print --help` or something similar.

recipes/openvdb/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+source:
+  url: https://github.com/AcademySoftwareFoundation/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 887a3391fbd96b20c77914f4fb3ab4b33d26e5fc479aa036d395def5523c622f
+
+
+build:
+  number: 0
+  skip: true  # [win or osx]
+
```
carterbox, line 16:
```suggestion
build:
  number: 0
  skip: true  # [win or osx]
  run_exports:
    - {{ pin_subpackage('openvdb', max_pin='x.x.x') }}
```
Run exports are required for packages that ship libraries which can be dynamically linked by other packages. Adjust the max_pin for the ABI compatability guarantees of your package.

recipes/openvdb/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - zlib
+    - tbb-devel
+    - jemalloc
+
+  run:
+    - boost-cpp
+    - blosc
+    - zlib
+    - tbb-devel
+    - jemalloc
```
carterbox, line 35:
```suggestion
  run:
    - boost-cpp
```
According to the logs, you are overdepending on jemalloc, zlib, and tbb-devel. If these are header-only libraries, you don't need to list them in the run requirements. If you are static linking, don't! static linking is not allowed.

Generally recipes should have run_exports, so you should not need to list C libraries in run requirements explicitly unless conda-build complains that something is missing.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=745563&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1493

carterbox (changes requested):

recipes/openvdb/meta.yaml
tetov: openvdb guarantees ABI compatability between version in the same major release, see [these references to `OPENVDB_ABI_VERSION_NUMBER`](https://github.com/search?q=repo%3AAcademySoftwareFoundation%2Fopenvdb%20OPENVDB_ABI_VERSION_NUMBER&type=code).

Meaning that the line should be:
```
    - {{ pin_subpackage('openvdb', max_pin='x') }}
 ```

carterbox: Is the ABI version separate from the API version? In that case, we can easily make a subpackage like `libvdb10` which would contain `libvdb.so.10`. That way the API and ABI can be depended on separately.

tetov: Major is incremented on ABI changes for the Grid class and its constituent classes or changes to file format. Minor on API. I think your suggestion would make a lot of sense. I'll have to educate myself a bit more on subpackages in the conda-forge ecosystem and then test this out :).

tetov: https://www.openvdb.org/documentation/doxygen/faq.html#sVersionNumbering

tetov: I looked into this where I used one output named libopenvdb, and added `    - {{ pin_subpackage('openvdb', max_pin='x') }}`  to the top run requires. (I also tested this with two outputs, libopenvdb and openvdb..). Didn't really get this to work. I have been trying to find more documentation on split packages / metapackages, but haven't really grokked it yet. I've also tried to find some other feedstock-recipes that does something similar but the ones I find are quite complicated.

Here are some more direct questions:

* Should there be one or two subpackages under `outputs:` ? If one, should I explicitly exclude the shared lib from the top level package?
* Should the library only package be named libopenvdb or libopenvdb10?
* Should I have no build steps in top leve / openvdb subpackage and only build in libopenvdb?

recipes/openvdb/meta.yaml
tetov: I added more tests in https://github.com/versatran01/staged-recipes/pull/1/commits/646e474ffa3b84ba8bfd4256495f38db99f6f88e but eventually we should run upstreams unit tests as well, right? Or would that waste too much CI time?

In that case, should we write a simple cpp file that includes headers from the lib? I've seen that in [conda-forge/opencv-feedstock. ](https://github.com/conda-forge/opencv-feedstock/).

recipes/openvdb/meta.yaml
tetov: Fixed in https://github.com/versatran01/staged-recipes/pull/1/commits/646e474ffa3b84ba8bfd4256495f38db99f6f88e.

recipes/openvdb/meta.yaml
carterbox: Just to reiterate, creating a `libopenvdb10` package isn't really needed unless the ABI and API versions are separate. But it looks like [you are treating them as the same](https://github.com/versatran01/staged-recipes/blob/12481bcf36e1c4f0cc9d285d0b1382f49a58ecfc/recipes/openvdb/meta.yaml#L34C41-L34C41).

For multiple outputs, think of two steps:

1. Build all the artifacts and install them to $PREFIX. The top level package should be called `openvdb-split`, and it does this. Just build everything here together.
2. Separate the artifacts from `openvdb-split` into multiple outputs. In the output section you list all of the package that will be published. Use the glob matching file lists to separate the files from $PREFIX into each output.

Probably, the outputs for this recipe are `openvdb` and `libopenvdb10`

https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#specifying-files-to-include-in-output

Here's an live example feedstock, but they use install scripts instead of file lists to separate the outputs. IMHO, file lists are more verbose, but simpler and easier to understand.

https://github.com/conda-forge/mlir-feedstock/blob/main/recipe/meta.yaml

tetov: Thanks! I'll do some more testing on this. And I would be interested in hearing if @versatran01 has any thoughts or opinions. 

It can always be a later enhancement, right? This would be better than no openvdb on conda-forge.

versatran01:
@tetov I’d be happy to merge what you have if it’s ready. Thanks for the work.

tetov:
@versatran01
> @tetov I’d be happy to merge what you have if it’s ready. Thanks for the work.

No worries! Just merge this PR https://github.com/versatran01/staged-recipes/pull/1 and it should show up in this PR.

tetov:
Fixed error from linter in https://github.com/versatran01/staged-recipes/pull/2

@versatran01  When that has been merged I think we should ask for a new review. I have worked more on tests with split packages but would like to do so moving forward in a branch on merged recipe, so I can test using CI instead of just locally.

recipes/openvdb/meta.yaml
carterbox: The purpose of the tests in this recipe is to check *packaging*. Unit tests are for upstream to check correctness. Here we only want to check that everything is installed and linked correctly.

recipes/openvdb/meta.yaml
carterbox: Please add an appropriate run_export. This is required so that downstream packages have a compatible version of this library at runtime.

carterbox (changes requested):
This is ready to merge once a run_export is added.

tetov:
Thanks for the review @carterbox!

> This is ready to merge once a run_export is added.

@versatran01: Merge this to add run_exports. https://github.com/versatran01/staged-recipes/pull/3

versatran01:
Merged. Thanks!

Issue #12537, opened 2020-09-02
Package dgl


hadim:
I am considering packaging [dgl](https://github.com/dmlc/dgl) into conda-forge but I don't know whether this is something possible.

See https://github.com/dmlc/dgl/issues/1855 for context.

They already have a conda channel at https://anaconda.org/dglteam

dgl is built against different versions of CUDA, see https://github.com/dmlc/dgl#using-anaconda and they provide different packages per CUDA version which I believe is not the best way to do it. I think it makes more sense to have a single package with different variants that are selected according to the presence or not of a specific cudatoolkit in the graph of deps. Please correct me if I am wrong (it seems to be similar to what pytorch is doing on their own channel).

Do you have an example of such packages in conda-forge I could start with to build dgl?

ping @conda-forge/staged-recipes

hadim:
Closing here in favor of https://github.com/conda-forge/staged-recipes/pull/12552 and https://github.com/dmlc/dgl/issues/1855

jakirkham:
Reopening since this appears not to have been completed

jakirkham:
There's a more recent attempt in PR ( https://github.com/conda-forge/staged-recipes/pull/18620 ), which has since been superseded by PR ( https://github.com/conda-forge/staged-recipes/pull/22691 )

mikemhenry:
Haha yes there are a trail of PRs for this project! Thank you @jakirkham for helping this along! 

Really the slow part is doing this the right way :tm: since in https://github.com/conda-forge/staged-recipes/pull/18620 we showed we can get it to build, but had some issues with vendored packages. @hmacdope is now working on trying to un-vendor as much as we can... 

I don't want trying to get this perfect to prevent us from getting this moved to a feedstock. We have improved things considerably re: modifcations to `dgl` build process to use system packages instead of vendoring everything.

jakirkham:
Yeah that seems pretty reasonable

Devendoring can help with cutting down build time and avoiding clobbering

Though yeah sometimes these issues need further iteration in the feedstock

mikemhenry:
Yeah I found in particular that getting CUDA stuff working tends to go easier in the feedstock. And I would like to play with that once we get it in a feedstock to work with the new cuda12 packaging 

Thanks!

jakirkham:
Would be happy to work with you on CUDA 12 builds once we get there 🙂

PR #22691, opened 2023-04-25
Add conda-forge recipie for `dgl`

hmacdope:
Tagging @mikemhenry and @hadim

Checklist
- [ ] Build number is 0.

Fixes https://github.com/conda-forge/staged-recipes/issues/12537


mikemhenry:
```
-- DLPACK_INCLUDE_DIR not defined, searching for dlpack header
CMake Error at /home/conda/staged-recipes/build_artifacts/dgl_1682466386125/_build_env/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find DLPACK (missing: DLPACK_INCLUDE_DIR)
  /home/conda/staged-recipes/build_artifacts/dgl_1682466386125/_build_env/share/cmake-3.26/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindDLPACK.cmake:39 (find_package_handle_standard_args)
  CMakeLists.txt:279 (find_package)
```

hmacdope:
@mikemhenry this is the same issue I was seeing locally, I will continue to investigate. :smiley_cat:

mikemhenry:
Few hints:
https://github.com/conda-forge/dlpack-feedstock/blob/main/recipe/meta.yaml#L26
https://github.com/conda-forge/libtvm-feedstock/blob/main/recipe/build.sh#L48

hmacdope:
> Few hints:
> 
> https://github.com/conda-forge/dlpack-feedstock/blob/main/recipe/meta.yaml#L26
> 
> https://github.com/conda-forge/libtvm-feedstock/blob/main/recipe/build.sh#L48

Ah awesome should help a lot

bartekkuncer:
Hi @hmacdope, @mikemhenry! First of all thanks for driving the initiative of adding dgl to conda-forge channel. I have two questions:
1. how are the works going - is there any major blocker atm?
2. I noticed that in build.sh you are setting USE_LIBXSMM flag to 'OFF' even though it is available on conda-forge channel and by default set to 'ON'. Why is that?

mikemhenry:
👋 Hey @bartekkuncer Check out this thread here: https://github.com/dmlc/dgl/issues/1855 (you may have seen it) @hmacdope is in Australia so there is a bit of a time delay. We are still in the middle of working on this and will post an update soon. Right now I think that flag controls if to build with libxsmm, and doesn't control using the vendored lib or the lib from conda-forge. Before we merge in this feedstock, we will have it using every lib that is on codna-forge and we will add the missing ones as well.

hmacdope:
@mikemhenry this will compile now at least (import failing with some linker error) which means the  mechanism for inserting CF deps into the build is working. I think I will now try and make this work with the latest version + **vendored** thrust + up to date deps.

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/dmlc/dgl
+  git_rev: {{ version }}
+  patches:
+    - conda-build.patch
+
+build:
+  skip: true  # [(not linux) or (cuda_compiler_version in (undefined, 'None'))]
```
bgawrych, line 16: Are you skipping version without GPU on purpose? Because linux without GPU is skipped with this line

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+  number: 0
+  rpaths:
+    - lib/
+  missing_dso_whitelist:
+    - '*/libcuda.*'  # [linux64]
```
bgawrych, line 21:
```suggestion
    - '*/libcuda.*'  # [linux64]
    - "$RPATH/libtorch_cpu.so"
    - "$RPATH/libc10.so"
    - "$RPATH/libtorch.so"
```

If PyTorch added to build requirments

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,83 @@
...
+  string: cuda{{ cuda_compiler_version | replace('.', '') }}py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [cuda_compiler_version not in (undefined, 'None')]
+    - sysroot_linux-64  2.17  # [linux64]    
+    - cmake
```
bgawrych, line 30:
```suggestion
    - cmake
    - make
```

I couldn't run build (only CPU) without make

recipes/dgl/build.sh
```diff
@@ -0,0 +1,37 @@
...
+
+mkdir build
+cd build
+
+cmake -DUSE_CUDA=ON \
+  -DUSE_CONDA_INCLUDES=ON \
+  -DEXTERNAL_DLPACK_PATH=$PREFIX/include \
+  -DEXTERNAL_DMLC_PATH=$PREFIX/include \
+  -DEXTERNAL_PHMAP_PATH=$PREFIX/include \
+  -DEXTERNAL_NANOFLANN_PATH=$PREFIX/include \
```
bgawrych, line 23:
```suggestion
  -DEXTERNAL_DLPACK_PATH=$BUILD_PREFIX/include \
  -DEXTERNAL_DMLC_PATH=$BUILD_PREFIX/include \
  -DEXTERNAL_PHMAP_PATH=$BUILD_PREFIX/include \
  -DEXTERNAL_NANOFLANN_PATH=$BUILD_PREFIX/include \
```

bgawrych: $PREFIX path doesnt containt exterrnal dependencies - it is the installation directory

recipes/dgl/build.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+set -euxo pipefail
+
+rm -rf build || true
+
+CUDA_CMAKE_OPTIONS="-DCMAKE_CUDA_COMPILER=${CUDA_HOME}/bin/nvcc -DCMAKE_CUDA_HOST_COMPILER=${CXX}"
+
+# Remove -std=c++17 from CXXFLAGS for compatibility with nvcc
+export CXXFLAGS="$(echo $CXXFLAGS | sed -e 's/ -std=[^ ]*//')"
+
```
bgawrych, line 10:
```suggestion
export CFLAGS="$(echo $CFLAGS | sed -e 's/ -mtune=[^ ]*//')"
```

When compiling without cuda on GCC 12 I got error from third_party library (tensorpipe) - removing mtune flag fixed my issue

bgawrych (commented):
Hi guys @hmacdope @mikemhenry ,
I run locally your recipe and I've got some conclusion shared below. I still have some issues, but I hope it might be helpful for you

hmacdope:
Thanks so much @bgawrych for the help! I will keep working to get this through :+1:

hmacdope:
@mikemhenry @bgawrych @conda-forge/help-c-cpp I cannot for the life of me to get `BUILD_PREFIX ` to expand properly in the conda-build env with CMake.

while `ls` reports the correct dir in the build script. 

```bash
ls $BUILD_PREFIX/lib 
> /home/conda/staged-recipes/build_artifacts/dgl_1686114006279/_build_env/lib`
> # correct contents
...
```

it refuses to expand in the CMAKE and we end up with literal.

This appears to be because cmake reports a circular environment variable 

```bash
cmake -E environment
...
BUILD_PREFIX=$BUILD_PREFIX
...
```

and we end up with a literal in the CMake. 
```
-- looking for dlpack headers in $BUILD_PREFIX/include
-- looking for dmlc headers in $BUILD_PREFIX/include
-- looking for dmlc library in $BUILD_PREFIX/lib
...
> fails
```

I have been tearing my hair out over this. Am I missing something blindingly obvious (probably) ?

recipes/dgl/conda-build.patch
```diff
@@ -0,0 +1,222 @@
...
++                          )
++      else(REBUILD_LIBXSMM)
++        add_custom_target(libxsmm COMMAND make -j ECFLAGS="-Wno-error=deprecated-declarations" BLAS=0 CC=${LIBXSMM_COMPILER}
++                          WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/third_party/libxsmm
++                          )
++      endif(REBUILD_LIBXSMM)
++      add_dependencies(dgl libxsmm)
++      list(APPEND DGL_LINKER_LIBS -L${CMAKE_SOURCE_DIR}/third_party/libxsmm/lib/ xsmm)
++    endif(NOT MSVC)
++  endif(EXTERNAL_LIBXSMM_PATH)
```
bgawrych, line 162: @hmacdope regarding libxsmm I think it won't work as DGL uses newer version of libxssm which is not officialy released (latest is 1.17)
My idea is to build libxsmm from sources, but using header-only version - this probably would be compatible with conda-forge requirements

Please, don't commit this suggestion as it won't work with git patch 


To make it work changes in 'src/array/cpu/spmm_blocking_libxsmm.h' and 'src/runtime/config.cc' are required -
#include <libxsmm.h> ===> #include <libxsmm_source.h>
#include <libxsmm_cpuid.h> ===> #include <libxsmm_source.h>

To build in header-only mode replace all the code in USE_LIBXSMM condition with this single line

```suggestion
+ target_include_directories(dgl PRIVATE "third_party/libxsmm/include")
```

hmacdope: Thanks so much again @bgawrych, I will make the required changes and combine them in one patch.

hmacdope:
@bgawrych I have added in the header only libxsmm build, thanks for the tip.

@mikemhenry  The build seems to be timing out in the separate pytorch compilation part (confirming it will finish on my local box now), do we know if there is a workaround for that? Other than that I think all we need is `METIS` and `GKLib` and then we should be good to go.

mikemhenry:
https://github.com/conda-forge/gklib-feedstock is on conda-forge v5.1.1
and I am working on `METIS` 5.1.1 here https://github.com/conda-forge/metis-feedstock/pull/32 (might use 5.1.2)

mikemhenry:
@hmacdope 
> Other than that I think all we need is METIS and GKLib and then we should be good to go.

Do you mean just getting the stuff added in cmake so that it pulls them from conda forge or something else?

mikemhenry:
What are we building here:
`Building CXX object CMakeFiles/dgl_sparse_pytorch_2.0.0.post200.dir/src/elemenwise_op.cc.o`

hmacdope:
> What are we building here: `Building CXX object CMakeFiles/dgl_sparse_pytorch_2.0.0.post200.dir/src/elemenwise_op.cc.o`

I'm fairly sure these are triggered by https://github.com/dmlc/dgl/blob/master/CMakeLists.txt#L247 which does a libtorch recompile.

hmacdope:
> https://github.com/conda-forge/gklib-feedstock is on conda-forge v5.1.1 and I am working on `METIS` 5.1.1 here [conda-forge/metis-feedstock#32](https://github.com/conda-forge/metis-feedstock/pull/32) (might use 5.1.2)

Both! Sorry I should have been clearer,  checking the tags seems like GKLib is up to date enough on CF already and METIS will be soon. Thanks for pushing that forward @mikemhenry 

 I will make a start on the CMake integration but, we will likely need them both on CF to really figure out how it will work because GKLib is built inside of the METIS build in the current setup, and we don't want that so we may need to experiment in how to separate them.

bgawrych:
Hi @hmacdope @mikemhenry, what is the current status? Do you need help with anything?

hmacdope:
Hey @bgawrych ! Sorry I have been a bit slow, currently seeing the following linking issue for `libdgl_sparse_pytorch`that is I think causing CI to hang and has me a bit stumped. Any help would be massively appreciated.

```
[100%] Linking CXX shared library libdgl_sparse_pytorch_1.12.1.post201.so
/home/conda/staged-recipes/build_artifacts/dgl_1687909279052/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/7.5.0/../../../../x86_64-conda-linux-gnu/bin/ld: cannot find -ldmlc
make[5]: *** [CMakeFiles/dgl_sparse_pytorch_1.12.1.post201.dir/build.make:289: libdgl_sparse_pytorch_1.12.1.post201.so] Error 1
make[4]: *** [CMakeFiles/Makefile2:83: CMakeFiles/dgl_sparse_pytorch_1.12.1.post201.dir/all] Error 2
make[3]: *** [Makefile:91: all] Error 2
make[2]: *** [CMakeFiles/dgl_sparse.dir/build.make:70: CMakeFiles/dgl_sparse] Error 2
make[1]: *** [CMakeFiles/Makefile2:211: CMakeFiles/dgl_sparse.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
``` 

Seems to be an issue with the  `torch` sub-build as seems to work with `-DBUILD_TORCH=OFF`.

hmacdope:
Perhaps some of the dgl devs can weigh in also @jermainewang @BarclayII ?

bgawrych:
@hmacdope I was able to overcome your problem with following steps:
- dmlc library need to be linked from external path in dgl_sparse library, so dgl_sparse/CMakeLists.txt needs something like this:
changing `target_link_libraries(${LIB_DGL_SPARSE_NAME} dmlc dgl)` to:
```
if (EXTERNAL_DMLC_LIB_PATH)
   # external dmlc requires OpenMP link
   include(FindOpenMP)
   if(OPENMP_FOUND)
        set(CMAKE_C_FLAGS "${OpenMP_C_FLAGS} ${CMAKE_C_FLAGS}")
        set(CMAKE_CXX_FLAGS "${OpenMP_CXX_FLAGS} ${CMAKE_CXX_FLAGS}")
   endif(OPENMP_FOUND)	
   message(STATUS "looking for dmlc library in ${EXTERNAL_DMLC_LIB_PATH}")
   find_package(dmlc REQUIRED HINTS ${EXTERNAL_DMLC_LIB_PATH})
   target_link_libraries(${LIB_DGL_SPARSE_NAME} dmlc::dmlc dgl)
else (EXTERNAL_DMLC_LIB_PATH)
   target_link_libraries(${LIB_DGL_SPARSE_NAME} dmlc dgl)
endif()
```

In DGL's CMakeLists.txt we need to pass `EXTERNAL_DMLC_LIB_PATH=${EXTERNAL_DMLC_LIB_PATH}` in line trigger dgl_sparse build (CMakeLists.txt:450)
image

And finally in `dgl_sparse/build.sh` file we need to pass `-DEXTERNAL_DMLC_LIB_PATH=$EXTERNAL_DMLC_LIB_PATH` alongside with USE_CUDA and other variables 


With these changes i was able to run successfully build-locally.py script, so i hope it helps

Let me know if I there will be any other blockers

hmacdope:
Thanks @bgawrych! I will implement the changes. 

@bgawrych I have added you as a collaborator on my branch so you can push directly. @mikemhenry you also have push privileges already.  This will hopefully make things smoother and means we can work together more effectively! Use responsibly :)

hmacdope:
Brilliant, it seems to work, just GKlib, METIS CMake and some overlinking errors to go.

mikemhenry:
Awesome!

recipes/dgl/conda-build.patch
```diff
@@ -0,0 +1,291 @@
...
+       BINDIR=${CMAKE_CURRENT_BINARY_DIR}
+       INCLUDEDIR="${DGL_INCLUDE_DIRS}"
+       CFLAGS=${CMAKE_C_FLAGS}
+diff --git a/cmake/modules/FindMETIS.cmake b/cmake/modules/FindMETIS.cmake
+new file mode 100644
+index 00000000..67182a58
+--- /dev/null
++++ b/cmake/modules/FindMETIS.cmake
+@@ -0,0 +1,28 @@
++# Find the METIS includes and library
```
hmacdope, line 201: @bgawrych any ideas how to get this to work in the conda build env? Seems to work fine locally but doesn't detect in the build env.

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - make
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
+    - dlpack ==0.7
+    - dmlc ==0.5
+    - gklib ==5.1.1
+    - metis ==5.1.0
+    - nanoflann
+    - parallel-hashmap
+    - pytorch * *{{ torch_proc_type }}*
```
jakirkham, line 43: These should all be in `host`

```suggestion
    - dlpack ==0.7
    - dmlc ==0.5
    - gklib ==5.1.1
    - metis ==5.1.0
    - nanoflann
    - parallel-hashmap
    - pytorch * *{{ torch_proc_type }}*
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - git 
+
+
+  host:
+    - python
+    - setuptools
+    - cython
```
jakirkham, line 50: Copied from above

```suggestion
    - cython
    - dlpack ==0.7
    - dmlc ==0.5
    - gklib ==5.1.1
    - metis ==5.1.0
    - nanoflann
    - parallel-hashmap
    - pytorch * *{{ torch_proc_type }}*
```

jakirkham (commented):
Thanks for the work here! 🙏

Looks like there are some dependencies that need to be moved to the `host` section. Have indicated these below

jakirkham:
Sorry meant to remove those lines from `build`. Have updated the suggestion above

mikemhenry:
@jakirkham Ah I see what you mean, let me know if it looks right now

jakirkham:
It does. Thanks Mike! 🙏

recipes/dgl/bld.bat
```diff
@@ -0,0 +1,9 @@
+REM Needs vcvars64.bat to be called
+git submodule init
+git submodule update --recursive
+md build
+cd build
+COPY %TEMP%\dgl.dll .
```
jakirkham, line 6: Could someone please provide some more context on what is happening here?

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - "$RPATH/libc10.so"
+    - "$RPATH/libtorch.so"
+  string: cuda{{ cuda_compiler_version | replace('.', '') }}py{{ CONDA_PY }}h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [cuda_compiler_version not in (undefined, 'None')]
+    - sysroot_linux-64  2.17  # [linux64]
```
jakirkham, line 32: Would suggest generalizing this for all Linuxes (like so)

```suggestion
    - sysroot_{{ target_platform }} 2.17  # [linux]
```

mikemhenry:
Okay, this should work once https://github.com/conda-forge/metis-feedstock/pull/32 is merged

mikemhenry:
since our error now is just coming from the wrong metis version:

```
[ 81%] Building CXX object CMakeFiles/dgl.dir/src/graph/metis_partition.cc.o
/home/conda/staged-recipes/build_artifacts/dgl_1688769855228/work/src/graph/metis_partition.cc: In function 'dgl::IdArray dgl::MetisPartition(dgl::GraphPtr, int, dgl::runtime::NDArray, bool)':
/home/conda/staged-recipes/build_artifacts/dgl_1688769855228/work/src/graph/metis_partition.cc:49:11: error: 'METIS_OPTION_ONDISK' was not declared in this scope; did you mean 'METIS_OPTION_CONTIG'?
   49 |   options[METIS_OPTION_ONDISK] = 1;
      |           ^~~~~~~~~~~~~~~~~~~
      |           METIS_OPTION_CONTIG
/home/conda/staged-recipes/build_artifacts/dgl_1688769855228/work/src/graph/metis_partition.cc:51:11: error: 'METIS_OPTION_NIPARTS' was not declared in this scope; did you mean 'METIS_OPTION_TPWGTS'?
   51 |   options[METIS_OPTION_NIPARTS] = 1;
      |           ^~~~~~~~~~~~~~~~~~~~
      |           METIS_OPTION_TPWGTS
/home/conda/staged-recipes/build_artifacts/dgl_1688769855228/work/src/graph/metis_partition.cc:52:11: error: 'METIS_OPTION_DROPEDGES' was not declared in this scope; did you mean 'METIS_OPTION_TPWGTS'?
   52 |   options[METIS_OPTION_DROPEDGES] = 1;
      |           ^~~~~~~~~~~~~~~~~~~~~~
      |           METIS_OPTION_TPWGTS
[ 83%] Building CXX object CMakeFiles/dgl.dir/src/graph/network.cc.o
make[2]: *** [CMakeFiles/dgl.dir/build.make:1448: CMakeFiles/dgl.dir/src/graph/metis_partition.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [CMakeFiles/Makefile2:166: CMakeFiles/dgl.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

mikemhenry:
Okay as a quick test I wanted to see if my changes to the metis recipe would allow this to build

mikemhenry:
Oh come on conda bot, I am just testing

mikemhenry:
Okay please do not merge this without removing `recipes/metis-plus-plus` :smiling_imp:

jakirkham:
Ok converting to draft while the `metis` portion is worked on

mikemhenry:
image
Weird, I am not sure what "this repository" refers to, since I have write access to the repository that made this PR (as seen with the bump ci commit) but can't mark this PR as ready for review. I know that I and @hmacdope don't have write access to `conda-forge/staged-recipes` so I might need a core member to mark this as ready, but I will wait until CI passes.

jakirkham:
Think it means this repo: https://github.com/hmacdope/staged-recipes

In particular this branch: [`hmacdope:dgl_v2`]( https://github.com/hmacdope/staged-recipes/tree/dgl_v2 )

mikemhenry:
@jakirkham is is possible to have write access to the branch in the PR but not the repo? Anyway, this can be marked as ready @hmacdope since now `micromamba --dry-run create -n foo-bar "metis=5.1.1"` works (I must have bumped the CI too soon before the package landed on the CDN)

jakirkham:
Thanks Mike! 🙏

Not really. Working with Hugo to get permissions sorted out on his repo is probably the best path given how close this is (or submitting a new PR from a repo you have access to)

Great! Didn't realize that PR was already merged 😄 

Marked as ready 🙂

jakirkham:

mikemhenry:
:crossed_fingers: Hopefully it builds.....

mikemhenry:
:tada: `[100%] Built target dgl`
But we also need to build `dgl_sparse` which doesn't seem to have an error, but all the builds just stop here:
`[ 92%] Building CXX object CMakeFiles/dgl_sparse_pytorch_1.12.1.post201.dir/src/spspmm.cc.o`

@jakirkham could we be hitting a timeout or something?

jakirkham:
Certainly possible. Could also be the error is getting swallowed. Maybe it is worth trying a verbose build?

mikemhenry:
@jakirkham what do I need to do for that?

recipes/dgl/build.sh
```diff
@@ -0,0 +1,40 @@
...
+  -DEXTERNAL_METIS_PATH=${BUILD_PREFIX}/include \
+  -DEXTERNAL_METIS_LIB_PATH=${BUILD_PREFIX}/lib \
+  -DUSE_LIBXSMM=ON \
+  -DUSE_OPENMP=ON \
+  -DCUDA_ARCH_NAME=All \
+  ${CMAKE_FLAGS} \
+  ${CUDA_CMAKE_OPTIONS} \
+  ${SRC_DIR}
+
+make -j$CPU_COUNT
```
jakirkham, line 34:
```suggestion
make -j$CPU_COUNT VERBOSE=1
```

mikemhenry:
haha thanks for that :) I will try a local build as well to see if it fails and the same spot or not

mikemhenry:
Locally I was able to get further `[100%] Built target dgl_sparse` For me locally now I am seeing this:

```
Error compiling Cython file:
------------------------------------------------------------
...
        """
        cdef DLManagedTensor* dltensor
        if self.c_is_view != 0:
            raise ValueError("to_dlpack do not work with memory views")
        CALL(DGLArrayToDLPack(self.chandle, &dltensor, alignment))
        return pycapsule.PyCapsule_New(dltensor, _c_str_dltensor, _c_dlpack_deleter)
                                                                  ^
------------------------------------------------------------

dgl/_ffi/_cython/./ndarray.pxi:80:66: Cannot assign type 'void (object) except *' to 'PyCapsule_Destructor'
```

which happens after `python setup.py install --single-version-externally-managed --record=record.txt` and I think is coming from this `[1/1] Cythonizing dgl/_ffi/_cython/core.pyx`

hadim:
Cython 3 got released recently (this week I think). Could that be this?

mikemhenry:
I think so? I'm trying another local build with `cython <3` to see if it works

hmacdope:
For future reference if anyone ever revisits this https://github.com/scipy/scipy/issues/17234 seems to be the same issue with Cython3, but we will see.

mikemhenry:
image
Local build is still going, but IDK what is going here

mikemhenry:
:point_up: this is from the most recent build log

jakirkham:
Think this is still the same issue as before ( https://github.com/conda-forge/staged-recipes/pull/22691#issuecomment-1644641276 )

Would look at the "Run docker build for CUDA <X>.<Y>" step for any job. They appear to stop somewhere close to the end of the build

Something else worth trying might be dialing down the number of threads used to build. Perhaps this is an oversubscription issue that hangs the jobs

mikemhenry:
IT WORKS! :tada: 

Locally the cython pin was all that I needed.

Could we get this merged in and then I can do a https://github.com/conda-forge/cfep/blob/main/cfep-03.md ?

recipes/dgl/build.sh
```diff
@@ -0,0 +1,40 @@
...
+  -DEXTERNAL_METIS_PATH=${BUILD_PREFIX}/include \
+  -DEXTERNAL_METIS_LIB_PATH=${BUILD_PREFIX}/lib \
+  -DUSE_LIBXSMM=ON \
+  -DUSE_OPENMP=ON \
+  -DCUDA_ARCH_NAME=All \
+  ${CMAKE_FLAGS} \
+  ${CUDA_CMAKE_OPTIONS} \
+  ${SRC_DIR}
+
+make -j$CPU_COUNT VERBOSE=1
```
jakirkham, line 34: Could we try something like this?

```suggestion
make -j2
```

jakirkham:
That's encouraging

Would like to see if we can get this to build on CI. Maybe by cutting down the number of threads?

It seems like this should be possible and should be less maintenance effort on a go forward basis

mikemhenry:
Yes it is worth a shot! 

> It seems like this should be possible and should be less maintenance effort on a go forward basis

I was under the impression that `staged-recipes` isn't exactly the same as `-feedstock` when it came to resources and build env (I remember having an issue with a GPU build of a package that went away when it became a feedstock) so I thought CI might work better in the feedstock. 

Anyway, yes I would rather see this :heavy_check_mark: in the CI so hopefully we can get it to work

mikemhenry:
Bummer, looks like it dies at the same spot. If anyone wants to test the packages, I have them uploaded here: https://anaconda.org/mmh/dgl/files (make sure strict repo priority is set up since I've got some junk in that channel, including an older `dgl` build so specify the `dgl` version like: `micromamba create -n espaloma-dgl -c conda-forge -c mmh "dgl=1.1.0"`

mikemhenry:
@jakirkham Any other ideas? Or know anyone we can tag to get some more eyes on this? One Idea I had was to maybe only do a CPU version of the build? I am not sure if that would help or not. I'm thinking about also making the matrix smaller with like, a python version and a cuda version.

mikemhenry:
Of course once we get it as a feedstock we can bump the matrix back up (or at least test it there).

jakirkham:
We dropped some content from the Docker images ( https://github.com/conda-forge/docker-images/pull/238 ) to free up space. Though Idk if that was an issue here (since we didn't get any indication of it)

Doing a CPU only job makes sense. We could iterate in the feedstock on the rest

Another option would be to trim the number of GPU archs we build on. For [example]( https://github.com/conda-forge/libmagma-feedstock/blob/81c92288644466356bab536f5fdf7d6bc604e02d/recipe/build.sh#L8-L26 )

Perhaps we could try turning off parallel builds if even using 2 threads is too much

jakirkham:
There's some additional cleanup of Azure Linux VM content in PR ( https://github.com/conda-forge/staged-recipes/pull/23437 ), which is now merged into here

Also the old CUDA versions were dropped ( https://github.com/conda-forge/staged-recipes/pull/23436 ). So this is now just on CUDA 11.2 (and 12 when the feedstock gets migrated)

mikemhenry:
I'll give this a shot:

```
# Trim down arches 
export CUDA_ARCH_LIST="sm_75,sm_80"
export CUDAARCHS="75-real;80"
````

And if that doesn't work, I will see about a CPU build

**Comment by mikemhenry** [comment]
```
[ 41%] Building NVCC (Device) object CMakeFiles/dgl.dir/src/array/cuda/dgl_generated_negative_sampling.cu.o
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
[ 41%] Building NVCC (Device) object CMakeFiles/dgl.dir/src/array/cuda/dgl_generated_rowwise_sampling.cu.o
```
Looks like that didn't work (or at least with how I tried it)

**Comment by mikemhenry** [comment]
It looks like it was able to build a python version but then died in the middle of doing another, I will see about now building one python and one cuda and see if that does the trick

**Comment by mikemhenry** [comment]
`2023-07-22T23:26:09.7690591Z ##[section]This job was abandoned. We have detected that logs from the agent may have not finished uploading. We have included our in-memory record of all log lines uploaded before we lost contact with the agent:` I noticed this at the top of the logs, maybe we are hitting some resource limit

**Comment by carterbox** [comment]
> I'll give this a shot:
> 
> ```
> # Trim down arches 
> export CUDA_ARCH_LIST="sm_75,sm_80"
> export CUDAARCHS="75-real;80"
> ```
> 
> And if that doesn't work, I will see about a CPU build

Looks like this project [implements its own CMake CUDA support](https://github.com/dmlc/dgl/blob/master/cmake/modules/CUDA.cmake), so you shouldn't expect the CUDAARCHS (a CMake specific variable) environment variable to work. CUDA_ARCH_LIST is also specific to magma, so it wouldn't either. Please check dgl project CMake modules if you want to try and control which CUDA archs are targeted.

**Comment by mikemhenry** [comment]
> > I'll give this a shot:
> > ```
> > # Trim down arches 
> > export CUDA_ARCH_LIST="sm_75,sm_80"
> > export CUDAARCHS="75-real;80"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And if that doesn't work, I will see about a CPU build
> 
> Looks like this project [implements its own CMake CUDA support](https://github.com/dmlc/dgl/blob/master/cmake/modules/CUDA.cmake), so you shouldn't expect the CUDAARCHS (a CMake specific variable) environment variable to work. CUDA_ARCH_LIST is also specific to magma, so it wouldn't either. Please check dgl project CMake modules if you want to try and control which CUDA archs are targeted.

Yes I noticed that and updated that here: https://github.com/conda-forge/staged-recipes/pull/22691/commits/2136d74a0604200acff748dbf290dfa830940971#diff-c1683d8daaef0e549ca7cdc0ab36bcdab83a31f6e6cba96b27156e029335db95R29

**Comment by mikemhenry** [comment]
I'm going to give a CPU build a shot since it seems to still have some sort of resource problem

**Comment by mikemhenry** [comment]
Could we be running out of memory or something? Locally (which works fine) when I do the CPU build I get:
```
Resource usage summary:

Total time: 0:10:37.8
CPU usage: sys=0:00:23.0, user=0:08:10.3
Maximum memory usage observed: 8.8G
Total disk usage observed (not including envs): 1.3M
```

**Comment by mikemhenry** [comment]
Looks like we ran out of space on the GPU build, I will go back to building a single arch.

**Comment by mikemhenry** [comment]
If we run out of space again, we can skip python 3.8

**Comment by mikemhenry** [comment]
Seeing things like:

```
[ 48%] Building NVCC (Device) object CMakeFiles/dgl.dir/src/graph/sampling/randomwalks/dgl_generated_randomwalk_gpu.cu.o
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
nvcc warning : The 'compute_35', 'compute_37', 'compute_50', 'sm_35', 'sm_37' and 'sm_50' architectures are deprecated, and may be removed in a future release (Use -Wno-deprecated-gpu-targets to suppress warning).
```

Makes me think there is an issue with https://github.com/dmlc/dgl/blob/master/cmake/modules/CUDA.cmake somewhere, since it looks like if I specify "Maxwell" it should only build 

```
    set(__cuda_arch_bin "50")
    set(__cuda_arch_ptx "50")
```

So we will have to see if the build runs out of space again

**Comment by mikemhenry** [comment]
Looks like 3.8 and 3.9 are done, just building 3.10 now for cuda 11.2...

File: .azure-pipelines/azure-pipelines-linux.yml [hunk 14]
```diff
@@ -21,6 +21,19 @@ jobs:
       done
     displayName: Manage disk space
 
+  - script: |
```
mikemhenry, line 24: This is more vorbose then it needs to be, see https://github.com/conda-forge/conda-smithy/pull/1751

mikemhenry:
@jakirkham this is ready for review! The only funky thing left is the swap file I create, we probably want to remove that before merge, but it won't build without it and I can sort that out when it becomes a feedstock. I didn't want to remove it until this gets reviewed so it keeps a :heavy_check_mark:

mikemhenry:
@bgawrych This should run locally fine, I've uploaded the packages to my conda channel https://anaconda.org/mmh/dgl there is some junk in there so make sure you use strict repo priority

jakirkham:
Thanks Mike! 🙏

Nice work 😄

@conda-forge-admin, please ping team (so we can get some more eyes on this)

carterbox (changes requested):

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+{% set version = "1.1.0" %}
+{% set torch_proc_type = "cuda" if cuda_compiler_version != "None" else "cpu" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/dmlc/dgl
+  git_rev: {{ version }}
```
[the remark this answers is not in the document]
mikemhenry: image

We have pulled what we can, but tensorpipe isn't on conda-forge and it also isn't actively developed anymore and has been removed from newever versions of `dgl` so we can't build from a tarball right now. Getting this error when we build from the tarball and not repo:

```
CMake Error at CMakeLists.txt:317 (add_subdirectory):
  The source directory

    /home/conda/staged-recipes/build_artifacts/dgl_1690385414680/work/third_party/tensorpipe

  does not contain a CMakeLists.txt file.
```

jakirkham:
Appears to have issues finding `GKlib.h` [on CI]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=751682&view=logs&j=eb6ecf2b-0363-5c1a-baa7-2d771889d197&t=b342d271-df1b-539d-7106-021ce5ffca81&l=2252 ):

```C
/home/conda/staged-recipes/build_artifacts/dgl_1690388553230/work/src/graph/gk_ops.cc:8:10: fatal error: GKlib.h: No such file or directory
    8 | #include <GKlib.h>
      |          ^~~~~~~~~
compilation terminated.
```

mikemhenry:
@jakirkham Yah I removed it to see if we would get lucky and it would get picked up from METIS which I know vendors it. 

RE (from @carterbox review): 

> There's an overdepending warning on gklib. Are you statically linking?

What can we do to fix that? I assumed it came from METIS shipping it, but clearly we need it it to build so I am not sure how to resolve the issue: `WARNING (dgl): run-exports library package conda-forge::gklib-5.1.1-h7f98852_2 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)`

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - git 
+
+
+  host:
+    - python =3.10
+    - setuptools
+    - cython <3
+    - dlpack ==0.7
+    - dmlc ==0.5
+    - gklib ==5.1.1
```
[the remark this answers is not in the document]
jakirkham: Appears this can't be removed otherwise the build fails. Please see this comment for details: https://github.com/conda-forge/staged-recipes/pull/22691#issuecomment-1652408390

jakirkham:
Makes sense. Added this observation to that thread 🙂

recipes/dgl/meta.yaml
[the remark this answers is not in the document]
carterbox: The trick is that when you have git submodules, each submodule becomes its own source. [example](https://github.com/conda-forge/tomopy-feedstock/blob/f314424e93713e49c93333a22a858e4fe390b5d5/recipe/meta.yaml#L33-L42)

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+    - dgl
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/dmlc/dgl
+  license: Apache-2.0
+  license_file: LICENSE
```
carterbox, line 75: Since apparently you are statically linking libraries from the third-party folder, you need to list the license files for those libaries here in addition to the DGL license.

carterbox (changes requested):

recipes/dgl/meta.yaml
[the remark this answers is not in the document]
carterbox: Any statically linked libraries need to have their license included in the license section of this recipe.

There are some possibilities I can think of:

1. DGL is linking to the static library object produced by the METIS submodule instead of the GKlib conda package (but you shouldn't need this submodule because you are using the conda package?)
2. DGL is linking to a header-only portion of GKlib?
3. DGL is missing a GKlib argument during the linking step of the build, so conda cannot detect any links to from DGL to GKlib, but at runtime an undefined symbol error will occur
4. DGL doesn't link to GKlib. It only links to METIS, but METIS mentions GKlib in its headers, so the headers have to be there at build time but not runtime.

recipes/dgl/meta.yaml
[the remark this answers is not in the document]
mikemhenry: > Alternatively, update to the next patch release which doesn't depend on tensorpipe?

We would need to re-do our big cmake patch, so I rather get this version on conda-forge, then release a new version ASAP with these improvements (hopefully after we can get the cmake changes up streamed so we don't need to that patch) 

I do like the idea of adding the modules in that way! I will switch to that, thanks!

recipes/dgl/meta.yaml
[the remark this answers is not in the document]
mikemhenry: What I will do is switch to making the sub modules explicit as recommended here https://github.com/conda-forge/staged-recipes/pull/22691#discussion_r1275477390

Then be sure to package the license files as asked to here: https://github.com/conda-forge/staged-recipes/pull/22691#discussion_r1275479206

recipes/dgl/meta.yaml
mikemhenry: Yes, thank you! I forgot I needed to do that :smiley:

mikemhenry:
This is going to get silly as I start adding deps on deps on deps

mikemhenry:
@conda-forge-admin, please ping team (so we can get some more eyes on this)

mikemhenry:
@carterbox ready for another review! Thank you for all of your help!

recipes/dgl/build.sh
```diff
@@ -0,0 +1,46 @@
...
+else
+CUDA_CMAKE_OPTIONS=""
+USE_CUDA=OFF
+fi
+
+# Remove -std=c++17 from CXXFLAGS for compatibility with nvcc
+export CXXFLAGS="$(echo $CXXFLAGS | sed -e 's/ -std=[^ ]*//')"
+export CFLAGS="$(echo $CFLAGS | sed -e 's/ -mtune=[^ ]*//')"
+CMAKE_FLAGS="${CMAKE_ARGS} -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_BUILD_TYPE=Release -DPython_EXECUTABLE=${PYTHON}"
+
```
carterbox, line 18:
```suggestion
if [[ ${cuda_compiler_version} != "None" ]]; then
    if [[ ${cuda_compiler_version} == 9.0* ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;7.0+PTX"
    elif [[ ${cuda_compiler_version} == 9.2* ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0+PTX"
    elif [[ ${cuda_compiler_version} == 10.* ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5+PTX"
    elif [[ ${cuda_compiler_version} == 11.0* ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0+PTX"
    elif [[ ${cuda_compiler_version} == 11.1 ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX"
    elif [[ ${cuda_compiler_version} == 11.2 ]]; then
        export TORCH_CUDA_ARCH_LIST="3.5;5.0;6.0;6.1;7.0;7.5;8.0;8.6+PTX"
    else
        echo "unsupported cuda version. edit build.sh"
        exit 1
    fi
fi
```
Please target the [same archs as the pytorch package](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/229b44730ae4e9b1cb144910f0a3a25cee0e7391/recipe/build_pytorch.sh#L128C32-L128C32). It looks like you're targeting [3.5,5.0,8.0,8.6+PTX](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=753751&view=logs&j=eb6ecf2b-0363-5c1a-baa7-2d771889d197&t=b342d271-df1b-539d-7106-021ce5ffca81&l=2418), which probably doesn't make sense because the package wouldn't run on any 6.x,7.x devices.

carterbox (changes requested):
Please target the same CUDA archs as the pytorch package. ignore run_exports from gklib if they aren't used.

recipes/dgl/build.sh
mikemhenry: We can try this but we might run out of disk space/system memory building all of these different arches

carterbox: If you run out of space, then use two strategies:

1. build for major versions only: 3.5;5.0;6.0;7.0;8.0+PTX
2. Cut newer versions: 3.5;5.0;6.0;7.0+PTX

mikemhenry: `All` looks pretty close to this logic 
https://github.com/dmlc/dgl/blob/master/cmake/modules/CUDA.cmake#L13-L26

mikemhenry:
@carterbox added the ignore run_exports from gklib and I think I've set the cuda arch building correctly. I was trying to build just enough arches I could test on my GPU since I was having issues with the build running out of resources building everything.

I've got it set to build "All" using the dgl cuda cmake module, which looks close to the same values as the pytorch build. I'll have to grok that cmake module more to figure out of there is an easy way to pass in `TORCH_CUDA_ARCH_LIST` directly. I thought "Manual" would do it but I couldn't quite get it to work.

recipes/dgl/build.sh
```diff
@@ -32,7 +49,7 @@ cmake -DUSE_CUDA=${USE_CUDA} \
   -DEXTERNAL_METIS_LIB_PATH=${BUILD_PREFIX}/lib \
   -DUSE_LIBXSMM=ON \
   -DUSE_OPENMP=ON \
-  -DCUDA_ARCH_NAME=Maxwell \
+  -DCUDA_ARCH_NAME=All \
```
carterbox, line 52: non-cuda build are failing tests now. Maybe add CUDA_ARCH_NAME to CUDA_CMAKE_OPTIONS so it is only used when CUDA is enabled.

carterbox (approved):
Except for builds failing now. LGTM.

recipes/dgl/build.sh
mikemhenry: Yes that was a good surprise to see. Good idea!

mikemhenry:
@carterbox should I remove https://github.com/conda-forge/staged-recipes/pull/22691/files#diff-db71f153817776d1878eba60d2c71512969cca05e702cdc382ac178b6fb316df or will that sort of thing get removed by a bot action?

mikemhenry:
@carterbox I am still working on getting the CPU builds working again, but I need that swap file for the builds to work, so after I get a green check on the build, do I just remove those changes and then we can get it merged, even if the builds don't pass?

mikemhenry:
Hmmm, checking out an older the commit works fails now, suggesting that we have an issue with our pins for the CPU version of the package.

mikemhenry:
But it is odd that the GPU version of the code is still working, since `OSError: /home/conda/staged-recipes/build_artifacts/dgl_1690904733296/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/lib/python3.10/site-packages/dgl/libdgl.so: undefined symbol: dgemm_`

seems to point to an issue with LAPACK since dgemm is a LAPACK routine

mikemhenry:
`libdgl.so` is something we build, so a symbol missing (IMHO) means that we are missing a runtime dependency now...

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+  host:
+    - python =3.10
+    - setuptools
+    - cython <3
+    - dlpack ==0.7
+    - dmlc ==0.5
+    - gklib ==5.1.1
+    - metis ==5.1.1
+    - mkl
```
jakirkham, line 77: Wonder if we could follow [this BLAS doc]( https://conda-forge.org/docs/maintainer/knowledge_base.html#blas ) to allow user BLAS selection

```suggestion
    - libblas
    - libcblas
    - liblapack
```

hmacdope: Yep, I'll give it a try. Adding MKL as a dep seems to fix things locally in test stage but not upstream so far.

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - nanoflann
+    - parallel-hashmap
+    - pytorch  # duplicate dep to pick up conda-forge pinnings
+    - pytorch * *{{ torch_proc_type }}*
```
jakirkham, line 81: Have found this syntax a bit more reliable in a range of situations

```suggestion
    - pytorch =*=*{{ torch_proc_type }}*
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+
+  run:
+    - python =3.10
+    - numpy
+    - scipy
+    - networkx
+    - requests
+    - tqdm
+    - pytorch * *{{ torch_proc_type }}*
```
jakirkham, line 90:
```suggestion
    - pytorch =*=*{{ torch_proc_type }}*
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - psutil
+    - tensorflow
+    - mkl
```
jakirkham, line 93: Would suggest using the BLAS portion mentioned above and then this can be dropped

```suggestion
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}  # [cuda_compiler_version not in (undefined, 'None')]
+    - sysroot_{{ target_platform }} 2.17  # [linux]
+    - cmake
+    - make
+    - llvm-openmp  # [osx]
+    - libgomp      # [linux]
+    - git 
+
+
```
jakirkham, line 68: nit: Would drop the extra spaces

```suggestion
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - dlpack ==0.7
+    - dmlc ==0.5
+    - gklib ==5.1.1
+    - metis ==5.1.1
+    - mkl
+    - nanoflann
+    - parallel-hashmap
+    - pytorch  # duplicate dep to pick up conda-forge pinnings
+    - pytorch * *{{ torch_proc_type }}*
+
```
jakirkham, line 82: nit: Would drop the extra spaces

```suggestion
```

recipes/dgl/meta.yaml
jakirkham: Yeah we can constrain the tests to use `mkl` separately. Please see [this example]( https://github.com/conda-forge/opencv-feedstock/blob/891639f983996e8f7cae4595b53d7293cd31058b/recipe/meta.yaml#L242 ) (though again would use `liblapack =*=*mkl` as noted elsewhere (since the `=` syntax works for more kinds of globs)

hmacdope:
Thanks for the feedback @jakirkham much appreciated. I will make the changes asap, just waiting for CI to finish so we have a record that this commit worked fully

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+    - scipy
+    - networkx
+    - requests
+    - tqdm
+    - pytorch =*=*{{ torch_proc_type }}*
+    - psutil
+    - tensorflow
+
+
+test:
```
jakirkham, line 94: If the tests need MKL, may want to add something like this 

```suggestion
test:
  requires:
    - liblapack =*=*mkl
```

recipes/dgl/meta.yaml
mikemhenry: So the tests don't need `mkl` explicitly, but we need a runtime BLAS implementation. It isn't clear to me if upstream supports different implementations. Looking here https://github.com/dmlc/dgl/blob/a1a3ce89ed752d52fdb8681d266c486d40d78007/docker/install/ubuntu_install_core.sh#L4 it appears they use libopenblas for their docker image. If I don't need to pin an implementation, can I just use `libblas` as a run time dep?

recipes/dgl/meta.yaml
mikemhenry: I am running it locally right now to save some CI time, but all the tests do is an import, so if I don't need to pin a blas implementation, what do I need to add `run`? Right now I am testing to see if `- libblas` is enough.

mikemhenry: See my question here: https://github.com/conda-forge/staged-recipes/pull/22691#discussion_r1284720073

mikemhenry:
I'm just worried that without `liblapack =*=*mkl` or something in `run` users will run into the same error, unless there is something different about the way the testing env is created and when a user installs this package, some blas implementation will be pulled in automatically

recipes/dgl/meta.yaml
jakirkham: Yeah the way this works is we build against NETLIB and then change the actual library used at runtime. OpenBLAS is used by default

Also all these libraries have `run_exports`. So adding the in `host` is sufficient

mikemhenry:
@jakirkham thank you for your answer here: https://github.com/conda-forge/staged-recipes/pull/22691#discussion_r1284740195 I will reply here since I don't want to loose this conversation.

Okay so I just pushed the commit that adds mkl to the test env. Do I need to make any changes to `run` or `host`? Or should it all be good now?

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - python =3.10
+    - setuptools
+    - cython <3
+    - dlpack ==0.7
+    - dmlc ==0.5
+    - gklib ==5.1.1
+    - metis ==5.1.1
+    - libblas
+    - libcblas
+    - liblapack
```
jakirkham, line 77: Sorry had missed this one

```suggestion
    - liblapack
    - liblapacke
```

jakirkham:
We can try relaxing the MKL constraint after. Am trying to get us back to a place where CI is passing first

There are ways to constrain the BLAS implementations supported if it comes to that (kind of hoping that won't be needed)

Had one change above. Sorry for missing that. Otherwise think we should be good to wait for CI

mikemhenry:
@jakirkham no worries! I saw that it was missing from the docs you linked but I wasn't sure if it was an intentional omission.

mikemhenry:
I am now trying locally a build where I don't have anything added to the test env to see if adding `liblapacke` is enough

mikemhenry:
Without `- liblapack =*=*mkl` in `test` we still get `libdgl.so: undefined symbol: dgemm_`

So I know that having `- liblapack =*=*mkl` in `test` gets CI to pass, but I am worried about users running into this issue.

I uploaded the built package and then installed it, and get `libdgl.so: undefined symbol: dgemm_` locally. 

Then if I install `liblapack =*=*mkl`

```
   - liblapack=*[build=*mkl]


  Package            Version  Build                        Channel           Size
───────────────────────────────────────────────────────────────────────────────────
  Install:
───────────────────────────────────────────────────────────────────────────────────

  + libhwloc           2.9.2  nocuda_h7313eea_1008         conda-forge        3MB
  + libiconv            1.17  h166bdaf_0                   conda-forge     Cached
  + libxml2           2.11.4  h0d562d8_0                   conda-forge     Cached
  + llvm-openmp       16.0.6  h4dfa4b3_0                   conda-forge     Cached
  + mkl             2022.2.1  h84fe81f_16997               conda-forge     Cached
  + rocm-smi           5.6.0  h59595ed_1                   conda-forge        4MB
  + tbb            2021.10.0  h00ab1b0_0                   conda-forge      186kB

  Remove:
───────────────────────────────────────────────────────────────────────────────────

  - nomkl                1.0  h5ca1d4c_0                   conda-forge     Cached

  Change:
───────────────────────────────────────────────────────────────────────────────────

  - _openmp_mutex        4.5  2_gnu                        conda-forge     Cached
  + _openmp_mutex        4.5  2_kmp_llvm                   conda-forge     Cached
  - libblas            3.9.0  17_linux64_openblas          conda-forge     Cached
  + libblas            3.9.0  16_linux64_mkl               conda-forge     Cached
  - libcblas           3.9.0  17_linux64_openblas          conda-forge     Cached
  + libcblas           3.9.0  16_linux64_mkl               conda-forge     Cached
  - liblapack          3.9.0  17_linux64_openblas          conda-forge     Cached
  + liblapack          3.9.0  16_linux64_mkl               conda-forge     Cached
  - liblapacke         3.9.0  17_linux64_openblas          conda-forge     Cached
  + liblapacke         3.9.0  16_linux64_mkl               conda-forge     Cached
  - pytorch            2.0.0  cpu_generic_py310h7ffd2bf_1  conda-forge     Cached
  + pytorch            2.0.0  cpu_mkl_py310h402c8e3_101    conda-forge       71MB
```

The error goes away.

mikemhenry:
So I don't know why, but it looks like we are dependent on the mkl build of blas, despite not having that in our host env explicitly.

jakirkham:
Am confused by the `_`. The BLAS symbol name is `dgemm` (no `_`). Do we know where it is coming up with that?

mikemhenry:
If you want to play around with this, this command will install `dgl` (the build that came from CI)
`micromamba create -n dlg-cpi -c conda-forge -c "mmh/label/cpu-test" dgl` 

and `python -c "import dgl"` should give you an error unless you first do `micromamba install "liblapack =*=*mkl"`

hmacdope:
Can we just use MKL for now and then unwind the issues in a feedstock @mikemhenry  @jakirkham?

mikemhenry:
I am going to try moving `liblapack =*=*mkl` to `run:` and see if that works (locally) and if it does, push it to CI and we can figure out a better way after we get this merged

mikemhenry:
Okay so once CI passes, I will remove the stuff I did for the swapfile, that might break the build but then it should be good to go...

mikemhenry:

mikemhenry:
Okay so this might fail without the swap file stuff, but this is ready to merge :smile: @carterbox

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - liblapacke
+    - metis ==5.1.1
+    - nanoflann
+    - parallel-hashmap
+    - python =3.10
+    - pytorch  # duplicate dep to pick up conda-forge pinnings
+    - pytorch =*=*{{ torch_proc_type }}*
+    - setuptools
+  run:
+    - liblapack =*=*mkl
```
jakirkham, line 84: Maybe we should add a comment here?

Can also raise an issue in the feedstock to track

```suggestion
    # Currently tests fail with other BLASes
    # Pin to a known working BLAS.
    - liblapack =*=*mkl
```

recipes/dgl/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - dmlc ==0.5
+    - gklib ==5.1.1
+    - libblas
+    - libcblas
+    - liblapack
+    - liblapacke
+    - metis ==5.1.1
+    - nanoflann
+    - parallel-hashmap
+    - python =3.10
```
jakirkham, line 79: Was this just due to CI limits?

mikemhenry: Good catch! Yes! I will fix that now

mikemhenry:
@jakirkham this is ready to merge, we won't get a green check without the swap fix

jakirkham:
Thanks Mike! 🙏

LGTM

@carterbox did you have any more thoughts here?

Also Mike it might be worth sending a separate PR to add the swapfile changes to `staged-recipes` if you are up for it 🙂

mikemhenry:
> Also Mike it might be worth sending a separate PR to add the swapfile changes to staged-recipes if you are up for it 

Will do!

jakirkham:
Thanks all! 🙏

PR #22830, opened 2023-05-16
Dymoval: first attempt ever in making a conda-forge package

ubaldot:
Checklist
- [x] Build number is 0.


ubaldot:
One more thing: I stretching a bit over adding pytest. 
I have tests/ folder included in the sdist (tar.gz file) and I added in the meta.yml file the following:

```
test:
  imports:
    - dymoval
  commands:
    - pip check
    - pytest --pyargs dymoval
  requires:
    - pip
    - pytest
```

However, looking at the logs zero tests are executed. How to fix that?

ocefpaf:
> but that has been specified in dymoval's dependencies.
> I am a bit confused: where such a dependency shall be specified ?

pathlib is in the standard library since Python 3.4, right? I think the package you added is aimed for Python <3.4.
For conda-forge you don't need to add that b/c we build >=3.7 only. With that said, upstream lists pathlib as a dependency and pip check will try to find it but miss it b/c it won't be installed as a third party library. You can:

1. skip pip check for now
2. Remove pathlib from https://github.com/VolvoGroup/dymoval/blob/b233037b125411e112c21d1dcf5d8ebd9fd83b28/pyproject.toml#L19 upstream

ubaldot:
@ocefpaf Thanks Felipe. I have removed the dependency of pathlib. However, there is stil something not working. 
One test includes in ./tests failed in different ways which is strange because it went through the github workflow, see [here](https://github.com/ubaldot/dymoval/actions/runs/4996581810).

The errors that I get on the Azure pipelines are the following: 

Linux-64:
FileNotFoundError: [Errno 2] No such file or directory: 'xdg-open'

Windows-64:
conda.CondaMultiError: Downloaded bytes did not match Content-Length

MacOSX:
Everything run fine.

You can verify yourself. :) 

How to solve these issues?

ocefpaf:
> Linux-64: FileNotFoundError: [Errno 2] No such file or directory: 'xdg-open'

This library is relying on `xdg-utils`, a Linux tool, for that test. You should skip that test in the CI b/c it is not available in the image nor as a conda package (at least that I know of).

> Windows-64: conda.CondaMultiError: Downloaded bytes did not match Content-Length

This seems like a temporary network issue and it is passing now.

ubaldot:
@ocefpaf Thanks again.
Very good new for Windows, but I am a bit puzzled for Linux. 
I think the runner should have xdg-open, don't you think? :) 

ocefpaf:
> I think the runner should have xdg-open, don't you think? :)

Nope. CI runners must be as "slim" as possible to avoid assuming stuff and/or conflicts with links and calls from pre-existing packages. Ideally, everything should be explicitly requested for a CI runner.

ubaldot:
@ocefpaf Ok! I though that xdg-open was a quite standard command for  Linux, otherwise I agree that it is better to keep the runners as slim as possible :) 
Ok, now it seems that all the checks have passed!

PR #22903, opened 2023-05-24
BUG: Missing cuda keys

carterbox:
The keys `cuda_compiler` and `cuda_compiler_version_min` were added to the conda-forge pinnings, but they are not added here in staged recipes. Without them, CUDA build jobs will try to install the `none_linux-64` package instead of `nvcc_linux-64`. Also added `cuda_compiler_version_min` to be consistent with the pinnings, and since I don't know how it is used.

EDIT: I now realize that the keys in staged recipes override the ones from the pinnings, so the error is due to the overrided keys here disrupting the key zipping from the pinnings. Thus, I should include all cuda related keys that zip with `cuda_compiler_verion`.

~~Also added a line to the build steps to squash an error from running the build-locally script.~~
EDIT: Removed docker permissions related changes

Related to: https://github.com/conda-forge/staged-recipes/pull/22898 , https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/4407

### DIFF
--- a/.ci_support/linux64_cuda111.yaml
+++ b/.ci_support/linux64_cuda111.yaml
@@ -22,3 +22,7 @@ docker_image:
 - quay.io/condaforge/linux-anvil-cuda:11.1
 cuda_compiler_version:
 - 11.1
+cuda_compiler:
+- nvcc
+cuda_compiler_version_min:
+- 10.2
--- a/.ci_support/linux64_cuda111.yaml
+++ b/.ci_support/linux64_cuda111.yaml
@@ -26,3 +26,5 @@ cuda_compiler:
 - nvcc
 cuda_compiler_version_min:
 - 10.2
+cudnn:
+- 8
--- a/.ci_support/linux64.yaml
+++ b/.ci_support/linux64.yaml
@@ -18,3 +18,5 @@ cuda_compiler:
 - None
 cuda_compiler_version_min:
 - None
+cudnn:
+- undefined
--- a/.ci_support/linux64.yaml
+++ b/.ci_support/linux64.yaml
@@ -14,3 +14,7 @@ docker_image:
 - quay.io/condaforge/linux-anvil-cos7-x86_64
 cuda_compiler_version:
 - None
+cuda_compiler:
+- None
+cuda_compiler_version_min:
+- None
--- a/.ci_support/linux64_cuda102.yaml
+++ b/.ci_support/linux64_cuda102.yaml
@@ -24,3 +24,5 @@ cuda_compiler:
 - nvcc
 cuda_compiler_version_min:
 - 10.2
+cudnn:
+- 7
--- a/.ci_support/linux64_cuda102.yaml
+++ b/.ci_support/linux64_cuda102.yaml
@@ -20,3 +20,7 @@ docker_image:
 - quay.io/condaforge/linux-anvil-cos7-cuda:10.2
 cuda_compiler_version:
 - 10.2
+cuda_compiler:
+- nvcc
+cuda_compiler_version_min:
+- 10.2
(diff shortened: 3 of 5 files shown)


.ci_support/linux64_cuda111.yaml
```diff
@@ -22,3 +22,7 @@ docker_image:
 - quay.io/condaforge/linux-anvil-cuda:11.1
 cuda_compiler_version:
 - 11.1
+cuda_compiler:
+- nvcc
+cuda_compiler_version_min:
+- 10.2
```
jakirkham, line 28:
```suggestion
- 10.2
cudnn:
- 8
```

carterbox: Thanks!

.scripts/build_steps.sh
```diff
@@ -38,6 +38,7 @@ pushd /home/conda/staged-recipes/recipes > /dev/null
 if [ "${AZURE}" == "True" ]; then
     git fetch --force origin main:main
 fi
+git config --global --add safe.directory /home/conda/staged-recipes
```
jakirkham, line 41: Why is this needed?

carterbox:
```
+ echo 'Finding recipes merged in main and removing them from the build.'
Finding recipes merged in main and removing them from the build.
+ pushd /home/conda/staged-recipes/recipes
+ '[' False == True ']'
+ git ls-tree --name-only main -- .
+ xargs -I '{}' sh -c 'rm -rf ~/staged-recipes-copy/recipes/{} && echo Removing recipe: {}'
fatal: detected dubious ownership in repository at '/home/conda/staged-recipes'
To add an exception for this directory, call:

	git config --global --add safe.directory /home/conda/staged-recipes
  File "/home/dching/Documents/staged-recipes/build-locally.py", line 101, in <module>
    main()
  File "/home/dching/Documents/staged-recipes/build-locally.py", line 95, in main
    run_docker_build(ns)
  File "/home/dching/Documents/staged-recipes/build-locally.py", line 33, in run_docker_build
    subprocess.check_call([script])
  File "/usr/lib64/python3.11/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['.scripts/run_docker_build.sh']' returned non-zero exit status 1.
```
I get the following error message when using the `build-locally.py`

jakirkham: Thanks for the clarification 🙏

Could we please split this out as a separate issue or PR?

It seems unrelated to the changes here (and we may want to evaluate our options for solving it including whether the Docker images themselves need changes)

carterbox: Yes

jakirkham:
Thanks Daniel! 🙏

.scripts/build_steps.sh
jakirkham: Sounds good. Please feel free to link an issue here (once opened). Agree we should figure out some kind of solution. Think it may even be needed in feedstocks

vicentebolea: How did you guys get around to this issue I am hitting the same thing when using the build_locally.py script

carterbox: Add the topic line of this conversation to the build_steps script.

vicentebolea: What I wonder is that this change cannot make it to the feedstock project, thus I would have to add it anytime I want to build locally, would there be an alternative?

jakirkham: If the feedstock hasn't been updated in a while, please make sure to [re-render]( https://conda-forge.org/docs/maintainer/updating_pkgs.html#dev-update-rerender )

PR #22982, opened 2023-05-31
Add kalpy recipe

mmcauliffe:
Checklist
- [x] Build number is 0.


mmcauliffe:
@h-vetinari This is for pybind11 bindings on the shared libraries of kaldi, if you could take a look over the recipe and https://github.com/mmcauliffe/kalpy, definitely appreciate any feedback! I have the low level bindings working along with some higher level pythonic classes/functions that I'll be developing out further, but wanted to get the ball rolling on distribution.

recipes/kalpy/bld.bat
```diff
@@ -0,0 +1,7 @@
+@echo on
+
+
+set "CXXFLAGS=%CXXFLAGS% /D_LIBCPP_DISABLE_AVAILABILITY"
```
h-vetinari, line 4: We don't need this on windows, that's OSX-only

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,65 @@
+{% set name = "kalpy" %}
+{% set version = "0.0.4" %}
+
+package:
+  name: kalpy
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/k/kalpy-kaldi/kalpy-kaldi-{{ version }}.tar.gz
+  sha256: 43cf28d35a97751d4f1b40f133897c67f66dfd81bb0d8d010c826998cca2b49e
```
h-vetinari, line 10: Can you explain the situation a little. On the README of , I see:
> Kalpy is also available on pip via the mfa-kaldi package

what's the relationship between kalpy and mfa-kaldi? I guess kalpy is the more fully-featured one (includes the lib)? Should we add an output `mfa-kaldi` here that just wraps `kalpy` to ensure that no-one gets the idea of packaging `mfa-kalpi` separately?

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+build:
+  script:
+    - CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY" {{ PYTHON }} -m pip install . --no-deps -vv  # [not win]
+  number: 0
+  skip: True  # [py<38]
+  script_env:
+   - KALDI_ROOT={{ environ.get('LIBRARY_PREFIX') }}  # [win]
+   - KALDI_ROOT={{ environ.get('PREFIX') }}  # [not win]
```
h-vetinari, line 19: I think this is easier - less indirections.
```suggestion
   - KALDI_ROOT=%LIBRARY_PREFIX%  # [win]
   - KALDI_ROOT=$PREFIX           # [unix]
```

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - python                                  # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}      # [build_platform != target_platform]
+    - cmake
+    - ninja
+    - m2-patch  # [win]
+    - libtool   # [not win]
+    - automake  # [not win]
+    - make  # [not win]
```
h-vetinari, line 32: I see you have a CMakeLists.txt in the repo. What are all these autotools-deps for? What do we need to patch on windows?

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - pybind11
+    - setuptools
+    - setuptools_scm
+  host:
+    - python
+    - pip
+    - kaldi * cpu*
```
h-vetinari, line 39: Why does this depend on the CPU builds? What would go wrong if someone tried to use the CUDA builds? FTR, I don't think we should constrain this.
```suggestion
    - kaldi
```

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - openfst
+  run:
+    - python
+    - pynini
+    - kaldi * cpu*
+    - openfst
+    - numpy
+    - librosa
```
h-vetinari, line 47: openfst has a [run-export](https://github.com/conda-forge/openfst-feedstock/blob/main/recipe/meta.yaml#L24), so this gest done by the build. Kaldi doesn't have one yet, but we should (eventually) do that.

Numpy should either have a lower bound or a `*`, otherwise conda smithy will think this recipe needs to be built for different numpy versions (but I don't see it as a host-dependence, so that shouldn't be necessary).

```suggestion
    - kaldi
    - numpy >=1.x
    - librosa
```

h-vetinari (commented):
Cool, thanks a lot for the PR & the ping!

Some comments

recipes/kalpy/meta.yaml
mmcauliffe: Oh this is just a typo left over from a previous one.  Just plain `kalpy` is too close to some other package on pypi, I think `kaipy` is the conflicting one.  So I was struggling with what to name it, but at the moment it's `kalpy-kaldi` which should help with discoverability considering the number of other binding libraries on pip at various stages of abandonment.  In this case, should the conda recipe output both `kalpy` and `kalpy-kaldi`?  Should I figure out another name that allows for more transparency and consistency across both systems?

I'm also on the fence about having it on pypi to begin with, because it does basically require having the shared libraries from the kaldi conda package installed anyway.

The three options I was considering were:

1. (Current set up) Have just sdist on Pypi, installation requires having kaldi shared libraries where KALDI_ROOT is pointing.  Upside is that I can list kalpy in the dependencies for montreal-forced-aligner, but pip installation is not as easy or fully packaged as I would like.  I haven't really looked into the work required to get wheels working for each platform that would not depend on the conda package, but that would simplify pip installation.
2. No package on pypi, just a conda package, which would make more sense in my mind because having a pip package depend on conda packages seems like a broken experience, also doesn't get the autodetection of new versions.
3. Have kalpy be an output of the kaldi feedstock, since they are largely just bindings for the kaldi libraries, but I think this option probably complicates that feedstock too much.

Do you have any thoughts on best practices here?

recipes/kalpy/meta.yaml
mmcauliffe: Yeah that makes sense, I was getting some errors when I didn't constrain it, but I can try another pass with them now that I've fixed the other blocking errors.

recipes/kalpy/meta.yaml
h-vetinari: > Do you have any thoughts on best practices here?

I can only really comment on conda-forge, distribution outside of that is up to the respective maintainers. I agree with much of your reasoning (pypi packages depending on conda-forge makes no sense, "better" [for PyPI] would be vendoring libkaldi in the kalpy wheel), but it doesn't really matter for our purposes here.

Also, IIUC `mfa-kaldi` in the readme was a leftover from something else (it's not there anymore in the repo), so that sounds fine as well.

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+build:
+  script:
+    - CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY" {{ PYTHON }} -m pip install . --no-deps -vv   # [not win]
+    - {{ PYTHON }} -m pip install . --no-deps -vv                                                         # [win]
+  number: 0
+  skip: True  # [py<38]
+  script_env:
+   - KALDI_ROOT={{ environ.get('LIBRARY_PREFIX') }}  # [win]
+   - KALDI_ROOT=$PREFIX           # [not win]
```
h-vetinari, line 20: `script_env` is not really so much in use. rather than work around bugs there (e.g. if `%LIBRARY_PREFIX%` doesn't work out of the box), it would IMO be easier to just add it to the script section.

```suggestion
  script:
   - export KALDI_ROOT="$PREFIX"        # [unix]
   - set "KALDI_ROOT=%LIBRARY_PREFIX%"  # [win]
    - CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY" {{ PYTHON }} -m pip install . --no-deps -vv   # [not win]
    - {{ PYTHON }} -m pip install . --no-deps -vv                                                         # [win]
  number: 0
  skip: True  # [py<38]
```

recipes/kalpy/meta.yaml
h-vetinari: It's still there in the host requirements...

mmcauliffe:
@h-vetinari I'm still working on figuring out what's going wrong with windows builds, but linux is giving the same message `ImportError: libcuda.so.1: cannot open shared object file: No such file or directory`.  It doesn't look like it is in the cudatoolkit, so that makes sense, but I'm confused how it gets linked to it in that case.  Do we need to have a cuda build matrix for kalpy like for kaldi to have libcuda.so on the build machine?  Is there something wrong in how Kaldi is linking to the cuda libraries (ref: https://github.com/kaldi-asr/kaldi/blob/master/CMakeLists.txt#L188)?

h-vetinari:
> linux is giving the same message `ImportError: libcuda.so.1: cannot open shared object file: No such file or directory`.

That looks like it needs to load the cuda driver, which we cannot distribute, but which is part of the cuda image that gets used when we build something with `{{ compiler('cuda') }}`. There are IMO two choices:
* use the cpu-version of kaldi in the test phase
* use a "fake" `{{ compiler('cuda') }}` in the build section, so that we use the cuda image by default.

I think the first approach is easier...

mmcauliffe:
@h-vetinari ok looks like everything's passing again if you want to take another look!

recipes/kalpy/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_compiler:                    # [win]
+  - vs2019                     # [win]
+cxx_compiler:                  # [win]
+  - vs2019                     # [win]
```
h-vetinari, line 4: Delete this file (we've moved to default vs2019 already across all of conda-forge)

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - python
+    - pip
+    - kaldi
+    - openfst
+  run:
+    - python
+    - pynini
+    - kaldi
+    - numpy >=1.x
```
h-vetinari, line 37: Numpy is _only_ a run-dependence? Also, you cannot use "1.x" (or, at least, I have no idea how conda would interpret that, it's not a valid version). Conda-forge is already globally building only for numpy>=1.21 anymore, anything below that would not be installed without forcing it anyway.
```suggestion
    - numpy
```

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - setuptools_scm
+  host:
+    - python
+    - pip
+    - kaldi
+    - openfst
+  run:
+    - python
+    - pynini
+    - kaldi
```
h-vetinari, line 36: Kaldi has a run-export, so you can delete this line here.

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - numpy >=1.x
+    - librosa
```
h-vetinari, line 38: I find it really confusing that a python package has a name starting with "lib", but whatever...

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+test:
+  requires:
+    - kaldi * cpu*
```
h-vetinari, line 42: Please add a comment why we're adding this requirement. Also, prefer `=` over semantically relevant whitespace.
```suggestion
  requires:
    # the cuda variant of kaldi requires loading libcuda.so, which we don't have in regular CI
    - kaldi =*=cpu*
```

recipes/kalpy/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -ex
+
+export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+export KALDI_ROOT=$PREFIX
+$PYTHON -m pip install . --no-deps -vv
```
h-vetinari, line 6: Missing linebreak

recipes/kalpy/bld.bat
```diff
@@ -0,0 +1,6 @@
+@echo on
+
+set "KALDI_ROOT=%LIBRARY_PREFIX:\=/%"
+
+%PYTHON% -m pip install . --no-deps -vv
+if %ERRORLEVEL% neq 0 exit 1
```
h-vetinari, line 6: Missing line break

h-vetinari (commented):
Looking pretty good, some remaining nits

recipes/kalpy/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  run:
+    - python
+    - pynini
+    - numpy
+    - librosa
+
+test:
+  requires:
+    # the cuda variant of kaldi requires loading libcuda.so, which we don't have in regular CI
+    - kaldi * cpu*
```
h-vetinari, line 42:
```suggestion
    - kaldi =*=cpu*
```

h-vetinari (approved):
LGTM, thanks a lot!

PTAL @conda-forge/help-python-c

h-vetinari:
@conda-forge/help-python-c, this should be ready (I had a look already, though of course a second review would help, and I don't have the commit bit here), PTAL[^1]!

[^1]: I've noticed that the content of the review-summary often does not cause the same actions as a regular comment, e.g. no xref created, no bot triggered etc.; sorry for the duplicated ping

mmcauliffe:
@conda-forge/help-python-c reminder ping to see if I can get this reviewed and merged.  Thanks!

h-vetinari:
I already approved, but recently, I gained the power to push the button here. :)
This one is good to go.

mmcauliffe:
Awesome, thanks so much!!

PR #23000, opened 2023-06-02
Adding cuda-sanitizer-api recipe

adibbley:
xref: #21382

Checklist
- [x] Build number is 0.


recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  sha256: 80e23d82bf4a0bb79c564104121467bd0d15f0142e1dc4fff36bb491bc4b3d4f  # [linux64]
+  sha256: 50d66cab58c7b8aa8473b36844669ee2ea7fc85ad92d49473b30214de22c2d25  # [ppc64le]
+  sha256: c872f56fa4a0ce50f8acf5f154d73d0c7a3290e9fa7ffa0b368fbdae3d1435b9  # [aarch64]
+  sha256: cdb48694de35aed5576aaf81f2443d12791a346d7a186e2c37e7606e252ef4e4  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - '$RPATH/libInterceptorInjectionTarget.so'
```
jakirkham, line 26: As these libraries are not in the package, can we please try dropping this?

```suggestion
```

recipes/cuda-sanitizer-api/bld.bat
```diff
@@ -0,0 +1,3 @@
+if not exist %PREFIX% mkdir %PREFIX%
+
```
jakirkham, line 2: Think this is needed on Windows too (like what was done on Linux)

```suggestion

rmdir /q /s compute-sanitizer\x86

```

jakirkham (commented):
Thanks Alex! 🙏

Had a couple questions below

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - cuda-version {{ cuda_version }}
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/compute-sanitizer/compute-sanitizer                   # [linux]
```
bdice, line 42: Is this the right location for these files? Does anything need to be in `$PREFIX/bin/`?

Should we try to execute this with something like `compute-sanitizer --version`, or does that require a GPU?

Are there other files we should test that would go in `targets/.../`?

bdice (commented):
A few questions, but generally LGTM.

recipes/cuda-sanitizer-api/meta.yaml
adibbley: I had tried this without and it gives an error. I assume the issue is the library being installed in `$PREFIX/compute-sanitizer`. We don't want this library on any PATH either, as it is only used by this tool (and it knows where to look).

recipes/cuda-sanitizer-api/meta.yaml
adibbley: There is a `bin/compute-sanitizer` being installed, I'll add a test for that as well.

`bin/compute-sanitizer` searches for `../compute-sanitizer`, so this will need to be the layout of this component.

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/compute-sanitizer                                 # [linux]
+    - test -f $PREFIX/compute-sanitizer/compute-sanitizer                   # [linux]
+    - if not exist %PREFIX%\compute-sanitizer\compute-sanitizer.exe exit 1  # [win]
+    - compute-sanitizer --version
```
jakirkham, line 46: Do we need to add some soft (or hard) links in the install scripts above for these?

adibbley: Which are you referring to here? If you were looking at the windows failure, I updated the test to call the `.exe`.

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/compute-sanitizer                                 # [linux]
+    - test -f $PREFIX/compute-sanitizer/compute-sanitizer                   # [linux]
+    - compute-sanitizer --version                                           # [linux]
+    - if not exist %PREFIX%\compute-sanitizer\compute-sanitizer.exe exit 1  # [win]
+    - compute-sanitizer\compute-sanitizer.exe --version                     # [win]
```
jakirkham, line 47: Would it be possible to do something like this? May require adding a hard link in the Windows case

```suggestion
    - compute-sanitizer --version
```

recipes/cuda-sanitizer-api/meta.yaml
jakirkham: Added a comment on that below ( https://github.com/conda-forge/staged-recipes/pull/23000#discussion_r1230181880 ). Let's discuss there

recipes/cuda-sanitizer-api/meta.yaml
adibbley: Added a hardlink in the latest commit.

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - '$RPATH/libInterceptorInjectionTarget.so'  # [linux]
+    - '$RPATH/InterceptorInjectionTarget.dll'    # [win]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
vyasr, line 32: Do we need these? There's no actual compilation happening at the build step, is there?

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+  host:
+    - cuda-version {{ cuda_version }}
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
vyasr, line 39: Is `pin_compatible` supported in `run_constrained`, and if so should we be using it, or do we not care about a specific version and only that the right ARM variant is present?

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - compute-sanitizer --version
+
+about:
+  home: https://docs.nvidia.com/compute-sanitizer/SanitizerApiGuide/index.html
+  license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: Provides a set of APIs to enable third party tools to write GPU sanitizing tools
+  description: |
+    Provides a set of APIs to enable third party tools to write GPU sanitizing
```
vyasr, line 56:
```suggestion
    Provides a set of APIs to enable third-party tools to write GPU sanitizing
```

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  host:
+    - cuda-version {{ cuda_version }}
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/compute-sanitizer                                         # [linux]
```
vyasr, line 43: Is this a symlink to `compute-sanitizer/compute-sanitizer`, or a separate executable? If it's a link should we be using `test -L`?

vyasr (changes requested):
Couple of small questions.

recipes/cuda-sanitizer-api/meta.yaml
vyasr: Hmm it looks like it's a two-line bash script that relies on the relative path, but isn't a direct symlink. Not sure why it's designed that way but looks like `test -f` is indeed the right check here.

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - '$RPATH/libInterceptorInjectionTarget.so'  # [linux]
+    - '$RPATH/InterceptorInjectionTarget.dll'    # [win]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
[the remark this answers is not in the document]
adibbley: We set the supported `arm_variant_type` in [conda_build_config.yaml](https://github.com/conda-forge/staged-recipes/pull/23000/files#diff-a490e74b94b7b5543425cac3521e8c749e5cce478e191dcad0b50b718d7dbe59)

John can probably explain better, but I would expect this to match any version including the [build string](https://github.com/conda-forge/arm-variant-feedstock/blob/main/recipe/meta.yaml#L7)

recipes/cuda-sanitizer-api/meta.yaml
adibbley: `arm-variant` only has 1 version [available](https://anaconda.org/conda-forge/arm-variant/files), and we only care about the  type AFAICT.

recipes/cuda-sanitizer-api/meta.yaml
adibbley: They followed some other devtools design, which use the bash script for more involved install layouts which aren't used by `compute-sanitizer`.

recipes/cuda-sanitizer-api/meta.yaml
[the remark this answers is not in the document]
vyasr: @jakirkham could you clarify here?

vyasr (approved):
I don't have permissions to resolve threads here, but aside from one open question regarding build strings this looks good to me.

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/compute-sanitizer                                         # [linux]
+    - test -f $PREFIX/compute-sanitizer/compute-sanitizer                           # [linux]
+    - if not exist %LIBRARY_PREFIX%\compute-sanitizer\compute-sanitizer.exe exit 1  # [win]
+    - if not exist %LIBRARY_BIN%\compute-sanitizer exit 1                           # [win]
+    - compute-sanitizer --version
```
bdice, line 47: This test appear to be failing on Windows.

```
(%PREFIX%) %SRC_DIR%>compute-sanitizer --version 
'compute-sanitizer' is not recognized as an internal or external command,
operable program or batch file.
```

recipes/cuda-sanitizer-api/bld.bat
```diff
@@ -0,0 +1,7 @@
+if not exist %PREFIX% mkdir %PREFIX%
```
jakirkham, line 1:
```suggestion
if not exist %PREFIX% mkdir %PREFIX%
if not exist %LIBRARY_PREFIX% mkdir %LIBRARY_PREFIX%
if not exist %SCRIPTS% mkdir %SCRIPTS%
```

recipes/cuda-sanitizer-api/bld.bat
```diff
@@ -0,0 +1,7 @@
...
+
+rmdir /q /s compute-sanitizer\x86
+
+move compute-sanitizer %LIBRARY_PREFIX%
+
+mklink /h %LIBRARY_BIN%\compute-sanitizer %LIBRARY_PREFIX%\compute-sanitizer\compute-sanitizer.exe
```
jakirkham, line 7:
```suggestion
:: Create bat links for all exe files in %LIBRARY_PREFIX%\compute-sanitizer
cd %SCRIPTS%
for /r "%LIBRARY_PREFIX%\compute-sanitizer" %%f in (*.exe) do (
    echo @echo off > %%~nf.bat
    echo %%~dp0.\compute-sanitizer\%%~nf.exe %%* >> %%~nf.bat
    if errorlevel 1 exit 1
)
```

jakirkham (commented):
Thanks Alex! 🙏

Maybe we can try adding wrapper scripts to call the `.exe`s (as below). This is a strategy we have used for similar situations in the past (though it is always possible the code below needs additional tweaks)

recipes/cuda-sanitizer-api/meta.yaml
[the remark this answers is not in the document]
jakirkham: Yeah the syntax above works in `meta.yaml`. It is equivalent to `=` based syntax one would use when calling on the command line

recipes/cuda-sanitizer-api/meta.yaml
jakirkham: Yeah think we may need a different approach than adding links. Made a suggestion below to write wrapper scripts to call each executable

recipes/cuda-sanitizer-api/bld.bat
```diff
@@ -0,0 +1,15 @@
+if not exist %PREFIX% mkdir %PREFIX%
+if not exist %LIBRARY_PREFIX% mkdir %LIBRARY_PREFIX%
+if not exist %SCRIPTS% mkdir %SCRIPTS%
+
+rmdir /q /s compute-sanitizer\x86
+
+move compute-sanitizer %LIBRARY_PREFIX%
```
jakirkham, line 7:
```suggestion
move compute-sanitizer %LIBRARY_PREFIX%
if errorlevel 1 exit 1
```

recipes/cuda-sanitizer-api/bld.bat
```diff
@@ -0,0 +1,15 @@
+if not exist %PREFIX% mkdir %PREFIX%
+if not exist %LIBRARY_PREFIX% mkdir %LIBRARY_PREFIX%
+if not exist %SCRIPTS% mkdir %SCRIPTS%
+
+rmdir /q /s compute-sanitizer\x86
```
jakirkham, line 5:
```suggestion
rmdir /q /s compute-sanitizer\x86
if errorlevel 1 exit 1
```

recipes/cuda-sanitizer-api/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/compute-sanitizer                                         # [linux]
+    - test -f $PREFIX/compute-sanitizer/compute-sanitizer                           # [linux]
+    - if not exist %LIBRARY_PREFIX%\compute-sanitizer\compute-sanitizer.exe exit 1  # [win]
+    - if not exist %LIBRARY_BIN%\compute-sanitizer exit 1                           # [win]
```
jakirkham, line 46: Looks like just this test is failing. As the location changed, that is expected. Maybe we can do this?

```suggestion
    - if not exist %SCRIPTS%\compute-sanitizer.bat exit 1                           # [win]
```

jakirkham:
Thanks all! 🙏

Let's get this in. We can follow up on anything else in the feedstock

recipes/cuda-sanitizer-api/meta.yaml
jakirkham: Resolved in recent fixes

PR #23001, opened 2023-06-02
Add nsight-compute recipe

adibbley:
xref: #21382

Checklist
- [x] Build number is 0.


recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+build:
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - "*"
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
jakirkham, line 31: Think we need the `sysroot` package here

```suggestion
    - {{ compiler("cxx") }}
    - sysroot_{{ target_platform }} 2.17
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+about:
+  home: https://developer.nvidia.com/nsight-compute
+  license_file: LICENSE
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: NVIDIA Nsight Compute is an interactive kernel profiler for CUDA applications
+  description: |
+    NVIDIA Nsight Compute is an interactive kernel profiler for CUDA
+    applications on x86_64 platforms. It provides detailed performance
```
bdice, line 55: "on x86_64 platforms": Does this mean that the Nsight Compute program can run on linux64/aarch64/ppc64le/win64 but it can only profile applications running on linux64? Or is this comment inaccurate, and we _can_ profile on aarch64/etc.?

adibbley: Probably best to leave these details up to their docs, which are better maintained. Their support matrix has the details: https://developer.nvidia.com/tools-overview/nsight-compute/get-started

recipes/nsight-compute/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+mkdir -p ${PREFIX}/nsight-compute
+
```
jakirkham, line 6: Should we drop the 32-bit libraries here?

```suggestion
# Remove 32bit libraries
[[ -d "nsight-compute/2022.4.0/target/linux-desktop-glibc_2_11_3-x86" ]] && rm -rf nsight-compute/2022.4.0/target/linux-desktop-glibc_2_11_3-x86
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - "*"
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - sysroot_{{ target_platform }} 2.17
```
jakirkham, line 32: Think this needs a Linux selector (since this builds on Windows where `sysroot`s don't exist)

```suggestion
    - sysroot_{{ target_platform }} 2.17  # [linux]
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  host:
+     - cuda-version {{ cuda_version }}
+  run:
+     - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
+    - test -f $PREFIX/bin/ncu                      # [linux]
```
jakirkham, line 43: Should we test this on Windows too?

jakirkham (commented):
Thanks Alex! 🙏

Had a couple questions below

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+  host:
+     - cuda-version {{ cuda_version }}
+  run:
+     - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+  run_constrained:
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+
+test:
+  commands:
```
bdice, line 42: It'd be nice to show the executable can be run, like `ncu --version`.

bdice (commented):
Small request for testing the executable, otherwise please apply @jakirkham's suggestions.

recipes/nsight-compute/meta.yaml
adibbley: Replaced the directory tests with `ncu --version` as it validates the full structure of the host platform. This was very helpful to discover the default layout under `$PREFIX/nsight-compute` was not working.

recipes/nsight-compute/meta.yaml
adibbley: Actually, there is no `bin/` on Windows.

recipes/nsight-compute/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+
+nsightVersion=$(echo $PKG_VERSION | cut -d. -f1-3)
+nsightLib32="nsight-compute/${nsightVersion}/target/linux-desktop-glibc_2_11_3-x86/"
+
+# Remove 32bit libraries
+[[ -d "${nsightLib32}" ]] && rm -rf ${nsightLib32}
```
vyasr, line 10: FWIW this could probably be generalized using something like
```
set +e
rm -rf nsight-compute/*/target/*x86
set -e
```
The set would only be needed if a nonzero exit code is problematic.

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  sha256: f084e05eb4d2ba32aceb25e1dcfe03f2a50127630973722b65219cf9e986a139  # [linux64]
+  sha256: 20e58ce79681bc8fd39394cfb8f8316c177fe4175af3ae95c025996f45904732  # [ppc64le]
+  sha256: 7d0b3d4d01ce36657fa739496c7b0a9c627f5fa42021c1696ddd15e119bb05a4  # [aarch64]
+  sha256: 958da9986841c49cb5a2885d1e14e4c673ba94e4b404ef9389d083b7a0095d84  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
+    - "*"
```
vyasr, line 27: I assume conda-build fails without this? Just double-checking that we're not adding to the list unless really necessary.

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
```
vyasr, line 32: Similar to the question I had on another recipe, do we need the compilers?

recipes/nsight-compute/meta.yaml
adibbley: Yes, the libraries internal to `nsight-compute` linking each other are being flagged as missing, and the list is very long.

recipes/nsight-compute/meta.yaml
adibbley: With the whitelist this is probably not needed. But there are libraries linking compiler files, so having these available has been preferred in the past.

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "nsight-compute" %}
+{% set version = "2022.4.0.15" %}
+{% set version_split = version.split(".")[0]+"."+version.split(".")[1]+"."+version.split(".")[2] %}
+{% set cuda_version = "12.0" %}
```
vyasr, line 4: Tests are failing because this is hardcoded. Should we be skipping when the CUDA version < 12?

bdice (approved):
LGTM once open conversations are resolved.

recipes/nsight-compute/bld.bat
```diff
@@ -0,0 +1,9 @@
+setlocal enableDelayedExpansion
```
jakirkham, line 1:
```suggestion
if not exist %PREFIX% mkdir %PREFIX%
if not exist %LIBRARY_PREFIX% mkdir %LIBRARY_PREFIX%
if not exist %SCRIPTS% mkdir %SCRIPTS%

setlocal enableDelayedExpansion
```

recipes/nsight-compute/bld.bat
```diff
@@ -0,0 +1,9 @@
...
+for /F "tokens=1,2,3 delims=. " %%a in ("%PKG_VERSION%") do (
+   set "version=%%a.%%b.%%c"
+)
+rmdir /q /s nsight-compute\!version!\lib
+
+move nsight-compute %LIBRARY_PREFIX%
+
+mklink /h %LIBRARY_BIN%\ncu %LIBRARY_PREFIX%\nsight-compute\!version!\ncu.bat
```
jakirkham, line 9:
```suggestion
:: Create bat links for all exe files in %LIBRARY_PREFIX%\nsight-compute
cd %SCRIPTS%
for /r "%LIBRARY_PREFIX%\nsight-compute" %%f in (*.exe) do (
    echo @echo off > %%~nf.bat
    echo %%~dp0.\nsight-compute\%%~nf.exe %%* >> %%~nf.bat
    if errorlevel 1 exit 1
)
```

jakirkham (commented):
Thanks Alex! 🙏

Maybe we can try adding wrapper scripts to call the `.exe`s (as below). This is a strategy we have used for similar situations in the past (though it is always possible the code below needs additional tweaks)

recipes/nsight-compute/meta.yaml
jakirkham, line 27: Could we try dropping this again? Want to see if we can handle this with the list dependencies at this point

```suggestion
```

jakirkham: Yeah let's try dropping this again. Think this may already be solved by dependencies added below. Even if that is not the case, would like to see if we can fix the dependencies up to make this unnecessary

jakirkham (commented):
Suggesting dropping the `missing_dso_whitelist` portion since this may already be addressed by dependencies since added. Even if not, would be good if we can solve this with adding dependencies instead to gain some confidence that this will work in a user's environment

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  skip: true  # [osx]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - sysroot_{{ target_platform }} 2.17  # [linux]
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+  host:
+     - cuda-version {{ cuda_version }}
```
jakirkham, line 34:
```suggestion
     - cuda-version {{ cuda_version }}
     - boost-cpp
     - icu
     - qt-main
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  run:
+     - {{ pin_compatible("cuda-version", max_pin="x.x") }}
```
jakirkham, line 36:
```suggestion
     - {{ pin_compatible("cuda-version", max_pin="x.x") }}
     - boost-cpp
```

jakirkham (commented):
Thanks Alex! 🙏

Think we need to add these dependencies based on the linking errors in CI. Though we may need more. Will need to look at CI based on the results after adding these

recipes/nsight-compute/bld.bat
```diff
@@ -0,0 +1,19 @@
...
+)
+rmdir /q /s nsight-compute\!version!\lib
+
+move nsight-compute %LIBRARY_PREFIX%
+
+:: Create bat links for all exe files in %LIBRARY_PREFIX%\nsight-compute
+cd %SCRIPTS%
+for /r "%LIBRARY_PREFIX%\nsight-compute" %%f in (*.exe) do (
+    echo @echo off > %%~nf.bat
+    echo %%~dp0.\nsight-compute\%%~nf.exe %%* >> %%~nf.bat
```
jakirkham, line 17:
```suggestion
    echo %%~dp0.\Library\nsight-compute\%%~nf.exe %%* >> %%~nf.bat
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+  url: https://developer.download.nvidia.com/compute/cuda/redist/nsight_compute/{{ platform }}/nsight_compute-{{ platform }}-{{ version }}-archive.{{ extension }}
+  sha256: f084e05eb4d2ba32aceb25e1dcfe03f2a50127630973722b65219cf9e986a139  # [linux64]
+  sha256: 20e58ce79681bc8fd39394cfb8f8316c177fe4175af3ae95c025996f45904732  # [ppc64le]
+  sha256: 7d0b3d4d01ce36657fa739496c7b0a9c627f5fa42021c1696ddd15e119bb05a4  # [aarch64]
+  sha256: 958da9986841c49cb5a2885d1e14e4c673ba94e4b404ef9389d083b7a0095d84  # [win]
+
+build:
+  number: 0
+  skip: true  # [osx]
+  missing_dso_whitelist:
```
adibbley, line 26: This list is still WIP, just narrowed down to the unique errors. Haven't been able to find packages which provide libraries in this list yet.

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - sysroot_{{ target_platform }} 2.17  # [linux]
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+  host:
+    - cuda-version {{ cuda_version }}
+    - boost-cpp
+    #- icu #not helping?
+    - dbus
+    - wayland  # [linux]
+    - libxkbcommon
+    - libexpat
```
jakirkham, line 137:
```suggestion
    - expat
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ compiler("cxx") }}
+    - sysroot_{{ target_platform }} 2.17  # [linux]
+    - arm-variant * {{ arm_variant_type }}  # [aarch64]
+  host:
+    - cuda-version {{ cuda_version }}
+    - boost-cpp
+    #- icu #not helping?
```
jakirkham, line 133:
```suggestion
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - xorg-libxcomposite
+    - xorg-libxdamage
+    - xorg-libxfixes
+#requires krb5
+    #- qt-main
+    #- libcurl
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    - {{ pin_compatible("boost-cpp", max_pin="x.x") }}
+    - {{ pin_compatible("libexpat", max_pin="x.x") }}
```
jakirkham, line 171: Since `expat` has a `run_exports`, which will handle this for us

```suggestion
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - libexpat
+    - freetype
+    - fontconfig
+    - alsa-lib
+    - libglib
+    - libxcb
+    - libzlib
+    - nspr
+    - nss
+    - xorg-libxrender
```
jakirkham, line 146: This was later renamed

```suggestion
    - libxrender
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - xorg-libxdamage
+    - xorg-libxfixes
+#requires krb5
+    #- qt-main
+    #- libcurl
+  run:
+    - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+    - {{ pin_compatible("boost-cpp", max_pin="x.x") }}
+    - {{ pin_compatible("libexpat", max_pin="x.x") }}
+    - {{ pin_compatible("xorg-libxrender", max_pin="x.x") }}
```
jakirkham, line 172: As `libxrender` should handle this for us

```suggestion
```

recipes/nsight-compute/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+    - freetype
+    - fontconfig
+    - alsa-lib
+    - libglib
+    - libxcb
+    - libzlib
+    - nspr
+    - nss
+    - xorg-libxrender
+    - xcb-util-image
```
jakirkham, line 147: All of the X11 packages will need selectors for Linux. For example

```suggestion
    - xcb-util-image  # [linux]
```

recipes/nsight-compute/meta.yaml
jakirkham: It looks like tests have since been working. Marking this resolved

recipes/nsight-compute/meta.yaml
jakirkham: Looks like we need to keep this given the nature of the package. Marking resolved

jakirkham:
Thanks all! 🙏

Let's get this one out. We can follow up on anything else in the feedstock 🙂

PR #23003, opened 2023-06-02
Add cuda-compiler recipe

adibbley:
xref: #21382

Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/cuda-compiler/README.md
+++ b/recipes/cuda-compiler/README.md
@@ -1,12 +1,17 @@
 # CUDA Metapackage Versioning
 
-The version of a CUDA Toolkit metapackage corresponds to the CUDA release label. For example, the release label of CUDA 12.0 Update 1 is 12.0.1.
-This does not include the `cuda-version` metapackage which is versioned only by the MAJOR.MINOR of a release label.
+The version of a CUDA Toolkit metapackage corresponds to the CUDA release
+label. For example, the release label of CUDA 12.0 Update 1 is 12.0.1.  This
+does not include the `cuda-version` metapackage which is versioned only by the
+MAJOR.MINOR of a release label.
 
 # Metapackage dependency versions
 
-Installing a metapackage at a specific version should install all dependent packages at the exact version from that CUDA release.
+Installing a metapackage at a specific version should install all dependent
+packages at the exact version from that CUDA release.
 
 # Metapackage dependencies on cuda-version
 
-Metapackages do not directly constrain to a specific `cuda-version` as their version is more precise. Dependent packages will still install an appropriate `cuda-version`.
+Metapackages do not directly constrain to a specific `cuda-version` as their
+version is more precise. Dependent packages will still install an appropriate
+`cuda-version`.
--- a/recipes/cuda-compiler/README.md
+++ b/recipes/cuda-compiler/README.md
@@ -1,0 +1,12 @@
+# CUDA Metapackage Versioning
+
+The version of a CUDA Toolkit metapackage corresponds to the CUDA release label. For example, the release label of CUDA 12.0 Update 1 is 12.0.1.
+This does not include the `cuda-version` metapackage which is versioned only by the MAJOR.MINOR of a release label.
+
+# Metapackage dependency versions
+
+Installing a metapackage at a specific version should install all dependent packages at the exact version from that CUDA release.
+
+# Metapackage dependencies on cuda-version
+
+Metapackages do not directly constrain to a specific `cuda-version` as their version is more precise. Dependent packages will still install an appropriate `cuda-version`.
--- a/recipes/cuda-compiler/meta.yaml
+++ b/recipes/cuda-compiler/meta.yaml
@@ -12,15 +12,15 @@ build:
 
 requirements:
   run:
+    - c-compiler
+    - cxx-compiler
+    - {{ compiler("cuda") }}
     - cuda-cuobjdump 12.0.76
     - cuda-cuxxfilt 12.0.76
     - cuda-nvcc 12.0.76
     - cuda-nvprune 12.0.76
 
 test:
-  requires:
-    - {{ compiler("c") }}
-    - {{ compiler("cxx") }}
   commands:
     - which cuobjdump
     - which cu++filt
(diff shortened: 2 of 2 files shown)


recipes/cuda-compiler/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_documentation/LICENSE.txt
+  sha256: 5db25d4fd138013b563f9a3d1d87f7de7df1dac014fd4cccdfbb3435a5cff761
+
+build:
+  number: 0
+  skip: true  # [osx]
+
+requirements:
+  run:
+    - cuda-cuobjdump 12.0.76
```
bdice, line 15: Will these component versions always be in sync? If so, should we use a variable that is defined once?

recipes/cuda-compiler/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - cuda-cuxxfilt 12.0.76
+    - cuda-nvcc 12.0.76
+    - cuda-nvprune 12.0.76
```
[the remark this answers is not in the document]
adibbley: Right I would expect a version `12.0.1` of this package to bump these component versions, while `cuda-version` is only tracking `12.0`. I don't think there is a use to add `cuda-version` directly here, but can add it if there is.

recipes/cuda-compiler/meta.yaml
adibbley: No, all components are versioned independently.

jakirkham (commented):
Thanks Alex! 🙏

Think we should make this a fully baked compiler package (similar to those in [the `compilers` feedstock]( https://github.com/conda-forge/compilers-feedstock )). This would allow developers to have a package to install and get a fully functional compilation environment (with the C & C++ compilers they need)

Have made a couple suggestions below along these lines. Please let me know what you think 🙂

recipes/cuda-compiler/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  name: cuda-compiler
+  version: 12.0.0
+
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_documentation/LICENSE.txt
+  sha256: 5db25d4fd138013b563f9a3d1d87f7de7df1dac014fd4cccdfbb3435a5cff761
+
+build:
+  number: 0
+  skip: true  # [osx or win]
```
jakirkham, line 11: Think we need this `skip` to fix CI

```suggestion
  skip: true  # [osx or win]
  skip: true  # [cuda_compiler_version != "None"]
```

recipes/cuda-compiler/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+requirements:
+  run:
+    - c-compiler
+    - cxx-compiler
+    - {{ compiler("cuda") }}
```
jakirkham, line 17:
```suggestion
    - {{ compiler("cuda") }} 12.0.76
```

jakirkham, line 17: Since we already have `cuda-nvcc` below, that should be good enough

```suggestion
```

recipes/cuda-compiler/meta.yaml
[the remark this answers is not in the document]
bdice: Great. I agree this is the right strategy. Should we document someplace how the metapackages are versioned, and note that it's independent of `cuda-version`? Not sure where this info should go.

recipes/cuda-compiler/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - cuda-cuxxfilt 12.0.76
+    - cuda-nvcc 12.0.76
+    - cuda-nvprune 12.0.76
+
+test:
+  commands:
+    - which cuobjdump
+    - which cu++filt
+    - which nvcc
+    - which nvprune
```
bdice, line 27: Maybe pre-emptively guard these, for when Windows support is added. I think we've done this kind of selector for tests on other packages where no Windows support is available yet.
```suggestion
    - which cuobjdump  # [linux]
    - which cu++filt   # [linux]
    - which nvcc       # [linux]
    - which nvprune    # [linux]
```

bdice (approved):
LGTM, with one minor suggestion.

recipes/cuda-compiler/meta.yaml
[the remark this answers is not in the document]
jakirkham: We could add a `README.md` to the recipe. Have done this in a few other cases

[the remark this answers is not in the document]
adibbley: Added a README, we can use this for other metapackages as well.

recipes/cuda-compiler/README.md
```diff
@@ -0,0 +1,12 @@
...
+The version of a CUDA Toolkit metapackage corresponds to the CUDA release label. For example, the release label of CUDA 12.0 Update 1 is 12.0.1.
+This does not include the `cuda-version` metapackage which is versioned only by the MAJOR.MINOR of a release label.
+
+# Metapackage dependency versions
+
+Installing a metapackage at a specific version should install all dependent packages at the exact version from that CUDA release.
+
+# Metapackage dependencies on cuda-version
+
+Metapackages do not directly constrain to a specific `cuda-version` as their version is more precise. Dependent packages will still install an appropriate `cuda-version`.
```
jakirkham, line 12: Would suggest wrapping the lines here

```suggestion
# CUDA Metapackage Versioning

The version of a CUDA Toolkit metapackage corresponds to the CUDA release
label. For example, the release label of CUDA 12.0 Update 1 is 12.0.1.  This
does not include the `cuda-version` metapackage which is versioned only by the
MAJOR.MINOR of a release label.

# Metapackage dependency versions

Installing a metapackage at a specific version should install all dependent
packages at the exact version from that CUDA release.

# Metapackage dependencies on cuda-version

Metapackages do not directly constrain to a specific `cuda-version` as their
version is more precise. Dependent packages will still install an appropriate
`cuda-version`.
```

jakirkham:
Thanks all! 🙏

PR #23023, opened 2023-06-07
Add openms and pyopenms recipe

jpfeuffer:
Checklist

We do compile other packages like sqlitecpp and nlohmann-json into our library. How to package the licenses? Is there any documentation?

- [x] Build number is 0.

TODOS:
- [X] Solve overlinking. We probably need to add coin, qt-main to run dependencies. But why libcpp?? Should we really list this there? Solved with help of @hmaarrfk 
- [x] set MACOSX_DEPLOYMENT_TARGET for qt: `/lib/libQt5Network.5.15.8.dylib) was built for newer macOS version (10.13) than being linked (10.9)`

- [X] Qt recipe is bugged. See https://github.com/conda-forge/qt-feedstock/issues/242. -> Switch to centos7 solved it.
- [X] switch from coinmp (not working on win) to coin-or-xyz recipes. Probably won't help, they are also deactivated on win. So we have to switch to glpk for Win?? See https://github.com/conda-forge/coin-or-cbc-feedstock/issues/3
- [x] Enable osx_arm64 in conda-forge.yml


recipes/openms/activate.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+
+export OPENMS_DATA_PATH=${CONDA_PREFIX}/share/OpenMS/
```
hmaarrfk, line 3: I strongly suggest you find an other way to do this. This kind of environment variable is very hard to keep consistent especially when nesting environments

recipes/openms/build.sh
```diff
@@ -0,0 +1,51 @@
...
+  -DHAS_XSERVER=OFF \
+  -DENABLE_TUTORIALS=OFF \
+  -DWITH_GUI=OFF \
+  -DBOOST_USE_STATIC=OFF \
+  -DBoost_NO_BOOST_CMAKE=ON \
+  -DBoost_ARCHITECTURE="-x64" \
+  -DBUILD_EXAMPLES=OFF
+
+# limit concurrent build jobs due to memory usage on CI
+make -j4 OpenMS TOPP UTILS
```
hmaarrfk, line 49:
```suggestion
make -j${CPU_COUNT} OpenMS TOPP UTILS
```

recipes/openms/meta.yaml
```diff
@@ -0,0 +1,138 @@
...
+    host:
+      - llvm-openmp  # [osx]
+      - zlib
+      - libtool
+      - xerces-c
+      - boost-cpp
+      - eigen >=3.3.2
+      - hdf5 >=1.10
+      - bzip2
+      - qt
```
hmaarrfk, line 42:
```suggestion
      - qt-main
```

recipes/openms/meta.yaml
```diff
@@ -0,0 +1,138 @@
...
+        host:
+          - llvm-openmp  # [osx]
+          - zlib
+          - libtool
+          - xerces-c
+          - boost-cpp
+          - eigen >=3.3.2
+          - hdf5 >=1.10
+          - bzip2
+          - qt
```
hmaarrfk, line 70:
```suggestion
          - qt-main
```

recipes/openms/meta.yaml
```diff
@@ -0,0 +1,138 @@
...
+          - libsvm
+          - coinmp
+    build:
+      run_exports:
+          - {{ pin_compatible('llvm-openmp', max_pin='x.x') }}  # [osx]
```
hmaarrfk, line 75: You only need to add run exports for your own package. Not the others

hmaarrfk (commented):
A few comments in hopes of helping

recipes/openms/meta.yaml
```diff
@@ -0,0 +1,138 @@
...
+      # see https://github.com/conda/conda-build/issues/3478
+      # so specify everything again, such that EXACTLY the same versions are used!
+      # I could not get the "resolved_packages" jinja function to gather dependencies from other subpackages.
+        - llvm-openmp  # [osx]
+        - zlib
+        - libtool
+        - xerces-c
+        - boost-cpp
+        - eigen >=3.3.2
+        - hdf5 >=1.10
```
hmaarrfk, line 106: I would just leave the version pin empty here. 1.10 is quite old and Fonda forge is on 1.12 and 1.14 mow

recipes/pyopenms/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - {{ compiler('cxx') }}
+    - make
+    - cmake
+  host:
+    - libopenms {{ version }}
+    - llvm-openmp  # [osx]
+    - libsvm
+    - xerces-c
+    - boost-cpp
+    - qt
```
hmaarrfk, line 33: If you aren't linking to qt, which i doubt you are in a python package, just remove this. Let the c package drag this in if it needs to.

recipes/pyopenms/conda_build_config.yaml
```diff
@@ -0,0 +1,16 @@
+python:
```
hmaarrfk, line 1: Just remove this file. I understand you might be excited for 3.11 but it is better to let the conda-forge tooling do it's work after the feedstock is accepted rather than trying to trick it to do it on the first merge.

recipes/openms/build.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/sh
+
+# useless default include directory that is silently added by the compiler packages "to help"...
+# it is not even added with -isystem https://github.com/AnacondaRecipes/aggregate/blob/master/clang/activate-clang%2B%2B.sh#L87
+USELESS="-I${PREFIX}/include"
+export CXXFLAGS=${CXXFLAGS//${USELESS}/}
```
hmaarrfk, line 6: What's the point of all of this? Can we remove these environment variable defintions?

recipes/openms/build.sh
```diff
@@ -0,0 +1,51 @@
...
+if [[ $(uname -s) == Darwin ]]; then
+  RPATH='@loader_path/../lib'
+else
+  ORIGIN='$ORIGIN'
+  export ORIGIN
+  RPATH='$${ORIGIN}/../lib'
+fi
+
+# not sure if needed. CMake should take care of that.
+LDFLAGS='-Wl,-rpath,${RPATH}'
```
hmaarrfk, line 25: Correct, let's let, as a first pass, CMake do its job. I'm hoping it will resolve things.

Conda and CMake should be able to work together to avoid this kind of manual intervention

recipes/openms/build.sh
jpfeuffer: The point is that the compiler packages silently add this to the CXXFLAGS variable.
But this should be taken care of by CMake. Additionally, it adds it with "-I" and not "-isystem" which means that there will be millions of compiler warnings slowing down the build to a point where CI did not finish on bioconda.

jpfeuffer: Maybe this is different for conda-forge though.

recipes/openms/build.sh
jpfeuffer: Ok, will try without it.

jpfeuffer:
@hmaarrfk Thanks a lot for your very helpful comments so far.

One question: Do you know why qt needs/ is built on cos7? And shouldn't that propagate transitively as soon as you have a dependency on a cos7 package? Is there ever a case where it would work if the dependent package does not build on cos7, too?

hmaarrfk:
> cos7

I just don't think qt works with older cos. I don't think our tooling detects the correct version of cos yet. But i might be wrong.

hmaarrfk:
I think reading this will help

jpfeuffer:
Looks exactly like what I need!

jpfeuffer:
@hmaarrfk I cannot get the openmp dependency right. I have it in the pyopenms meta.yml run section but it *still* complains that it wants openms_mutex. I've tried multiple combinations now.

recipes/openms/meta.yaml
```diff
@@ -0,0 +1,153 @@
+{% set name = "OpenMS" %}
+{% set version = "2.9.1" %}
+package:
+  name: {{ name|lower }}-meta
+  version: {{ version }}
+
+{% if 'dev' in version %}
+source:
+  git_url: https://github.com/OpenMS/OpenMS.git
+  git_rev: develop
```
hmaarrfk, line 10: could we please remove this condition, it really isn't useful for conda-forge and just adds challenges to maintanability.

recipes/pyopenms/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  git_rev: develop
+  git_depth: 1 # (Defaults to -1/not shallow)
+{% else %}
+source:
+  url: https://abibuilder.cs.uni-tuebingen.de/archive/openms/OpenMSInstaller/release/{{ version }}/OpenMS-{{ version }}-src.tar.gz
+  sha256: 2d186f35fe42573a19283605ccfb491ca6da585880fa3725a013975df26bc008
+{% endif %}
+
+build:
+  skip: True  # [py2k]
```
hmaarrfk, line 19:
```suggestion
```

hmaarrfk:
lets see if these changes do it.

recipes/openms/meta.yaml
jpfeuffer: Yes I think we can. It made it a bit easier to have nightly builds but once this is close to finished I can get rid of it.

hmaarrfk:
you should at the very least add a "skip" for windows.

jpfeuffer:
Yep definitely. We will investigate building coinor on windows sometime later.

jpfeuffer:
@hmaarrfk Where can I see if osx arm is built and if it passes?
From the [blog](https://conda-forge.org/blog/posts/2020-10-29-macos-arm64/#how-to-add-a-osx-arm64-build-to-a-feedstock), it sounds like this should do it:
https://github.com/conda-forge/staged-recipes/pull/23023/files#diff-478786365dd93f740eb520c2faa03d7a0623273663b75472272c8ef94297bbe2R7

But I cannot find related CI output.

hmaarrfk:
typically OSX arm64 is done in a subsequent step, once the recipe is accepted.

jpfeuffer:
Ok, I think last question:
We do compile and link two smaller libraries that are available on conda-forge, namely sqlitecpp and nlohmann-json to have full control over the build parameters. They will not be exposed in our public API.
Are we allowed to do this until we find time to allow an arbitrary external dynamic library to be linked?
If so, how do we need to "package the licenses" as requested in the Pull Request template (= first comment)?

hmaarrfk:
yes you should add them to your license file.

I would personally prefer if you patched out the linkage and use the shared libraries.

It is easy to have the symbols "leak" and cause more problems down the road.

recipes/openms/build.sh
```diff
@@ -0,0 +1,51 @@
...
+
+mkdir build
+cd build
+
+if [[ $(uname -s) == Darwin ]]; then
+  RPATH='@loader_path/../lib'
+else
+  ORIGIN='$ORIGIN'
+  export ORIGIN
+  RPATH='$${ORIGIN}/../lib'
```
hmaarrfk, line 24: do you really need to do RPATH stuff? can we remove this.

recipes/pyopenms/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+build:
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - make
+    - cmake
+    - {{ compiler('cxx') }}
```
timosachsenberg, line 20: duplicate of line 17?

timosachsenberg, line 20:
```suggestion
```

recipes/openms/build.sh
jpfeuffer: we have added this in CMake a while ago. I removed it and try to test it.

timosachsenberg:
btw: I accept being an additional maintainer of this recipe

recipes/openms/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
+# Otherwise libraries won't be found during linking on Linux
+export LIBRARY_PATH=${PREFIX}/lib
+export LD_LIBRARY_PATH=${PREFIX}/lib
+
+# Otherwise Error: dyld: Symbol not found: _iconv Referenced from: /usr/lib/libarchive.2.dylib
+# Expected in: /Users/runner/mambaforge/conda-bld/openms-meta_1686666744655/_h_env_placehold_/lib/libiconv.2.dylib
+#export DYLD_LIBRARY_PATH=${PREFIX}/lib
```
hmaarrfk, line 9:
```suggestion
```

can we try removing these?

recipes/openms/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh
+
```
hmaarrfk, line 2:
```suggestion
set -ex
```

recipes/pyopenms/build.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+# Otherwise libraries won't be found during linking on Linux
+export LIBRARY_PATH=${PREFIX}/lib
+export LD_LIBRARY_PATH=${PREFIX}/lib
+
+# Otherwise Error: dyld: Symbol not found: _iconv Referenced from: /usr/lib/libarchive.2.dylib
+# Expected in: /Users/runner/mambaforge/conda-bld/openms-meta_1686666744655/_h_env_placehold_/lib/libiconv.2.dylib
+#export DYLD_LIBRARY_PATH=${PREFIX}/lib
```
hmaarrfk, line 9:
```suggestion
set -ex
```

conda-forge.yml
```diff
@@ -3,3 +3,18 @@ azure:
   settings_win:
     pool:
       vmImage: windows-2022
+build_platform:
```
hmaarrfk, line 6: please revert changes to this file.

hmaarrfk: this will be done in the feedstock after merging

recipes/openms/build.sh
jpfeuffer: We can. I think I did before but we can have a look at the errors together.

recipes/openms/activate.sh
hmaarrfk: Can you please print me to exactly where in your codebase this is used. I generally feel like there is often a better way and I would like to help avoid this activate script. It is often unmaintainable

jpfeuffer: https://github.com/OpenMS/OpenMS/blob/947b5b966299300049cb6b9f9239b2462a244480/src/openms/source/SYSTEM/File.cpp#L497

We actually also have other fallbacks. We just thought it was a good idea to make it explicit. We could try to remove it and see if the relative default locations work.

hmaarrfk: Please let's try the fallbacks. Setting and unsettling environment variables is error prone especially in nested environments.

jpfeuffer:
Is the share not relative to the bin folders on conda?

hmaarrfk:
Are you confident that we can now remove the activate /deactivate scripts?

Now remains the bundled third party software. How feasible is it for you to use the conda forge versions?

jpfeuffer:
Yes I removed them. We support every lib that exists in condos now. We need to incorporate into the release.

hmaarrfk:
I'm curious if we can 

- [ ] try to remove the lines with RPATH in them in the CMake configuration.
- [ ] wait for https://github.com/OpenMS/OpenMS/pull/6928 or backport
- [ ] remove the other comments
- [ ] add tests

bernt-matthias:
Hi. What is the state here? I recently tried to get the bioconda build running again .. [unsuccessfully](https://github.com/bioconda/bioconda-recipes/pull/47219). 

Having a working 3.0 (or better 3.1?) in conda would be really cool.

jpfeuffer:
There was a lack of support/seeing the need from other OpenMS developers to move to conda-forge and I currently do not have the time to move that forward alone.
For me it is the only viable solution to keep OpenMS on a public conda channel though.
The bioconda situation is just annoying in many aspects (no macos-arm, no win, no support for finding dependencies in multipackages, no support for packages larger than 1GB, no support in merging PRs to their build tool to fix some of the issues, an additional confusing layer of pinnings, etc.)

I am happy to give guidance on reviving that PR though. In theory there is nothing missing (except maybe CoinOR on windows). You could probably even remove some hacks for the licenses or MacOS deployment targets by now.

bernt-matthias:
My primary interest is to finally get the Galaxy tools for 3.x done. For this I only need Linux. Can you reopen? I could try to bump to 3.1. Then we see what happens....

timosachsenberg:
Would be great to move this forward. I am a bit lost in regard what can be done from our side, but we would like to help.

PR #23033, opened 2023-06-08
add ncvis (renamed ncvis-climate due to conflict in conda-forge)

mahf708:
Checklist
- [x] Build number is 0.


recipes/ncvis-climate/build.sh
```diff
@@ -0,0 +1,20 @@
...
+WXFLAGS=`wx-config --cxxflags --libs --cppflags`
+
+# get the netCDF build flags
+NCFLAGS=`nc-config --cflags --libs`
+
+# infer the RPATH needed for dynamic linking to wxwidgets
+RPATH=`wx-config --prefix`/lib
+
+# build the executable
+cd src && $CXX -std=c++11 -fpermissive -Wl,-rpath,${RPATH} -o ${PREFIX}/ncvis ncvis.cpp kdtree.cpp wxNcVisFrame.cpp wxNcVisOptionsDialog.cpp wxNcVisExportDialog.cpp wxImagePanel.cpp GridDataSampler.cpp ColorMap.cpp netcdf.cpp ncvalues.cpp Announce.cpp TimeObj.cpp ShpFile.cpp schrift.cpp lodepng.cpp ${WXFLAGS} ${NCFLAGS}
```
xylar, line 20:
```suggestion
cd src && $CXX -std=c++11 -fpermissive -Wl,-rpath,${RPATH} -o ${PREFIX}/bin/ncvis ncvis.cpp kdtree.cpp wxNcVisFrame.cpp wxNcVisOptionsDialog.cpp wxNcVisExportDialog.cpp wxImagePanel.cpp GridDataSampler.cpp ColorMap.cpp netcdf.cpp ncvalues.cpp Announce.cpp TimeObj.cpp ShpFile.cpp schrift.cpp lodepng.cpp ${WXFLAGS} ${NCFLAGS}
```

recipes/ncvis-climate/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  host:
+    - libnetcdf
+    - wxwidgets
+  run:
+    - libnetcdf
+    - wxwidgets
+
+test:
+  commands:
+    - test -f ${PREFIX}/ncvis
```
xylar, line 33:
```suggestion
    - test -f ${PREFIX}/bin/ncvis
```

recipes/ncvis-climate/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  description: 'NetCDF Visualizer'
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+  doc_url: https://github.com/SEATStandards/ncvis
+  dev_url: https://github.com/SEATStandards/ncvis
+
+extra:
+  recipe-maintainers:
+    - mahf708
```
xylar, line 47:
```suggestion
    - mahf708
    - xylar
```

xylar:
@mahf708, I think the problem is just that the `bin` directory is missing.  You could add me as a maintainer if you want but I'd lean on you pretty heavily here since I'm really only good at maintaining python recipes.

xylar:
@mahf708, I'm seeing an error when I try to run this:
```
ERROR: Cannot open resource directory "/home/xylar/mambaforge/envs/test/bin/resources"
Set environment variable NCVIS_RESOURCE_DIR instead
```

I think we need to copy the `resources` directory into `$PREFIX/share/ncvis/resources` or something like that and then have a series of activation scripts like these to set `$NCVIS_RESOURCE_DIR`:

https://github.com/conda-forge/cartopy_offlinedata-feedstock/tree/main/recipe/scripts

I can help if you're not able to do that but it will be easier for you to give it a go since this is your branch.

recipes/ncvis-climate/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  skip: True  # [win]
+  skip: True  # [osx]
+
+requirements:
+  build:
+    - wxwidgets
+    - libnetcdf
+    - make
+    - {{ compiler('cxx') }}
+    - {{ cdt('libxxf86vm-devel') }}      # [linux]
```
hmaarrfk, line 23: See the qt main repository. Please use conda forge x11 packages.

recipes/ncvis-climate/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  sha256: a522926739b2a05ef0b436fe67a2014557f9e5fecf3b7d7700964e9006a4bf3e
+
+build:
+  number: 0
+  skip: True  # [win]
+  skip: True  # [osx]
+
+requirements:
+  build:
+    - wxwidgets
```
hmaarrfk, line 19: You should only need this in host

recipes/ncvis-climate/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - make
+    - {{ compiler('cxx') }}
+    - {{ cdt('libxxf86vm-devel') }}      # [linux]
+  host:
+    - libnetcdf
+    - wxwidgets
+  run:
+    - libnetcdf
+    - wxwidgets
+
```
hmaarrfk, line 30: You can remove these from run. The packages above declare themselves as runtime dependencies

hmaarrfk (commented):
Are you sure you need the activate réactive scripts?

xylar:
> Are you sure you need the activate réactive scripts?

Yes, I think so.  In my testing, we need to set `$NCVIS_RESOURCE_DIR` to point to the `resources` directory or the application doesn't run and instead gives the error above https://github.com/conda-forge/staged-recipes/pull/23033#issuecomment-1583612188

xylar:
@mahf708, your latest test package worked for me but I wasn't able to add the continents.  The shape files are being packaged so I'm not sure why not.  Maybe an upstream issue.

xylar (approved):
Thanks @mahf708

recipes/ncvis-climate/meta.yaml
hmaarrfk: no, you likely want the xxfvm line in the host section.

hmaarrfk:
I would almost be inclined to patch:
https://github.com/SEATStandards/ncvis/blob/main/src/ncvis.cpp#L97

instead of using activate /deactivate scripts.

Basically, they don't nest well in multiple environments, and thus are prone to breakage.

Maybe look for

```
../share/ncvis/resources
```
instead of 
```
resources/
```

xylar:
@hmaarrfk, that's a great suggestion.

hmaarrfk:
> I have one question: Do we still need the yum_requirements.txt file or should we also delete it?

you can delete it.

Are we waiting for the patch too?

recipes/ncvis-climate/patches/fix-resources-path.patch
```diff
@@ -0,0 +1,14 @@
+diff --git a/src/ncvis.cpp b/src/ncvis.cpp
+--- src/ncvis.cpp
++++ src/ncvis.cpp
+@@ -83,9 +83,9 @@
+ 
+ 	wxString wxstrNcVisResourceDir = wxString(std::getenv("NCVIS_RESOURCE_DIR"));
+ 	if (wxstrNcVisResourceDir.length() == 0) {
+ 		wxFileName wxfn(wxStandardPaths::Get().GetExecutablePath());
+-		wxfn.AppendDir(_T("resources"));
++		wxfn.AppendDir(_T("../share/ncvis/resources"));
```
hmaarrfk, line 10: I feel like this patch is not using wx path opérations correctly. It seems to be pretty specific for Linux. Is this really what you want? Is there a better way to do this?

hmaarrfk:
I merged but I want to give a few words of caution on your use case:

However fast conda-forge moves, it won't be able to keep up with your own internal needs.
Beyond your immediate users, we have to consider the needs of the conda-forge community at large.

In this particular case, you were able to resolve the majority of the "blockers" quickly and hit your deadline.
This won't always be the case.

I strongly suggest you start to investigate how to maintain your own channel:

```
channels:
  - doe_e3sm
  - conda-forge
channel_priority: strict
```
this allows you to leverage the conda-forge know-how tooling, updates, while having a great deal of customization for missing packages and "emergency" releases.

The linux HPC usecase is quite "unique" in the sense that you have access to many powerful machines helping you with compilation, and the open source tooling (that we use at conda-forge) can readily be used without any additional machines or licenses.

Goodluck with your deadline!

xylar:
@hmaarrfk, thanks for sharing your thoughts.  We do have an `e3sm` channel already that we've been using for many years but we have moved most packages to `conda-forge`.  The slower development time is worth it for the automation.  @mahf708 pushed hard for this one and I appreciate it but the June 15th "deadline" is just for the beginning of testing.  We didn't need to push quite so quickly here since we don't actually intend to deploy our next environment until August.

In our E3SM-Unified environment(s), we use a mix of conda and spack packages.  Building a python-based workflow is a nightmare on spack but anything that uses MPI basically can't come from conda (because our HPC machines are not configured in a compatible way with conda-forge's MPI packages, and anyway performance would likely be poor).  This has worked surprisingly well for us for a couple of years now.

hmaarrfk:
> (because our HPC machines are not configured in a compatible way with conda-forge's MPI packages, and anyway performance would likely be poor)

has writing MPI software gotten easier? it was pretty hard in 2010 when i wrote some....

rajeeja:
Worked, 

gives a ton of warnings such as: 

```
...
...
./src/common/object.cpp(239): assert "classTable->Get(m_className) == __null" failed in Register(): Class "wxSashLayoutWindow" already in RTTI table - have you used wxIMPLEMENT_DYNAMIC_CLASS() multiple times or linked some object file twice)?
./src/common/object.cpp(239): assert "classTable->Get(m_className) == __null" failed in Register(): Class "wxXmlDocument" already in RTTI table - have you used wxIMPLEMENT_DYNAMIC_CLASS() multiple times or linked some object file twice)?
Usage: ncvis [options] <filename> [filename] ...

(ncvis-test) ➜  ~
$ which ncvis
/opt/homebrew/anaconda3/envs/ncvis-test/bin/ncvis
```

A fresh environment was easy, on my existing with a few other packages:

`Solving environment: failed with initial frozen solve. Retrying with flexible solve.`

rajeeja:
Yes. I didn't get the gui to launch.

rajeeja:
I think so - on osx, does it work for you? 
what platform?

PR #23178, opened 2023-06-26
Add cuda metapackage recipes

adibbley:
Checklist
- [x] Build number is 0.


recipes/cuda/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  sha256: 5db25d4fd138013b563f9a3d1d87f7de7df1dac014fd4cccdfbb3435a5cff761
+
+build:
+  number: 0
+  skip: true  # [osx or win]
+
+requirements:
+  run:
+    - cuda-toolkit {{ version }}
+    - cuda-runtime {{ version }}
```
bdice, line 18: Alphabetize.
```suggestion
    - cuda-runtime {{ version }}
    - cuda-toolkit {{ version }}
```

bdice, line 18: I'm confused by the structure here. `cuda` requires `cuda-runtime`, which requires `cuda-libraries`. But `cuda-toolkit` also requires `cuda-libraries`. Why do we need to list `cuda-runtime` here, and why do we need `cuda-runtime` at all?

bdice (commented):
I have some questions like "why does `cuda-nvml-dev` not belong to `cuda-libraries-dev`?" and similarly for `gds-tools` not being a part of `cuda-command-line-tools` but I would guess that those questions are beyond the scope of conda-forge and are just part of the broader CUDA deployment decisions. This looks generally fine.

recipes/cuda-command-line-tools/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+test:
+  commands:
+    - exit 0
+
+about:
+  home: https://developer.nvidia.com/cuda-toolkit
+  license_file: LICENSE.txt
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: Meta-package containing the command line tools to debug CUDA applications
```
vyasr, line 33: I'm assuming that this is copy-pasted from NVIDIA docs? Otherwise I'd recommend aligning on the unhyphenated spelling of metapackage.

recipes/cuda-toolkit/README.md
vyasr, line 1: Will conda-build follow symlinks (I would hope so)? If so we could use a single README and symlink it into all five recipes rather than copy-pasting it everywhere.

recipes/cuda-command-line-tools/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - cuda-cupti-dev 12.0.90
+    - cuda-gdb 12.0.90
+    - cuda-nvdisasm 12.0.76
+    - cuda-nvprof 12.0.90
+    - cuda-nvtx 12.0.76
+    - cuda-sanitizer-api 12.0.90
+
+test:
+  commands:
+    - exit 0
```
vyasr, line 26: Can we just omit the test section if it's not doing anything? Is the trivial installation of a metapackage worth testing?

vyasr: Same question applies to all the other recipes too.

recipes/cuda-command-line-tools/meta.yaml
vyasr, line 1: Is there a source of truth for what metapackages should exist and what goes in each? Are these just rough approximations of what we've previously done for e.g. apt or something?

vyasr (changes requested):

recipes/cuda-toolkit/README.md
bdice: These will become separate feedstocks so symlinks are not viable. This is a single PR but every directory in recipes/ becomes its own feedstock.

vyasr: Oh missed that this was staging new ones. Never mind, you're right.

recipes/cuda-command-line-tools/meta.yaml
jakirkham: The linter complains if there are no tests

With metapackages, we use this approach

recipes/cuda/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+  license_url: https://docs.nvidia.com/cuda/eula/index.html
+  summary: Meta-package containing all the available packages for native CUDA development
+  description: |
+    Meta-package containing all the available packages required for native CUDA
+    development
+  doc_url: https://docs.nvidia.com/cuda/index.html
+
+extra:
+  recipe-maintainers:
+    - adibbley
```
jakirkham, line 37: Let's go ahead and update this to include all the people we want here. Would suggest the following

```suggestion
extra:
  recipe-maintainers:
    - adibbley
    - bdice
    - jakirkham
    - leofang
    - vyasr
```

recipes/cuda-command-line-tools/meta.yaml
adibbley: Yes the summaries of all formats have used "Meta-package", although I agree that "metapackage" would be better.

recipes/cuda-command-line-tools/meta.yaml
adibbley: We don't ship any metadata specific to metapackages the way we do for components. It does exist internally, and was used to generate these.

bdice (approved):
I think everything looks good here, just need to apply the maintainers list suggestion. Are there other action items needed before merging?

recipes/cuda/meta.yaml
jakirkham: cc @kmittman

jakirkham:
Thanks all! 🙏

Let's follow up on anything else on the feedstock

PR #23182, opened 2023-06-27
Adding photochem

Nicholaswogan:
This is a recipe for the `photochem` Python package.

Checklist
- [x] Build number is 0.


Nicholaswogan:
@conda-forge/help-python-c , I am having some issues figuring out how to make a recipe that will build. Do you have any tips?

Nicholaswogan:

bkpoon:
When I download the source zip file, I get a `__MACOSX` directory as well as the `photochem_0.4.1_withdata`. Can you replace the source zip file with one that does not have `__MACOSX` and try again? Or try the GitHub generated source tarball or zip file to see if that works?

Nicholaswogan:
@bkpoon thanks for the tip! I will retry soon a tar without that weird mac directory.

Nicholaswogan:
@bkpoon I updated the source to tar.gz, and the linux build still fails. I can not use the Github Release tar.gz, because I need to add data in a submodule that is not encompassed in the release.

bkpoon:
The GitHub source file would just be a test to see if those automatically generated source tarballs work. Another step could be to use a `build.sh` script and add commands like `pwd` and `ls` to see exactly what the directory looks like and debug that way. It's not clear why your `setup.py` and `pyproject.toml` files are not being found. The contents of the tarball should be extracted into some standard directory without the `photochem-0.4.1_withdata` directory name.

Nicholaswogan:
@bkpoon I added the `ls` command to `build.sh`. In the Linux build, it looks like the source is uncompressed, but the server does not `cd` into the folder `photochem-0.4.1_withdata`. Do you know why this might be the case? It seems like Linux and MacOS are uncompressing the tar.gz in different ways.

```sh
BUILD STARTS HERE
ls
build_env_setup.sh
conda_build.sh
metadata_conda_debug.yaml
photochem-0.4.1_withdata

```

bkpoon:
I'm not sure why you have the `photochem-0.4.1_withdata` directory in the linux build, but not in the macOS build. And it looks like the GitHub source tarball failed as well. Someone else might be able to answer that.

A workaround would be to `cd` into the `photochem-${PKG_VERSION}_withdata` directory if it exists.
`${PKG_VERSION}` should be already be defined (https://docs.conda.io/projects/conda-build/en/stable/user-guide/environment-variables.html#id2)
Something like
```sh
if [ -d photochem-${PKG_VERSION}_withdata ]; then
  cd photochem-${PKG_VERSION}_withdata
fi 
```
or move the contents up one level so that the directory structure on linux is fixed
```sh
if [ -d photochem-${PKG_VERSION}_withdata ]; then
  mv photochem-${PKG_VERSION}_withdata/* ..
  rmdir photochem-${PKG_VERSION}_withdata
fi
```

Nicholaswogan:
@bkpoon I think everything is nearly working out for MacOS and Linux. Do you think this is possible for this build to work with windows? I see that scipy uses `- {{ compiler('m2w64_fortran') }}`.

For this build to work, I need a fortran compiler compatible with most all of Fortran. I think `flang` will not work.

bkpoon:
I would try 'flang` first and then 'm2w64_fortran`. Generally, you want to see if the regular compilers work first. One other note is that for OpenMP, the suggestion is to use 'libgomp` as the dependency (https://conda-forge.org/docs/maintainer/knowledge_base.html#openmp).

As for Apple Silicon builds, they are added after the recipe is merged (https://conda-forge.org/docs/maintainer/knowledge_base.html#apple-silicon-builds) by requesting a migration. It's a longer migration so it cannot be generally assumed that all the dependencies are available for Apple Silicon.

Nicholaswogan:
@bkpoon It seems like Flang is not working with the MSVC compiler. I get the error below from CMake. How could I use`clang` on windows instead?

```
ABI-incompatible: C compiler MSVC 19.29.30151.0 and Fortran compiler Flang 99.99.1
```

bkpoon:
You can try `m2w64_fortran` instead and see if that works. You may need to copy the `conda_build_config.yaml` file from the `scipy` feedstock since that recipe is choosing `vs2019` as their default Microsoft compiler. I'm not too familiar with Fortran on Windows, though. Have you gotten your Python package to build on Windows on a local system?

Another fix is to move the OpenMP dependencies to the `build` section of the requirements. Then you can remove the `_openmp_mutex` dependency.

Nicholaswogan:
i think my package should compile with windows because I have build older versions of the code on windows. I have a windows computer at home that I can use to test different ways of building.

Nicholaswogan:
@bkpoon I'm having issues building `photochem` on my local windows machine. I can only get it to work with mingw GNU C and Fortran compilers. But various issues pop up when using other compilers.

For now, I would like to skip windows builds on conda-forge. Someday, I might re-visit the problem, and add windows support.

Nicholaswogan:
Note that my code is linking to and `hdf5`, `zlib`. I put this dependencies in the `host` section. Everything seems to build OK and passes the import tests. Does this mean everything is ready?

bkpoon:
I'm looking through the build logs and it looks like you have additional dependencies that you are building with `CPMAddPackage` in `photochem/src/CMakeLists.txt`. Some from your repository and something from `sundials`. There is already a `sundials` package on conda-forge. Can you link to the libraries built from that package instead of rebuilding them yourself? From the most recent build log of that package (https://github.com/conda-forge/sundials-feedstock), it looks like `cvode` should exist in the package. If your code only works with version 5.7.0, then you can pin that version in your recipe. There are packages for linux, macOS (Intel and Apple Silicon), and Windows (https://anaconda.org/conda-forge/sundials).

For your other repositories, do you plan on making them available on `conda-forge` as well? Then you will want to add those first before adding `photochem`. You can add multiple feedstocks to the same pull request and the builds will sort out the dependencies (https://conda-forge.org/docs/maintainer/adding_pkgs.html#adding-multiple-packages-at-once). So if you add a `clima` dependency to `photochem` and add a `clima` recipe to this PR in a separate directory, the builds should work.

Generally, you will want to link to existing packages in the `conda-forge` channel. You do not want your shared/static libraries clobbering other packages. And it will be easier to test newer versions of `sundials`.

Nicholaswogan:
The problem with the `sundials` on conda-forge is that it does not contain the Fortran interface to `cvode`, and `photochem` needs the fortran interface. So I think I can not link to the conda-forge sundials.

`photochem` also depends on [`fortran-yaml-c`](https://github.com/Nicholaswogan/fortran-yaml-c), which is my fortran interface to `libyaml`. I use CMake Package Manager to download and then build `libyaml`. I see that conda-forge contains `libyaml`, but the problem is that the conda-forge libyaml is not findable by CMake, so I'm not sure how to link to it.

I don't want to add `clima` as a separate conda-forge package. I will access clima through the photochem package via `from photochem import clima`. The reason I want this is because `photochem` and `clima` both depend on quite a large chunk of package data (~25 Mb). It seems bad practice to have two separate packages both with the same 25 Mb of data.

Nicholaswogan:
Is it ok if this pull request is accepted as is? Or should I figure out how to link to libyaml on conda forge?

bkpoon:
For `sundials`, would adding `-DBUILD_FORTRAN_MODULE_INTERFACE=ON` to the `build.sh` (https://github.com/conda-forge/sundials-feedstock/blob/main/recipe/build.sh) make the package work for you? If so, can you make a PR to that feedstock?

For `libyaml`, can you modify your `fortran-yaml-c` to use the `yaml` package from `conda-forge`? Looking at it's recipe (https://github.com/conda-forge/yaml-feedstock/tree/main/recipe) and package (https://anaconda.org/conda-forge/yaml/files), it will have the `pkg-config` directory in `${PREFIX}/lib`, so you would need to tell `cmake` to look for that instead of a `cmake` module.

And for `clima` and `futils`, if you do not plan on making them separately available and you maintain them, then it should be okay to include them in this package.

I do not have merging permissions, so final approval will need to come from members of the`help-python` team. But using existing dependencies in `conda-forge` is important.

Nicholaswogan:
I do not think sundials will work without more effort than the patch you describe. I will experiment, however.

I will work on using the conda-forge libyaml.

bkpoon:
I built a version of `sundials` with the Fortran interface. You can get them from my custom channel for testing (https://anaconda.org/cctbx-dev/sundials).

The build logs are here (https://dev.azure.com/cctbx-release/feedstock-builds/_build/results?buildId=8167&view=results) and the changes are in my fork (https://github.com/bkpoon/sundials-feedstock/tree/fortran). The change is only for linux and macOS.

I did have to set the `SUNDIALS_INDEX_SIZE` back to the default because the Fortran interface requires it.

You can try building your package using a conda environment with your dependencies and the `sundials` package from my channel.

Nicholaswogan:
Ok I understand how to link to libyaml.

I need sundials version 5.7.0. Unfortunately, v6.5.1 will not work. One option is that I could change `photochem` to work with sundials v6.5.1. This will be a bit of work as sundials reorganized stuff quite a bit since v5.7.

There is a potential further problem. Fortran module files are not standardized. This means that different compilers produce different .mod files which are not compatible with each other. Even `gfortran` has changed its .mod file layout before and they may do it again (there is nothing stopping them from doing so). So it is conceivable that any Fortran module built today with e.g. gfortran 11 will not work with gfortran 16 or something like that (see this thread: https://fortran-lang.discourse.group/t/where-to-package-mod-files/5024/2).

The true solution to the sundials issue is if the Fortran interface was taken out of Sundials and put in a separate repository. Then I could just build the interface with CMake Package Manager, and link to conda-forge Sundials. But it seems hard to convince sundials to make this change in the near future.

Note that the `photochem` package will only contain Python c-extensions (like `_photochem.cpython-310-darwin.so`). My non-conda-forge cvode v5.7.0 will be fully contained within this binary. No one will be trying to link to these python extensions, so there won't be any issues with clobbering, I think?

Nicholaswogan:
If a package depends on `photochem`, it will be fine if it also depends on sundials v6.5.1. Unless for some reason the package links to a Python c-extension in `photochem`, in which case there will be clobbering. But I have never heard of someone linking to a Python c-extension.

Nicholaswogan:
What do you think @bkpoon. Are you OK with `photochem` building cvode v5.7.0 via CMake Package Manager?

Nicholaswogan:
Given the big barrier to linking to conda-forge sundials, and the very unlikely chance of clobbering, is it OK if `photochem` does not link to the conda-forge sundials and instead uses CMake Package Manager?

bkpoon:
Maybe? Were you able to link against the existing `yaml` package? I also built version 5.7.0 of `sundials` with the Fortran interface if you want to do more testing. It's available in the same channel as before (https://anaconda.org/cctbx-dev/sundials).

For ABI and API compatibility, `conda-forge` maintains a list of versions of different dependencies so that packages are generally able to link against one another (https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml). You can reasonably expect (if packages you're interested in are regularly maintained), that any current package can link to another current package if needed.

 It does look like you are statically linking against your dependencies so there should not be any clobbering. You can ping the `help-python` team to see if that is acceptable and if this can be merged.

Nicholaswogan:
@help-python can you see if this is ready to merge?

PR #23297, opened 2023-07-07
azure-sdk-for-cpp: core, identity, storage

jdblischak:
Checklist
- [x] Build number is 0.

https://github.com/Azure/azure-sdk-for-cpp

* This PR contains recipes for the core, identity, and the storage-* C++ libraries

  https://github.com/Azure/azure-sdk-for-cpp#vcpkg

* I've separated them into separate recipes (as opposed to a single recipe with multiple outputs) because even though it is a mono-repo, the release tags are per library

  https://github.com/Azure/azure-sdk-for-cpp/releases

* ~~I haven't used `run_exports` because the dependencies in the various `vcpkg.json` files are specified with only a minimum version bound~~ **UPDATE:** the recipes now use `run_exports` to pin the libraries to their major version

* I purposefully avoided the alternative core library with AMQP support ([`azure-core-amqp`](https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/azure-core-amqp)) because it depends on `azure-uamqp-c`, and we haven't yet been able to build it and its dependencies for all platforms (`azure-uamqp-c` -> `azure-c-shared-utility` -> `azure-macro-utils-c` & `umock-c`), eg https://github.com/conda-forge/azure-c-shared-utility-feedstock/pull/4

  https://github.com/conda-forge/azure-uamqp-c-feedstock

### DIFF
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -30,6 +30,10 @@ requirements:
     - wil  # [win]
 
 test:
+  requires:
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
   commands:
     # shared libs
     - test -f ${PREFIX}/lib/libazure-core${SHLIB_EXT}   # [unix]
@@ -43,6 +47,11 @@ test:
     # CMake metadata
     - test -f ${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake                     # [unix]
     - if not exist %LIBRARY_PREFIX%\share\azure-core-cpp\azure-core-cppTargets.cmake exit 1  # [win]
+
+    # link
+    - cmake $CMAKE_ARGS -G Ninja -S $RECIPE_DIR/test-linking-src -B test-linking-bld
+    - cmake --build test-linking-bld
+    - cmake --build test-linking-bld --target test
 
 about:
   home: https://github.com/Azure/azure-sdk-for-cpp/
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -49,7 +49,8 @@ test:
     - if not exist %LIBRARY_PREFIX%\share\azure-core-cpp\azure-core-cppTargets.cmake exit 1  # [win]
 
     # link
-    - cmake $CMAKE_ARGS -G Ninja -S $RECIPE_DIR/test-linking-src -B test-linking-bld
+    - cmake $CMAKE_ARGS -G Ninja -S $RECIPE_DIR/test-linking-src -B test-linking-bld    # [unix]
+    - cmake %CMAKE_ARGS% -G Ninja -S %RECIPE_DIR%\test-linking-src -B test-linking-bld  # [win]
     - cmake --build test-linking-bld
     - cmake --build test-linking-bld --target test
 
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -48,7 +48,7 @@ test:
     - test -f ${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake                     # [unix]
     - if not exist %LIBRARY_PREFIX%\share\azure-core-cpp\azure-core-cppTargets.cmake exit 1  # [win]
 
-    # link
+    # CMake find_package() and target_link_libraries()
     - cmake $CMAKE_ARGS -G Ninja -S $RECIPE_DIR/test-linking-src -B test-linking-bld    # [unix]
     - cmake %CMAKE_ARGS% -G Ninja -S %RECIPE_DIR%\test-linking-src -B test-linking-bld  # [win]
     - cmake --build test-linking-bld
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -32,8 +32,7 @@ requirements:
 test:
   commands:
     # shared libs
-    - test -f ${PREFIX}/lib/libazure-core.so            # [linux]
-    - test -f ${PREFIX}/lib/libazure-core.dylib         # [osx]
+    - test -f ${PREFIX}/lib/libazure-core${SHLIB_EXT}   # [unix]
     - if not exist %LIBRARY_BIN%\azure-core.dll exit 1  # [win]
     - if not exist %LIBRARY_LIB%\azure-core.lib exit 1  # [win]
 
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -16,7 +16,7 @@ source:
 build:
   number: 0
   run_exports:
-    - {{ pin_subpackage("azure-core-cpp", min_pin="x", max_pin="x") }}
+    - {{ pin_subpackage("azure-core-cpp", max_pin="x") }}
 
 # https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/vcpkg.json
 requirements:
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -62,4 +62,5 @@ about:
 extra:
   recipe-maintainers:
     - jdblischak
+    - jjerphan
     - Shelnutt2
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -1,0 +1,58 @@
+{% set name = "azure-core-cpp" %}
+{% set version = "1.10.0" %}
+{% set sha256 = "e6ae26ad30376bef6272024564764ba7bf5756063dc94cc4b74180cd004ab75e" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Azure/azure-sdk-for-cpp/archive/refs/tags/azure-core_{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
+  host:
+    - libcurl
+    - openssl  # [not win]
+    - wil  # [win]
+  run:
+    - libcurl
+    - openssl  # [not win]
+    - wil  # [win]
+
+test:
+  commands:
+    - test -f "${PREFIX}/include/azure/core.hpp"  # [not win]
+    - test -f "${PREFIX}/lib/libazure-core${SHLIB_EXT}"  # [not win]
+    - test -f "${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake"  # [not win]
+    - if not exist %PREFIX%\\Library\\bin\\azure-core.dll exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\include\\azure\\core.hpp exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\lib\\azure-core.lib exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\share\\azure-core-cpp\\azure-core-cppTargets.cmake exit 1  # [win]
+
+about:
+  home: https://github.com/Azure/azure-sdk-for-cpp/
+  summary: Azure SDK Core Library for C++
+  description: >-
+    Azure::Core (`azure-core`) provides shared primitives, abstractions, and
+    helpers for modern Azure SDK client libraries written in the C++
+  license: MIT
+  license_family: MIT
+  license_file:
+    - LICENSE.txt
+    - sdk/core/azure-core/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/azure-core/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -34,7 +34,7 @@ test:
 
     # headers
     - test -f ${PREFIX}/include/azure/core.hpp    # [unix]
-    - if not exist %LIBRARY_INC%\core.hpp exit 1  # [win]
+    - if not exist %LIBRARY_INC%\azure\core.hpp exit 1  # [win]
 
     # CMake metadata
     - test -f ${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake                     # [unix]
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -14,6 +14,7 @@ build:
   run_exports:
     - {{ pin_subpackage("azure-core-cpp", min_pin="x", max_pin="x") }}
 
+# https://github.com/Azure/azure-sdk-for-cpp/blob/main/sdk/core/azure-core/vcpkg.json
 requirements:
   build:
     - {{ compiler("cxx") }}
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -1,9 +1,8 @@
-{% set name = "azure-core-cpp" %}
 {% set version = "1.10.0" %}
 {% set sha256 = "e6ae26ad30376bef6272024564764ba7bf5756063dc94cc4b74180cd004ab75e" %}
 
 package:
-  name: {{ name|lower }}
+  name: azure-core-cpp
   version: {{ version }}
 
 source:
@@ -12,31 +11,34 @@ source:
 
 build:
   number: 0
+  run_exports:
+    - {{ pin_subpackage("azure-core-cpp", min_pin="x.x.x", max_pin="x.x") }}
 
 requirements:
   build:
     - {{ compiler("cxx") }}
     - cmake
-    - make  # [not win]
-    - ninja  # [win]
+    - ninja
   host:
-    - libcurl
-    - openssl  # [not win]
-    - wil  # [win]
-  run:
     - libcurl
     - openssl  # [not win]
     - wil  # [win]
 
 test:
   commands:
-    - test -f "${PREFIX}/include/azure/core.hpp"  # [not win]
-    - test -f "${PREFIX}/lib/libazure-core${SHLIB_EXT}"  # [not win]
-    - test -f "${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake"  # [not win]
-    - if not exist %PREFIX%\\Library\\bin\\azure-core.dll exit 1  # [win]
-    - if not exist %PREFIX%\\Library\\include\\azure\\core.hpp exit 1  # [win]
-    - if not exist %PREFIX%\\Library\\lib\\azure-core.lib exit 1  # [win]
-    - if not exist %PREFIX%\\Library\\share\\azure-core-cpp\\azure-core-cppTargets.cmake exit 1  # [win]
+    # shared libs
+    - test -f ${PREFIX}/lib/libazure-core.so            # [linux]
+    - test -f ${PREFIX}/lib/libazure-core.dylib         # [osx]
+    - if not exist %LIBRARY_BIN%\azure-core.dll exit 1  # [win]
+    - if not exist %LIBRARY_BIN%\azure-core.lib exit 1  # [win]
+
+    # headers
+    - test -f ${PREFIX}/include/azure/core.hpp    # [unix]
+    - if not exist %LIBRARY_INC%\core.hpp exit 1  # [win]
+
+    # CMake metadata
+    - test -f ${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake                     # [unix]
+    - if not exist %LIBRARY_PREFIX%\share\azure-core-cpp\azure-core-cppTargets.cmake exit 1  # [win]
 
 about:
   home: https://github.com/Azure/azure-sdk-for-cpp/
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   number: 0
   run_exports:
-    - {{ pin_subpackage("azure-core-cpp", min_pin="x.x.x", max_pin="x.x") }}
+    - {{ pin_subpackage("azure-core-cpp", min_pin="x", max_pin="x") }}
 
 requirements:
   build:
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -30,7 +30,7 @@ test:
     - test -f ${PREFIX}/lib/libazure-core.so            # [linux]
     - test -f ${PREFIX}/lib/libazure-core.dylib         # [osx]
     - if not exist %LIBRARY_BIN%\azure-core.dll exit 1  # [win]
-    - if not exist %LIBRARY_BIN%\azure-core.lib exit 1  # [win]
+    - if not exist %LIBRARY_LIB%\azure-core.lib exit 1  # [win]
 
     # headers
     - test -f ${PREFIX}/include/azure/core.hpp    # [unix]
--- a/recipes/azure-core-cpp/meta.yaml
+++ b/recipes/azure-core-cpp/meta.yaml
@@ -8,6 +8,10 @@ package:
 source:
   url: https://github.com/Azure/azure-sdk-for-cpp/archive/refs/tags/azure-core_{{ version }}.tar.gz
   sha256: {{ sha256 }}
+  patches:
+    # https://github.com/Azure/azure-sdk-for-cpp/issues/4784
+    # https://github.com/Azure/azure-sdk-for-cpp/pull/4785
+    - 0001-remove-wil-from-exported-config.patch
 
 build:
   number: 0
--- a/recipes/azure-identity-cpp/bld.bat
+++ b/recipes/azure-identity-cpp/bld.bat
@@ -1,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\identity\azure-identity
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G "Ninja" ^
+  -D BUILD_SHARED_LIBS=ON ^
+  -D BUILD_TRANSPORT_WINHTTP=ON ^
+  ..
+if %ERRORLEVEL% neq 0 exit 1
+
+cmake --build . --target install --config Release
+if %ERRORLEVEL% neq 0 exit 1
--- a/recipes/azure-identity-cpp/bld.bat
+++ b/recipes/azure-identity-cpp/bld.bat
@@ -7,7 +7,7 @@ cd sdk\identity\azure-identity
 mkdir build
 cd build
 cmake %CMAKE_ARGS% ^
-  -G "Ninja" ^
+  -G Ninja ^
   -D BUILD_SHARED_LIBS=ON ^
   -D BUILD_TRANSPORT_WINHTTP=ON ^
   ..
--- a/recipes/azure-identity-cpp/build.sh
+++ b/recipes/azure-identity-cpp/build.sh
@@ -12,20 +12,13 @@ then
   CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
 fi
 
-# For some reason CMAKE_BUILD_TYPE=Release isn't included in CMAKE_ARGS by
-# ctng-compiler-activation-feedstock, even though it is for vc-feedstock and
-# clang-compiler-activation-feedstock
-if [[ "${target_platform}" == linux-* ]]
-then
-  CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release"
-fi
-
 cd sdk/identity/azure-identity
 
 # https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#building-the-project
 mkdir build
 cd build
 cmake $CMAKE_ARGS \
+  -D CMAKE_BUILD_TYPE=Release \
   -G Ninja \
   -D BUILD_SHARED_LIBS=ON \
   -D BUILD_TRANSPORT_CURL=ON \
--- a/recipes/azure-identity-cpp/build.sh
+++ b/recipes/azure-identity-cpp/build.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -eux
+
+# Disable installing dependencies with vcpkg
+# https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#customize-the-vcpkg-dependency-integration
+# https://github.com/search?utf8=%E2%9C%93&q=AZURE_SDK_DISABLE_AUTO_VCPKG&type=code
+export AZURE_SDK_DISABLE_AUTO_VCPKG=ON
+
+# https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
+if [[ "${target_platform}" == osx-* ]]
+then
+  CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+cd sdk/identity/azure-identity
+
+# https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#building-the-project
+mkdir build
+cd build
+cmake $CMAKE_ARGS \
+  -D BUILD_SHARED_LIBS=ON \
+  -D BUILD_TRANSPORT_CURL=ON \
+  ..
+cmake --build . --target install --config Release
--- a/recipes/azure-identity-cpp/build.sh
+++ b/recipes/azure-identity-cpp/build.sh
@@ -26,6 +26,7 @@ cd sdk/identity/azure-identity
 mkdir build
 cd build
 cmake $CMAKE_ARGS \
+  -G Ninja \
   -D BUILD_SHARED_LIBS=ON \
   -D BUILD_TRANSPORT_CURL=ON \
   ..
--- a/recipes/azure-identity-cpp/build.sh
+++ b/recipes/azure-identity-cpp/build.sh
@@ -12,6 +12,14 @@ then
   CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
 fi
 
+# For some reason CMAKE_BUILD_TYPE=Release isn't included in CMAKE_ARGS by
+# ctng-compiler-activation-feedstock, even though it is for vc-feedstock and
+# clang-compiler-activation-feedstock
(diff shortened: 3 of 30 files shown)


Shelnutt2:
I am willing to maintain these packages. Thanks @jdblischak !

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - make  # [not win]
+    - ninja  # [win]
+  host:
+    - libcurl
+    - openssl  # [not win]
+    - wil  # [win]
+  run:
+    - libcurl
+    - openssl  # [not win]
+    - wil  # [win]
```
xhochy, line 29: This whole section shouldn't be needed. The first two come via run_exports and the last one is a header only library.

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
+{% set name = "azure-core-cpp" %}
```
xhochy, line 1: To be consistent with other packages, we should name this li azure-core

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Azure/azure-sdk-for-cpp/archive/refs/tags/azure-core_{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
```
xhochy, line 14: This should definitely have a ru_exports as having it in host leads to a run dependency on it.

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - cmake
+    - make  # [not win]
```
xhochy, line 20: For simplicity, build with ninja on all platforms.

xhochy (changes requested):
Comments apply to all recipes but I limited them to the first.

xhochy:
@h-vetinari please also have a look as this is third big cloud vendor C++ SDK

recipes/azure-identity-cpp/build.sh
```diff
@@ -0,0 +1,32 @@
...
+  CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+# For some reason CMAKE_BUILD_TYPE=Release isn't included in CMAKE_ARGS by
+# ctng-compiler-activation-feedstock, even though it is for vc-feedstock and
+# clang-compiler-activation-feedstock
+if [[ "${target_platform}" == linux-* ]]
+then
+  CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release"
+fi
```
h-vetinari, line 21: Just add `-DCMAKE_BUILD_TYPE=Release` to the CMake invocation, the duplication doesn't hurt. Having extra treatment should be reserved for things where that's not possible or reasonable for some reason. Also feel free to raise an issue about this on the ctng activation feedstock.

recipes/azure-identity-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - {{ compiler("cxx") }}
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
+  host:
+    - azure-core-cpp >=1.9.0
+    - openssl
+  run:
+    - azure-core-cpp >=1.9.0
+    - openssl
```
h-vetinari, line 27: More things to remove that are (to be) covered by run-exports.

recipes/azure-identity-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
+  host:
+    - azure-core-cpp >=1.9.0
```
h-vetinari, line 23: I think we'll have to add this to the pinning, and in this case the `>=` will be detrimental to having this handled by the bot.

recipes/azure-identity-cpp/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+test:
+  commands:
+    - test -f "${PREFIX}/include/azure/identity.hpp"  # [not win]
+    - test -f "${PREFIX}/lib/libazure-identity${SHLIB_EXT}"  # [not win]
+    - test -f "${PREFIX}/share/azure-identity-cpp/azure-identity-cppTargets.cmake"  # [not win]
+    - if not exist %PREFIX%\\Library\\bin\\azure-identity.dll exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\include\\azure\\identity.hpp exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\lib\\azure-identity.lib exit 1  # [win]
+    - if not exist %PREFIX%\\Library\\share\\azure-identity-cpp\\azure-identity-cppTargets.cmake exit 1  # [win]
```
h-vetinari, line 37: Please structure this a bit better (same for other recipes). Also, double back-slashes are not necessary.

```suggestion
    # shared libs
    - test -f "${PREFIX}/lib/libazure-identity.so"          # [linux]
    - test -f "${PREFIX}/lib/libazure-identity.dylib"       # [osx]
    - if not exist %LIBRARY_BIN%\azure-identity.dll exit 1  # [win]
    - if not exist %LIBRARY_BIN%\azure-identity.lib exit 1  # [win]

    # headers
    - test -f "${PREFIX}/include/azure/identity.hpp"  # [unix]
    - if not exist %LIBRARY_INC%\identity.hpp exit 1  # [win]

    # CMake metadata
    - test -f "${PREFIX}/share/azure-identity-cpp/azure-identity-cppTargets.cmake"                   # [unix]
    - if not exist %LIBRARY_PREFIX%\share\azure-identity-cpp\azure-identity-cppTargets.cmake exit 1  # [win]
```

h-vetinari: It's also very weird to me that the CMake metadata is not in `$PREFIX/lib/cmake/<project>`, which is ~the standard~ much more common as a location.

recipes/azure-storage-blobs-cpp/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "azure-storage-blobs-cpp" %}
+{% set version = "12.7.0" %}
+{% set sha256 = "a28a665c740ae510053afa2c9b794052aa4de11d122b3ac4eaa0bb3bccb8983e" %}
+
+package:
+  name: {{ name|lower }}
```
h-vetinari, line 6: Please don't use the name variables, and just set this directly
```suggestion
  name: azure-storage-blobs
```

recipes/azure-storage-common-cpp/build.sh
```diff
@@ -0,0 +1,32 @@
...
+# clang-compiler-activation-feedstock
+if [[ "${target_platform}" == linux-* ]]
+then
+  CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_BUILD_TYPE=Release"
+fi
+
+# https://github.com/Azure/azure-sdk-for-cpp/blob/main/CONTRIBUTING.md#building-the-project
+mkdir build
+cd build
+cmake $CMAKE_ARGS \
```
h-vetinari, line 28: I suggest to just use ninja as a backend everywhere
```suggestion
cmake -G Ninja \
    $CMAKE_ARGS \
```

recipes/azure-storage-common-cpp/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler("cxx") }}
+    - cmake
+    - make  # [not win]
+    - ninja  # [win]
```
h-vetinari, line 21:
```suggestion
    - cmake
    - ninja
```

h-vetinari (commented):
Thanks for the PR!

Aside from some general clean-up requests, I'd expect that we need some sort of `azure-sdk` metapackage that pulls everything together?

It would also be good to know what kind of ABI-stability (if any) these packages have. Migrating for every version is quite painful in the aws stack, for [example](https://github.com/conda-forge/aws-sdk-cpp-feedstock/issues/662).

To summarize the clean-ups:
* [ ] add run-exports
* [ ] don't pin (or lower-bound) packages where we will need to rebuild for new versions
* [ ] remove run-requirements where we can rely on run-exports
* [ ] don't special-case `-DCMAKE_BUILD_TYPE=Release` in build scripts
* [ ] better document & structure the test sections
* [ ] don't use `{{ name }}`
* [ ] [optional] use ninja universally

recipes/azure-core-cpp/meta.yaml
jdblischak: What are you suggesting the name should be exactly? We can't call it `azure-core` because that is used by the Python SDK: https://anaconda.org/conda-forge/azure-core

Similarly with the other SDK libraries: https://anaconda.org/conda-forge/azure-storage-common

recipes/azure-core-cpp/meta.yaml
jdblischak: Ok, I'll add

recipes/azure-identity-cpp/build.sh
jdblischak: I had already raised the issue: https://github.com/conda-forge/ctng-compiler-activation-feedstock/issues/94

recipes/azure-identity-cpp/meta.yaml
jdblischak: Is this a problem? Are you suggesting we override it for the conda installation?

jdblischak:
@xhochy and @h-vetinari Thanks for the thorough review! I've applied your suggestions to the `azure-core-cpp` recipe. Once that is in good shape, I'll apply the changes to the remaining recipes. My biggest uncertainty is around what to set for the `min_pin` and `max_pin` for the `run_exports`.

h-vetinari:
> My biggest uncertainty is around what to set for the `min_pin` and `max_pin` for the `run_exports`.

Leave out `min_pin` and set `max_pin` to `'x.x.x'`, unless you know that the ABI will not change between 1.2.x and 1.2.y.

teo-tsirpanis:
> unless you know that the ABI will not change between 1.2.x and 1.2.y.

They declare compatibility within the same major version. (https://github.com/Azure/azure-sdk-for-cpp/blob/2f3e5de931c3f1c19551f6f58418c85aa410745b/cmake-modules/AzureVcpkg.cmake#L152-L156)

recipes/azure-core-cpp/meta.yaml
teo-tsirpanis: I don't think 2 is an option because it would make `azure-core-cpp` unusable by itself.

jdblischak:
@xochy @h-vetinari I removed the runtime requirements, but now I am getting the following errors. This was my initial motivation for adding them to the run requirements:

```
   INFO (azure-storage-blobs-cpp,lib/libazure-storage-blobs.so): Needed DSO lib/libazure-storage-common.so found in home/conda/staged-recipes/build_artifacts::azure-storage-common-cpp-12.3.2-heddee37_0
  ERROR (azure-storage-blobs-cpp,lib/libazure-storage-blobs.so): Needed DSO lib/libazure-core.so found in ['azure-core-cpp']
  ERROR (azure-storage-blobs-cpp,lib/libazure-storage-blobs.so): .. but ['azure-core-cpp'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

jdblischak:
I've cleaned up the recipes:

- [x] add run-exports
- [x] don't pin (or lower-bound) packages where we will need to rebuild for new versions
- [x] remove run-requirements where we can rely on run-exports
- [x] don't special-case `-DCMAKE_BUILD_TYPE=Release` in build scripts
- [x] better document & structure the test sections
- [x] don't use `{{ name }}`
- [ ] [preferable] put CMake metadata in `$PREFIX/lib/cmake/<project>` (resp. `%LIBRARY_LIB%\cmake\<project>`)
- [x] [optional] use ninja universally

The only one I didn't do was moving the CMake metadata because I wasn't sure how best to accomplish this. Can I simply add a `mv` command after `cmake --build` to move those files, or do I need to do something more complex like patch their `CMakeLists.txt`?

jdblischak:
And some bigger picture questions to discuss:

1. Should I create a metapackage azure-sdk-for-cpp?
2. I didn't include every single SDK library in this PR. Do I need to add them all in this PR, or can I send them in a subsequent PR?
3. Are we ok with the run_exports being pinned to the major version?
4. Do we need to coordinate this PR with conda-forge-pinning? If yes, how?
5. Would anyone else like to be added as a recipe maintainer?

recipes/azure-core-cpp/meta.yaml
jdblischak: After discussing with an upstream Azure SDK developer, we decided to apply a patch to remove the exported dependency on `wil` in the config file for `azure-core-cpp`

For more context, see https://github.com/Azure/azure-sdk-for-cpp/issues/4784 and https://github.com/Azure/azure-sdk-for-cpp/pull/4785

recipes/azure-core-cpp/meta.yaml
jdblischak: > To be consistent with other packages, we should name this li azure-core

@xhochy Is your suggestion to name it `lib-azure-core`? If yes, what other packages follow this convention? The [AWS SDK for C++](https://github.com/conda-forge/aws-sdk-cpp-feedstock) is named `aws-sdk-cpp`

And in general, I think it would be potentially confusing if the conda-forge recipes all used different names than upstream

jdblischak:

recipes/azure-identity-cpp/build.sh
jdblischak: For potential future readers of this thread, the ctng-compiler-activation-feedstock now includes `CMAKE_BUILD_TYPE=Release` in the exported env var `CMAKE_ARGS` (https://github.com/conda-forge/ctng-compiler-activation-feedstock/pull/95/files)

carterbox:
> * Should I create a metapackage azure-sdk-for-cpp?

Only if you think people would actually want to install all of the components in one go.

> * I didn't include every single SDK library in this PR. Do I need to add them all in this PR, or can I send them in a subsequent PR?

I prefer that related recipes are clumped together because it's easier to review them for consistency and interdependency, but 10 recipes in a PR is alot. Make a future PR to add more packages in this ecosystem.

> * Are we ok with the run_exports being pinned to the major version?

As long as you (as the maintainer) trust the [aformentioned declaration of ABI compatability](https://github.com/conda-forge/staged-recipes/pull/23297#issuecomment-1629101717). You're the one who primarily need to clean up the mess if there are ABI breaks.

> * Do we need to coordinate this PR with conda-forge-pinning? If yes, how?

Yes. Ensure that there is a [global pinning](https://conda-forge.org/docs/maintainer/pinning_deps.html#globally-pinned-packages) added before you update to your next version of these packages. If there is only one version available, then the pinning isn't needed. Once there are multiple versions, that's when pinnings become relevant.

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  sha256: {{ sha256 }}
+  patches:
+    # https://github.com/Azure/azure-sdk-for-cpp/issues/4784
+    # https://github.com/Azure/azure-sdk-for-cpp/pull/4785
+    - 0001-remove-wil-from-exported-config.patch
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage("azure-core-cpp", min_pin="x", max_pin="x") }}
```
carterbox, line 19:
```suggestion
  run_exports:
    - {{ pin_subpackage("azure-core-cpp", max_pin="x") }}
```
Please update all your exports to unspecify the min_pin. SameMajorVersion means backwards compatability, not forward and backward compatability.

recipes/azure-core-cpp/meta.yaml
carterbox: I believe there is a typo in xhochy's comment. It should read

> we should name this libazure-core

If the package is just headers, config, and libraries (no CLI tools or other binaries), then it should be named libazure-core because the package contains libazure-core.so/dylib.

Examples:
 - libclang is libraries separate from clang the tool
 - libcurl is libraries separate from curl the tool
 - libopenvino just libs from the openvino project
 - libmad is libs from the mad project

carterbox (changes requested):

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/core/azure-core/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/azure-core/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 65: If you need another maintainer.
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-identity-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\identity\azure-identity
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-identity-cpp/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/identity/azure-identity/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/identity/azure-identity/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 63:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-storage-blobs-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\storage\azure-storage-blobs
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  host:
+    - libcurl
+    - openssl  # [not win]
+    - wil  # [win]
+
+test:
+  commands:
+    # shared libs
+    - test -f ${PREFIX}/lib/libazure-core.so            # [linux]
+    - test -f ${PREFIX}/lib/libazure-core.dylib         # [osx]
```
jjerphan, line 36: Nit: this can be rewritten using and `SHLIB_EXT`. 
```suggestion
    - test -f ${PREFIX}/lib/libazure-core${SHLIB_EXT}   # [unix]
```

This suggestion also applies to the other libraries' recipe.

recipes/azure-storage-blobs-cpp/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/storage/azure-storage-blobs/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/storage/azure-storage-blobs/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 63:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-storage-common-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\storage\azure-storage-common
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-storage-common-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/storage/azure-storage-common/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/storage/azure-storage-common/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 65:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-storage-files-datalake-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\storage\azure-storage-files-datalake
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-storage-files-datalake-cpp/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/storage/azure-storage-files-datalake/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/storage/azure-storage-files-datalake/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 66:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-storage-files-shares-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\storage\azure-storage-files-shares
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-storage-files-shares-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/storage/azure-storage-files-shares/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/storage/azure-storage-files-shares/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 65:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-storage-queues-cpp/bld.bat
```diff
@@ -0,0 +1,17 @@
+@echo on
+
+set "AZURE_SDK_DISABLE_AUTO_VCPKG=ON"
+
+cd sdk\storage\azure-storage-queues
+
+mkdir build
+cd build
+cmake %CMAKE_ARGS% ^
+  -G Ninja ^
```
jjerphan, line 10:
```suggestion
  -D CMAKE_BUILD_TYPE=Release ^
  -G Ninja ^
```

recipes/azure-storage-queues-cpp/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  license_file:
+    - LICENSE.txt
+    - sdk/storage/azure-storage-queues/NOTICE.txt
+  doc_url: https://azure.github.io/azure-sdk-for-cpp/
+  dev_url: https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/storage/azure-storage-queues/
+
+extra:
+  recipe-maintainers:
+    - jdblischak
+    - Shelnutt2
```
jjerphan, line 64:
```suggestion
    - Shelnutt2
    - jjerphan
```

recipes/azure-core-cpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - if not exist %LIBRARY_BIN%\azure-core.dll exit 1  # [win]
+    - if not exist %LIBRARY_LIB%\azure-core.lib exit 1  # [win]
+
+    # headers
+    - test -f ${PREFIX}/include/azure/core.hpp    # [unix]
+    - if not exist %LIBRARY_INC%\azure\core.hpp exit 1  # [win]
+
+    # CMake metadata
+    - test -f ${PREFIX}/share/azure-core-cpp/azure-core-cppTargets.cmake                     # [unix]
+    - if not exist %LIBRARY_PREFIX%\share\azure-core-cpp\azure-core-cppTargets.cmake exit 1  # [win]
```
jjerphan, line 46: To know if a library CMake configuration can be used, and to document its use, those tests could be completed with the build of a simple CMake project using this library.

Here is [a simple example of such tests](https://github.com/conda-forge/staged-recipes/pull/23539/files#diff-0287fef753ebf619329766f0ff34c4fe490a07b4d759d9b6e7063d6118661610R36-R42).

This suggestion also applies to the other libraries' recipe.

recipes/azure-core-cpp/meta.yaml
jjerphan: Also, shouldn't we pin the minor version instead?
```suggestion
  run_exports:
    - {{ pin_subpackage("azure-core-cpp", min_pin="x", max_pin="x.x") }}
```

This suggestion also applies to the other libraries' recipe.

jjerphan (commented):
Thank you for bringing Azure SDK for C++ on conda-forge, @jdblischak!

Here are a few comments, the main important one being about pinning on minor versions rather than on major ones.

I also think we must prefix the names of those libraries with `lib`.

I volunteer to maintain feedstocks.

recipes/azure-core-cpp/meta.yaml
jdblischak: I'd love another maintainer. Thanks for volunteering!

recipes/azure-core-cpp/meta.yaml
jdblischak: We currently have it set to pin to only the major version because the [upstream CMake file](https://github.com/Azure/azure-sdk-for-cpp/blob/2f3e5de931c3f1c19551f6f58418c85aa410745b/cmake-modules/AzureVcpkg.cmake#L152-L156) declares [SameMajorVersion](https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html) compatibility. See also the previous comments https://github.com/conda-forge/staged-recipes/pull/23297#issuecomment-1629101717 and https://github.com/conda-forge/staged-recipes/pull/23297#issuecomment-1659338135. @jjerphan what do you think of this choice? Pinning to the minor version would be the more conservative route, but it would also require more maintenance when migrating the pins

jdblischak:
@carterbox Thanks for the review!

> Only if you think people would actually want to install all of the components in one go.

I like that advice. I don't need a single meta-package myself. If someone else needs it, they can request it or create it themselves

> I prefer that related recipes are clumped together because it's easier to review them for consistency and interdependency, but 10 recipes in a PR is alot. Make a future PR to add more packages in this ecosystem.

Happy to comply with this. Updating 7 recipes at a time is cumbersome enough

> As long as you (as the maintainer) trust the https://github.com/conda-forge/staged-recipes/pull/23297#issuecomment-1629101717. You're the one who primarily need to clean up the mess if there are ABI breaks.

Ok, let's see if all the maintainers agree. I know that the pinning migrations for the similar `aws-*` packages are non-trivial, so I'd prefer to minimize this as much as possible

> Yes. Ensure that there is a [global pinning](https://conda-forge.org/docs/maintainer/pinning_deps.html#globally-pinned-packages) added before you update to your next version of these packages. If there is only one version available, then the pinning isn't needed. Once there are multiple versions, that's when pinnings become relevant.

Wonderful. Thanks for the concise explanation. I will need to add the pins after the initial feedstock creation but prior to the first updates

recipes/azure-core-cpp/meta.yaml
jdblischak: Applied. Thanks!

recipes/azure-core-cpp/meta.yaml
jdblischak: I'm open to adding the prefix `lib`. My main concerns are consistency and transparency. I think there is an advantage to having the conda recipe names exactly match the upstream names

> libclang is libraries separate from clang the tool

But for this example, the upstream maintainers have also named it [libclang](https://clang.llvm.org/doxygen/group__CINDEX.html), right? Same with [libcurl](https://curl.se/libcurl/). This seems like a slightly different situation than what we are discussing here, where the `lib` prefix name would only exist in conda-forge

My understanding is that the prefixes `-cpp` (and `-c`) communicate that the software contains no CLI tools or other binaries. Looking across the conda-forge ecosystem, the many various AWS libraries either [end in `-cpp`](https://github.com/search?q=org%3Aconda-forge+path%3Arecipe%2Fmeta.yaml+%2Fname%3A%5Csaws-.*-cpp%2F&type=code) or start with [`aws-c-`](https://github.com/search?q=org%3Aconda-forge+path%3Arecipe%2Fmeta.yaml+%2Fname%3A%5Csaws-c-.*%2F&type=code) (including the very similar [aws-sdk-cpp](https://github.com/conda-forge/aws-sdk-cpp-feedstock)). There are also Azure C libraries that either start with `azure-c` or end in `-c` (see the [recipe for azure-iot-sdk-c](https://github.com/conda-forge/azure-iot-sdk-c-feedstock/blob/main/recipe/meta.yaml) for various iterations on the naming scheme). I see that the `lib` prefix was added to libgoogle-cloud (https://github.com/conda-forge/google-cloud-cpp-feedstock/pull/83), but the `-cpp` was also dropped.

Given the existence of the recipes for `aws-c-*`, `aws-sdk-cpp`, etc. already in conda-forge, I'd argue that a name like `azure-core-cpp` already communicates that it is only a C++ library, with the added benefit that it matches the official upstream name.

Is anyone somewhat convinced by my argument above? I don't feel that strongly about it, but I wanted to at least present the counter-argument before we deviate the name from upstream. And if we do decide to add the prefix `lib`, should we drop the suffix `-cpp`?

teo-tsirpanis: I agree that the `lib` prefix is excessive. Names like `azure-core-cpp` unquestionably refer to C++ libraries. Moreover this is the official name of the libraries, and we use the official names in the other cloud SDKs.

recipes/azure-core-cpp/meta.yaml
jjerphan: Thank you!

In this case, I think you can use a simpler CMake setup such as, for `azure-identity-cpp`:

```cmake
cmake_minimum_required(VERSION 3.0)
project(azure-identity-cpp-dependent-project LANGUAGES C CXX)

find_package(azure-identity-cpp REQUIRED)
```

recipes/azure-core-cpp/meta.yaml
jjerphan: I see. I think we better wait for an actual administrator of conda-forge. If you and another maintainer agreed on pinning on the major version already, just do ignore my comment.

recipes/azure-core-cpp/meta.yaml
jdblischak: Ah, I see. You mainly want to test that `find_package()` can find the installed CMake configuration files. I'll leave my more comprehensive test of azure-core-cpp (ie that includes `target_link_libraries()`), but then use the simpler `CMakeLists.txt` for the other libraries.

jjerphan (approved):
Thank you for packaging those libraries from the Azure SDK for C++, @jdblischak! I believe this is useful for a lot of projects.

Regarding what's left on the todo list (https://github.com/conda-forge/staged-recipes/pull/23297#issuecomment-1631265485): I think we can figure moving the CMake metadata to the canonical path in the feedstocks since the test using CMake projects work.

Regarding packages names: prefixing C++ and C library with `lib` is conventional for conda-forge. Yet, the developers of the Azure SDK for C++ chose to suffix them with `-c` or `-cpp` (which is also explicit). I am OK with using names that are suffixed with `-c` and `cpp`, especially if this prevent discrepancies between with identifiers to use for CMake.

I think we should also be careful about some packages being renamed. For instance `azure-macro-utils-c` was [renamed `macro-utils-c` upstream](https://github.com/Azure/macro-utils-c) but [is published on conda-forge as `azure-macro-utils-c`](https://anaconda.org/conda-forge/azure-macro-utils-c). Republishing is an option I think, but I do not know if there's a better option for this situation

jdblischak:

jdblischak:
Higher-level questions that are still being actively discussed:

* Do we maintain the official upstream library names, eg azure-core-cpp, or rename them in conda-forge, eg `libazure-core-cpp` or `libazure-core`? I am currently leaning towards keeping the official names. However, I'm willing to be convinced if provided examples that this sort of renaming is common in conda-forge. Given that the existing recipes for AWS and Azure C/C++ libraries have not been renamed, I think it would make sense to continue this tradition

* Do we pin the `run_exports` to the major or minor pin? I'm currently leaning towards the major pin since 1) this is what the upstream maintainers say they will follow, and 2) this will reduce the maintenance burden from conda-forge-pinning migrations

Lastly, if anyone else would like to be added to the list of maintainers, please let me know

recipes/azure-core-cpp/meta.yaml
carterbox: It's not immediately apparent that the official name of the component is 'azure-core-cpp' when the components are bundled together under a project called 'azure-sdk-for-cpp'. 

[The upstream project's docs](https://github.com/Azure/azure-sdk-for-cpp/tree/main/sdk/core/azure-core) indicate that both the CMake package and vcpkg names are `azure-core-cpp`, so let's go with that because it is explicit evidence in the upstream projects docs/source that the upstream maintainers have named these packages in that way. In the future, let's lead with links instead of researching whether or not it's conda-forge convention to name C/CXX libraries with a `lib-` prefix.

jdblischak:
Thanks all for your reviews! The recipes are much improved thanks to all of your feedback

teo-tsirpanis:
Great work! Small question: I see the packages are built only for x64 architectures. Is there a way to add support for linux-aarch64, macos-arm64 and linux-ppc64le?

jjerphan:
We need to assess whether the libraries have been designed for the support of those architectures. We can try their support via individual PRs.

PR #23313, opened 2023-07-10
Add noiseflow v0.0.8

OUCyf:
Checklist
- [x] Build number is 0.


recipes/noiseflow/build_linux.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+git clone https://github.com/xtensor-stack/xtensor-fftw extern/xtensor-fftw
+git clone https://github.com/kfrlib/kfr extern/kfr
```
jaimergp, line 4: If these clones are to provide new dependencies, you'll need to provide these as conda-forge packages instead

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,79 @@
...
+source:
+  url: https://github.com/{{ org }}/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: ae0f060010667a924f7281976762d85cf5ba2402ccb404a0bab7a5fc893188fe
+
+build:
+  number: 0
+  skip: True  # [python_impl == 'pypy']
+  
+
+outputs:
```
jaimergp, line 18: I don't think you need `outputs` for this. There are several differences in behaviour that are very subtle and might be causing issues with your current setup. I'd try using a regular global level for this recipe and see if it fixes the Linux issues.

recipes/noiseflow/build_linux.sh
OUCyf: Thanks for your mention. Those 2 packages worked well, I have provided all dependencies for them.

recipes/noiseflow/meta.yaml
OUCyf: Thanks, I will delete `outputs` and try again. Actually, I want to add some other output in the future, so I just add the `outputs` here.

recipes/noiseflow/build_linux.sh
tdegeus: What @jaimergp means : there should not be a git clone here. Rather you should create (in the PR) two additional feedstocks (one for each dependency). 

Now. `xtensor-fftw` is simply available: https://github.com/conda-forge/xtensor-fftw-feedstock , so no need to clone or to create a new feedstock. Just add `xtensor-fftw` to `build`.

recipes/noiseflow/build_osx.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+export MACOSX_DEPLOYMENT_TARGET=10.13
```
tdegeus, line 2:
```suggestion
export MACOSX_DEPLOYMENT_TARGET=10.13
```
Needed?

recipes/noiseflow/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+python:
+  - 3.9
+  - 3.10
+  - 3.11
```
tdegeus, line 4: Usually you would not have to make this file.

recipes/noiseflow/build_osx.sh
tdegeus: (Or rather... What do you need it for? Can you avoid it?)

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - ninja  
+
+  host:
+    - pip
+    - numpy >=1.25.0
+    - fftw  # [unix]
+    - pybind11  # [unix]
+    - xtensor  # [unix]
+    - xtensor-python  # [unix]
+    - xsimd =8.0.5  # [unix]
```
tdegeus, line 41:
```suggestion
    - xsimd =8.0.5  # [unix]
```
It is not a good idea to enable simd: the hardware that you compile on and run on are not the same. So this will lead to errors

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - cmake
+    - numpy
+    - pip  
+    - make  # [unix]
+    - ninja  
+
+  host:
+    - pip
+    - numpy >=1.25.0
+    - fftw  # [unix]
```
tdegeus, line 37: I'm surprised that you have `# [unix[` for key dependencies. Why is that?

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - tqdm >=4.65.0
+    - obspy >=1.4.0
+    - faker >=18.13.0
+
+test:
+  imports:
+    - noiseflow
+  commands:
+    - pip check
+    - python -c "import noiseflow; ver = noiseflow.__version__; assert ver != '0.0.0'"
```
tdegeus, line 63:
```suggestion
    - python -c "import noiseflow; ver = noiseflow.__version__; assert ver = '{{ version }}'"
```

recipes/noiseflow/build_linux.sh
tdegeus: In addition, conda-build seems to fail over this

OUCyf: For conda build have be failed over there. Nope, you are looking for the new commit for same PR, which I have made some changes according to @jaimergp's suggestions to do the tests. In fact, they are all working well, I can literally install the package but just can't import it in the test stage

recipes/noiseflow/conda_build_config.yaml
OUCyf: Yes, now I don't know how to fix the Python versions for the package, so I add those for a try

recipes/noiseflow/meta.yaml
OUCyf: I just build the c++ lib for macos and linux, not for windows, cuz I can't compile it

recipes/noiseflow/meta.yaml
OUCyf: Gotcha

recipes/noiseflow/build_linux.sh
OUCyf: Yes!

recipes/noiseflow/meta.yaml
tdegeus: Do you mean: I will never be able to compile on Windows? Then you should skip it

OUCyf:
@conda-forge-admin Ready to review

OUCyf:
@conda-forge/help-python, ready to merge

recipes/noiseflow/build_linux.sh
OUCyf: Hey, sorry for the late reply, I lost my glasses 3 days ago, and now I am back to work. I think I need to use `git clone`, because I need to use `FindFFTW.cmake` script from `xtensor-fftw` in my cmake file. An easy approach is to just copy the `xtensor-fftw` repositories to my package, but I don't want to do that. So I use `git clone` to make everybody know that it's from `xtensor-fftw` repositories.

OUCyf: And finally i solved the problem, it's `poetry` problem. I must add the following lines in my `pyproject.toml` to let the building tool to copy the `*.py` file to package. And in `macos` and `win`, there are no problems when I didn't add those lines, but in `linux` (at least in `centos`, cuz conda-forge use the docker of `centos`) I must add those lines.

Thanks for your guys help!!! @jaimergp @tdegeus 

```toml
[tool.poetry]
include = ["CMakeLists.txt", ".gitmodules", "noiseflow/**/*.py"]
exclude = ["noiseflow/tests/**/*"]
```

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/{{ org }}/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 20736acea7dde31bbd3c6c6c3fbe922bedf242dd0b64bcd02a4557b4d566b427
+
+build:
+  number: 0
+  script: 
+    - git clone https://github.com/xtensor-stack/xtensor-fftw extern/xtensor-fftw  # [unix]
```
tdegeus, line 16:
```suggestion
    # to use the custom FindPackage CMake assistance
    - git clone https://github.com/xtensor-stack/xtensor-fftw extern/xtensor-fftw  # [unix]
```

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - git clone https://github.com/kfrlib/kfr extern/kfr  # [unix]
+    - export LLVM_OPENMP="1"  # [osx]
+    - export NOISEFLOW_USE_CPP="1"  # [linux]
```
tdegeus, line 19: I don't know about the package, so just to check: you don't want this on other platforms

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - export MACOSX_DEPLOYMENT_TARGET="10.13"  # [osx]
```
tdegeus, line 20: To check: this should be a measure of last resort, is that indeed the case ? See https://conda-forge.org/docs/maintainer/knowledge_base.html#requiring-newer-macos-sdks

recipes/noiseflow/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - {{ PYTHON }} -m pip install poetry-core@https://github.com/python-poetry/poetry-core/archive/refs/tags/1.6.1.zip
```
tdegeus, line 21: Is the hardcode version something that can be avoided?

recipes/noiseflow/meta.yaml
tdegeus: To have this dependencies on `unix` but enable CPP only in `linux` (above) seems a bit contradictory. Could you clarify ?

recipes/noiseflow/meta.yaml
tdegeus: Have you performed the tests on different hardware, with different OS (version) ? If not, passing tests are expected

recipes/noiseflow/meta.yaml
OUCyf: Yes, I don't want to use cpp_module in Win platform.

recipes/noiseflow/meta.yaml
OUCyf: Yes, I think it is necessary to set MACOSX_DEPLOYMENT_TARGET="10.13", otherwise I will get an error.

recipes/noiseflow/meta.yaml
OUCyf: Do you mean the `poetry-core v1.6.1`? 
I try to install the last version of poetry-core via `pip install poetry-core`, but it not works. I think I need to specify the version.

recipes/noiseflow/meta.yaml
OUCyf: Not yet, I need to remove the `xsimd` I think.

recipes/noiseflow/meta.yaml
OUCyf: The `fftw` is not used in win platform in my package. The c++ extension is only used in macos and linux.

OUCyf: I'm sorry, what do you mean `but enable CPP only in linux (above)`?

recipes/noiseflow/meta.yaml
tdegeus: And not on Mac either?

recipes/noiseflow/meta.yaml
OUCyf: It not works when I use `find_package(fftw)` directly, I need to specify the `cmake module path` which is from `xtensor-fftw` package. And that is why I should use git clone for `xtensor-fftw`
```cmake
set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_CURRENT_SOURCE_DIR}/extern/xtensor-fftw/cmake/Modules/findFFTW") # use FindFFTW.cmake to find fftw
find_package(FFTW REQUIRED COMPONENTS ${REQUIRE_FLOAT_LIB} ${REQUIRE_DOUBLE_LIB} ${REQUIRE_LONG_DOUBLE_LIB})
```

recipes/noiseflow/meta.yaml
OUCyf: Ohhhh, I see. It should be `- export NOISEFLOW_USE_CPP="1"  # [unix]`

OUCyf:
@conda-forge/help-python, ready to merge. 
Hey @tdegeus I have modified my package, do you have more suggestions?

PR #23354, opened 2023-07-14
Add gnss-sdr

ryanvolz:
Checklist
- [x] Build number is 0.


ryanvolz:
@conda-forge/help-c-cpp, ready for review. Thanks!

recipes/gnss-sdr/meta.yaml
```diff
@@ -0,0 +1,142 @@
...
+  url: https://github.com/gnss-sdr/gnss-sdr/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d5533370de188d50c214a5bb7832a433ceb09369c3e3296a4fa34fa1aefb685e
+  patches:
+    - 0001-Manually-specify-compiled-in-GNSSSDR_INSTALL_DIR.patch
+
+build:
+  number: 0
+  skip: true  # [win]
+
+outputs:
```
carterbox, line 17: SUGGESTION ONLY: It may be simpler to build once and [separate the output using glob name matching](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#specifying-files-to-include-in-output) instead of building multiple times.

recipes/gnss-sdr/meta.yaml
```diff
@@ -0,0 +1,142 @@
...
+        - gstreamer-orc  # [not win]
+      run:
+        - {{ pin_compatible('gstreamer-orc', max_pin='x.x') }}  # [not win]
+    test:
+      commands:
+        # verify that commands run
+        - volk_gnsssdr-config-info --version --prefix --cc --cflags --all-machines --avail-machines --machine --alignment --malloc
+        - volk_gnsssdr_profile --help
+
+        # verify libraries
```
carterbox, line 41: Please also verify existence of some headers / cmake files / config files.

recipes/gnss-sdr/meta.yaml
```diff
@@ -0,0 +1,142 @@
...
+  sha256: d5533370de188d50c214a5bb7832a433ceb09369c3e3296a4fa34fa1aefb685e
+  patches:
+    - 0001-Manually-specify-compiled-in-GNSSSDR_INSTALL_DIR.patch
+
+build:
+  number: 0
+  skip: true  # [win]
+
+outputs:
+  - name: volk-gnss-sdr
```
carterbox, line 18: Each output with libraries needs run_exports to ensure ABI compatability downstream.

carterbox (changes requested):

recipes/gnss-sdr/meta.yaml
ryanvolz: I did consider that because it's how I'd normally do outputs, but the way that their CMakeLists.txt is structured makes that more awkward. As it is, the build script for the volk-gnss-sdr output descends to the appropriate subdirectory and only builds that component, and then the subsequent build for gnss-sdr finds the already built volk-gnss-sdr library and just uses it, so there's no extra building happening.

recipes/gnss-sdr/meta.yaml
ryanvolz: Done!

recipes/gnss-sdr/meta.yaml
ryanvolz: Done!

ryanvolz:
Thanks for the review @carterbox!

@conda-forge/help-c-cpp, ready for re-review/merge. Thanks!

PR #23370, opened 2023-07-16
Add AutoROM package

ChristofKaufmann:
Checklist
  → partially, see my question below.
- [x] Build number is 0.

~@conda-forge/help-python: One question I could not figure out is how the license field should be handled for pure meta packages, that do not contain source code. So I like to add `AutoROM` to conda-forge. In addition I made a meta package `AutoROM-accept-rom-license` that has `AutoROM` as dependency and a post-link script to install the Atari ROMs, as if the user executed `AutoROM --accept-rom-license` afterwards. This mimics the behaviour of the corresponding [pip package](https://pypi.org/project/AutoROM/#description) with the extra: `autorom[accept-rom-license]`. I did not include any source code license information in the meta package `AutoROM-accept-rom-license`. In the real package `AutoROM` I added the information as usual.~

Sorry for the noise. I realized that the post-link script is also source code and needs a license. I copied the license from the source code to the recipe directory and added it in `meta.yaml`.

### DIFF
--- a/recipes/autorom-accept-rom-license/meta.yaml
+++ b/recipes/autorom-accept-rom-license/meta.yaml
@@ -1,0 +1,37 @@
+{% set name = "AutoROM-accept-rom-license" %}
+{% set version = "0.6.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+build:
+  noarch: python
+  number: 0
+
+requirements:
+  host:
+    - python >=3.6
+    - autorom
+  run:
+    - python >=3.6
+    - autorom
+
+test:
+  imports:
+    - AutoROM
+  commands:
+    - python -c "import ale_py.roms as roms; print(roms.__all__)"  # print ROMs the ALE can find
+    - python -c "from ale_py.roms import Pong; import sys; sys.exit(not Pong.exists())"  # test that it can be imported
+  requires:
+    - ale-py
+
+about:
+  home: https://github.com/Farama-Foundation/AutoROM
+  summary: Automated installation of Atari ROMs for Gymnasium/ALE-Py
+  license: MIT
+  license_file: LICENSE.txt
+
+extra:
+  recipe-maintainers:
+    - ChristofKaufmann
--- a/recipes/autorom-accept-rom-license/LICENSE.txt
+++ b/recipes/autorom-accept-rom-license/LICENSE.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Farama Foundation
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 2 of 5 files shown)


ChristofKaufmann:
@jjshoots: In the [readme](https://github.com/Farama-Foundation/AutoROM#packaging) its stated that AutoROM must be packaged as source code. I think, it is not violated, because the conda package is a zipped file and contains the `AutoROM.py`. However, I am not quite sure, if this is, what the readme is talking about. Maybe you could leave a comment here on that topic.

BTW: Do you want to be recipe maintainer?

ChristofKaufmann:
@pseudo-rnd-thoughts: Maybe you can clarify, what the statement in the readme is about? This could help to go forward here.

pseudo-rnd-thoughts:
If I understand correctly, AutoROM purely provides the ability to download the atari ROMs from a server but doesn't provide the "source code" to run the ROMs, i.e., ALE-py.

@jjshoots do you agree? 

FYI, we have been slow to make this change but in the future, ale-py will include the ROMs on install such that AutoROM will be archived. My hope is that this will be done by Christmas but I'm not sure

jjshoots:
@pseudo-rnd-thoughts yes, that's exactly what AutoROM does.

ChristofKaufmann:
Thank you for your responses.

> ale-py will include the ROMs on install such that AutoROM will be archived

Oh I didn't know... maybe the package could still be useful until then.

ChristofKaufmann:

ChristofKaufmann:

recipes/autorom-accept-rom-license/meta.yaml
```diff
@@ -0,0 +1,37 @@
+{% set name = "AutoROM-accept-rom-license" %}
```
carterbox, line 1: Why is this package needed? The README says that users should accept the license by calling

```bash
AutoROM --accept-license
```
It's not OK to have a package which silently accepts the license for the user. Downstream packages could depend on this package and accept the license for them automatically and silently.

carterbox (changes requested):

recipes/autorom-accept-rom-license/meta.yaml
ChristofKaufmann: I just tried to bring some dependencies of gymnasium to conda-forge. This is the equivalent of PyPI's [autorom[accept-rom-license]](https://pypi.org/project/AutoROM.accept-rom-license/). There it is stated, that you can accept the license either with `AutoROM --accept-license` or by installing `autorom[accept-rom-license]`. If it is used in a dependency, the dependency should also have the extra `[accept-rom-license]`, like with [gymnasium](https://github.com/Farama-Foundation/Gymnasium/blob/main/pyproject.toml#L38). There, `autorom[accept-rom-license]` is [explicitly](https://github.com/Farama-Foundation/Gymnasium/blob/main/pyproject.toml#L56) excluded from `[all]`, as this would indeed be not OK.

However, this will be obsolete quite soon, as the ROMs will be included in ALE-py, as @pseudo-rnd-thoughts stated. So, if you want, you can close it.

PR #23593, opened 2023-08-07
Add skprogs-python recipe

vanderhe:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/skprogs-python/meta.yaml
+++ b/recipes/skprogs-python/meta.yaml
@@ -1,0 +1,57 @@
+{% set name = "skprogs" %}
+{% set version = "0.2" %}
+
+package:
+  name: {{ name|lower }}-python
+  version: {{ version }}
+
+source:
+  url: https://github.com/dftbplus/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: a16cc91cfa702b2e40b61bbe218fd43d9225469c6e90b56ab04b89838916f9bf
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
+    - collectspinw = sktools.scripts.collectspinw:main
+    - collectwavecoeffs = sktools.scripts.collectwavecoeffs:main
+    - skdiff = sktools.scripts.skdiff:main
+    - skgen = sktools.scripts.skgen:main
+    - skmanip = sktools.scripts.skmanip:main
+  script: "{{ PYTHON }} -m pip install ./sktools -vv"
+
+requirements:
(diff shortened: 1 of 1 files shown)


vanderhe:
@aradi @bhourahine Could you post a comment here that you are willing to join as package maintainer?

bhourahine:
@vanderhe sure.

aradi:
@vanderhe I am in

vanderhe:
@aradi Shall we let this be a `noarch` package? It is kind of useless without `skprogs`, but what about cases where you let's say want to post-process a calculation (e.g. extract spin constants/wfc's) that has already been carried out on another machine or by someone else?

vanderhe:
@conda-forge/staged-recipes Ready for review.

recipes/skprogs-python/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "skprogs" %}
+{% set version = "0.2" %}
+
+package:
+  name: {{ name|lower }}-python
```
ocefpaf, line 5: We never formalized this and the change below is more os a suggestion but most of the python wrappers use a prefix instead of a suffix. 

```suggestion
  name: python-{{ name|lower }}
```

recipes/skprogs-python/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - sktools
+  commands:
+    - collectspinw -h
+    - collectwavecoeffs -h
+    - skdiff -h
+    - skgen -h
+    - skmanip -h
+
+about:
+  home: https://github.com/dftbplus/{{ name }}
```
ocefpaf, line 43: The jinja variable just obscures the URL. No need to use it here.

```suggestion
  home: https://github.com/dftbplus/skprogs
```

recipes/skprogs-python/meta.yaml
vanderhe: Thanks for the review, I guess we would like to keep this recipe consistent with e.g. the [dftbplus-python-feedstock](https://github.com/conda-forge/dftbplus-python-feedstock). If it is really just a loose convention, I would prefer to keep the current name.

recipes/skprogs-python/meta.yaml
vanderhe: Sure, no problem.

vanderhe:
@ocefpaf Feedback applied.

PR #23595, opened 2023-08-07
fix linting message

ocefpaf:
The message is misleading and may cause confusing like in https://github.com/conda-forge/staged-recipes/pull/23585.


BastianZim (approved):
Thank you! :)

PR #23610, opened 2023-08-08
adding htpolynet

cameronabrams:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/htpolynet/meta.yaml
+++ b/recipes/htpolynet/meta.yaml
@@ -15,6 +15,7 @@ build:
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
   skip: True  # [win or osx]
+  skip: true  # [py<39]
 
 requirements:
   host:
@@ -23,9 +24,9 @@ requirements:
     - pip
   run:
     - python
-    - numpy >=1.24
-    - pyyaml >=6
-    - requests >=2.28
+    - numpy
+    - pyyaml
+    - requests
     - pandas >=2
     - scipy >=1.10
     - parmed >=4
--- a/recipes/htpolynet/meta.yaml
+++ b/recipes/htpolynet/meta.yaml
@@ -1,5 +1,5 @@
-{% set name = "HTPolyNet" %}
-{% set version = "1.0.7.2" %}
+{% set name = "htpolynet" %}
+{% set version = "1.0.7.3" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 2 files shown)


cameronabrams:
This is ready for review @conda-forge-admin, please ping team

recipes/htpolynet/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/htpolynet-{{ version }}.tar.gz
+  sha256: fad2cd0699cf3a18537bf77962625ae1fbc669a266d698c4c51b49ce7d72bb59
+
+build:
+  entry_points:
+    - htpolynet = HTPolyNet.driver:cli
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: True  # [win or osx]
```
ocefpaf, line 17: I think this one can be `noarch`. What are the reasons to skip Windows and macOS?

cameronabrams: It will not run on Windows because it uses ambertools.  In theory it should work on OSX but I have not tried this myself.

ocefpaf: It is better to do noarch, not install on Windows due to a missing dependency, than making it "arch'ed" and build for several Python versions. Also, if ambertools is ever available on Windows, this package would be ready without the need for any change here.

ocefpaf: Here. on staged-recipes, you don't need to avoid it. Just let it error out. B/c it will be a noarch package only the Linux build will count and that is the only one that will appear in the feedstock when we merge this.

cameronabrams: The recipe I just pushed fails in a local linux build for a reason I can't figure out, which I quote in the commit message.

recipes/htpolynet/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - requests >=2.28
+    - pandas >=2
+    - scipy >=1.10
+    - parmed >=4
+    - networkx >=2.8
+    - gputil >=1.4
+
+test:
+  imports:
+    - htpolynet
```
ocefpaf, line 37: Looks like the import name doesn't follow Python naming standards and it capitalized.

```suggestion
    - HTPolyNet
```

cameronabrams:
Unfortunately, that suggested change didn't work.  Greyskull's recipe worked just fine on an early version of this PyPI package, and the current package on PyPI also works via pip.  I'm not sure what I did to break the conda recipe.

ocefpaf:
> Unfortunately, that suggested change didn't work. Greyskull's recipe worked just fine on an early version of this PyPI package, and the current package on PyPI also works via pip. I'm not sure what I did to break the conda recipe.

It did. You have a new error:

```
ModuleNotFoundError: No module named 'matplotlib'
```

You are missing `matplotlib-base` in your recipe. Maybe upstream made it optional but it seems to be required for the CLI interface.

cameronabrams:
it is still giving the following message when I do a local build:
 
The reported errors are:
-   - package htpolynet-1.0.7.3-pyhd8ed1ab_0 is excluded by strict repo priority

cameronabrams:
This is ready for review @conda-forge-admin, please ping team

PR #23615, opened 2023-08-08
Add holodeck-gw

davecwright3:
Checklist
- [x] Build number is 0.


davecwright3:

recipes/holodeck-gw/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - astropy
+    - cosmopy
+    - h5py
+    - matplotlib-base
+    - scipy
+    - pydoe
+    - psutil
+    - ipywidgets
+    - tqdm
+    - cython <3.0.0
```
ocefpaf, line 42: It is unlike that you need cython as a run requirement. Can you check this upstream?

recipes/holodeck-gw/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - schwimmbad
+    - emcee
+    - george
+    - {{ pin_compatible('numpy') }}
+    - {{ pin_compatible('numba') }}
+    - {{ pin_compatible('mpi4py') }}
```
ocefpaf, line 48: I'm not sure about those. @isuruf if you have some time can you check this here?

ocefpaf:
xref.: https://github.com/conda-forge/staged-recipes/pull/23446

recipes/holodeck-gw/meta.yaml
davecwright3: The upstream code requires `cython` as a runtime dependency https://github.com/nanograv/holodeck/blob/main/requirements.txt

recipes/holodeck-gw/meta.yaml
davecwright3: There were issues at build-time, and pinning the versions of the above packages fixed the issues.

recipes/holodeck-gw/meta.yaml
ocefpaf: It lists but is it used? Does this package generate code?

ocefpaf: I could not find a place where cython was used beyond the building step. Maybe I'm missing something.

recipes/holodeck-gw/meta.yaml
ocefpaf: It would be better to understand what is going  on before pinning blindly. The numpy one is probably needed. But maybe the other two can be dropped of you adjust the min numpy version as I showed in the first review.

recipes/holodeck-gw/meta.yaml
davecwright3: The upstream code hasn't finished migrating docs and such to `pyproject.toml`. The current `setup.py` based installation uses `requirements.txt` for both build and runtime requirements. The PyPI package therefore requires `cython` as a runtime dependency. If `cython` isn't listed as a runtime dependency, then the build will fail (I've confirmed with local builds).

In the future, the code will move to pure `pyproject.toml` and `cython` can be removed from runtime dependencies as you've pointed out.

recipes/holodeck-gw/meta.yaml
ocefpaf: You mean the `pip check` test will fail, not the build, right? If so it is preferred to remove cython and comment out the `pip check` test to avoid polluting user env with build tools.

PS: one can have cython as a build tool only without a pyproct.toml.

recipes/holodeck-gw/meta.yaml
ocefpaf: You mean the `pip check` test will fail, not the build, right? If so it is preferred to remove cython and comment out the `pip check` test to avoid polluting user env with build tools.

PS: one can have cython as a build tool only without a pyproct.toml.

davecwright3: I've removed the cython run time dependency and the pip check as requested

recipes/holodeck-gw/meta.yaml
ocefpaf: Can you try to remove the numba and mpi4py here so I can take a look at the error?

davecwright3: I've removed the pins and the checks have passed. I believe the previous issues were with the Windows builds, which I've since disabled because the upstream code doesn't currently support Windows.

ocefpaf: BTW, next time don't close and re-open the PR. It puts an extra burden on the reviewers b/c most of our actions here were already review comments in the previous one that ended up duplicating the effort for everybody.

PR #23625, opened 2023-08-09
Add pykerr 0.1.0

cdcapano:
@conda-forge-admin, please ping @conda-forge/help-python

This adds `pykerr` which is a package used for computing quasi-normal modes and spheroidal harmonics of perturbed black holes. This is used in the gravitational-wave data analysis community.

### DIFF
--- a/recipes/pykerr/meta.yaml
+++ b/recipes/pykerr/meta.yaml
@@ -33,9 +33,9 @@ test:
     - pip
 
 about:
-  home: https://github.com/cdcapano/pykerr.git
+  home: https://github.com/cdcapano/pykerr
   summary: QNM frequencies and spheroidal harmonics of Kerr black holes.
-  license: GPL-3.0
+  license: GPL-3.0-or-later
   license_file: LICENSE
 
 extra:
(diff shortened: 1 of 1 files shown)


recipes/pykerr/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+test:
+  imports:
+    - pykerr
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/cdcapano/pykerr.git
```
ocefpaf, line 36:
```suggestion
  home: https://github.com/cdcapano/pykerr
```

recipes/pykerr/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  summary: QNM frequencies and spheroidal harmonics of Kerr black holes.
+  license: GPL-3.0
```
ocefpaf, line 38: Looks like it is `GPL-3.0-or-later`.

```suggestion
  license: GPL-3.0-or-later
```

cdcapano: Thanks for reviewing so quickly. I've made the suggested changes.

PR #23662, opened 2023-08-12
Add pipestat recipe

mencian:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pipestat/meta.yaml
+++ b/recipes/pipestat/meta.yaml
@@ -22,11 +22,10 @@ requirements:
     - pip
   run:
     - python >=3.8
-    - setuptools
     - jsonschema
     - logmuse >=0.2.5
     - oyaml
-    - psycopg2-binary
+    - psycopg2
     - pydantic >=1.10.7,<2.0.0
     - sqlmodel >=0.0.8
     - ubiquerg >=0.6.1
--- a/recipes/pipestat/meta.yaml
+++ b/recipes/pipestat/meta.yaml
@@ -18,10 +18,10 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - pip
   run:
-    - python
+    - python >=3.8
     - jsonschema
     - logmuse >=0.2.5
     - oyaml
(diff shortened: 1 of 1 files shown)


mencian:

recipes/pipestat/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  entry_points:
+    - pipestat = pipestat.__main__:main
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - setuptools
```
beenje, line 25: `setuptools` isn't a run dependency

recipes/pipestat/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - jsonschema
+    - logmuse >=0.2.5
+    - oyaml
+    - psycopg2-binary
```
beenje, line 29: `psycopg2-binary` is a convenient wheel when using pip. With conda, I'd prefer to specify `psycopg2` directly.

Will work with `psycopg2-binary` as well. It's just an alternative name (will install `psycopg2`).

beenje (changes requested):

recipes/pipestat/meta.yaml
mencian: Unfortunately, the build fails and runs into a `ModuleNotFoundError: No module named 'pkg_resources'` error when I remove setuptools from the list of run dependencies.

recipes/pipestat/meta.yaml
mencian: Thank you for catching this, I have corrected psycopg2-binary to psycopg2 in my next commit.

recipes/pipestat/meta.yaml
beenje: I see. You are right. `setuptools` isn't used directly by `pipestat`.
`eido` is using [it](https://github.com/pepkit/eido/blob/2525940bb3b424f8fae7f3b5c8f68ebc5da00f60/eido/conversion.py#L6C6-L6C19). It should be defined in that package instead of here. It's fine for now. I opened an issue: https://github.com/conda-forge/eido-feedstock/issues/16

PR #23699, opened 2023-08-17
Add cryolobm

thorstenwagner:
@conda-forge/help-python

Hi, I build it with grayskull, I think it is ready for review

### DIFF
--- a/recipes/cryolobm/meta.yaml
+++ b/recipes/cryolobm/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "cryoloBM" %}
+{% set version = "1.5.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/cryoloBM-{{ version }}.tar.gz
+  sha256: e7552fad5f1e7f8f70af53592e9d09e6dd0ebdec4731960562c57bf8f6f516e4
+
+build:
+  entry_points:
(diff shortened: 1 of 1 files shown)


recipes/cryolobm/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+requirements:
+  host:
+    - python >=3.6
+    - pip
+  run:
+    - python >=3.6
+    - matplotlib-base >=3.4.2,<3.5
+    - numpy >=1.16.0,<1.19.0
+    - pystardb >=0.3.0
+    - pandas ==1.1.4
```
bgruening, line 29: Is there a reason for the hard pin here? It would be better to relax the pin.

thorstenwagner:
I am willing to be a maintainer of the package

recipes/cryolobm/meta.yaml
thorstenwagner: Relaxed it. I think its not necessary anymore.

recipes/cryolobm/meta.yaml
```diff
@@ -26,7 +26,7 @@ requirements:
     - matplotlib-base >=3.4.2,<3.5
     - numpy >=1.16.0,<1.19.0
     - pystardb >=0.3.0
-    - pandas ==1.1.4
+    - pandas <=1.4.4
```
ocefpaf, line 29: Should probably raise this upstream b/c it is in the package metadata.

thorstenwagner: ok, I will do that.

PR #23707, opened 2023-08-17
Adding recipe for my new package BubbleDet

og113:
@conda-forge/help-python, ready for review!

PR #23768, opened 2023-08-22
Added recipe for findiff, made by grayskull.

og113:
Adding recipe for Python package findiff, which is a requirement of my package BubbleDet, in PR #23707.

### DIFF
--- a/recipes/findiff/meta.yaml
+++ b/recipes/findiff/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "findiff" %}
+{% set version = "0.10.0" %}
+
+package:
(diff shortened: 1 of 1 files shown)


og113:
So, it turns out that there is already a version of findiff on conda-forge, with merged PR #14582. However, the version there (0.8.9) is out of date, the current version being 0.10.0.

og113:
Closed this PR because findiff feedstock already exists.

maroba:
Thanks for taking care of this! When I’m back from vacation next week, I will see how I can keep the pypi and conda versions in sync for future releases. So far, I only cared for pypi…

PR #23785, opened 2023-08-24
Adding filepattern

JesseMckinzie:
Checklist
- [x] Build number is 0.


JesseMckinzie:
@conda-forge/help-python
@conda-forge/help-c-cpp
Ready for review!

sameeul:
@@conda-forge/staged-recipes : This is ready for review.

recipes/filepattern/build-lib.sh
```diff
@@ -0,0 +1,14 @@
...
+mkdir build-lib
+cd build-lib
+
+if [[ "$OSTYPE" == "darwin"* ]]; then
+	export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=${PREFIX} -DCMAKE_INSTALL_PREFIX=${PREFIX} -Dfilepattern_SHARED_LIB=ON -DRUN_GTEST=OFF  ..
+
+cmake --build . --target install --parallel
```
ocefpaf, line 14: Is there a `ctest`, `make check` or similar that we can run here?

sameeul: We are doing some install check that are described in `meta.yaml`. Will that be sufficient?
Naive question: What will be the purpose of `ctest` or `make check`?

ocefpaf: If they exists, test the library. Do they? If not there isn't anything we can do here. I asked b/c most of the time compiled libraries do have them.

sameeul: We can change the `DRUN_GTEST=OFF` to `DRUN_GTEST=ON` and then run the tests.

sameeul: also, we have this following section in `meta.yaml` to check whether the files are installed correctly.
```
    test:
      commands:
        - test -f $PREFIX/include/filepattern.h  # [unix]
        - if not exist %LIBRARY_INC%\\filepattern.h exit 1  # [win]
        - test -f $PREFIX/lib/libfilepattern.so.2  # [linux]
        - test -f $PREFIX/lib/libfilepattern.2.dylib  # [mac]
        - if not exist %LIBRARY_BIN%\\filepattern.dll exit 1  # [win]
        - test -f $PREFIX/lib/cmake/filepattern/filepatternConfig.cmake  # [unix]
        - if not exist %LIBRARY_LIB%\\cmake\\filepattern\\filepatternConfig.cmake exit 1  # [win]
```

carterbox: Existence checks are typically sufficient. We don't need to run the entire CI suite for every package. We do need to check that all the artifacts are installed in the correct place, that link paths are set correctly, and that the ABI won't break.

sameeul: These checks should be sufficient then since they make sure all the artifacts are installed in correct location.

sameeul:
Looks like the checksum has changed. We will update that.

recipes/filepattern/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - pydantic
+
+    test:
+      imports:
+        - filepattern
+      commands:
+        - pip check
+      requires:
+        - pip
+        - pytest
```
carterbox, line 52: These tests are not being run, and the python packages are empty. Maybe the syntax is incorrect for this outputs section? It might be because this output shares the same name as the top level package.

recipes/filepattern/meta.yaml
```diff
@@ -0,0 +1,97 @@
+{% set name = "filepattern" %}
+{% set version = "2.0.1" %}
+
+package:
+  name: {{ name|lower }}
```
carterbox, line 5:
```suggestion
  name: filepattern-split
```

recipes/filepattern/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+    or more of the aforementioned variables.
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+  doc_url: https://filepattern.readthedocs.io/en/latest/
+  dev_url: https://github.com/PolusAI/filepattern
+
+extra:
+  recipe-maintainers:
+    - JesseMckinzie
```
carterbox, line 97:
```suggestion
extra:
  feedstock-name: filepattern
  recipe-maintainers:
    - JesseMckinzie
```

recipes/filepattern/meta.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - cmake
+        - make  # [not win]
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - git
+      host:
+        - python
+        - setuptools
+        - pip
+        - pybind11
```
carterbox, line 40:
```suggestion
      host:
        - python
        - setuptools
        - pip
        - pybind11
        - libfilepattern {{ version }}
```
Shouldn't this package depend on libfilepattern?

carterbox (changes requested):

recipes/filepattern/meta.yaml
sameeul: we are actually creating two separate packages. One python package (`filepattern`) and another C++ package `libfilepattern`. Should we create two separate recipes or use one to create both packages?

sameeul: @carterbox : Trying to get some clarification here. Our intention is to create two separate things. One `filepattern` Python package and a `libfilepattern` C++ library. Should we split this recipe into two to do that? Both the Python package and C++ library uses the same source code but different CMake options to produce them. There is no dependency between them.

carterbox: The python package should link dynamically to the c++ shared library and the two packages should live in separate feedstocks/recipes in order to reduce the number of redundant builds.

sameeul: We have created a new recipe (https://github.com/conda-forge/staged-recipes/pull/24433) for libfilepattern. We also updated the build process for the Python package so that if `libfilepattern` is present, it will use that instead of compiling the core package files. Once, the `libfilepattern` recipe is merged, we will rework this recipe.

recipes/filepattern/meta.yaml
```diff
@@ -20,60 +20,24 @@ requirements:
...
-    test:
-      commands:
-        - test -f $PREFIX/include/filepattern.h  # [unix]
-        - if not exist %LIBRARY_INC%\\filepattern.h exit 1  # [win]
-        - test -f $PREFIX/lib/libfilepattern.so.2  # [linux]
-        - test -f $PREFIX/lib/libfilepattern.2.dylib  # [mac]
-        - if not exist %LIBRARY_BIN%\\filepattern.dll exit 1  # [win]
-        - test -f $PREFIX/lib/cmake/filepattern/filepatternConfig.cmake  # [unix]
-        - if not exist %LIBRARY_LIB%\\cmake\\filepattern\\filepatternConfig.cmake exit 1  # [win]
+    - libfilepattern
```
sameeul, line 23: this should move to host section.

recipes/filepattern/meta.yaml
sameeul: @carterbox : we updated the recipe to depend on the C++ shared library (as suggested). Can you please take a look at see if it is good to be merged? Thanks :-)

sameeul:
@conda-forge/help-python
@conda-forge/help-c-cpp 

We updated the recipe based on the previous review comments and think that it is good to review :-)

PR #23904, opened 2023-09-05
Add Unity Test for conda-forge

argltuc:
This PR provide a recipe to build C unit test library `Unity Test`.
This library is a static only lib. But since it is a library for testing binaries, it is not inteded to be a run dependency for other packages.

@conda-forge/help-c-cpp, read for review

### DIFF
--- a/recipes/unity-test/bld.bat
+++ b/recipes/unity-test/bld.bat
@@ -11,7 +11,7 @@ cmake ../ -G "Ninja" ^
 
 if errorlevel 1 exit 1
 
-cmake --build . --target INSTALL --config Release --parallel %NUMBER_OF_PRECESSOR%
+cmake --build . --target install --config Release --parallel %NUMBER_OF_PRECESSOR%
 if errorlevel 1 exit 1
 
 popd
--- a/recipes/unity-test/bld.bat
+++ b/recipes/unity-test/bld.bat
@@ -1,0 +1,17 @@
+@echo on
+setlocal EnableDelayedExpansion
+
+mkdir "%SRC_DIR%"\build
+pushd "%SRC_DIR%"\build
+
+cmake ../ -G "Ninja" ^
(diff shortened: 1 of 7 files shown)


recipes/unity-test/bld.bat
```diff
@@ -0,0 +1,17 @@
...
+pushd "%SRC_DIR%"\build
+
+cmake ../ -G "Ninja" ^
+    -DCMAKE_INSTALL_PREFIX:PATH="%LIBRARY_PREFIX%" ^
+    -DCMAKE_PREFIX_PATH:PATH="%LIBRARY_PREFIX%" ^
+    -DCMAKE_BUILD_TYPE:STRING=Release
+
+if errorlevel 1 exit 1
+
+cmake --build . --target install --config Release --parallel %NUMBER_OF_PRECESSOR%
```
xhochy, line 14:
```suggestion
cmake --build . --target install --config Release --parallel %CPU_COUNT%
```

recipes/unity-test/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+set -e
+mkdir -p $PREFIX/include
+mkdir build
+cd build
+
+cmake "${SRC_DIR}" -G "Ninja" \
+    -DCMAKE_INSTALL_PREFIX:PATH="$PREFIX" \
```
xhochy, line 9:
```suggestion
    ${CMAKE_ARGS} \
```

recipes/unity-test/build.sh
```diff
@@ -0,0 +1,13 @@
...
+    -DCMAKE_PREFIX_PATH:PATH="$PREFIX" \
```
xhochy, line 10:
```suggestion
```

xhochy (changes requested):
Does this recipe generate any shared libraries? If so this should checked for their existence in the tests.

PR #24011, opened 2023-09-18
Adding REEF3D and DIVEMesh

mattkram:
Checklist
- [x] Build number is 0.


mattkram:
@conda-forge/help-c-cpp just a friendly ping to see if there was any chance someone could please check out this recipe 🙂? Thanks 🙏.

recipes/divemesh/meta.yaml
```diff
@@ -0,0 +1,41 @@
+{% set name = "divemesh" %}
+{% set version = "23.08" %}
+{% set sha256 = "d67c287793da9972e6d4a4dd0cc281816382ea325a308ef434522e8d71993ed4" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
```
xhochy, line 10: Not needed anymore 
```suggestion
```

recipes/divemesh/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  url: https://github.com/REEF3D/DIVEMesh/archive/refs/tags/{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+  patches:
+    - patches/0001-remove-cxx-in-makefile.patch
+
+build:
+  number: 0
+  skip: true  # [win]
+  ignore_run_exports_from:
+    - {{ compiler('cxx') }}  # [osx]
```
xhochy, line 20: Why? This is compiled code amd thus should need this.

recipes/divemesh/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+make -j "${CPU_COUNT}"
```
xhochy, line 3: You probably don't need a patch but could use this
```suggestion
make -j "${CPU_COUNT}" CXX=$CXX LDFLAGS=$LDFLAGS CXXFLAGS=$CXXFLAGS
```

recipes/reef3d/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+{% set sha256 = "9e114697b032615b16a4ce1459b91474944caf3b3ac182a0ddd5d733681c4549" %}
+
+{% set mpi = mpi or 'openmpi' %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
```
xhochy, line 12:
```suggestion
```

recipes/reef3d/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+make -j "${CPU_COUNT}"
```
xhochy, line 3: Sew above

xhochy (changes requested):

recipes/divemesh/meta.yaml
mattkram: Ahh, I remember now. @chenghlee recommended I try building with `--error-overlinking` and `--error-overdepending`, which raised this:
```    
conda_build.exceptions.OverDependingError: overdepending check failed 
    ['  ERROR (divemesh): dso library package conda-forge::libcxx-16.0.6-h4653b0c_0 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)']
```

recipes/divemesh/meta.yaml
mattkram: Removed

recipes/divemesh/build.sh
mattkram: I found errors when including `LDFLAGS` and `CXXFLAGS`, so I've kept the upstream versions of these. But I've dropped the patch in favor of `CXX=$CXX`, which is a much cleaner solution.

recipes/reef3d/meta.yaml
mattkram: Removed.

recipes/reef3d/build.sh
mattkram: In this case, we need `mpicxx`, which is called out in the upstream Makefile. But I've put `CXX=mpicxx` to be more explicit. I can't find whether there is a variable for this available.

recipes/divemesh/meta.yaml
carterbox: Ignoring run_exports from a compiler is fine if you're doing it to avoid depending on unused standard libraries.

recipes/reef3d/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  run:
+    - hypre
+    - {{ mpi }}
+    - divemesh
+    - libcxx
+
+test:
+  commands:
+    - test -f $PREFIX/bin/reef3d
+    - reef3d  # [osx]
```
carterbox, line 36:
```suggestion
# Only test on osx because of missing mpi ORTE
    - reef3d  # [osx]
```
Please add explanatory notes for strange selectors, so reviewers don't have to ask / review commits for the whys. The git history of this recipe will be not be available in the feedstock.

carterbox (changes requested):

recipes/reef3d/meta.yaml
mattkram: Thank you @carterbox. FWIW, I agree this is a useful comment, but I was trying to follow the FAQ to the letter, which says to remove all comments: https://github.com/conda-forge/staged-recipes/blob/main/README.md#faq

Potentially an opportunity for updating the FAQ?

Thanks again for helping me get this merged!

carterbox: Yes! The intention was to have you remove all the comments from the example template.

PR #24045, opened 2023-09-20
Add tophu package

gmgunter:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/tophu/meta.yaml
+++ b/recipes/tophu/meta.yaml
@@ -11,11 +11,11 @@ source:
 build:
   script: {{ PYTHON }} -m pip install --no-deps . -vv
   number: 0
-  skip: true  # [win or osx]
+  skip: true  # [win or osx or py<38]
 
 requirements:
   host:
-    - python >=3.8
+    - python
     - pip
     - setuptools >=42
   run:
@@ -23,7 +23,7 @@ requirements:
     - h5py >=3
     - isce3 >=0.12
     - numpy >=1.21
-    - python >=3.8
+    - python
     - rasterio >=1.3
     - scipy >=1.5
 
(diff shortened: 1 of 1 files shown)


gmgunter:

gmgunter:
@conda-forge-admin, please ping conda-forge/help-python, ready for review!

recipes/tophu/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/isce-framework/tophu/archive/v{{ version }}.tar.gz
+  sha256: 691b0cae0bde670a4b430bcb1a6d2ec030e99f34c45e655e50c82c102bcda0d1
+
+build:
+  script: {{ PYTHON }} -m pip install --no-deps . -vv
+  number: 0
+  skip: true  # [win or osx or py<38]
```
ocefpaf, line 14: Why not noarch? If it is b/c of missing dependencies we still prefer to make it noarch to avoid CI waste and let the solver deal with the missing dependencies.

gmgunter: Unfortunately, the package is broken on macOS right now due to some issues with an upstream dependency that I haven't sorted out yet. It and its dependencies will install on macOS, but then will encounter errors at runtime. So the package could become noarch in the future if/when this issue is resolved, but for now I'd prefer to make it available on Linux only.

ocefpaf: You can still make it "some-arch" for Linux only. See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-packages-with-os-specific-dependencies

gmgunter: TIL, thanks.

The examples in the docs you linked seem to indicate that that I should use the appropriate preprocessor selector along with the `__linux` virtual package, but the linter doesn't seem to like that change. Should I remove the selector?

ocefpaf: > the linter doesn't seem to like that change. Should I remove the selector?

That should be fixed by now but no, don't remove it. Let's ignore the linter for now.

PR #24058, opened 2023-09-22
Add pytest-pyvista

tkoyama010:
Checklist
- [x] Build number is 0.


tkoyama010:
> GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there.

@MatthewFlamm @bjlittle Could you post if you are willing to be listed here?
https://github.com/conda-forge/staged-recipes/blob/758212d1171be6c301ff2b55f0ed5d71b16fcc8f/recipes/pytest-pyvista/meta.yaml#L50-L53

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+  run:
```
MatthewFlamm, line 21: Does this list come from pytest-pyvista[tests]?  Only pytest (and pyvista) is necessary to run this package.  I believe pyvista was removed due to problems with vtk installation (vtk vs. vtk-osmesa and so on).  I'm not familiar enough with conda to comment on whether it should go in here.

MatthewFlamm: `pytest-pyvista[tests]` is what is used for testing `pytest_pyvista` itself.

MatthewFlamm:
I am willing.  I don't know much about conda, but willing to learn. Thanks!

recipes/pytest-pyvista/meta.yaml
tkoyama010: Thank you. I will only use `pytest` and `pyvista` as dependencies, I looked at `geovista`'s feedstock and it looks fine to have `pyvista` as a dependency.

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - codecov
+    - coverage
+    - numpy
+    - python >=3.8
+    - pytest >=3.5.0
+    - pytest-cov
```
tkoyama010, line 27:
```suggestion
    - python >=3.8
    - pytest >=3.5.0
    - pyvista
```

recipes/pytest-pyvista/meta.yaml
tkoyama010: See 92cd072d9da658761fde4790fdbc7dbd4aa9e4c6 .

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+    - pyvista
+
```
tkoyama010, line 25: See https://github.com/pyvista/pytest-pyvista/blob/597da3865bb9f548fb66242ef37ed72cfa4a4af7/pyproject.toml#L2

```suggestion
    - flit_core >=3.2,<4

```

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+requirements:
+  host:
+    - pip
+    - python >=3.8
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+    - pyvista
+    - flit_core >=3.2,<4
```
MatthewFlamm, line 25: I think this needs to be flit
```suggestion
    - flit >=3.2,<4
```

MatthewFlamm: It also might have to go in host?  `setuptools` is mentioned as being put in host in the documentation.  https://conda-forge.org/docs/maintainer/adding_pkgs.html#build

bjlittle:
@tkoyama010 Yes, I'm more than happy to help out and be a maintainer of this package 👍

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
```
bjlittle, line 20: @tkoyama010 As you're using `flit` to build and package `pytest-pyvista`, I'm guessing that you're going to have to add it here as a dependency.

This should be enough as you're using `flit` instead of `setuptools` for the build system and backend. 

BTW I've never used `flit` myself and I've never seen a conda-forge recipe using `flit` before, but I'll see if I can find an example somewhere (we only ever use `setuptools`).

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/p/pytest_pyvista/pytest_pyvista-{{ version }}.tar.gz
+  sha256: 43e6f0a85b6ff23aff51b7f750528de0fdda6da2011a94b636f05d8498fe38a0
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
bjlittle, line 14: @tkoyama010 Aren't you instead using `flit` to build and install the package?

bjlittle: @ocefpaf Go some spare cycles for some sage advice?

I've never seen `flit` being used on conda-forge before to build a package, is there support to do this, and if so what pattern do we adopt here? Thanks, as always 😄

recipes/pytest-pyvista/meta.yaml
ocefpaf, line 25:
```suggestion
    - flit-core >=3.2,<4
```

recipes/pytest-pyvista/meta.yaml
bjlittle: No problem, found a tonne of examples, see [here](https://github.com/search?q=org%3Aconda-forge+%2Fflit-core%2F+path%3A%2F%5Erecipe%5C%2F%2F&type=code)

bjlittle: Yup, so the `script` call is fine with `pip`, so don't change that 👍

recipes/pytest-pyvista/meta.yaml
bjlittle: @tkoyama010 Okay, so you do need to include `flit_core >=3.2,<4` here in the `host` section, along with `pip` and `python` 👍

recipes/pytest-pyvista/meta.yaml
bjlittle: @tkoyama010 This is no need to have `flit` or `flit_core` in the `run` section 👍

recipes/pytest-pyvista/meta.yaml
bjlittle: Given the other `flit` package examples I found ([here](https://github.com/search?q=org%3Aconda-forge+%2Fflit-core%2F+path%3A%2F%5Erecipe%5C%2F%2F&type=code)) there is no need to include `setuptools` in the `host` section.

recipes/pytest-pyvista/meta.yaml
bjlittle: Awesome, thanks @ocefpaf. Legend as always 🍻

tkoyama010:
```suggestion
```

recipes/pytest-pyvista/meta.yaml
tkoyama010:
```suggestion
    - python >=3.8
    - flit-core >=3.2,<4
```

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+    - flit-core >=3.2,<4
```
tkoyama010, line 21:
```suggestion
    - flit-core >=3.2,<4
```

recipes/pytest-pyvista/meta.yaml
tkoyama010: Thanks for your thought!

recipes/pytest-pyvista/meta.yaml
tkoyama010: Thank for your comment!

tkoyama010:
Hmm.. So we have to solve the problem of win.

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+requirements:
+  host:
+    - pip
+    - python >=3.8
+    - flit >=3.2,<4
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+
+test:
```
tkoyama010, line 26: Skipped import test to avoid windows vtk import errors.
Please comment if you have any suggestions for improvement.

bjlittle: You can perform the import test for all other platforms other than windows with `import pytest_pyvista # not win`

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+    - flit >=3.2,<4
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+
```
MatthewFlamm, line 25: Do we need to have PyVista in the `run` section? It will be needed for import, and I'm not sure if this is a problem conda-forge recipe.  It isn't in the pip install dependencies because of the weirdness of vtk installation having multiple possible package names.

bjlittle: If `pytest-pyvista` imports pyvista, and it seems that it does https://github.com/pyvista/pytest-pyvista/blob/main/pytest_pyvista/pytest_pyvista.py#L8, then you need `pyvista` in the `run` section 👍

bjlittle: There shouldn't be a problem if `pyvista` is missing in the pip dependencies, as when`pip` is used to install the built `pytest-pyvista` package here it uses the `--no-deps` option, delegating the responsibility to `conda-build` to ensure that all the necessary runtime dependencies are pre-installed.

Conda will figure out the appropriate build of `vtk` to install given the install environment e.g., qt or egl et al

I don't think this is a problem ... are we willing to try and see?

bjlittle: Also, I pretty much exclusively use conda to install `pyvista` and next to never with `pip`.

I'm guessing you are the other way around?

MatthewFlamm: I don't know about conda, but I will lay out the issue with pip and vtk.  pyvista imports vtk.  You can install `vtk` or `vtkosmesa`.  If you install `vtkosmesa` first, it might be a surprise when you install `pytest-pyvista` that it will install `pyvista` which will install `vtk`.  We had decided to remove `pyvista` from dependencies as a result.  Ideally python would instead provide default optional_dependencies, which would prevent this problem.  We could have `pip install pyvista[openmesa]` to get `vtk-osmesa` or just `pip install pyvista` to get regular vtk.  Similar for this package.  How does conda handle this?

recipes/pytest-pyvista/meta.yaml
tkoyama010: Nice expression. Thanks!
```suggestion
test:
  import:
    pytest_pyvista # not win
```

recipes/pytest-pyvista/meta.yaml
bjlittle: @MatthewFlamm Thanks for explaining, this is really helpful insight, as I'm pretty new to `pyvista` and `vtk`.

Okay, so typically I leave the `conda` resolver to determine the `vtk` build bindings, but you can also be explicit as per your example with `pip` e.g., if you know that you want the `osmesa` build of `vtk` then you can do `conda install -c conda-forge pyvista "vtk=*=osmesa_*"`

The `vtk=*=osmesa_*` spec is of the form `<package>=<version>=<build>`, so here we're requesting for the latest version of `vtk` with the latest `osmesa` build.

Again, if you wanted the QT build you'd similarly do `vtk=*=qt_*`.

Does this help?

MatthewFlamm: I see now, whoever packaged vtk for conda-forge thankfully did not follow the model vtk uses for pip.  So it seems like the problem doesn't exist here.  Should be no problem to include then.

bjlittle: Yeah, I *think* so ...

Personally, I say we give it a whirl and see how it behaves.

What do you guys think?

ocefpaf: > I see now, whoever packaged vtk for conda-forge thankfully did not follow the model vtk uses for pip.

BTW, folks usually compare conda with pip but it is important to remind that conda is a language agnostic package manager. Think apt, yum, choco, brew, etc, not pip.

tkoyama010: Thank you for your comments. First, let's add PyVista and see what happens.
```suggestion
    - pyvista

```

bjlittle (approved):
@tkoyama010 LGTM

Thanks for making this happen 💯

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - python >=3.8
+    - flit >=3.2,<4
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+    - pyvista
+
+test:
+  imports:
+    pytest_pyvista # not win
```
bjlittle, line 29:
```suggestion
    pytest_pyvista  # [not win]
```

bjlittle, line 29:
```suggestion
    pytest_pyvista  # [not win]
```

bjlittle: Oops! My mistake in recalling the syntax to use here, see [conda-build preprocessing selectors](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#preprocessing-selectors).

tkoyama010: Thanks!

bjlittle:
@tkoyama010 @MatthewFlamm 

The `test` section is still failing on Windows with `pip check`

image

Given the explanation by @MatthewFlamm (above) regarding the behaviour of `pip` and `vtk`, would we be comfortable also skipping the `pip check` for Windows?

i.e., `pip check  # [not win]`

Or is there a better alternative here?

recipes/pytest-pyvista/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  run:
+    - python >=3.8
+    - pytest >=3.5.0
+    - pyvista
+
+test:
+  imports:
+    - pytest_pyvista  # [not win]
+  commands:
+    - pip check
```
tkoyama010, line 31: Personally, I'm +1 to adding `# [not win]`. `pytest-pyvista` is primarily envisioned for a Linux environment, so I don't think it needs to be perfected for a Windows environment in my mind.
```suggestion
    - pip check  # [not win]
```

MatthewFlamm: It is weird that we get this problem presumably by importing pyvista.  Yet pyvista does not skip any checks here on their recipe.

https://github.com/conda-forge/pyvista-feedstock/blob/cd2b71c6d642a3a745fe6eeaa46a9ba52db711f0/recipe/meta.yaml#L30-L36

tkoyama010: Also `geovista` has no problem about import `pyvista`. I have no idea why it happens 🤷
https://github.com/conda-forge/geovista-feedstock/blob/89bfd5fe9f8807d9a171d1be3c4a53d61928e5b1/recipe/meta.yaml#L40-L47

MatthewFlamm: Both have `yum_requirements.txt`?

tkoyama010: Oh! Good point. Maybe that is the reason.

tkoyama010:
🤞

recipes/pytest-pyvista/meta.yaml
ocefpaf: yum requirements should affect only Linux. That is probably not the reason for it. Most of the time, when pip check fails, it is a problem in the package metadata (a conda dependency or this package) and 90% of the time it is harmless to skip.

tkoyama010: @ocefpaf Thanks for your comment. We will skip it.

tkoyama010:
@bjlittle @MatthewFlamm @ocefpaf It is ready to go! Thanks for your help!

MatthewFlamm (approved):

PR #24090, opened 2023-09-28
adding neddy

thespacedoctor:


recipes/neddy/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+requirements:
+  host:
+    - pip
+    - python >=3.9
+  run:
+    - fundamentals
+    - python >=3.9
+    - pyyaml
+    - astrocalc
```
synapticarbors, line 27: It looks like `astrocalc` isn't available on conda-forge. If you want to add it in this PR, that would solve the issue. See the following for some info on submitting a PR with two packages:
https://conda-forge.org/docs/maintainer/adding_pkgs.html#adding-multiple-packages-at-once

Or you can open another PR for `astrocalc`.

recipes/neddy/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - eventlet
+    - numpy
+    - unicodecsv
+
+test:
```
synapticarbors, line 32: I recommend adding a `pip check` test:
https://github.com/conda-forge/staged-recipes/blob/main/recipes/example/meta.yaml#L61-L64

as well as adding a `commands` test for the entry point:
https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#test-commands

PR #24102, opened 2023-09-29
add Einsums

loriab:
Einsums is a C++ library that does tensor operations and determines at compile time the best blas pattern to call. It is being used in some new computational chemistry development.

A possible concern is Einsums's handling of two dependencies: h5cpp (header-only) and hptt (compiled; 11 files). Ordinarily, I'd package these up with new recipes. In this case, I haven't because of the following:
* upstream hptt (https://github.com/springer13/hptt) hasn't been worked on since 2018 and isn't merging fixes. upstream h5cpp (https://github.com/steven-varga/h5cpp) hasn't been worked on since 2021 and isn't merging fixes.
* h5cpp and hptt aren't packaged by Fedora, Debian, or OpenSuse.
* Einsums needs modifications to both dependencies, both for build and content.

Instead, Einsums:
* maintains forks of the upstream dependencies (https://github.com/Einsums) and pulls in the src at build-time, checking the hash.
* h5cpp (the header-only) is consumed at Einsums compile-time,  but in the Einsums install, its headers are installed under the Einsums namespace for downstream packages to compile against Einsums.
* hptt gets compiled as an object library into Einsums and doesn't appear in an Einsums install.
* the licenses for both h5cpp and hptt are included in the THIRD_PARTY_LICENSES file.

I hope these reasons and practices are sufficient to allow vendoring these particular dependencies in Einsums.

Checklist
  * see notes above. yes, licenses are packaged.
  * see notes above. yes, license is packaged. 
- [x] Build number is 0.

### DIFF
--- a/recipes/einsums/meta.yaml
+++ b/recipes/einsums/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "Einsums" %}
 {% set version = "0.1" %}
-{% set sha256 = "bd1fe2ba41053194164a063ff5e011d4f840b32bdab4540635a717a5ac1020ab" %}
+{% set sha256 = "76bd1ee576cba7b8af026011999bd9594adc74fdc897635ba7861571c059c75c" %}
 {% set build_number = 0 %}
 
 package:
--- a/recipes/einsums/meta.yaml
+++ b/recipes/einsums/meta.yaml
@@ -1,6 +1,6 @@
 {% set name = "Einsums" %}
 {% set version = "0.1" %}
-{% set sha256 = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855" %}
+{% set sha256 = "1baffca67ed789f79813cfe100140d7a86c147d23d7072f291098b8b887448bf" %}
 {% set build_number = 0 %}
 
 package:
--- a/recipes/einsums/meta.yaml
+++ b/recipes/einsums/meta.yaml
@@ -29,8 +29,7 @@ requirements:
     - {{ compiler('cxx') }}
     - cmake
     - ninja
-    - libgomp                                                              # [linux]
-    - llvm-openmp                                                          # [osx]
+    - llvm-openmp                                                          # [unix]
   host:
     - catch2
     - elfutils                                                             # [linux]
@@ -38,15 +37,14 @@ requirements:
     - hdf5
     - range-v3 0.12
     - zlib
-    - blas-devel * *_{{ blas_impl }}
+    - mkl-devel {{ mkl }}                                                  # [blas_impl == "mkl"]
+    - blas-devel * *_{{ blas_impl }}                                       # [blas_impl == "openblas"]
     - fftw                                                                 # [blas_impl == "openblas"]
     - openblas * openmp_*                                                  # [blas_impl == "openblas"]
-    - libgomp                                                              # [linux]
-    - llvm-openmp                                                          # [osx]
+    - llvm-openmp                                                          # [unix]
   run:
     - libblas * *_{{ blas_impl }}
-    - {{ pin_compatible("libblas") }}
-    - llvm-openmp                                                          # [osx]
+    - llvm-openmp                                                          # [unix]
 
 #####
 
--- a/recipes/einsums/meta.yaml
+++ b/recipes/einsums/meta.yaml
@@ -54,7 +54,7 @@ test:
     - range-v3 0.12
     - zlib
     - mkl-devel                                                            # [blas_impl == "mkl"]
-    - openblas                                                             # [blas_impl == "openblas"]
+    - blas-devel * *_openblas                                              # [blas_impl == "openblas"]
   files:
     - timing/CMakeLists.txt
   source_files:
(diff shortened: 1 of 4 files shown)


jturney:
I'm willing to be a maintainer for this package.

loriab:

recipes/einsums/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+  host:
+    - catch2
+    - elfutils                                                             # [linux]
+    - fmt 10.1
+    - hdf5
+    - range-v3 0.12
+    - zlib
+    - mkl-devel {{ mkl }}                                                  # [blas_impl == "mkl"]
+    - blas-devel * *_{{ blas_impl }}                                       # [blas_impl == "openblas"]
+    - fftw                                                                 # [blas_impl == "openblas"]
```
jakirkham, line 41: Is this meant to be constrained by `openblas`?

jakirkham (commented):
Thanks Lori! 🙏

Had a question below

recipes/einsums/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - openblas * openmp_*                                                  # [blas_impl == "openblas"]
+    - llvm-openmp
```
jakirkham, line 43: As [`llvm-openmp` has `run_exports/strong`]( https://github.com/conda-forge/openmp-feedstock/blob/62f174e38a237b9be6fd86d7553ae55fbbee6505/recipe/meta.yaml#L30-L32 ), there is no need to list it in `host`

```suggestion
```

recipes/einsums/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+  run:
+    - libopenblas * openmp_*                                               # [blas_impl == "openblas"]
+    - llvm-openmp
```
jakirkham, line 46: Same as above with `run`

```suggestion
```

recipes/einsums/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+    - elfutils                                                             # [linux]
+    - libzlib
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - llvm-openmp
```
jakirkham, line 31: As LLVM OpenMP is used with macOS builds and GNU OpenMP is used on Linux, would change these like so

```suggestion
    - llvm-openmp                                                          # [osx]
    - libgomp                                                              # [linux]
```

jakirkham (commented):
Sorry for the second review. Noticed some things with OpenMP usage and added them below

recipes/einsums/meta.yaml
loriab: Are you asking about fftw in particular being constrained by openblas? In that case, yes. einsums uses the internal FFT when MKL is selected, and MKL & OpenBLAS are the only implementations allowed at present.

Thanks for the review! Sorry I wasn't monitoring this PR last week. I'll fix up the OpenMP items you raised.

loriab:
Ok, the suggestions didn't work as proposed for mkl:
```
Packaging einsums-0.2-mkl_h09027b7_0
...
  ERROR (einsums,lib/libeinsums.so.0.2.0): Needed DSO lib/libomp.so found in ['llvm-openmp']
  ERROR (einsums,lib/libeinsums.so.0.2.0): .. but ['llvm-openmp'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
WARNING (einsums): run-exports library package conda-forge::_openmp_mutex-4.5-2_kmp_llvm in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
```
probably b/c `mkl` has depends `_openmp_mutex * *_llvm, _openmp_mutex >=4.5, llvm-openmp >=16.0.6` but the analyzer wants openmp to be explicit. So I added back `llvm-openmp` only for mkl+linux.

The osx issue is the new `libxcc` that's popped up a lot lately.

recipes/einsums/meta.yaml
```diff
@@ -0,0 +1,107 @@
...
+    - zlib
+    - mkl-devel                                                            # [blas_impl == "mkl"]
+    - blas-devel * *_openblas                                              # [blas_impl == "openblas"]
+  files:
+    - timing/CMakeLists.txt
+  source_files:
+    - timing/main.cpp
+  commands:
+    # Verify library
+    - test -f $PREFIX/lib/libeinsums$SHLIB_EXT                             # [unix]
```
carterbox, line 64:
```suggestion
    # Verify library
    - test -f $PREFIX/lib/libeinsums.so.{{ version }}                      # [linux]
    - test -f $PREFIX/lib/libeinsums$SHLIB_EXT                             # [unix]
```

carterbox: My only suggestion is to start checking the explicit so version to prevent ABI breaks in the future.

loriab: I'll add in the SO check. Thanks for the merge!

PR #24154, opened 2023-10-05
NEW: Add libgloo

carterbox:
One of the third_party dependecies of pytorch.

Checklist
- [x] Build number is 0.

https://github.com/conda-forge/pytorch-cpu-feedstock/issues/108

### DIFF
--- a/recipes/gloo/meta.yaml
+++ b/recipes/gloo/meta.yaml
@@ -1,0 +1,126 @@
+# To generate the version number
+{%- if false %}
+git clone git@github.com:facebookincubator/gloo.git
+cd gloo
+git pull
+echo 0.5.0.$(date  +%Y%m%d -d "$(git show -s --format=%ci HEAD)").$(git rev-list HEAD --count).$(git rev-parse --short HEAD)
+# Get the git sha
+git rev-parse HEAD
+wget --quiet https://github.com/facebookincubator/gloo/archive/$(git rev-parse HEAD).tar.gz
+openssl sha256 $(git rev-parse HEAD).tar.gz
+{%- endif %}
+
+{% set gitrev = "cf1e1abc95d0b961222ee82b6935f76250fbcf16" %}
+{% set version = "0.5.0.20230719.418.cf1e1ab" %}
+{% set build = 0 %}
+
+# Set CUDA related variables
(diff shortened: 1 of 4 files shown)


carterbox:
Please ignore the linter. It wasn't designed to handle pinnings with build strings.

recipes/gloo/meta.yaml
```diff
@@ -0,0 +1,140 @@
+# To generate the version number
+{%- if false %}
+git clone git@github.com:facebookincubator/gloo.git
+cd gloo
+git pull
+echo 0.5.0.$(date  +%Y%m%d -d "$(git show -s --format=%ci HEAD)").$(git rev-list HEAD --count).$(git rev-parse --short HEAD)
```
hmaarrfk, line 6: how do you choose 0.5.0?

hmaarrfk:
is the goal to help pytorch build better? if so I would really like an example that shows the resulting simplifications. I never convinced myself that they actually "simplified" the process.

recipes/gloo/meta.yaml
carterbox: This is the version stated in the CMake files! https://github.com/facebookincubator/gloo/blob/main/CMakeLists.txt

carterbox:
The goal is to reduce pytorch build times, not necessarily make it simpler. If your're no longer persuing de-static linking of pytorch dependencies could you please close all of the issues and PRs related to that?

hmaarrfk:
> The goal is to reduce pytorch build times,

The challenge is that I don't really think "gloo" on its own gets us under the 6 hour limit.

if you see any related PRs of mine, it is fine to close them, I'll close them as I come across them.

The challenge was that I needed to:

1. build up tooling to build the exact versions needed by pytorch.
2. build up tooling to adjust the pytorch build process to use pre-compiled libraries.

nothing I did got under the 6 hour limit.

carterbox:
OK. I might try one other thing after re-reading the existing discussions.

hmaarrfk:
don't give up, but I went as far as https://anaconda.org/mark.harfouche/repo#deletePackages

i had "gloo" working, but i had to hack around to get pytorch to detect it , and thats where I stopped.

PR #24158, opened 2023-10-05
split fenics pure Python dependencies

minrk:
split these out of the multi-output `fenics` feedstock, which now builds only dolfin

counterpart to https://github.com/conda-forge/fenics-feedstock/pull/187 removing publishing of these packages, which haven't ben updated in years, but keep getting rebuilds due to the migrations for dolfin dependencies (petsc, etc.).

- feedstock-outputs: https://github.com/conda-forge/feedstock-outputs/pull/53
- associated update to `fenics`, which will continue building dolfin and stop unnecessarily rebuilding these packages: https://github.com/conda-forge/fenics-feedstock/pull/187

Note that build numbers start at 39 instead of 0, because these packages had been previously published from the monolithic `fenics` feedstock.

Checklist
- [x] Build number is 0.
- [x] GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there. (list unchanged from the split feedstock)

### DIFF
--- a/recipes/fenics-ufl-legacy/meta.yaml
+++ b/recipes/fenics-ufl-legacy/meta.yaml
@@ -2,7 +2,7 @@
 {% set tag = "ufl_legacy-" + version %}
 
 package:
-  name: fenics-ufl_legacy
+  name: fenics-ufl-legacy
   version: {{ version }}
 
 source:
@@ -20,6 +20,7 @@ requirements:
     - pip
   run:
     - python >=3.7
+    - setuptools
     - numpy
 test:
   imports:


minrk:
I think this is ready to go Windows builds fail, but Windows won't be used. Not sure how to skip Windows for noarch.

cc @ocefpaf who merged the https://github.com/conda-forge/feedstock-outputs/pull/53 part

ocefpaf:
> I think this is ready to go Windows builds fail, but Windows won't be used

I'm on mobile only and it is hard to read the logs. If Windows is failing due to the absence of a dependency, then it is fine as is. If the failure is something we need to use the "semi-arch" approach to avoid the noarch package getting installed on Windows and not working there. See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-packages-with-os-specific-dependencies for an example of noarch packages that won't be installable on Windows.

minrk:
These are pure Python packages that don't support Windows, but since they are pure Python, I'm not sure there's a benefit to _preventing_ installation on Windows (you can `pip install` them from PyPI just fine). They just won't be very useful. I'm sure some subset of operations would work just fine.

ocefpaf:
> These are pure Python packages that don't support Windows, but since they are pure Python, I'm not sure there's a benefit to _preventing_ installation on Windows (you can `pip install` them from PyPI just fine). They just won't be very useful. I'm sure some subset of operations would work just fine.

If that is the case, it is fine as is. If you really want to prevent Windows users from installing them, just add the `__unix` in the run requeriments and the solver will ensure this package is unsolvable on Windows but still noarch.

minrk:
Checking back in, I think these are ready to go (I just made a name change to match the pypi name for fenics-ufl-legacy). I don't think preventing installation on Windows is helpful for these pure Python packages, despite the fact that Windows is "unsupported," so the tests won't pass, since they won't be run again once the feedstock is created.

recipes/fenics-fiat/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+package:
+  name: fenics-fiat
+  version: {{ version }}
+
+source:
+  url: https://bitbucket.org/fenics-project/fiat/downloads/fiat-{{ version }}.tar.gz
+  sha256: 341a1046cbe0f5f2eb26630c2f71f378b0dca51daf9892a54a2ff193970371e9
+
+build:
+  number: 39
```
ocefpaf, line 12: Should it be 0?

recipes/fenics-ffc/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  version: {{ version }}
+
+source:
+  url: https://bitbucket.org/fenics-project/ffc/downloads/ffc-{{ version }}.tar.gz
+  sha256: 4ff821a234869d8b9aaf8c5d7f617d42f9c134a2529e76c9519b681dff35affd
+  patches:
+    - ufc-include-path.patch
+
+build:
+  number: 39
```
ocefpaf, line 14: 0?

ocefpaf:
> I don't think preventing installation on Windows is helpful for these pure Python packages, despite the fact that Windows is "unsupported," so the tests won't pass, since they won't be run again once the feedstock is created.

I'm not sure I follow. So, even though Windows is unsupported and the tests fails, you are ok with Windows users being able to install these?

recipes/fenics-ffc/meta.yaml
minrk: As mentioned in the description, these packages used to be built as part of multiple outputs in fenics-feedstock, so they are continuing their build numbers as standalone feedstocks (with the exception of fenics-ufl-legacy, which hasn't yet been on conda-forge, so starts at 0.)

minrk:
> So, even though Windows is unsupported and the tests fails, you are ok with Windows users being able to install these?

Yes. Just like on PyPI for pure Python packages like these, it solves no problems at all to artificially _prevent_ installation on Windows, as far as I can tell. And while it's "unsupported" that doesn't actually mean they can't be used at all.

recipes/fenics-ffc/meta.yaml
ocefpaf: I thought there was a name change along the way.  Hard to follow all the changes here. Sorry.

ocefpaf:
> unsupported" that doesn't actually mean they can't be used at all.

Ok, that is what I wanted to know. Folks tend to be quite aggressive when opening issues upstream when conda-forge packages are involved, and upstream usually complains about it. If you are aware of what works and what doesn't, we can merge it.

recipes/fenics-ffc/meta.yaml
minrk: Not a problem, it's a weird situation. Thanks for reviewing!

PR #24166, opened 2023-10-06
Add orange3-fairness

ZanMervic:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/example/orange3-fairness/meta.yaml
+++ b/recipes/example/orange3-fairness/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "Orange3-Fairness" %}
+{% set version = "0.1.6" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 18f85525bf1a4b1cf6286a92051b6053ad745826311c037cd40482411b95824b
+
+build:
+  number: 0
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - aif360 ==0.5.0
+    - numpy 1.23.*
+    - orange3
+    - pip
+    - python
+    - setuptools
+  run:
+    - aif360 ==0.5.0
+    - numpy 1.23.*
+    - orange3
+    - python
+    - setuptools
+
+test:
+  imports:
+    - orangecontrib
+    - orangecontrib.fairness
+    - orangecontrib.fairness.evaluation
+    - orangecontrib.fairness.modeling
+    - orangecontrib.fairness.widgets
+    - orangecontrib.fairness.widgets.tests
+
+about:
+  home: https://orangedatamining.com/
+  license: GPL-3.0-only
+  license_family: GPL3
+  license_file: LICENSE
+  summary: "Orange3 add-on for fairness-aware machine learning."
+  doc_url: https://orange3-fairness.readthedocs.io/en/latest/
+  dev_url: https://github.com/biolab/orange3-fairness
+
+extra:
+  recipe-maintainers:
+    - zanmervic
--- a/recipes/orange3-fairness/meta.yaml
+++ b/recipes/orange3-fairness/meta.yaml
@@ -18,14 +18,14 @@ requirements:
   host:
     - aif360 ==0.5.0
     - numpy 1.23.*
-    - orange3
+    - orange3 >=3.36
     - pip
     - python >=3.9
     - setuptools
   run:
     - aif360 ==0.5.0
     - numpy 1.23.*
-    - orange3
+    - orange3 >=3.36
     - python >=3.9
     - setuptools
 
--- a/recipes/orange3-fairness/meta.yaml
+++ b/recipes/orange3-fairness/meta.yaml
@@ -21,9 +21,9 @@ requirements:
     - setuptools
   run:
     - aif360 >=0.5.0
-    - numpy 1.23.*
+    - numpy <=1.23
     - orange3 >=3.36
-    - python >=3.9
+    - python >=3.9,<3.12
     - setuptools
 
 test:
--- a/recipes/orange3-fairness/meta.yaml
+++ b/recipes/orange3-fairness/meta.yaml
@@ -32,11 +32,10 @@ requirements:
 test:
   imports:
     - orangecontrib
-    - orangecontrib.fairness
-    - orangecontrib.fairness.evaluation
-    - orangecontrib.fairness.modeling
-    - orangecontrib.fairness.widgets
-    - orangecontrib.fairness.widgets.tests
+  commands:
+    - pip check
+  requires:
+    - pip
 
 about:
   home: https://orangedatamining.com/
--- a/recipes/orange3-fairness/meta.yaml
+++ b/recipes/orange3-fairness/meta.yaml
@@ -17,7 +17,7 @@ build:
 requirements:
   host:
     - pip
-    - python >=3.9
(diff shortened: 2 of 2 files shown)


recipes/example/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  noarch: python
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - aif360 ==0.5.0
+    - numpy 1.23.*
+    - orange3
+    - pip
+    - python
```
PrimozGodec, line 23:
```suggestion
    - python >=3.9
```
Since it is noarch, it needs to have a lower bound. 3.9 is fine since Orange supports only >=3.9

recipes/example/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - setuptools
+  run:
+    - aif360 ==0.5.0
+    - numpy 1.23.*
+    - orange3
+    - python
```
PrimozGodec, line 29:
```suggestion
    - python >=3.9
```
Same here

recipes/example/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  license: GPL-3.0-only
+  license_family: GPL3
+  license_file: LICENSE
+  summary: "Orange3 add-on for fairness-aware machine learning."
+  doc_url: https://orange3-fairness.readthedocs.io/en/latest/
+  dev_url: https://github.com/biolab/orange3-fairness
+
+extra:
+  recipe-maintainers:
+    - zanmervic
```
PrimozGodec, line 52:
```suggestion
    - zanmervic
    - primozgodec
```

It is good that each addon recipe is maintained by more people in case one is not available when it is needed. I am adding myself, but you can also add anyone that you think can help or is included in the project.

PrimozGodec (commented):
Besides comments look good. Make sure also lint is happy.

ZanMervic:

ZanMervic:

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - setuptools
+
+test:
+  imports:
+    - orangecontrib
+    - orangecontrib.fairness
+    - orangecontrib.fairness.evaluation
+    - orangecontrib.fairness.modeling
+    - orangecontrib.fairness.widgets
+    - orangecontrib.fairness.widgets.tests
```
HeavyTony2, line 39: You can't have elaborate test with orange packages because of the absence of any graphical interface.

HeavyTony2 (changes requested):

recipes/orange3-fairness/meta.yaml
HeavyTony2:
```
test:
  imports:
    - orangecontrib
  commands:
    - pip check
  requires:
    - pip
```

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
```
HeavyTony2, line 40:
```suggestion
test:
  imports:
    - orangecontrib
  commands:
    - pip check
  requires:
    - pip

```
You can't have elaborate test with orange packages because of the absence of any graphical interface.

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - aif360 ==0.5.0
+    - numpy 1.23.*
+    - orange3 >=3.36
+    - pip
+    - python >=3.9
+    - setuptools
```
HeavyTony2, line 24:
```suggestion
    - pip
    - python >=3.9
    - setuptools
```
It's a pure python package, so even Orange isn't required as hosting dependancy.

HeavyTony2:
The error message makes me think it might need a a `conda_build_config.yaml` file with this code
```
channel_sources:
  - conda-forge/label/python_rc,conda-forge
```

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - pip
+    - python >=3.9
+    - setuptools
+  run:
+    - aif360 >=0.5.0
+    - numpy 1.23.*
```
xhochy, line 24: Are you sure you want to pin so tight?

xhochy:
The error message is misleading. It tries to build by default with Python 3.12 amd your numpy pin is for a version that is not supported by 3.12

PrimozGodec:
> The error message is misleading. It tries to build by default with Python 3.12 amd your numpy pin is for a version that is not supported by 3.12

A better pin for numpy would be numpy <=1.23. It is here because aif360 doesn't support numpy >=1.24. It would be correct to put this pin at aif360-feedstock, but then we have the same problem. The maintainers of aif360 already proposed a fix but they didn't make a release on Pypi. We asked them to do that, so we hope it is done soon.

xhochy:
Then add that as a pin and limit python also to `<3.12`

PrimozGodec:
I am afraid that this build still wants to use build 0 of aif360 0.5.0 even though build 1 is available (changes in this PR: https://github.com/conda-forge/aif360-feedstock/pull/5). Build 0 had a `python>=3.6<3.9` limit, which is also marked in the conflict:

> │  │  └─ python >=3.6,<3.9 , which conflicts with any installable versions previously reported;

Do you have any idea why it is happening? Build 1 is available at https://anaconda.org/conda-forge/aif360/files.

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  script: "{{ PYTHON }} -m pip install . -vv"
+
+requirements:
+  host:
+    - pip
+    - python >=3.9,<3.12
+    - setuptools
+  run:
+    - aif360 >=0.5.0
+    - numpy <=1.23
```
PrimozGodec, line 24: I think it should be changed to 
```suggestion
    - numpy <1.24
```
to also include versions between 1.23.1 and 1.23.5.

recipes/orange3-fairness/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  host:
+    - pip
+    - python >=3.9,<3.12
+    - setuptools
+  run:
+    - aif360 >=0.5.0
+    - numpy <1.24
+    - orange3 >=3.36
+    - python >=3.9,<3.12
+    - setuptools
```
HeavyTony2, line 27:
```suggestion
    - pip
    - python <3.12
    - setuptools
  run:
    - aif360 >=0.5.0
    - numpy <1.24
    - orange3 >=3.32.0
    - python <3.12
```
I tried it on two different spaces and it should work.

* Setuptools is clearly not needed for the run.
* According to the tox file, minimum **Orange3** required minimum version is 3.32.0. I suppose the intention was to have the most recent version installed. However, the conda recipe is supposed to be consistent with the actual requirements from the source code. It also suppresses the inherent flexibility of pure Python packages which is supposed to be the point. https://github.com/biolab/orange3-fairness/blob/8ae2f385962a9ebd911a62ab0e805d3c55f3fcfe/tox.ini#L27 
*Neither Oange3 nor Orange3-fairness require a minimum Python3 version according to Pypi. Aif360 requires 3.7 or higher, but  Conda currently support 3.8 to 3.11, so let's keept that simple.

HeavyTony2: @ZanMervic

PR #24176, opened 2023-10-06
Adding kernax

bstaber:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/kernax/meta.yaml
+++ b/recipes/kernax/meta.yaml
@@ -3,7 +3,7 @@ package:
   version: 0.0.1
 
 source:
-  git_url: # https://gitlab.com/drti/kernax.git
+  git_url: https://gitlab.com/drti/kernax.git
 
 build:
   script: {{ PYTHON }} -m pip install . -vv
--- a/recipes/kernax/meta.yaml
+++ b/recipes/kernax/meta.yaml
@@ -1,0 +1,33 @@
+package:
+  name: kernax
+  version: 0.0.1
+
+source:
+  git_url: # https://gitlab.com/drti/kernax.git
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  build:
+    - python
+    - setuptools
+
+  run:
+    - python>=3.9
+    - jax
+    - blackjax==0.9.6
+    - numpy
+    - scipy
+    - tqdm
+    - pandas
+    - matplotlib
+    - scikit-learn
+    - typing_extensions
+
+test:
+  imports:
+    - kernax
+
+about:
+  home: https://gitlab.com/drti/kernax
--- a/recipes/kernax/meta.yaml
+++ b/recipes/kernax/meta.yaml
@@ -7,6 +7,7 @@ source:
 
 build:
   script: {{ PYTHON }} -m pip install . -vv
+  number: 0
 
 requirements:
   build:
@@ -14,7 +15,7 @@ requirements:
     - setuptools
 
   run:
-    - python >=3.9
+    - python
     - jax
     - blackjax ==0.9.6
     - numpy
--- a/recipes/kernax/meta.yaml
+++ b/recipes/kernax/meta.yaml
@@ -3,7 +3,9 @@ package:
   version: 0.0.1
 
 source:
-  git_url: https://gitlab.com/drti/kernax.git
+  # git_url: https://gitlab.com/drti/kernax.git
+  url: https://gitlab.com/drti/kernax/-/archive/main/kernax-main.tar.gz
+  sha256: bd3896bbb15120ebc3c7732899451f924dba8367d1b1119a61018295664f8fd1
 
 build:
   script: {{ PYTHON }} -m pip install . -vv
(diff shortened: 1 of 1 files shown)


bstaber:

bstaber:
I'm willing to be a maintainer

bstaber:
@conda-forge/staged-recipes, sorry I'm not sure if I had to ping you with this!

recipes/kernax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  version: 0.0.1
+
+source:
+  # git_url: https://gitlab.com/drti/kernax.git
+  url: https://gitlab.com/drti/kernax/-/archive/main/kernax-main.tar.gz
+  sha256: bd3896bbb15120ebc3c7732899451f924dba8367d1b1119a61018295664f8fd1
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [win]
```
synapticarbors, line 12: Does this package not build under windows? If it's just a missing dependency, I'd say we should make this `noarch: python`:

recipes/kernax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  url: https://gitlab.com/drti/kernax/-/archive/main/kernax-main.tar.gz
+  sha256: bd3896bbb15120ebc3c7732899451f924dba8367d1b1119a61018295664f8fd1
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [win]  
+  number: 0
+
+requirements:
+  build:
```
synapticarbors, line 16: This should be `host` instead of `build`. Also, I believe `setuptools-scm` is missing (it's in the `[build-system]` upstream).

recipes/kernax/meta.yaml
```diff
@@ -0,0 +1,51 @@
+package:
+  name: kernax
+  version: 0.0.1
+
+source:
+  # git_url: https://gitlab.com/drti/kernax.git
+  url: https://gitlab.com/drti/kernax/-/archive/main/kernax-main.tar.gz
```
synapticarbors, line 7: I'm concerned that the url isn't really a release tied to the `version`. I'm not familiar with how gitlab does releases, but is there a less generic tarball available that clearly is tied to one of the tagged releases?

recipes/kernax/meta.yaml
bstaber: Yes, it didn't build under windows because the latest versions of jax (and jaxlib) are not available yet under that platform, if I understood clearly. So it's only a dependency issue, as you mentioned. I changed the recipe to `noarch: python`.

recipes/kernax/meta.yaml
bstaber: Fixed, thank you.

recipes/kernax/meta.yaml
bstaber: I found where the tagget tarballs are located on gitlab. Thank you for spotting that.

bstaber:
I think the pipelines are failling because the docker container can't infer the package version due to the absence of the .git folder (version is inferred from the current git tag). How can I fix this ? Should I change `git_url` instead of `url` ? Thanks for your help.

bstaber:
It's not failing anymore thanks to `git_url`, but the package indeed doesn't build under windows because `jaxlib >= 3.0.10` is missing.

synapticarbors:
The windows build failing due to the missing dependency is fine (we see this for torch and tensorflow all of the time). I'd still like to get away from the git. See:
https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos

What was the problem with https://gitlab.com/drti/kernax/-/archive/0.1.9/kernax-0.1.9.tar.gz, which is currently commented out in the recipe?

bstaber:
I think the issue is that I manage the version of the package with git tags : setuptools_scm gets this information and generates a `_version.py` file that contains the string `__version__`. Is this not recommended ? I would be happy to implement a better solution if needed. Thank you.

synapticarbors:
Using setuptools_scm shouldn't be an issue. If you are running into problems with the version, this might help resolve it:
https://conda-forge.org/docs/maintainer/knowledge_base.html#using-setuptools-scm

recipes/kernax/meta.yaml
```diff
@@ -0,0 +1,53 @@
+package:
+  name: kernax
+  version: 0.0.1
```
synapticarbors, line 3: The `version` here should match the package version (`0.1.9`) and then you can use the jinja tags in the `url`

recipes/kernax/meta.yaml
```diff
@@ -0,0 +1,53 @@
+package:
```
synapticarbors, line 1: You need to set the version as a jinja variable. See:
https://github.com/conda-forge/staged-recipes/blob/main/recipes/example/meta.yaml#L9

I would then use the variable `{{ version }}` everywhere that it appears (i.e. in the `url` under `source` and `version` under `package`).

bstaber: Thanks for helping me out. It should be corrected.

bstaber:
Thank you!

synapticarbors:
No problem. If you have any issues once it gets to the feedstock, just ping me or one of the relevant conda-forge teams.

PR #24187, opened 2023-10-09
Add mcstas

tkittel:
Checklist
- [x] Build number is 0.


willend:
@tkittel willing and able to participate, listing me is OK

tkittel:
@conda-forge/staged-recipes ready for review! :-)

tkittel:

zklaus:
I am not a core member of conda-forge, just a fellow packager with a bit of experience, but perhaps I can help get this started.
I appreciate that McStas is a venerable package with a 25 year history. It's all the more impressive that you have a modern build system.

Having said that, two things stand out with this recipe that may make people shy away from a review: Its non-standard handling of mpi and the use of Python for the build script even though it seems to be essentially a standard Cmake script.

For mpi, have you read https://conda-forge.org/docs/maintainer/knowledge_base.html#message-passing-interface-mpi yet?

For the Python build script, are you open to converting this into a standard shell script?

tkittel:
Thanks a lot for looking at this @zklaus, it is much appreciated! For the usage of MPI @willend is the expert, although note that McStas works by translating domain-specific input files ("instrument files") into C-code which is then compiled and executed (hence why we have a compiler as a `run` section dependency). Only during this latter compilation step is MPI support actually enabled and linked, which means there is no actual build-time dependency on MPI from the McStas packages themselves: MPI merely needs to be present in the users environment at the point McStas is being used. For that reason we have only two real packages in the recipe: mcstas-data (big noarch data-files) and mcstas-core (everything else which makes up McStas, but without optional dependencies). The other packages merely add in optional dependencies on things like GUI's and MPI. Most users are expected to simply install the top-level kitchen-sink package "mcstas", but for things like CI and integration with other codes, it is going to be tremendously useful to have a smaller and more lean mcstas-core package.

The reason for writing build scripts in python rather than as shell scripts, is that we in the near future plan to enable support for windows as well, and having to keep duplicated `.sh` and `.bat` build scripts seems not only like a maintenance hassle down the road, but also as a constant source of bugs (esp. we the developers and maintainers are mostly unix-users).

zklaus:
I appreciate your idea about minimizing future maintenance. However, this is an issue that you share with many other packages (as you may imagine), and people are used to certain ways, so it may be easier for other people to deal with it in another way. Maybe we can still adopt your approach and learn from it for other packages, but just to get a better understanding for myself, I have opened #24410 to try and move this forward.

tkittel:
So I plan to try to implement the suggestions from @zklaus, although first I thought I would rebase the PR on the latest developments, since it has been a while :-)

willend:
Nice @tkittel 👍 🚀

tkittel:
So I have now updated the recipe to implement the suggestions from @zklaus . We no longer use `.py` build scripts, and we no longer have any `mpi` packages as dependencies (they were not worth the hassle as any mcstas users needing them can simply install them manually).

It would be great if someone with merge rights could now review this :-)

tkittel:
@conda-forge/help-python-c  @conda-forge/staged-recipes we really would love a review!

willend:
@conda-forge/help-python-c @conda-forge/staged-recipes it would be great if someone could initiate a review! :-)

recipes/mcstas/install-files-core.sh
```diff
@@ -0,0 +1,65 @@
...
+for i in $(seq 1 100000); do
+    #Find a unique build dir:
+    BLDDIR="$PWD/build_mcstas_core_${i}"
+    if [ ! -d "${BLDDIR}" ]; then
+        break
+    fi
+done
+
+mkdir "${BLDDIR}"
+cd "${BLDDIR}"
```
dhirschfeld, line 25: I'm curious what the point of creating a `BLDDIR` is? It doesn't seem to be specified elsewhere and the build happens in an isolated environment already so, it doesn't appear necessary?

NB: I don't have merge rights - just a casual observer.

tkittel: Thanks for the feedback/question @dhirschfeld! Well, in my experience it always pays off to have the CMake build dir completely clean and isolated from the source directory, otherwise there might be hard to debug issues coming up some day. And in fact, the source code is located in a sub-directory of the working directory (`./src/`), so it is indeed not completely clean (imagine if CMake some day started using a cache file named `src`). And as lines 16-22 testify, we did actually run into clashy not-completely-clean issues. With the lines you comment on, we are iron-clad sure to have a nice clean directory, even if something would change in either CMake or conda-forge code that we would not be aware of. Seems like a small price to pay :-)

recipes/mcstas/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+  sha256: {{ sha_value }}
+  folder: src
+
+build:
+  number: 0
+
+outputs:
+
+  - name: mcstas-core
+    script: install-files-core.sh
```
dhirschfeld, line 20: What's the motivation for the complex script over just copying over the source folder verbatim - i.e. 
```bash
cp -a "${SRCDATADIR}/." "${DESTDATADIR}/"
```

recipes/mcstas/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+      commands:
+        - python test_pkg.py core
+      source_files:
+        - src/mcstas-comps/examples/BNL_H8.instr
+        - src/mcstas-comps/examples/Union_manual_example.instr
+
+  - name: mcstas-data
+    build:
+      noarch: generic
+    script: install-files-data.sh  # [not win]
```
dhirschfeld, line 51: What's the motivation for the complex script over just copying over the source folder verbatim - i.e. 
```bash
cp -a "${SRCDATADIR}/." "${DESTDATADIR}/"
```

tkittel: Basically most of that script is mostly just to be defensive in case some unexpected files appears in that folder in the upstream project (it happened in the past) which could cause a headache if simply copying blindly. If the upstream project one day acquires some automated safeguards against that, I will be happy to perform the simplification :-)

recipes/mcstas/install-files-core.sh
dhirschfeld: Why not just `mkdir build_mcstas_core`?

Since the build is in an isolated environment this folder will only exist if it exists in the upstream repository. It seems unnecessary to check folder names first?

recipes/mcstas/meta.yaml
dhirschfeld: Having complex scripts that aren't well motivated is a recipe for scaring away reviewers.

I still have no idea why you're writing a complex function to only copy 2 levels deep? It's more to review, and if I don't understand why you're doing it I probably won't hit the merge button (if I could). I'd include comments in the script stating why you're doing these very non-standard things.

I also think it would be better to split into an install and test script. The install script would just be a simple directory copy and after installation you can then run `check_no_forbidden_chars` as a test. Making the install script as bog-standard as possible will likely improve your chances of review from someone who can merge this.

tkittel: Ok thanks, that is a good point! I will try next week to move the sanity checking to the test script instead.

recipes/mcstas/install-files-core.sh
tkittel: Well, that was what I did the first time around! :-)

Then the tests failed with "directory already exists", and I started to wonder if perhaps some of the smart conda-magic was trying to use the same build dir to somehow reuse the CMake cache. Since the main mcstas developer claim to not support such usage, I came up with the logic above.

dhirschfeld (approved):
Looks fine, though it might be bettered structured (and easier to review) if you split the `check_no_forbidden_chars` part into a separate, post-install test.

tkittel:
Thanks a lot @dhirschfeld ! Next week I will do an update for the `check_no_forbidden_chars` part as you suggest.

tkittel:
The suggested changes have now been implemented, and all tests are passing again. 

@conda-forge/help-python-c @conda-forge/staged-recipes we really would love a review!

tkittel:
I also did a rebase on upstream yesterday, to prevent the PR from becoming outdated.

dhirschfeld (approved):
LGTM

tkittel:
Hurray! Thanks a lot @beckermr and also thanks to @dhirschfeld and @zklaus for very helpful feedback!

PR #24225, opened 2023-10-12
Add napari-cursor-tracker recipe

goanpeca:
Checklist
- ~[ ] If static libraries are linked in, the license of the static library is packaged.~
- [x] Build number is 0.


goanpeca:
We are in the process of creating conda packages for [napari plugins](https://www.napari-hub.org/) and related dependencies. Once this recipe passes all the checks, every time a new release goes into PyPI, a PR will be created automatically, with usually very minor need for changes so the process of maintaining this moving forward is small. Also the `@conda-forge/napari` team will be available to help as needed.

You can read a bit more about conda-forge at https://conda-forge.org/#contribute

---

@jaimergp, do you agree with being added as maintainer?

goanpeca:
This one is ready for review @conda-forge/staged-recipes :)

recipes/napari-cursor-tracker/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+test:
+  imports:
+    - napari_cursor_tracker
+  commands:
+    - pip list
+    - pip check
+  requires:
+    - pip
+    - pyqt
```
xhochy, line 39: Same qt issue again. I have probably overlooked this in previous reviews but we should fix this. In contrast to pip we have the means with conda and thus should always provide the user with a fully functional setup.

xhochy (changes requested):

recipes/napari-cursor-tracker/meta.yaml
goanpeca: @xhochy again, same comment from previous recipe (and this is something that has been  decided for napari). Napari by default does not provide a binding also on pip, so a user needs to install napari and then a binding, pyside or pyqt, hence why all these plugins will need to include a binding so the imports or tests pass.

This is not something we need to fix. Is a packaging decision. See https://napari.org/stable/tutorials/fundamentals/installation.html

goanpeca: The variant option already discussed and voted against.

See:
- https://github.com/conda-forge/napari-feedstock/pull/48
- https://github.com/conda-forge/napari-feedstock/issues/47

The napari team is ok with having this as is for both conda and pip.

xhochy: Fine with me then.

xhochy (approved):

PR #24241, opened 2023-10-13
Add r-bigsnpr from CRAN using conda_r_skeleton_helper

pettyalex:
This package depends on several packages that are not yet present in conda-forge, including r-bigsparser, r-bigutilsr, and r-runonce. I've made PRs to add each of those individally:

https://github.com/conda-forge/staged-recipes/pull/24240
https://github.com/conda-forge/staged-recipes/pull/24239
https://github.com/conda-forge/staged-recipes/pull/24238

Hopefully all three of these can be added, at which point r-bigsnpr can be added to conda-forge. I will be working through the checklists for all four of these PRs. Each of the four packages recipes was generated by the helper script https://github.com/bgruening/conda_r_skeleton_helper as directed in the conda-forge contribution guidelines.

Edit: I just noticed that conda-forge policies allow multiple packages in a single PR. I could move all four into this PR if preferred. Given that I've already made the 4 PRs and have the three dependencies passing CI checks, I'm also OK waiting for them to go in before merging this.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -56,6 +56,7 @@ requirements:
     - r-ggplot2
     - r-magrittr
     - r-quadprog
+    - r-r.utils
     - r-rmio
     - r-rmutil 
     - r-roptim >=0.1.6
@@ -87,6 +88,7 @@ requirements:
     - r-ggplot2
     - r-magrittr
     - r-quadprog
+    - r-r.utils
     - r-rmio
     - r-rmutil 
     - r-roptim >=0.1.6
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -56,6 +56,9 @@ requirements:
     - r-roptim >=0.1.6
     - r-runonce >=0.2.3
     - r-vctrs
+    - libblas
+    - liblapack
+    - libzlib
   run:
     - r-base
     - {{ native }}gcc-libs         # [win]
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -34,6 +34,8 @@ requirements:
     - {{ posix }}coreutils         # [win]
     - {{ posix }}zip               # [win]
     - cross-r-base {{ r_base }}    # [build_platform != target_platform]
+    - libgomp                      # [linux]
+    - llvm-openmp                  # [osx]
   host:
     - r-base
     - r-matrix >=1.3.0
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -82,9 +82,20 @@ requirements:
     - r-vctrs
 
 test:
+  requires:
+    - r-testthat
+    - r-dplyr
+    - r-dbplyr
+    - r-gaston
+    - r-hmisc
+    - r-rsqlite
+  source_files:
+    - tests/
   commands:
-    - $R -e "library('bigsnpr')"           # [not win]
+    - $R -e "library('bigsnpr')"  # [not win]
+    - $R -e "testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)"  # [not win]
     - "\"%R%\" -e \"library('bigsnpr')\""  # [win]
+    - "\"%R%\" -e \"testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)\""  # [win]
 
 about:
   home: https://privefl.github.io/bigsnpr/
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -21,6 +21,8 @@ build:
   # This is needed to suppress a warning about R.dll not being found
   missing_dso_whitelist:
     - '*/R.dll'            # [win]
+    - '*/Rblas.dll'        # [win]
+    - '*/Rlapack.dll'      # [win]
 
 requirements:
   build:
--- a/recipes/r-bigsnpr/meta.yaml
+++ b/recipes/r-bigsnpr/meta.yaml
@@ -1,0 +1,120 @@
+{% set version = '1.12.2' %}
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: r-bigsnpr
+  version: {{ version|replace("-", "_") }}
+
+source:
+  url:
+    - {{ cran_mirror }}/src/contrib/bigsnpr_{{ version }}.tar.gz
+    - {{ cran_mirror }}/src/contrib/Archive/bigsnpr/bigsnpr_{{ version }}.tar.gz
+  sha256: 443a5529ae2596eb3cf8f19760893b96beba4f12a4decbf6bea645db4d2187f6
+
+build:
+  merge_build_host: True  # [win]
+  number: 0
+  rpaths:
+    - lib/R/lib/
+    - lib/
+
+requirements:
+  build:
+    - {{ compiler('c') }}              # [not win]
+    - {{ compiler('m2w64_c') }}        # [win]
+    - {{ compiler('cxx') }}            # [not win]
+    - {{ compiler('m2w64_cxx') }}      # [win]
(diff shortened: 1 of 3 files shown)


pettyalex:
Once all three dependencies have gone in, this should be buildable.

mfansler:
This package wraps some PLINK functionality, which is available on the Bioconda channel. Would it make more sense for this package to be on Bioconda instead of Conda Forge?

pettyalex:
Good call. I started off adding all of these to Bioconda, saw their guidance that if things from CRAN can easily be packaged for conda-forge instead that is preferred. 

I didn't think about `plink` because it's not a dependency on CRAN, and not required for this package to work. I know that anecdotal evidence is weak, but my research group hasn't ever used the plink-wrapping functions and I wasn't familiar with them. This package also seems to wrap `beagle` as well, hrm: https://privefl.github.io/bigsnpr/reference/snp_beagleImpute.html
https://privefl.github.io/bigsnpr/reference/download_beagle.html

Given that these are very optional, and I don't see a way to configure the package to easily use conda-installed versions of plink or beagle anyway, I'd prefer to see this in conda-forge with its wider platform support if possible. That said, if you or any other conda-forge maintainers feel that this belongs in bioconda I can easily move it over there.

mfansler:
Okay. If "*very optional*", then lets just keep it here. I concur that there doesn't appear to be option-like specifications for the PLINK and beagle PATHs, so adding them as dependencies wouldn't automate connecting them.

I just figured I'd ask since this is specifically a key advantage over CRAN - that we can additionally guarantee non-R libraries.

mfansler (changes requested):
Thanks for submitting this! Review notes:

- linked libraries need to be added
- win build needs https://github.com/conda-forge/r-runonce-feedstock/pull/2
- run testthat tests (suggestion)

pettyalex:
@mfansler thanks so much for all of the fixups. Really appreciated.

This may not be the specific place to ask general questions, but is there a way for recipes to specifically request the MKL implementations of BLAS if they're available, without requiring it? We've got a whole host of tools that run much better with the MKL but when installed with default settings from conda-forge they get OpenBLAS.

pettyalex:
Hmm, it looks like it uses https://anaconda.org/conda-forge/quadprog for its tests?

recipes/r-bigsnpr/meta.yaml
```diff
@@ -38,6 +38,9 @@ requirements:
     - llvm-openmp                  # [osx]
   host:
     - r-base
+    - r-dbplyr
+    - r-dplyr
+    - r-fdrtool
```
mfansler, line 43: If these are not in `Depends` nor `Imports` then the intent of the author(s) is presumed to be that they are optional. Since a primary use of Conda packages is for minimal deployments (e.g., a specific step in a processing pipeline) we don't want to add on unnecessary components to installation. Let's move all these to only `test` requirements.

```suggestion
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -52,15 +55,22 @@ requirements:
     - r-foreach
     - r-ggplot2
     - r-magrittr
+    - r-quadprog
     - r-rmio
+    - r-rmutil 
     - r-roptim >=0.1.6
+    - r-rsqlite
```
mfansler: Move to `test` only.

```suggestion
    - r-rmio
    - r-roptim >=0.1.6
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -52,15 +55,22 @@ requirements:
...
     - r-runonce >=0.2.3
     - r-vctrs
+    - r-xgboost
```
mfansler, line 66: Move to `test` only.

```suggestion
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -52,15 +55,22 @@ requirements:
...
     - libblas
     - liblapack
     - libzlib
   run:
     - r-base
+    - r-dbplyr
+    - r-dplyr
+    - r-fdrtool
```
mfansler, line 74: Move to `test` only.

```suggestion
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -76,19 +86,19 @@ requirements:
...
     - r-ggplot2
     - r-magrittr
+    - r-quadprog
     - r-rmio
+    - r-rmutil 
     - r-roptim >=0.1.6
+    - r-rsqlite
     - r-runonce >=0.2.3
     - r-vctrs
+    - r-xgboost
```
mfansler: Move to `test` only.

```suggestion
    - r-rmio
    - r-roptim >=0.1.6
    - r-runonce >=0.2.3
    - r-vctrs
```

mfansler (changes requested):
Still waiting for an answer on how to deal with the OSX missing `libcxx` error. In the meantime, please implement these changes, which should get the Windows builds working and minimizes the dependencies.

recipes/r-bigsnpr/meta.yaml
```diff
@@ -0,0 +1,152 @@
...
+    - r-rsqlite
+    - r-runonce >=0.2.3
+    - r-vctrs
+    - r-xgboost
+
+test:
+  requires:
+    - r-gaston
+    - r-hmisc
+    - r-testthat
```
mfansler, line 104: Moving to `test`. Also adding `r-pcadapt`, since that is now available.

```suggestion
    - r-testthat
    - r-dbplyr
    - r-dplyr
    - r-fdrtool
    - r-gaston
    - r-hmisc
    - r-mutil
    - r-pcadapt
    - r-quadprog
    - r-rsqlite
    - r-xgboost
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -0,0 +1,152 @@
...
+    - r-magrittr
+    - r-quadprog
+    - r-r.utils
+    - r-rmio
+    - r-rmutil 
+    - r-roptim >=0.1.6
+    - r-rsqlite
+    - r-runonce >=0.2.3
+    - r-vctrs
+    - r-xgboost
```
mfansler, line 98:
```suggestion
    - r-rmio
    - r-roptim >=0.1.6
    - r-runonce >=0.2.3
    - r-vctrs
```

recipes/r-bigsnpr/meta.yaml
```diff
@@ -0,0 +1,152 @@
...
```
mfansler, line 66: Move to `test` only:

```suggestion
    - r-rmio
    - r-roptim >=0.1.6
    - r-runonce >=0.2.3
    - r-vctrs
```

mfansler:
> This may not be the specific place to ask general questions, but is there a way for recipes to specifically request the MKL implementations of BLAS if they're available, without requiring it? We've got a whole host of tools that run much better with the MKL but when installed with default settings from conda-forge they get OpenBLAS.

Most of the time<sup>**1**</sup> this should be left to the end user, not something to be specified in the recipe. Using `libblas` and `liblapack` in the recipe will build against the generic reference implementation (netlib). This is then flexible to run atop any of the implementations. MKL is great for Intel x86_64 users, but Apple Silicon users highly benefit from the Accelerate and AMD users might prefer OpenBLAS.

For end users to specify MKL in an environment, add the specification:

**YAML**
```yaml
  - blas=*=mkl
```

or, with the **CLI**
```bash
conda install 'blas=*=mkl'
```

[This StackOverflow answer](https://stackoverflow.com/a/70241376/570918) has some related content (though focus is Python's NumPy).

---

<sub>**[1]**: The exception here is if the package actually will use library-specific functions at compile-time. In that case one can use variants to build against different implementations. However, this is rare and does not sound like the situation here.</sub>

recipes/r-bigsnpr/meta.yaml
```diff
@@ -0,0 +1,149 @@
...
+
+test:
+  requires:
+    - r-testthat
+    - r-dbplyr
+    - r-dplyr
+    - r-fdrtool
+    - r-gaston
+    - r-hmisc
+    - r-mutil
```
mfansler, line 96: Should be:

```suggestion
    - r-rmutil
```

PR #24280, opened 2023-10-16
Add dynare preprocessor executable only.

albop:
Background:

- dynare is a matlab toolbox which contains a C++ executable for preprocessing model
- there are opensource alternatives to the matlab part (including one in octave and another in-dev in python) which keep the preprocessor. This package will only contain the preprocessor and other package(s) will contain the python logic.
- the goal is ultimately to run everything in-browser using jupyter-lite as a for a port to emscripten-forge.

Checklist
    - I made a static build. 
- [x] Build number is 0.


albop:
@conda-forge/help-c-cpp

recipes/dynare_preprocessor/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+export BOOST_ROOT=$CONDA_PREFIX
```
xhochy, line 3:
```suggestion
export BOOST_ROOT=$PREFIX
```

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+requirements:
+  build:
+    - '{{ compiler("cxx") }}'
+    - git-lfs
+    - meson
+    - gsl
+    - openlibm
+    - boost-cpp
+    - boost
+    - pkgconfig
```
xhochy, line 22:
```suggestion
    - pkg-config
```

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  url: https://git.dynare.org/Dynare/preprocessor/-/archive/master/preprocessor-master.tar.gz
+  sha256: 25e1b0f30d49b24daf520223e1d1919e9a93dae32d6156f8269014ef2f5aa85e
+
+build:
+  number: 0
+  skip: true  # [win or osx]
+
+requirements:
+  build:
+    - '{{ compiler("cxx") }}'
```
xhochy, line 15: Meson, bison, flex and the compiler should be in build, the rest belongs into host.

xhochy (changes requested):

recipes/dynare_preprocessor/build.sh
albop: Thank you! I've done just that

recipes/dynare_preprocessor/meta.yaml
albop: Thank you @xhochy !
I followed exactly your suggestions, but made separate commits instead of using the github interface. Sorry for that. Can you mark the issues as resolved ?

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+    - meson
+    - bison
+    - flex
+  host:
+    - git-lfs
+    - gsl
+    - openlibm
+    - boost-cpp
+    - boost
```
carterbox, line 24:
```suggestion
    - libboost-devel
```
Unless you need an old version of boost, the development package now has this name. If you need python, then list it explicitly instead of depending on the "boost" package (which depends on python).

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: dynare_preprocessor
+  version: 0.0.1
```
carterbox, line 3: Please place this package into a pre-release label until it has a stable release. Users of our channel should get stable non-experimental builds by default.

From the downloaded artifacts:
```
project('dynare-preprocessor', 'cpp',
        version : '6-unstable',
        default_options : [ 'cpp_std=gnu++20', 'warning_level=2' ],
        meson_version : '>=0.64.0')
```

carterbox, line 3: For packages without official releases, please ask the upstream project to make an official release. Otherwise, the version here may become out of sync with the upstream (canonical) version number.

If the upstream package is abandoned (it doesn't seem to be), then we have a convention that the package version should be something like "0.0.0.dev{{ build }}{{ git hash }}". I'd need to look up what we've done in the past, but the special build number + a run export prevents ABI breaks for non-versioned packages while maintaining version ordering.

recipes/dynare_preprocessor/run_test.sh
```diff
@@ -0,0 +1,2 @@
+$PREFIX/bin/dynare-preprocessor | grep -q "Missing model file!"
+$PREFIX/bin/dynare-preprocessor $RECIPE_DIR/example1.mod | grep -q "Preprocessing completed."
```
carterbox, line 2:
```suggestion
dynare-preprocessor | grep -q "Missing model file!"
dynare-preprocessor $RECIPE_DIR/example1.mod | grep -q "Preprocessing completed."
```
You shouldn't need to list the full path of the installed binary. If it's not on the path in the test environment, then you package is broken!

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,40 @@
+package:
+  name: dynare_preprocessor
```
carterbox, line 2:
```suggestion
  name: dynare-preprocessor
```

recipes/dynare_preprocessor/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+export BOOST_ROOT=$PREFIX
+meson setup build_preproc -Dcpp_link_args='-pthread -static'
```
carterbox, line 4: > I made a static build.

We strongly prefer that packages are dynamically linked on our channel in order to avoid the same packages/code being shipped, compiled, stored, and downloaded multiples times. If you need static linking of non-header-only packages, please provide justification.

carterbox (changes requested):

recipes/dynare_preprocessor/meta.yaml
albop: Yes, here I choose a very small number, to avoid any clash, but I'm happy to oblige, if there is 

There are two factors to consider here:
- last released version of upstream was 5.5 but hasn't switched to the new build system (meson). Based on this it would make sense to use number 5.6-dev .
- there is no separate release of the preprocessor which is a part of the whole distribution, so I don't know if it would follow a separate numbering.

I have no personal preference here, so I'll ask the package authors (I've been in touch with them before).

recipes/dynare_preprocessor/build.sh
albop: It didn't work before I tried to add the static option ;-) On my computer it was still linking to some of my personal libraries outside of conda.
Let me try again without the static argument.

albop: @carterbox : the build passes the online test.
If I try to install the artifact from CI and install locally it also work. But I get:
```
ldd $(which dynare-preprocessor)                                                                                                                                                                                        (base) 
        linux-vdso.so.1 (0x00007ffcc87ed000)
        libstdc++.so.6 => /home/pablo/micromamba/bin/../lib/libstdc++.so.6 (0x00007f662fc1d000)
        libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f6630243000)
        libgcc_s.so.1 => /home/pablo/micromamba/bin/../lib/libgcc_s.so.1 (0x00007f6630228000)
        libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f6630223000)
        libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f662fa00000)
        librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f663021c000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f6630353000)
```
Is that ok ?
It is not totally clear to me how much of the shared libraries must be provided by conda, vs expected to be present on a reasonable system. Can you enlighten me here?
On the other hand the static build is really small in size ;-)

recipes/dynare_preprocessor/meta.yaml
albop: Thank you @carterbox for your feedback.
Just to make sure: do you make a difference between the experimental build and the experimental package ?
If I look at what upstream is doing, 5.5 is stable, but is using the old build system. Hypothetically, could a conda package eventually provide a stable release for 5.5, backporting only the new build system that is much simpler ?

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - '{{ compiler("cxx") }}'
+    - meson
+    - bison
+    - flex
+  host:
+    - git-lfs
+    - gsl
+    - openlibm
+    - libboost-devel
+    - pkg-config
```
SylvainCorlay, line 24: Move pkg-config and git-lfs to the build section.

albop: I'm moved pkg-config and removed git-lfs as it appears it is not needed anymore (we download an archive instead of cloning)

recipes/dynare_preprocessor/meta.yaml
svillemot: Dynare upstream here.

We actually don’t publish releases of the Dynare preprocessor because it is only a component of Dynare, and so far we have never distributed it independently.

I think that using the version number of the whole Dynare could lead to problems, because you may want to update the preprocessor in Conda without waiting for a new Dynare release.

My suggestion would be to use a version number specific to Conda forge, that would incorporate the git hash. The scheme proposed by @carterbox looks good to me: `0.0.0.dev{{ build }}{{ git hash }}`. A slight variation could be `0.0.0.{{ date }}.{{ git hash }}`, which looks more explicit to me. But either is fine.

recipes/dynare_preprocessor/build.sh
svillemot: The Dynare preprocessor is only linked against standard C and C++ libraries. I’m not familiar with Conda, but I would be surprised if those were not shipped as shared library by the package system.

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  host:
+    - gsl
+    - openlibm
+    - libboost-devel
+    - pthread-stubs
+
+about:
+  home: https://www.dynare.org/
+  summary: Dynare preprocessor
+  description: A platform for handling a wide class of economic models.
```
svillemot, line 29: I would mention in the description that it’s only the preprocessor that’s packaged.

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,38 @@
...
+  doc_url: https://www.dynare.org/manual/
+  dev_url: https://git.dynare.org/Dynare
```
svillemot, line 31: I would put a more specific URL, namely https://git.dynare.org/Dynare/preprocessor

recipes/dynare_preprocessor/meta.yaml
albop: yep, I took that from your preprocessor archive, but of course, that makes sense

recipes/dynare_preprocessor/meta.yaml
albop: I'm deeply indifferent between both versions. I'll go for the latest proposal by @svillemot unless @carterbox (or @SylvainCorlay) has a strong opinion.

recipes/dynare_preprocessor/meta.yaml
albop: done

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+requirements:
+  build:
+    - '{{ compiler("cxx") }}'
+    - meson
+    - bison
+    - flex
+    - pkg-config
+  host:
+    - gsl
+    - openlibm
```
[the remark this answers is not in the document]
albop: Not sure. Maybe it is needed by other parts of dynare other than the preprocessor. @svillemot ?

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - libboost-devel
```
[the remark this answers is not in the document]
albop: Done as suggested above.

recipes/dynare_preprocessor/meta.yaml
albop: @carterbox : New version is `0.0.0.dev0.2a63840eb534c3dcd1f50477d935ecc92294f00b` . All good ?

recipes/dynare_preprocessor/build.sh
albop: Thank you @carterbox for looking into it. Now I understand well. I'll wait for @svillemot for confirmation that we shouldn't use libm or openlibm and remove the deps

recipes/dynare_preprocessor/meta.yaml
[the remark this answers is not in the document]
carterbox: I don't see anywhere in the logs that the build searches for or uses gsl. The linker logs show that sysroot libm is being used instead of openlibm, so if you want to use openlibm, you'll need to tweak the build to find the correct libm.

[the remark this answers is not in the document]
albop: Ok, then I'll try to remove it.

recipes/dynare_preprocessor/meta.yaml
carterbox: 0.0.0.dev{{ build }}.{{ git hash }} should be fine.

recipes/dynare_preprocessor/meta.yaml
albop: Actually, I tried with `libboost-headers`, and got an error so I reverted to `libboost-devel`. Is that OK?

recipes/dynare_preprocessor/run_test.sh
```diff
@@ -0,0 +1,2 @@
+dynare-preprocessor | grep -q "Missing model file!"
+dynare-preprocessor $RECIPE_DIR/example1.mod | grep -q "Preprocessing completed."
```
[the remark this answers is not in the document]
albop: done

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,39 @@
+package:
+  name: dynare-preprocessor
+  version: 0.0.0.20231103.2a63840eb534c3dcd1f50477d935ecc92294f00b
```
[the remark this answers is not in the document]
albop: done as requested

recipes/dynare_preprocessor/meta.yaml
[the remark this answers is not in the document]
albop: removed

recipes/dynare_preprocessor/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  name: dynare-preprocessor
+  version: 0.0.0.dev0.2a63840eb534c3dcd1f50477d935ecc92294f00b
+
+source:
+  url: https://git.dynare.org/Dynare/preprocessor/-/archive/6235e6907526d468e2c35f14c38db1e308698bc9/preprocessor.tar.gz
+  sha256: 06f76245ff5c6ac0811438cea6b817a321dc702c188b680e2a3191b54837c4ab
+
+build:
+  number: 0
+  skip: true  # [win or osx]
```
carterbox, line 11:
```suggestion
{% set build = 1 %}
{% set commit = 6235e6907526d468e2c35f14c38db1e308698bc9 %}

package:
  name: dynare-preprocessor
  version: 0.0.0.dev{{ build }}.{{ commit }}

source:
  url: https://git.dynare.org/Dynare/preprocessor/-/archive/{{ commit }}/preprocessor.tar.gz
  sha256: 06f76245ff5c6ac0811438cea6b817a321dc702c188b680e2a3191b54837c4ab

build:
  number: {{ build }}
  skip: true  # [win or osx]
```
The build number and the version string need to by programatically linked.

recipes/dynare_preprocessor/meta.yaml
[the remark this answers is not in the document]
carterbox: Looks like you're still using libboost-devel, though.

recipes/dynare_preprocessor/run_test.sh
```diff
@@ -0,0 +1,4 @@
+set -ex
+
+dynare-preprocessor | grep -q "Missing model file!"
+dynare-preprocessor example1.mod | grep -q "Preprocessing completed."
```
carterbox, line 4: Nothing is being grepped. Doesn't that mean your test is failing?

carterbox (changes requested):

recipes/dynare_preprocessor/run_test.sh
albop: No. When you run the preprocessor without modem arguments, it should print that model file is missing. When the right kind of model file is supplied the output message contains "preprocessing completed". In both cases it returns an error if output is not what I'd expected.

recipes/dynare_preprocessor/meta.yaml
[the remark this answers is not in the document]
svillemot: `gsl` and `openlibm` are not needed and can be dropped. `boost` is used, but only its headers, not the shared libraries.

albop:
@carterbox , @svillemot : I think all points have been adressed save for the boost library. In theory it should depend only the headers. In practice, it fails if I depend only no libboost-headers (see last two commits).

carterbox:
Please explain why the build is failing, and I may know the correct solution. You may need to increase the verbosity of your build script.

albop:
@carterbox: I have a satisfying explanation, I think. My script was assuming that `$PREFIX/bin`  exists. I suppose it doesn' t by default and that it was created for `boost-devel`. Creating that directory (if needed) in the build script solves the issue.

PR #24310, opened 2023-10-19
PR For Stable Release of NTNU SmallSat HYPSO

DevAlvaroF:
Checklist
- [x ] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/5eddbd7fc9d1502169089da06c3688d9759be978/recipes/example/meta.yaml#L64-L73) for an example).
- [x ] If static libraries are linked in, the license of the static library is packaged.
- [ x] Build number is 0.
- [x ] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details).


DevAlvaroF:
@conda-forge/help-python  Ready for review!

recipes/hypso/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - pip
+    - setuptools >=61.0
+    - setuptools_scm  >=3.4
+  run:
+    - python >=3.9
+    - numpy >=1.24
+    - gdal =3.6.2
+    - cartopy >=0.21.1
+    - pandas
+    - matplotlib
```
ocefpaf, line 30: Unless you explicitly needs qt consider using just matplotlib-base instead of the heavier package with the qt bindings.

```suggestion
    - matplotlib-base
```

DevAlvaroF: I did the modifications in the meta.yaml and the pyproject.toml (https://github.com/NTNU-SmallSat-Lab/hypso-package/blob/v1.0.1/pyproject.toml) but got this error from the test in the meta.yml file. Any ideas? 

```
import: 'hypso'
import: 'hypso'
+ pip check
hypso 1.0.1 requires matplotlib-base, which is not installed.
```

ocefpaf: @DevAlvaroF there are some misconceptions here you should not change the name in the Python package metadata. There is no such thing as "matplotlib-base" in the Python world.

Conda is an agnostic package manager and, in the world of diverse packages, we can break them down to make them easier to install. In this case, the python package named matplotlib was broken down into the "full matplotlib", that few people really needs it, and the matplotlib-base that removes the qt dependency making it easier to install, lighter, and taking less space. B/c virtually everyone uses notebooks theses days, installing qt to have matplotlib is not desired. 

TL;DR you don't need a new release or any change upstream, just here you need to change matplotlib to matplotlib-base to avoid download qt b/c you probably don't need it.

DevAlvaroF: Thanks for the explanation! I learn something new every time. All tests passed <3

PR #24378, opened 2023-10-30
Adding Medcoupling

Krande:
Checklist
- [x] Build number is 0.


Krande:

Krande:
@conda-forge/help-python-c

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,53 @@
...
+    -DMEDCOUPLING_PARTITIONER_PTSCOTCH=${on_mpi} \
+    -DMPI_C_COMPILER:PATH="$(which mpicc)" \
+    -DPYTHON_EXECUTABLE:FILEPATH="$PYTHON" \
+    -DHDF5_ROOT_DIR="${PREFIX}" \
+    -DSWIG_ROOT_DIR="${PREFIX}" \
+    -DMEDFILE_ROOT_DIR="${PREFIX}" \
+    -DSCOTCH_ROOT_DIR="${PREFIX}" \
+    -DMETIS_ROOT_DIR="${PREFIX}" \
+    -DPARMETIS_ROOT_DIR="${PREFIX}" \
+    -DCMAKE_PREFIX_PATH="${PREFIX}" \
```
hmaarrfk, line 49: Check the logs for the contents of `CMAKE_ARGS` many of these might be duplicated.

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,53 @@
...
+    export FCFLAGS="-g -O0 ${FCFLAGS}"
+    build_type="Debug"
+else
+    build_type="Release"
+    echo "Debugging Disabled"
+fi
+
+cmake .. \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DPYTHON_ROOT_DIR="${PREFIX}" \
```
hmaarrfk, line 26: Try to keep the variables that start with `PYTHON` together.

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,53 @@
...
+if [[ "${PKG_DEBUG}" == "True" ]]; then
+    echo "Debugging Enabled"
+    export CFLAGS="-g -O0 ${CFLAGS}"
+    export CXXFLAGS="-g -O0 ${CXXFLAGS}"
+    export FCFLAGS="-g -O0 ${FCFLAGS}"
+    build_type="Debug"
+else
+    build_type="Release"
+    echo "Debugging Disabled"
+fi
```
hmaarrfk, line 22: Please remove this.

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
+export CLICOLOR_FORCE=1
```
hmaarrfk, line 2: Please remove this

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - "{{ compiler('c') }}"
+    - "{{ compiler('cxx') }}"
+    - make                                      # [not win]
+    - ninja                                     # [win]
+    - cmake
+    - swig >=4.0.2
+
+  host:
+    - python
+    - numpy >=1.22
```
hmaarrfk, line 38:
```suggestion
    - numpy
```

Numpy will build with an "old numpy" to ensure forward compatibility

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - boost
+    - libxml2
+    - astor
+    - pybind11-stubgen
+    - libmed 4.1.1
+    - petsc 3.17.1                              # [mpi != 'nompi']
```
hmaarrfk, line 44:
```suggestion
    - libmed
    - petsc                               # [mpi != 'nompi']
```

recipes/medcoupling/stubs/.gitignore
```diff
@@ -0,0 +1 @@
+.stubs/
```
hmaarrfk, line 1: Please remove this file.

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,41 @@
...
+    -D CMAKE_PREFIX_PATH="%LIBRARY_PREFIX%" ^
+    %CMAKE_ARGS%
+
+if errorlevel 1 exit 1
+ninja
+if errorlevel 1 exit 1
+ninja install
+
+:: Generate pyi stubs for binary module
+%PYTHON% %RECIPE_DIR%/stubs/custom_stubs_gen.py
```
hmaarrfk, line 40: Why does this get called for windows but not unix.

recipes/medcoupling/LICENSE.txt
```diff
@@ -0,0 +1,458 @@
+                  GNU LESSER GENERAL PUBLIC LICENSE
```
hmaarrfk, line 1: Is this file included in the distribution? if so, just refer to it directly.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,73 @@
+{% set name = "medcoupling" %}
+{% set version = "9.10.0" %}
+{% set petsc_ver = "3.17.1" %}
```
hmaarrfk, line 3:
```suggestion
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    folder: deps/config
+    sha256: 53cbdf2def53e32b2237e948b7dbe8e4051e69e96319a786cd51aba0c915703d
+
+
+build:
+  detect_binary_files_with_prefix: true
+  # string: aster_py{{ CONDA_PY }}_mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}   # [mpi != "nompi"]
+  # string: aster_py{{ CONDA_PY }}_nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}           # [mpi == "nompi"]
+  number: 0
+  skip: true  # [py2k or osx or py>=312 or win]
```
hmaarrfk, line 25:
```suggestion
  skip: true  # [osx or win]
```

hmaarrfk: Python 3.12 support should be trivial to add.

py2k has been dropped outright from conda-forge

recipes/medcoupling/LICENSE.txt
Krande: Unfortunately no. As far as I've seen it is only specified inside the top of the CMAKELists.txt file.

recipes/medcoupling/bld.bat
Krande: oh, then that's a mistake on my part. The stub generator isn't quite finished yet. I'll remove it and the related dependencies for now and return to it once I've made more progress on it

recipes/medcoupling/meta.yaml
hmaarrfk: If you will include a `bld.bat` file, then the windows build should not be skipped.

Lets get windows fixed before you merge.


If you want to skip windows, then delete the bld.bat file (but that seems like more work)

recipes/medcoupling/meta.yaml
Krande: Sure, we can do this first. And if we later find that Code_Aster fails with these changes we'll deal with it then.

recipes/medcoupling/bld.bat
hmaarrfk: Honestly, this is the kind of feature that "seems nice for all" but I am not sure it should be added until attempts to reach out to upstream are.

There is a slippery slope between "adding a nice feature" and "forking" a package.

If that is ok, please remove the stub generator at this stage, and make attempts to reach out to upstream and contribute it.

Otherwise, maybe live without type hints for this package?

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+test:
+  imports:
+    - medcoupling
+
+about:
+  home: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=summary
+  license: LGPL-2.1-or-later
+  license_family: GPL
+  license_file: LICENSE.txt
```
hmaarrfk, line 63:
```suggestion
  license_file: COPYING
```

recipes/medcoupling/LICENSE.txt
hmaarrfk: The file is called `COPYING`

hmaarrfk: see suggestion below and delete this file please.

recipes/medcoupling/build.sh
Krande: Good tip! I managed to remove all ROOT_DIR reference

Krande:
@hmaarrfk I really appreciate the review! Also I believe your suggestions and comments should be implemented/answered by now. I tested this locally and it worked, so hopefully the CI agrees.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    folder: deps/config
+    sha256: 53cbdf2def53e32b2237e948b7dbe8e4051e69e96319a786cd51aba0c915703d
+
+
+build:
+  detect_binary_files_with_prefix: true
+  # string: aster_py{{ CONDA_PY }}_mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}   # [mpi != "nompi"]
+  # string: aster_py{{ CONDA_PY }}_nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}           # [mpi == "nompi"]
+  number: 0
+  skip: true  # [osx or win]
```
hmaarrfk, line 24:
```suggestion
  skip: true  # [osx]
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - ninja                                     # [win]
+    - cmake
+    - swig >=4.0.2
+
+  host:
+    - python
+    - numpy
+    - boost
+    - libxml2
+    - libmed
```
hmaarrfk, line 40: Lets wait for https://github.com/conda-forge/libmed-feedstock/pull/50 to get merged

Krande: agreed. Let me know if I need to do anything else

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+{% set version = "9.10.0" %}
+
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  - fn: 'medcoupling_{{ version }}.tar.gz'
+    url: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=snapshot;h=fe2e38d301902c626f644907e00e499552bb2fa5;sf=tgz # 9.10
```
hmaarrfk, line 11:
```suggestion
    url: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=snapshot;h=V{{ version.replace('.', '_') }} ;sf=tgz
```

hmaarrfk: similarly below

Krande: Awesome! Thanks for the tip 👍

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+
+package:
+  name: '{{ name|lower }}'
+  version: '{{ version }}'
+
+source:
+  - fn: 'medcoupling_{{ version }}.tar.gz'
+    url: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=snapshot;h=V{{ version.replace('.', '_') }};sf=tgz
+    sha256: e727bf8b8bfe848a0e80605e269b8be818bebf9ee098cbb5d6168a920b210c46
```
hmaarrfk, line 12:
```suggestion
    sha256: 02087a4e59ffcdab6fc68460ffd86b03a80b5c0faa695b0bfb2142f1d0d3e5ea
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - libmed
+    - petsc                                     # [mpi != 'nompi']
+    - {{ mpi }}                                 # [mpi != 'nompi']
+
+test:
+  imports:
+    - medcoupling
+
+about:
+  home: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=summary
```
hmaarrfk, line 58:
```suggestion
  home: https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  license: LGPL-2.1-or-later
+  license_family: GPL
```
hmaarrfk, line 60:
```suggestion
  license_family: LGPL
```

Krande:

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,37 @@
+mkdir build
+cd build
+
+:: -fdefault-integer-8 is not supported on msvc. Should consider moving to clang/msys2
+set FCFLAGS=-fdefault-integer-8 %FCFLAGS%
+set FFLAGS=-fdefault-integer-8 %FFLAGS%
+set CMAKE_CXX_FLAGS=-fvisibility=hidden %CMAKE_CXX_FLAGS%
+
+cmake -G "Ninja" %SRC_DIR% ^
```
hmaarrfk, line 9:
```suggestion
cmake -G "Ninja" .. ^
```

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,37 @@
...
+    -D MEDCOUPLING_INSTALL_PYTHON=%SP_DIR% ^
+    -D XDR_INCLUDE_DIRS="" ^
+    -D MEDCOUPLING_PARTITIONER_PARMETIS=OFF ^
+    -D MEDCOUPLING_PARTITIONER_METIS=OFF ^
+    -D MEDCOUPLING_PARTITIONER_SCOTCH=OFF ^
+    -D MEDCOUPLING_PARTITIONER_PTSCOTCH=OFF ^
+    -D HDF5_ROOT_DIR="%LIBRARY_PREFIX%" ^
+    -D SWIG_ROOT_DIR="%LIBRARY_PREFIX%" ^
+    -D MEDFILE_ROOT_DIR="%LIBRARY_PREFIX%" ^
+    -D CMAKE_PREFIX_PATH="%LIBRARY_PREFIX%" ^
```
hmaarrfk, line 31:
```suggestion
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+    folder: deps/config
+    sha256: d58f6749514ce49b00590a96d6dc9090a76a07c13690e2453403643afb40346c
+
+
+build:
+  detect_binary_files_with_prefix: true
+  # string: aster_py{{ CONDA_PY }}_mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}   # [mpi != "nompi"]
+  # string: aster_py{{ CONDA_PY }}_nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}           # [mpi == "nompi"]
+  number: 0
+  skip: true  # [osx]
```
hmaarrfk, line 24:
```suggestion
```

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+requirements:
+  build:
+    - "{{ compiler('c') }}"
+    - "{{ compiler('cxx') }}"
```
hmaarrfk, line 29:
```suggestion
    - {{ compiler('c') }}
    - {{ compiler('cxx') }}
```

hmaarrfk (commented):
nothing really jumps out immediately...

just a few nits.

hmaarrfk:
Maybe we have to wait for https://github.com/conda-forge/libmed-feedstock/pull/52

hmaarrfk:
```
 error: cannot convert 'int*' to 'med_int*' {aka 'long int*'}
```

hmaarrfk:
Linux:

```
error: cannot convert 'int*' to 'med_int*' {aka 'long int*'}
```

OSX:
```
2023-11-19T19:11:19.1570370Z -- Python libraries: /System/Library/Frameworks/Python.framework/Versions/2.7/lib/libpython3.9.dylib
2023-11-19T19:11:19.1626160Z -- Python include dir: /System/Library/Frameworks/Python.framework/Versions/2.7/include/python3.9
2023-11-19T19:11:19.1718750Z -- Found PythonLibs: /System/Library/Frameworks/Python.framework/Versions/2.7/lib/libpython3.9.dylib (Required is at least version "3.9.18") 
2023-11-19T19:11:19.1779540Z CMake Error at deps/config/cmake/FindSalomePythonLibs.cmake:66 (SALOME_EXTRACT_VERSION):
2023-11-19T19:11:19.1813370Z   SALOME_EXTRACT_VERSION Macro invoked with incorrect arguments for macro
2023-11-19T19:11:19.1865650Z   named: SALOME_EXTRACT_VERSION
2023-11-19T19:11:19.1903800Z Call Stack (most recent call first):
2023-11-19T19:11:19.1936160Z   CMakeLists.txt:201 (FIND_PACKAGE)
2023-11-19T19:11:19.1985840Z 
```


Windows:
```
%SRC_DIR%\build\src\MEDCoupling_Swig\CMakeFiles\_MEDCouplingCompat.dir\MEDCouplingPYTHON_wrap.cxx : fatal error C1128: number of sections exceeded object file format limit: compile with /bigobj
```

hmaarrfk:
I'm not sure what the error on windows is, they have `bigobj` enabled on line 109
https://git.salome-platform.org/gitweb/?p=tools/medcoupling.git;a=blob;f=src/PyWrapping/CMakeLists.txt;h=a897bb63dc53405546e114d74e71922853debc68;hb=HEAD#l109

As for the `MED_INT_IS_LONG` I thought that was supposed to help, but it seems like the following still appears in the CMake error log:
```
-- MEDFile: size of med_int is UNKNOWN
```

it is alot easier (for me) to debug on a local machine when you have a way to compile things.

hmaarrfk:
I think you might need to patch up https://git.salome-platform.org/gitweb/?p=tools/configuration.git;a=blob;f=cmake/FindSalomeMEDFile.cmake;h=960291aa92f1c14399345bbf80a5de7eb7feb8bf;hb=HEAD#l49 a little bit

hmaarrfk:
I'm not too sure where the SYSROOT stuff is defined. there has been this clash between `lib` and `lib64` for a long time.

I'm generally not sure where to look, but I feel like it should be fixed at a configuration level, and not by using a symlink.

Krande:
@hmaarrfk Thank you again for your invaluable supportI 


> it is alot easier (for me) to debug on a local machine when you have a way to compile things.

You mean a good local setup for debugging conda cmake packages such as medcoupling? Yeah, I am still trying to find a good compilation debugging workflow when I working on conda packaging. If you have any suggestions for good workflows for local debugging I am all ears?

My first steps is usually to do `conda mambabuild .` to compile locally (I have access to both windows and linux machines). But usually I find that a bit cumbersome (and I don't see an easy way of creating patches that way). So I find it easier to compile using an IDE (I use primarily CLION).

For `medcoupling` I tried making github copies of the relevant repo's [medcoupling-copy](https://github.com/Krande/medcoupling-copy) and a [medcoupling-config-copy](https://github.com/Krande/medcoupling-config-copy). And in the `medcoupling-copy` I made a "conda/dev" branch that contains a `CMakePresets.json` a `conda_developer.md` file with instructions for installing the pre-requisites in a conda environment. I then use a `.env.json` file to store user-specific paths to `medcoupling-config-copy` and the conda env containing the pre-requisites. 

Does this sound similar to the workflow you use (you seem be creating patches quite easily in your setup)?

hmaarrfk:
are you on windows or Linux?

Krande:
The screenshot describing the error I pasted in my previous comment was from windows (which is my primary platform).

I also have a Linux machine which when I tried failed with a missing libpthread.so error (which I reckon is related to the sysroot issue you are describing?) right before it would start to compile

hmaarrfk:
1. Install Visual Studio 2019 (with the C and C++ compiler -- about 20 GB)
2. run `conda debug recipes/medcoupling` inside this repo
3. follow instructions to activate the test environment
4. Inside the `conda-bld.bat` that is generated in the `work` folder, remove the `if error exit 1`
5. then run `call conda-bld.bat`

carterbox:
Threads (pthread) is a target that [CMake can locate for you](https://cmake.org/cmake/help/latest/module/FindThreads.html) with find_package.

Krande:
@carterbox Thanks for the tip!

I have managed to recreate the linux issues locally in my IDE (clion) using the conda build environment.

Using `find_package(Threads REQUIRED)` seems to find something, or at least when I am printing 

```
message(STATUS "CMAKE_THREAD_LIBS_INIT=${CMAKE_THREAD_LIBS_INIT}")
message(STATUS "CMAKE_USE_WIN32_THREADS_INIT=${CMAKE_USE_WIN32_THREADS_INIT}")
```

I get

```
-- CMAKE_THREAD_LIBS_INIT=-lpthread
-- CMAKE_USE_WIN32_THREADS_INIT=
```

However, I'm unsure where to go from here.

I tried unsuccessfully to add the library to the target linking of `medloader` ([original source](https://github.com/Krande/medcoupling-copy/blob/master/src/MEDLoader/CMakeLists.txt#L96)) like this

```
TARGET_LINK_LIBRARIES(medloader medcouplingcpp ${MEDFILE_C_LIBRARIES} ${HDF5_LIBRARIES} ${MEDLOADER_XDR_LIBRARIES} ${CMAKE_THREAD_LIBS_INIT})
```
which yielded the same error

```yaml
    kind: "try_run-v1"
    backtrace:
      - "config/cmake/FindSalomeMEDFile.cmake:39 (TRY_RUN)"
      - "CMakeLists.txt:142 (FIND_PACKAGE)"
    directories:
      source: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp"
      binary: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp"
    cmakeVariables:
      CMAKE_CXX_FLAGS: " -fno-operator-names"
      CMAKE_MODULE_PATH: "/home/kristoffer/code/medcoupling-copy/config/cmake"
      CMAKE_WARN_DEPRECATED: "FALSE"
    buildResult:
      variable: "_med_int_compile_results"
      cached: true
      stdout: |
        Change Dir: '/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp'
        
        Run Build Command(s): /home/kristoffer/miniforge3/envs/medc-deps/bin/ninja -v cmTC_61cbb
        ninja: error: '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/libmed_1703697193818/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so', needed by 'cmTC_61cbb', missing and no known rule to make it
        
      exitCode: 1
    runResult:
      variable: "_med_int_run_result"
      cached: true
```

Which is kind of strange, seeing that the sysroot path mentioned in the ninja error is the build path from the `libmed` compilation I made locally. If I use the libmed from conda-forge the sysroot path is the ci azure deveops build runner path. Any idea's what might be causing this?


I also noticed that for linux during the cmake build step there is an error with finding the HDF5 libraries 

`-- HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.`

[source](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=850296&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=261)

When re-running this locally I checked the debug logs and found this:

```yaml
 -
    kind: "try_compile-v1"
    backtrace:
      - "/home/kristoffer/.cache/JetBrains/RemoteDev/dist/c041d69a96617_CLion-2023.3.2/bin/cmake/linux/x64/share/cmake-3.27/Modules/FindHDF5.cmake:238 (try_compile)"
      - "/home/kristoffer/.cache/JetBrains/RemoteDev/dist/c041d69a96617_CLion-2023.3.2/bin/cmake/linux/x64/share/cmake-3.27/Modules/FindHDF5.cmake:608 (_HDF5_test_regular_compiler_C)"
      - "config/cmake/FindSalomeHDF5.cmake:40 (FIND_PACKAGE)"
      - "CMakeLists.txt:141 (FIND_PACKAGE)"
    directories:
      source: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeScratch/TryCompile-9bEq6t"
      binary: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeScratch/TryCompile-9bEq6t"
    cmakeVariables:
      CMAKE_C_FLAGS: ""
      CMAKE_MODULE_PATH: "/home/kristoffer/code/medcoupling-copy/config/cmake"
      CMAKE_WARN_DEPRECATED: "FALSE"
    buildResult:
      variable: "HDF5_C_COMPILER_NO_INTERROGATE"
      cached: true
      stdout: |
        Change Dir: '/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeScratch/TryCompile-9bEq6t'
        
        Run Build Command(s): /home/kristoffer/miniforge3/envs/medc-deps/bin/ninja -v cmTC_f01ac
        [1/2] /home/kristoffer/miniforge3/envs/medc-deps/bin/cc    -o CMakeFiles/cmTC_f01ac.dir/cmake_hdf5_test.c.o -c /home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/hdf5/cmake_hdf5_test.c
        FAILED: CMakeFiles/cmTC_f01ac.dir/cmake_hdf5_test.c.o 
        /home/kristoffer/miniforge3/envs/medc-deps/bin/cc    -o CMakeFiles/cmTC_f01ac.dir/cmake_hdf5_test.c.o -c /home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/hdf5/cmake_hdf5_test.c
        /home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/hdf5/cmake_hdf5_test.c:1:10: fatal error: hdf5.h: No such file or directory
            1 | #include <hdf5.h>
              |          ^~~~~~~~
        compilation terminated.
        ninja: build stopped: subcommand failed.
        
      exitCode: 1
  -
    kind: "message-v1"
    backtrace:
      - "/home/kristoffer/.cache/JetBrains/RemoteDev/dist/c041d69a96617_CLion-2023.3.2/bin/cmake/linux/x64/share/cmake-3.27/Modules/FindHDF5.cmake:360 (message)"
      - "/home/kristoffer/.cache/JetBrains/RemoteDev/dist/c041d69a96617_CLion-2023.3.2/bin/cmake/linux/x64/share/cmake-3.27/Modules/FindHDF5.cmake:658 (_HDF5_invoke_compiler)"
      - "config/cmake/FindSalomeHDF5.cmake:39 (FIND_PACKAGE)"
      - "CMakeLists.txt:141 (FIND_PACKAGE)"
    message: |
      HDF5 C compiler wrapper is unable to compile a minimal HDF5 program.
      
      /home/kristoffer/miniforge3/envs/medc-deps/bin/h5cc: 1: eval: x86_64-conda-linux-gnu-cc: not found
  -
```

I have tried a few different ways of defining the HDF5 include paths in order to get the compilation to find the hdf5.h (which I've made sure exists in the `<CONDA_PREFIX>/include` directory. I'll try some more in case this is 

In the last error regarding `MED_INT` I checked the build logs and found that the test script fails due to not finding pthread 

```yaml
    kind: "try_run-v1"
    backtrace:
      - "config/cmake/FindSalomeMEDFile.cmake:39 (TRY_RUN)"
      - "CMakeLists.txt:142 (FIND_PACKAGE)"
    directories:
      source: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp"
      binary: "/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp"
    cmakeVariables:
      CMAKE_CXX_FLAGS: " -fno-operator-names"
      CMAKE_MODULE_PATH: "/home/kristoffer/code/medcoupling-copy/config/cmake"
      CMAKE_WARN_DEPRECATED: "FALSE"
    buildResult:
      variable: "_med_int_compile_results"
      cached: true
      stdout: |
        Change Dir: '/home/kristoffer/code/medcoupling-copy/cmake-build-gcc-debug/CMakeFiles/CMakeTmp'
        
        Run Build Command(s): /home/kristoffer/miniforge3/envs/medc-deps/bin/ninja -v cmTC_283e5
        ninja: error: '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/libmed_1703697193818/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so', needed by 'cmTC_283e5', missing and no known rule to make it
        
      exitCode: 1
    runResult:
      variable: "_med_int_run_result"
      cached: true
```

carterbox:
Can you please try using the `Threads::Threads` CMake target instead of adding `-lpthreads` to the target libraries.

If you need hdf5, then why is hdf5 not listed in the host requirements of this recipe?

carterbox:
Conda build sets up its own environments for the build. If you're seeing those environments in your build logs from your own builds in a different environment it's probably because you haven't cleared the CMake cache and you're sharing source directory with conda-build.

Krande:
Thanks for getting back to me so quickly, @carterbox!

The `hdf5` library is indirectly imported by the `libmed` package (which acts like a wrapper lib for hdf5). 

Regarding the ninja build error referring to a non-existent libpthread path from the `libmed` dependency.

I am actually seeing the same issue on the conda forge linux build runner: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=850319&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=490

```
make[2]: *** No rule to make target '/home/conda/feedstock_root/build_artifacts/libmed_1700316022914/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so', needed by 'src/MEDLoader/libmedloader.so'.  Stop.
make[2]: *** Waiting for unfinished jobs....
```

When I do a local re-running of this recipe using `conda mambabuild . --python=3.11 --numpy=1.25` this is what is reported in `CMakeConfigureLog.yaml`:

```yaml
  -
    kind: "try_compile-v1"
    backtrace:
      - "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/share/cmake-3.28/Modules/FindHDF5.cmake:238 (try_compile)"
      - "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/share/cmake-3.28/Modules/FindHDF5.cmake:608 (_HDF5_test_regular_compiler_C)"
      - "deps/config/cmake/FindSalomeHDF5.cmake:39 (FIND_PACKAGE)"
      - "CMakeLists.txt:140 (FIND_PACKAGE)"
    directories:
      source: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeScratch/TryCompile-SiddKl"
      binary: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeScratch/TryCompile-SiddKl"
    cmakeVariables:
      CMAKE_C_FLAGS: "-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work=/usr/local/src/conda/medcoupling-9.10.0 -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix"
      CMAKE_MODULE_PATH: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/deps/config/cmake"
      CMAKE_WARN_DEPRECATED: "FALSE"
    buildResult:
      variable: "HDF5_C_COMPILER_NO_INTERROGATE"
      cached: true
      stdout: |
        Change Dir: '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeScratch/TryCompile-SiddKl'
        
        Run Build Command(s): /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/cmake -E env VERBOSE=1 /usr/bin/gmake -f Makefile cmTC_d53a8/fast
        /usr/bin/gmake  -f CMakeFiles/cmTC_d53a8.dir/build.make CMakeFiles/cmTC_d53a8.dir/build
        gmake[1]: Entering directory '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeScratch/TryCompile-SiddKl'
        Building C object CMakeFiles/cmTC_d53a8.dir/cmake_hdf5_test.c.o
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/x86_64-conda-linux-gnu-cc   -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work=/usr/local/src/conda/medcoupling-9.10.0 -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix  -o CMakeFiles/cmTC_d53a8.dir/cmake_hdf5_test.c.o -c /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/hdf5/cmake_hdf5_test.c
        Linking C executable cmTC_d53a8
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/cmake -E cmake_link_script CMakeFiles/cmTC_d53a8.dir/link.txt --verbose=1
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/x86_64-conda-linux-gnu-cc -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work=/usr/local/src/conda/medcoupling-9.10.0 -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix  -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib -Wl,-rpath-link,/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib -L/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/lib  -rdynamic CMakeFiles/cmTC_d53a8.dir/cmake_hdf5_test.c.o -o cmTC_d53a8 
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/13.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: CMakeFiles/cmTC_d53a8.dir/cmake_hdf5_test.c.o: in function `main':
        cmake_hdf5_test.c:(.text.startup.main+0x15): undefined reference to `H5check_version'
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/13.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: cmake_hdf5_test.c:(.text.startup.main+0x1b): undefined reference to `H5open'
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/../lib/gcc/x86_64-conda-linux-gnu/13.2.0/../../../../x86_64-conda-linux-gnu/bin/ld: cmake_hdf5_test.c:(.text.startup.main+0x31): undefined reference to `H5Fcreate'
        gmake[1]: *** [CMakeFiles/cmTC_d53a8.dir/build.make:99: cmTC_d53a8] Error 1
        gmake[1]: Leaving directory '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeScratch/TryCompile-SiddKl'
        gmake: *** [Makefile:127: cmTC_d53a8/fast] Error 2
        
      exitCode: 2
  -
    kind: "try_run-v1"
    backtrace:
      - "deps/config/cmake/FindSalomeMEDFile.cmake:39 (TRY_RUN)"
      - "CMakeLists.txt:141 (FIND_PACKAGE)"
    directories:
      source: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp"
      binary: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp"
    cmakeVariables:
      CMAKE_CXX_FLAGS: "-fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work=/usr/local/src/conda/medcoupling-9.10.0 -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix -fno-operator-names"
      CMAKE_MODULE_PATH: "/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/deps/config/cmake"
      CMAKE_WARN_DEPRECATED: "FALSE"
    buildResult:
      variable: "_med_int_compile_results"
      cached: true
      stdout: |
        Change Dir: '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp'
        
        Run Build Command(s): /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/cmake -E env VERBOSE=1 /usr/bin/gmake -f Makefile cmTC_071f6/fast
        /usr/bin/gmake  -f CMakeFiles/cmTC_071f6.dir/build.make CMakeFiles/cmTC_071f6.dir/build
        gmake[1]: Entering directory '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp'
        Building CXX object CMakeFiles/cmTC_071f6.dir/check_med_int_size.cxx.o
        /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_build_env/bin/x86_64-conda-linux-gnu-c++   -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho/include -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work=/usr/local/src/conda/medcoupling-9.10.0 -fdebug-prefix-map=/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placeho=/usr/local/src/conda-prefix -fno-operator-names  -o CMakeFiles/cmTC_071f6.dir/check_med_int_size.cxx.o -c /home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp/check_med_int_size.cxx
        gmake[1]: *** No rule to make target '/home/conda/feedstock_root/build_artifacts/libmed_1700316008636/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so', needed by 'cmTC_071f6'.  Stop.
        gmake[1]: Leaving directory '/home/kristoffer/miniforge3/envs/mamba-build/conda-bld/medcoupling_1703840036426/work/build/CMakeFiles/CMakeTmp'
        gmake: *** [Makefile:127: cmTC_071f6/fast] Error 2
        
      exitCode: 2
    runResult:
      variable: "_med_int_run_result"
      cached: true
```

I am not sure if the preceding hdf5 errors are related to the libpthread issues, but I included them nonetheless.


Any idea what might be going on?

Krande:
@carterbox & @hmaarrfk 

Hey guys! I finally got the linux build working!

For some reason `libmed` installs a couple of cmake files in `$PREFIX/share/cmake/medfile-4.1.1` that overrides the pthread path. By simply removing the `medfile-4.1.1` directory, it successfully installs,

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,40 @@
...
+cd build
+
+if [[ "$mpi" == "nompi" ]]; then
+  on_mpi="OFF"
+else
+  on_mpi="ON"
+fi
+
+# remove the share cmake files
+find ${PREFIX}/share/cmake -type d -name "medfile-*" -exec rm -rf {} +
```
hmaarrfk, line 15: these files are typically important since they include fact about how the medfile library was compiled.

Otherwise, you are just "guessing".

Can we:

* remove the matching files from the source instead?
or 
* fix things in medfile?

Krande: Sure, changing medfile would mean making changes to https://github.com/conda-forge/libmed-feedstock. I'll spend some time tinkering with windows compilation of medcoupling. And after that I can dig into where in libmed I would need to make changes. If you got any ideas where I might start to look I would of course appreciate it :)

Krande: It seems this is where the path is set in the libmed-part:

https://github.com/Krande/med-copy/blob/main/CMakeLists.txt#L104

Krande:
I added a patch that fixes the /bigobj issues on windows. 

I noticed that the compiled `.dll` files where installed in the `%PREFIX%/Library/Lib` directory instead of `%PREFIX%/Library/bin` directory which caused ImportErrors when trying to do `import medcoupling`.

I couldn't immediately find where these paths are set within the cmakelists.txt files. So to fix it I added a command that moves the dll files to the appropriate folder after the compilation.  If you know of any better way of handling this, please let me know.

Here is the list of files moved from the %LIBRARY_LIB% to %LIBRARY_BIN%

```
[5/6] Install the project...
%PREFIX%\Library\lib\interpkernel.dll
%PREFIX%\Library\lib\medcoupling.dll
%PREFIX%\Library\lib\medcouplingremapper.dll
%PREFIX%\Library\lib\medicoco.dll
%PREFIX%\Library\lib\medimport.dll
%PREFIX%\Library\lib\medloader.dll
%PREFIX%\Library\lib\medpartitionercpp.dll
%PREFIX%\Library\lib\renumbercpp.dll
```

hmaarrfk:
in your cmake install commands you need to specify the specific location of the `ARCHIVE` 
https://cmake.org/cmake/help/latest/command/install.html#signatures

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - {{ mpi }}                                 # [mpi != 'nompi']
+
+  run:
+    - python
+    - numpy
+    - boost
+    - libxml2
+    - zlib
+    - hdf5
+    - libmed
```
hmaarrfk, line 54: None of these are required since they should be "exported" as runtime requirements since they were build time requirements.

Can you make a PR to `libmed` to add a `run_export`:
https://conda-forge.org/docs/maintainer/pinning_deps.html#specifying-run-exports

Then finally, we should add a global pinning to libmed by adding the current version of libmed to:
https://conda-forge.org/docs/maintainer/pinning_deps.html#updating-package-pins

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - petsc                                     # [mpi != 'nompi']
```
hmaarrfk, line 55: I venture to say that petsc is also not required here.

Krande: Well, you might be onto something. 

The mpi variant of Medcoupling needs `PTscotch` which currently is dependency of `petsc`. So you are probably right that including PETSc isnt really necessary for `medcoupling`

Krande:
> in your cmake install commands you need to specify the specific location of the `ARCHIVE` https://cmake.org/cmake/help/latest/command/install.html#signatures

Is that in connection with?

> However, I noticed that the compiled .dll files where installed in the %PREFIX%/Library/Lib directory instead of %PREFIX%/Library/bin directory which caused ImportErrors when trying to do import medcoupling.

Krande:
I was just wondering. For the openmpi variants. Should we add?

```
  export CC=mpicc
  export CXX=mpicxx
  export FC=mpif90
  export F77=mpif77
  export F90=mpif90
```

Or is that automatically handled by conda build?

recipes/medcoupling/meta.yaml
hmaarrfk: see suggestion on 54. if that doesn't clear things up, please just say so and I'll try to explain in a different way.

hmaarrfk:
you've only included a nompi variant.

conda-forge tooling does alot, so the best way is to try and check the logs.

recipes/medcoupling/meta.yaml
hmaarrfk: not required on this specific line of yaml because each of the packages above used at build time will forward declare and "inject themselves" into the runtime dependencies

recipes/medcoupling/meta.yaml
Krande: Your comments on 54 helped a lot!

One final thing that I wouldn't mind clarifying regarding `run_export` with respect to which compile time dependencies libmed brings with it to another package using it.

Will adding `run_export` to libmed add the compile time requirements such as `zlib` and `hdf5` to medcoupling only because `zlib` and `hdf5` also have defined `run_export` (and not because they are `host` requirements?).

Or would adding packages without `run_export` to libmed `host` requirements also become compile time dependencies for `medcoupling`. The difference being that any updates of the packages without run_export would NOT automatically trigger a rebuild of medcoupling?

hmaarrfk: not entirely.
see https://github.com/conda-forge/conda-forge.github.io/issues/1880#issuecomment-1408733984

hmaarrfk:
You might as well also remove libmed;)

recipes/medcoupling/meta.yaml
Krande: Okay, I might not fully grasp all aspects of it yet, but I think it is slowly sinking in:) Thank you for your patience!

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - boost
+    - zlib
+    - libxml2
+    - libmed
+    - petsc                                     # [mpi != 'nompi']
+    - {{ mpi }}                                 # [mpi != 'nompi']
+
+  run:
+    - python
+    - numpy
```
hmaarrfk, line 49:
```suggestion
    - {{ pin_compatible('numpy') }}
```

I think numpy finally got run_exports, but I think this still works.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,69 @@
+{% set name = "medcoupling" %}
+{% set version = "9.10.0" %}
+
+
+package:
+  name: '{{ name|lower }}'
```
hmaarrfk, line 6:
```suggestion
  name: medcoupling
```

I think the whole jinja name variable is overused, it leads to too much confusion for no advantage.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,69 @@
+{% set name = "medcoupling" %}
```
hmaarrfk, line 1:
```suggestion
```

recipes/medcoupling/build.sh
hmaarrfk: I still really want to see if we can avoid this line

hmaarrfk:
i mentionned in an old comment that I still want to explore not deleting the cmake files.

recipes/medcoupling/meta.yaml
Krande: Yeah, I see your point. Will update

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    url: https://git.salome-platform.org/gitweb/?p=tools/configuration.git;a=snapshot;h=V{{ version.replace('.', '_') }};sf=tgz
+    folder: deps/config
+    sha256: d58f6749514ce49b00590a96d6dc9090a76a07c13690e2453403643afb40346c
+
+
+build:
+  skip: True  # [osx]
+  detect_binary_files_with_prefix: true
+  # string: aster_py{{ CONDA_PY }}_mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}   # [mpi != "nompi"]
+  # string: aster_py{{ CONDA_PY }}_nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}           # [mpi == "nompi"]
```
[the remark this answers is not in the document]
Krande: yeah, I'm sorry. This is an old debug remnant. No problem with changing this.

recipes/medcoupling/build.sh
Krande: Yes, I'm taking another stab at this now.

So the reason why I ended up deleting the folder, was becuse the following file in the libmed package (`share/cmake/medfile-4.1.1/MEDFileTargets.cmake`) has the following `set_target_properties` : 

```
set_target_properties(medC PROPERTIES
  INTERFACE_LINK_LIBRARIES "/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libhdf5.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libcrypto.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libcurl.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libsz.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/libz.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libdl.so;/home/conda/feedstock_root/build_artifacts/libmed_1705327158179/_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libm.so"
)
```

So the medC paths are (I shortened the paths by replace the main dir and the placeholder stuff with ...)

```
INTERFACE_LINK_LIBRARIES 
"../_h_env_../lib/libhdf5.so;
../_h_env_../lib/libcrypto.so;
../_h_env_../lib/libcurl.so;
../_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so;
../_h_env_../lib/libsz.so;
../_h_env_../lib/libz.so;
../_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libdl.so;
../_build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/libm.so"
```

So the pthread-issue I was seeing before deleting the cmake file was seemingly caused by that the sysroot paths in `MEDFileTargets.cmake` were kept as absolute.

I guess I could simply delete the `MEDFileTargets.cmake` in the libmed as a post-cmake step in the libmed run.sh. 
Or maybe I can replace the absolute paths in `MEDFileTargets.cmake` with $PREFIX and $BUILD_PREFIX/x86_64-conda-linux-gnu/sysroot/usr/lib/libpthread.so as part in the libmed build script?

Any of the 2 options you would prefer? If not I can try to find other alternatives :)

hmaarrfk: if build_env appears in any linked path that is a bug in the libmed recipe.

From what it seems like, it seems that 3 libraries are wrongly linked:

* libpthread.so
* libdl.so
* libm.so

this should be fixed in the libmed feedstock

hmaarrfk: FYI, from the logs, it seems that these are being pulled in by cmake + hdf5. this shouldn't really happen.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - ninja                                     # [win]
+    - cmake
+    - swig >=4.0.2
+
+  host:
+    - python
+    - numpy
+    - boost
+    - zlib
+    - libxml2
```
hmaarrfk, line 41:
```suggestion
    - hdf5
    - libxml2
```
the static analysis on windows is finding a direct dependency betweeh medcoupling.dll and hdf5. this should be declared
```
 ERROR (medcoupling,Library/bin/medimport.dll): Needed DSO Library/bin/hdf5.dll found in ['conda-forge/win-64::hdf5==1.14.2=nompi_h73e8ff5_100']
  ERROR (medcoupling,Library/bin/medimport.dll): .. but ['conda-forge/win-64::hdf5==1.14.2=nompi_h73e8ff5_100'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
 ```

**Comment by hmaarrfk** [comment]
1. Merge from main. We dropped python 3.8 and this version of the CI still tries to run it.
2. Needs: https://github.com/conda-forge/libmed-feedstock/pull/57

File: recipes/medcoupling/build.sh [hunk 31]
```diff
@@ -0,0 +1,40 @@
...
+
+cmake .. \
+    -DCMAKE_BUILD_TYPE="Release" \
+    -DPYTHON_ROOT_DIR="${PREFIX}" \
+    -DPYTHON_EXECUTABLE:FILEPATH="$PYTHON" \
+    -Wno-dev \
+    -DCONFIGURATION_ROOT_DIR="${SRC_DIR}/deps/config" \
+    -DSALOME_CMAKE_DEBUG=ON \
+    -DMED_INT_IS_LONG=ON \
```
hmaarrfk, line 24: This line is in conflict with:

https://github.com/conda-forge/libmed-feedstock/blob/main/recipe/bld.bat#L7C14-L7C14

Krande: Yeah, I think we could add this to libmed cmake build `-DMED_MEDINT_TYPE=long `

Krande: > actually, libmed cmake file has that line, i'm going to try to cleanup a few things in [conda-forge/libmed-feedstock#57](https://github.com/conda-forge/libmed-feedstock/pull/57)

Thanks, I really appreciate your help!

hmaarrfk:
so i feel like somewhere in their code they don't respect the actual value of `med_int`. Thats the only thing I can think of anymore.

I've tried to build locally too in debug mode

```
conda debug medcoupling
```
then you can drop in and inspect errors.

hmaarrfk:
1. Please update from main.
2. Please review https://github.com/conda-forge/libmed-feedstock/pull/58 and try to communicate to upstream that they should specify these "private flags" (There are a few places too where it should be added to the static artifacts, but I didn't care so much since it increases the risk of conflicts in future versions)
3. Make a new build with the above patch
4. Remove the "deletion of the cmake files" 
5. profit (at least on linux)

Krande:
Okay, looking at the `MEDFileTargets.cmake` file it seems the hardcoded paths from medC are fixed! So this should work fine for linux. 

I guess your comment

> so i feel like somewhere in their code they don't respect the actual value of med_int. Thats the only thing I can think of anymore.

is related to the failing windows build? I'll start looking at the failing windows builds now.

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,45 @@
...
+::set FFLAGS=-fdefault-integer-8 %FFLAGS%
+:: set CMAKE_CXX_FLAGS=-fvisibility=hidden %CMAKE_CXX_FLAGS%
+
+cmake -G "Ninja" .. ^
+    -Wno-dev ^
+    -D CMAKE_BUILD_TYPE="Release" ^
+    -D PYTHON_ROOT_DIR="%PREFIX%" ^
+    -D PYTHON_EXECUTABLE:FILEPATH="%PYTHON%" ^
+    -D CONFIGURATION_ROOT_DIR="%SRC_DIR%/deps/config" ^
+    -D SALOME_CMAKE_DEBUG=ON ^
```
hmaarrfk, line 17:
```suggestion
    -D SALOME_CMAKE_DEBUG=ON ^
   -D MED_INT_IS_LONG=ON ^
```

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,46 @@
...
+   -D MED_INT_IS_LONG=ON ^
```
hmaarrfk, line 18:
```suggestion
    -D MED_INT_IS_LONG=ON ^
```

recipes/medcoupling/bld.bat
hmaarrfk: I really thought it was this one...

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,46 @@
...
+ninja
+if errorlevel 1 exit 1
+ninja install
+if errorlevel 1 exit 1
+
+:: Move dll files from %PREFIX%/Library/Lib to %PREFIX%/Library/Bin
+:: This is needed for the python bindings to work
+
+cd %LIBRARY_LIB%
+move *.dll %LIBRARY_BIN%
```
hmaarrfk, line 46: while not related to this directly, please try to patch the appropriate locations (INSTALL commands) with the ARCHIVE  location. this is documented in cmake and will make it "more correct" for windows.

hmaarrfk:
Was it your intention have int as the type on windows. 

I kinda changed that in 

https://github.com/conda-forge/libmed-feedstock/pull/57/files#diff-54cf74e113dd3f6d11e092fdb1d888ec82c69bdafbb15cfb6570c83ecad28f33R20

Krande:
> Was it your intention have int as the type on windows. 
> 
> I kinda changed that in 
> 
> https://github.com/conda-forge/libmed-feedstock/pull/57/files#diff-54cf74e113dd3f6d11e092fdb1d888ec82c69bdafbb15cfb6570c83ecad28f33R20

Well windows support for libmed and medcoupling is uncharted waters for me. 

We could keep the windows version at 32 bit integer support, but I guess we should probably be consistent and change that for libmed as well.

I will jump on a windows computer in just a few hours and try to dig a bit deeper on the 64 bit issues we're seeing on windows. I'll keep you updated

hmaarrfk:
I mean the whole int ba long is quite a huge change. So I tried to keep things “the same”. 

Good luck.

Krande:
Hey @hmaarrfk, just letting you know that I will resume working on this now (I have been busy with other tasks at work for the past weeks). I'll try to finalize the windows support ASAP.

Krande:
Hey @hmaarrfk, I ended up using 32 bit integers for the windows variant for both libmed (build 11) and medcoupling. Now all checks are passing for windows and Linux (we can sort osx support at a later stage after the PR is merged). 

Is there anything else you wanted me to do before merging the PR?

hmaarrfk:
I noticed @jschueller had opened https://github.com/conda-forge/staged-recipes/pull/23258

maybe you want to comment?

hmaarrfk:
@Krande I must apologize and admit that seeing typedefs change based on compilation options triggered horror stories I exprienced years ago when building some of my first software executables.

To be honest, conda(-forge) does not have a good to track these. We can understand "version numbers" quite well, but not "different compilation options".

So i would have two requests:

1. Are you sure you are OK with your windows compilations options?
2. Can you make any packages that were compiled with incompatible compilations as broken following https://conda-forge.org/docs/maintainer/updating_pkgs/#removing-broken-packages

Krande:
> @Krande I must apologize and admit that seeing typedefs change based on compilation options triggered horror stories I exprienced years ago when building some of my first software executables.
> 
> To be honest, conda(-forge) does not have a good to track these. We can understand "version numbers" quite well, but not "different compilation options".
> 


I understand your concerns. And I will make it a priority of keeping the conda-forge `libmed` and `medcoupling` packaging as simple as possible (ie. no extra variants). If there are any additional variants needed, I can upload those variants to a dedicated conda channel (either my "krande" channel or create a "code_aster" channel if needed).


> So i would have two requests:
> 
> 1. Are you sure you are OK with your windows compilations options?
> 2. Can you make any packages that were compiled with incompatible compilations as broken following https://conda-forge.org/docs/maintainer/updating_pkgs/#removing-broken-packages

1. To strengthen my confidence in the current compilation options on Windows, I'll make an effort to run all the unittests on windows I can find for both `med` and `medcoupling` to ensure that the code itself runs as expected on windows. If all tests are passing (skipping any "big int" tests if there are any) then I will assume it is OK. I guess all use-cases involving simulations using big integers should be applicable only for linux versions.

2. Yes, definitively. As of now, I do not know of any downstream packages that are relying on the libmed package. So what I can do once `medcoupling` is merged I can follow the steps in https://conda-forge.org/docs/maintainer/updating_pkgs/#removing-broken-packages and remove all the old `libmed` builds. If you want I can tag you in that PR so that you can comment on it.

recipes/medcoupling/swig_big_file_compile_flag.patch
```diff
@@ -0,0 +1,43 @@
...
+<+>UTF-8
+===================================================================
+diff --git a/src/MEDCoupling_Swig/CMakeLists.txt b/src/MEDCoupling_Swig/CMakeLists.txt
+--- a/src/MEDCoupling_Swig/CMakeLists.txt	(revision 5cb0b9f7852dbb4d68059c363cc38377ed5b3ddb)
++++ b/src/MEDCoupling_Swig/CMakeLists.txt	(date 1703937272130)
+@@ -123,6 +123,7 @@
+
+ IF(WIN32 OR CYGWIN)
+   SWIG_LINK_LIBRARIES(${MEDCouling_target_name} ${PYTHON_LIBRARIES} ${PLATFORM_LIBS} medcouplingcpp)
++  TARGET_COMPILE_OPTIONS(medcouplingcpp PRIVATE /bigobj)
```
jschueller, line 31: why not just add the flag to CMAKE_CXX_FLAGS instead of patching ?

Krande: @jschueller Yeah, I agree that would be better :) I can't recall why I ended up making a patch for this, but I'll just re-run this locally with adding the flag now to see that there wasn't a reason why I did end up patching.

recipes/medcoupling/bld.bat
```diff
@@ -0,0 +1,45 @@
+@echo off
+
+mkdir build
+cd build
```
jschueller, line 4: its cleaner to let cmake create the build dir with "cmake -B build"

recipes/medcoupling/build.sh
```diff
@@ -0,0 +1,37 @@
...
+    -DXDR_INCLUDE_DIRS="" \
+    -DMEDCOUPLING_PARTITIONER_PARMETIS=OFF \
+    -DMEDCOUPLING_PARTITIONER_METIS=OFF \
+    -DMEDCOUPLING_PARTITIONER_SCOTCH=OFF \
+    -DMEDCOUPLING_PARTITIONER_PTSCOTCH=${on_mpi} \
+    -DMPI_C_COMPILER:PATH="$(which mpicc)" \
+    ${CMAKE_ARGS}
+
+make -j$CPU_COUNT
+make install
```
jschueller, line 37: here too you can replace this with just `make install -j$COUNT`

recipes/medcoupling/swig_big_file_compile_flag.patch
Krande: Adding `-D CMAKE_CXX_FLAGS="/bigobj" ^` did the same trick as my patch. So I'm removing the patch. Thanks!

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+  host:
+    - python
+    - numpy
+    - boost
+    - zlib
+    - hdf5
+    - libxml2
+    - libmed
+    - petsc                                     # [mpi != 'nompi']
```
jschueller, line 43: it seems that I add metis enabled also in my pr

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+  # string: py{{ CONDA_PY }}_mpi_{{ mpi }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}   # [mpi != "nompi"]
+  # string: py{{ CONDA_PY }}_nompi_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}           # [mpi == "nompi"]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - make                                      # [not win]
+    - ninja                                     # [win]
```
jschueller, line 31: or, use ninja on both ?

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - cmake
+    - swig >=4.0.2
+
+  host:
+    - python
+    - numpy
+    - boost
```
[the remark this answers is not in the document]
Krande: yes, libboost-devel was enough to compile locally. Re-running ci now with libboost-devel now.

recipes/medcoupling/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - {{ mpi }}                                 # [mpi != 'nompi']
+
+  run:
+    - python
+    - {{ pin_compatible('numpy') }}
+    - {{ mpi }}                                 # [mpi != 'nompi']
+
+test:
+  imports:
+    - medcoupling
```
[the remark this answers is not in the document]
Krande: I included your imports which were all successful

recipes/medcoupling/meta.yaml
Krande: I see. As far as I know, METIS when used as a dependency for Code Aster, EDF uses a patched version of METIS that enables support for both 64 and 32 bit integers. Last I checked the conda-forge version of METIS did not have these changes.

As of now we are looking to support 64 bit integers on linux and 32 bit integers on windows.
Given that some of us wish to use `med` and `medcoupling` with Code Aster (see https://github.com/conda-forge/code-aster-feedstock where we currently use static compilation and embed the dependencies directly instead of relying on other conda-forge packages), I suggest we skip adding METIS as a dependency until we are certain that conda-forge versions of METIS contains the patches.

hmaarrfk:
> And I will make it a priority of keeping the conda-forge libmed and medcoupling packaging as simple as possible (ie. no extra variants). 

I think it is entirely ok to have variants. Just they are hard to maintain, and sometimes lead to more mistakes. Its hard to get them right, so its really simpler to not have them!

Thanks for your perseverance here!

recipes/medcoupling/meta.yaml
Krande: For reference, here is the metis-5.1.0 patch file

[metis-5.1.0_aster4.patch](https://github.com/conda-forge/staged-recipes/files/14601523/metis-5.1.0_aster4.patch)

recipes/medcoupling/meta.yaml
hmaarrfk: the pattern of using ninja on windows and make on unix is something quite common accross conda-forge.

I think its fine to leave it.

I've always wondered why "ninja isn't used for both" myself, I chalked it up to "a lighter dependency".

hmaarrfk (approved):

Krande:
It looks like all tests are passing after I made the last changes. @hmaarrfk merge whenever you are ready :)

PR #24408, opened 2023-11-02
Add simple-build-system pkg

tkittel:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/simple-build-system/meta.yaml
+++ b/recipes/simple-build-system/meta.yaml
@@ -13,9 +13,9 @@ source:
 
 build:
   number: 0
-  script: python -m pip install --no-deps --ignore-installed .
+  script: python -m pip install --no-deps -vv .
   skip: True  # [win]
-  skip: True  # [py<38]
+  skip: True  # [py<39]
 
 requirements:
   host:
@@ -26,6 +26,7 @@ requirements:
     - tomli  # [py<311]
     - pybind11 >=2.10.4
     - cmake >=3.24.2
+    - make
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
     - numpy
--- a/recipes/simple-build-system/meta.yaml
+++ b/recipes/simple-build-system/meta.yaml
@@ -21,7 +21,7 @@ requirements:
     - pip
   run:
     - python >=3.8
-    - tomli  # [py<311]
+    - tomli  # can go away when oldest supported python is 3.11+
     - pybind11 >=2.10.4
     - cmake >=3.24.2
     - {{ compiler('c') }}
--- a/recipes/simple-build-system/meta.yaml
+++ b/recipes/simple-build-system/meta.yaml
@@ -3,7 +3,7 @@
 #NB get sha256 with: curl -sL <URL> | openssl sha256
 
 package:
-  name: simple-build-system
+  name: simple-build-system-suite
   version: {{ version }}
 
 source:
@@ -11,32 +11,48 @@ source:
   sha256: {{ sha_value }}
 
 build:
-  noarch: python
   number: 0
-  script: python -m pip install --no-deps --ignore-installed .
-#  skip: True  # [win]
 
-requirements:
-  host:
-    - python
-    - pip
-  run:
-    - python
-    - tomli  # only needed temporarily (ships with py3.11+)
-    - pybind11 >=2.10.4
-    - cmake >=3.24.2
-    - {{ compiler('c') }}
-    - {{ compiler('cxx') }}
-    - numpy
-    - bash  # [unix]
+outputs:
+  - name: simple-build-system-core
+    script: python -m pip install --no-deps --ignore-installed .
+    build:
+      noarch: python
+    requirements:
+      host:
+        - python >=3.8
+        - pip
+      run:
+        - python >=3.8
+    test:
+      imports:
+        _simple_build_system._cli
+      commands:
+        - simplebuild --help
+        - sb --help
+        - unwrapped_simplebuild --help
 
-test:
-  imports:
-    _simple_build_system._cli
-  commands:
-    - simplebuild --help
-    - sb --help
-    - unwrapped_simplebuild --help
+  - name: simple-build-system
+    build:
+      skip: True  # [win]
+    requirements:
+      run:
+        - simple-build-system-core
+        - tomli  # [py<311]
+        - pybind11 >=2.10.4
+        - cmake >=3.24.2
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - numpy
+        - bash  # [unix]
+    test:
+      imports:
+        _simple_build_system._cli
+      commands:
+        - simplebuild --help
+        - sb --help
+        - unwrapped_simplebuild --help
+
 about:
   home: https://github.com/mctools/simplebuild
   summary: A very simple to use build system for scientific projects with C++/Python.
--- a/recipes/simple-build-system/meta.yaml
+++ b/recipes/simple-build-system/meta.yaml
@@ -1,5 +1,6 @@
 {% set version = "0.0.17" %}
-{% set sha_value = "7a5b313aae468078bf088d5065910ecad649ee995b5437f7fb51fe429113fb9f" %}
+{% set sha_value = "ba04e7abf56cd4b827ee79ef085ad34023102e537fcadb366c1bbabb5670b23f" }
+
 #NB get sha256 with: curl -sL <URL> | openssl sha256
 
 package:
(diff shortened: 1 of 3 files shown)


tkittel:
@conda-forge/help-python ready for review! :smile:

tkittel:
Since almost a month have passed, I rebased on main and updated the pkg version. I also took the opportunity to add a few more tests.

Still waiting for a review @conda-forge/help-python :-)

recipes/simple-build-system/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+source:
+  url: https://github.com/mctools/simplebuild/archive/v{{ version }}.tar.gz
+  sha256: {{ sha_value }}
+
+build:
+  number: 0
+  script: python -m pip install --no-deps -vv .
+  skip: True  # [win]
+  skip: True  # [py<39]
```
dhirschfeld, line 18:
```suggestion
build:
  skip: True  # [win or py<39]
  number: 0
  script: python -m pip install --no-deps -vv .
  entry_points:
    - sb = "_simple_build_system._cli:main"
    - unwrapped_simplebuild = "_simple_build_system._cli:unwrapped_main"
    - sbenv = "_simple_build_system._cli:sbenv_main"
    - sbrun = "_simple_build_system._cli:sbrun_main"
```

tkittel:
Thanks @dhirschfeld - I have added the entry_point changes (including a tiny syntax fix).

dhirschfeld:
ping @conda-forge/staged-recipes

recipes/simple-build-system/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+#NB get sha256 with: curl -sL <URL> | openssl sha256
+
+package:
+  name: simple-build-system
+  version: {{ version }}
+
+source:
+  url: https://github.com/mctools/simplebuild/archive/v{{ version }}.tar.gz
+  sha256: {{ sha_value }}
```
xhochy, line 12: Please don't use an indirection here
```suggestion
  sha256: f8135874cfb599f449363ce11ca45c691bbc7ea4407511c90367cc117dff9f96
```

recipes/simple-build-system/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  run:
+    - python
+    - tomli  # [py<311]
+    - pybind11 >=2.10.4
+    - cmake >=3.24.2
+    - make
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - numpy
+    - bash
```
xhochy, line 37: `bash` is mostly given by the system and having this a dependency led to problems with some packages on `osx` for me. I would prefer to omit this.
```suggestion
```

recipes/simple-build-system/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - tomli  # [py<311]
+    - pybind11 >=2.10.4
+    - cmake >=3.24.2
+    - make
+    - {{ compiler('c') }}
```
xhochy, line 34: Not that this will bind to the compiler version and will need a rebuild for every compiler release.

xhochy (changes requested):

recipes/simple-build-system/meta.yaml
tkittel: Ok, done!

recipes/simple-build-system/meta.yaml
tkittel: Ok, I removed it. I hope real bash is always available on both osx and ubuntu then.

recipes/simple-build-system/meta.yaml
tkittel: Yes, that is unfortunate, but given that we do need C/C++ compilers as a runtime dependency, what is the alternative? Is there already some meta-package available that I don't know about which we could rely on instead to solve this? If not, then I think the dependency will have to stay as it is unfortunately.

tkittel:
@xhochy thanks for the review! I have implemented the two straight-forward requests.

Concerning the `run:` dependency on `{{ compiler('c') }}` and `{{ compiler('cxx') }}`. Would it perhaps be more appropriate, and solve the issue, if we had a dependency on `c-compiler` and `cxx-compiler` ?

Background info: The compilers are not required at all to build the `simple-build-system` package (which is pure python + noarch data files), but 100% of users of the package are going to need a C and C++ compiler.

tkittel:
@xhochy are you happy with the status now, or do you wish me to change the runtime dependencies from `{{ compiler('c') }}` to `c-compiler` (and likewise for cxx)?

xhochy:
I would actually prefer to omit the compiler macros completely. They need to be mentioned explicitly in downstream recipes anyway for matrix expansion and `run_exports` to work.

tkittel:
@xhochy the problem with removing them completely is then that some users might forget to add them explicitly in their environment .yml files, leading to hard to support issues when the simple-build-system then inadvertently picks up a system compiler which might lead to hard to detect and weird failures. I would like to make 100% sure that all users always gets the correct compiler from conda-forge, to avoid a support hassle.

Of course, if you absolutely demand that I remove them I will. But I was hoping that adding `c-compiler` and `cxx-compiler` as a `run:` dependency, I would be able to get the best of both worlds. And to me, even keeping the current recipe with compilers as jinja functions would be preferable over ripping out all reference to compilers from the `run:` section (keeping in mind that the actual contents of the package are rather tiny).

In any case, I very much appreciate you taking the time to review this, just let me know what you decide :-)

tkittel:
@xhochy please let me know how you want me to proceed with this :-)

tkittel:
Ok @xhochy I caved in and have removed the compiler dependencies of the package. I guess it is not so bad to tell our users to add them in their conda env.yml files :-)

Could it please be merged then? :-)

xhochy:
Back from my Christmas break now and will have a look again. In general, the `compilers` package should not be used inside any package that is used inside a conda-recipe (i.e. it should never end up being pulled in by something). We need projects that explicitly reference the compiler via the compiler macro in the recipe to get the correct `run_exports`. If you want to have a convenience dependency on `compilers`, you should have two outputs here: One to use in a recipe and one for end-users to have in their `environment.yml`.

recipes/simple-build-system/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - tomli  # [py<311]
+    - pybind11 >=2.10.4
+    - cmake >=3.24.2
+    - make
+    - numpy
+
+test:
+  requires:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
```
xhochy, line 38: Can you also test that at least one entry point works?
```suggestion
    - {{ compiler('cxx') }}
    - pip
  commands:
    - pip check
    - sb --help
```

tkittel: The tests in `run_test.sh` actually exercise both the `sb` and `sbenv` entry points. Or did you mean something else?

tkittel: To elaborate: `sb --help` is already in `run_test.sh`, but `pip check` is not. Should I add `pip check` to `run_test.sh` as well?

I am afraid to simply commit your suggestion, since I thought (perhaps wrongly?) that one could either specify test commands in `meta.yaml` *or* have a `run_test.sh`, but not both?

tkittel:
Thanks @xhochy . I have answered about entry points in a comment above. As for having multiple outputs, that is something was considering. But at this point, it is perhaps just a needless complication - after having thought about it some more, I am getting more comfortable about not having `compilers` as a dependency. Most of our direct users will anyway paste the recommended conda env.yml from the documentation, so we can simply add them there :-)

recipes/simple-build-system/meta.yaml
tkittel: I have now pushed a change which adds `pip check` to run_test.sh as well, and which removes the duplicated (and I guess misleading) import line from `meta.yaml`. Instead all the actual tests are now in `run_test.sh`, and just the test requirements are listed in `meta.yaml`. I assume this resolves this.

tkittel:
@xhochy I have now added a pip check to the existing checks in `run_tests.sh`, I can see why that makes a lot of sense.

Are you otherwise happy with everything now?

tkittel:
A minor update, just to get the latest simple-build-system version from upstream + of course a rebase of this PR.

@xhochy is all OK now?

tkittel:
@xhochy I hope I have addressed all concerns with the latest changes, do you think you could finally accept this? :-)

xhochy (approved):

xhochy:
This got lost during the new year catch up 😱

tkittel:
Thanks a lot!! :-)

PR #24483, opened 2023-11-13
Adding pyflamegpu

Robadob:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/pyflamegpu/conda_build_config.yaml
+++ b/recipes/pyflamegpu/conda_build_config.yaml
@@ -1,0 +1,10 @@
+# https://docs.conda.io/projects/conda-build/en/3.21.x/resources/compiler-tools.html#using-your-customized-compiler-package-with-conda-build-3
+# https://github.com/rapidsai/cuml/blob/branch-23.12/conda/recipes/libcuml/conda_build_config.yaml
+# https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml
+cuda_compiler:
+  - cuda-nvcc # requires 'conda build -c nvidia conda' on Windows
+  
+c_compiler: # [win]
+  - vs2022 # [win]
+cxx_compiler: # [win]
+  - vs2022 # [win]
--- a/recipes/pyflamegpu/conda_build_config.yaml
+++ b/recipes/pyflamegpu/conda_build_config.yaml
@@ -3,6 +3,15 @@
 # https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml
 cuda_compiler:
   - cuda-nvcc # requires 'conda build -c nvidia conda' on Windows
+  #- nvcc # requires 'conda build -c nvidia conda' on Windows
+cuda_compiler_version:
+  #- 11.2                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12.0                       # [(linux or win64) and os.environ.get("CF_CUDA_ENABLED", "False") == "True"]
+  - 12.0                       
+cuda_compiler_version_min:
+  #- 11.2                       # [linux or win64]
+  - 12.0                       # [linux or win64]
+  - 12.0                       
   
 c_compiler: # [win]
   - vs2022 # [win]
--- a/recipes/pyflamegpu/meta.yaml
+++ b/recipes/pyflamegpu/meta.yaml
@@ -21,6 +21,7 @@ requirements:
     - swig ==4.0.2 # Have to manually direct swig to find this
     - git
     - cmake
+    - make  # [not win]
   host:
     - python
     - astpretty
@@ -30,6 +31,7 @@ requirements:
     - wheel
     - build
   run:
+    - cuda-version >=12.0
     - python
     - astpretty
     - cuda-cudart
--- a/recipes/pyflamegpu/meta.yaml
+++ b/recipes/pyflamegpu/meta.yaml
@@ -1,0 +1,77 @@
+{% set name = "pyflamegpu" %}
+{% set version_py = "2.0.0rc1" %}
+{% set version_cpp = "2.0.0-rc1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version_py }}
+
+source:
+  # Temporary, will switch to release, when conda recipe is working
+  url: https://github.com/FLAMEGPU/FLAMEGPU2/archive/refs/heads/rob_conda.tar.gz
+  # url: https://github.com/FLAMEGPU/FLAMEGPU2/archive/refs/tags/{{ version_cpp }}.tar.gz
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cuda-cudart-dev # [linux]
+    - cuda-nvrtc-dev # [linux]
+    - libcurand-dev # [linux]
+    - swig ==4.0.2 # Have to manually direct swig to find this
+    - git
+    - cmake
+  host:
+    - python
+    - astpretty
+    - python-devtools
+    - pip
+    - setuptools >=40.8.0
+    - wheel
+    - build
+  run:
+    - python
+    - astpretty
+    - cuda-cudart
+    - cuda-nvrtc
+
+build:
+  script_env:
+    - FLAMEGPU_CONDA_BUILD_THREADS
+    - FLAMEGPU_CONDA_CUDA_ARCHITECTURES
+  missing_dso_whitelist:
+    - $RPATH/libcuda.so.1 # [linux] Ignore as this comes from CUDA driver
+    - $RPATH/libnvrtc.so.12 # [linux] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/libcudart.so.12 # [linux] Conda-build can't find this in run reqs cuda-nvrtc
+    - $RPATH/nvrtc64_120_0.dll # [win64] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/cudart64_12.dll # [win64] Conda-build can't find this in run reqs cuda-nvrtc
+  ignore_run_exports:
+    - python # [linux] Not clear why conda thinks this isn't used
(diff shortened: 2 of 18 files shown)


Robadob:

*Questions primarily relate to CUDA and build variants/metapackages, this is a Python/C++ hybrid package that uses CUDA.*

I've been working through [creating recipes](https://conda-forge.org/docs/maintainer/adding_pkgs.html#creating-recipes), and have run into a few questions (although some are more general). I don't believe any of these are answered in the [knowledge base](https://conda-forge.org/docs/maintainer/knowledge_base.html)

1.   Our package has four build variants (2x2), each independently compiled from source, I've named them in what I believe to be a meta-package style but would like clarity on whether this is the correct process. I can't find any particularly clear documentation on formatting build variants.

2. Our package depends on CUDA, to get the local build to succeed I have currently been using the `linux64` config. Using `linux64_cuda112` prevents us from building with CUDA 12.0. Ideally we would like to have both CUDA 11.2 and CUDA 12.0 build variants for compatibility reasons, however CUDA 11.2 requires the `cudatoolkit` package and CUDA 12.0 has split CUDA packages. It's not clear how we can detect this to filter packages within `meta.yml`.

3.  The above linked "creating recipes" states:

> In case your project has tests included, you need to decide if these tests should be executed while building the conda-forge feedstock.

Our package relies on CUDA, and hence tests specified in `meta.yml` won't be able to execute on CI runners. It's not clear how the execution of tests on conda-forge should be disabled.

mondus:
I agree to be a maintainer.

carterbox:
> 1. Our package has four build variants (2x2), each independently compiled from source, I've named them in what I believe to be a meta-package style but would like clarity on whether this is the correct process. I can't find any particularly clear documentation on formatting build variants.

You can optionally add keywords to the build string of each build variant in order to make it easier to see which variant is installed in an environment. i.e. `pytorch-2.1.2-cuda120_py39h832322d_303.conda`. However, this is not strictly necessary because each variant will have a unique build string which is generated from hashing the dependencies used to build the package.

> 2. Our package depends on CUDA, to get the local build to succeed I have currently been using the `linux64` config. Using `linux64_cuda112` prevents us from building with CUDA 12.0. Ideally we would like to have both CUDA 11.2 and CUDA 12.0 build variants for compatibility reasons, however CUDA 11.2 requires the `cudatoolkit` package and CUDA 12.0 has split CUDA packages. It's not clear how we can detect this to filter packages within `meta.yml`.

Our infrastructure creates separate build jobs for each CUDA version. We have a tool called conda-smithy which parses recipes and decided which build variants to automatically insert into the build matrix. The same process is used for building for multiple python versions.

> 3. The above linked "creating recipes" states:
> 
> > In case your project has tests included, you need to decide if these tests should be executed while building the conda-forge feedstock.
> 
> Our package relies on CUDA, and hence tests specified in `meta.yml` won't be able to execute on CI runners. It's not clear how the execution of tests on conda-forge should be disabled.

We do not offer build runners with GPUs, so you need to skip any tests that require loading the CUDA driver.

> 4. Our Windows builds are currently failing to detect CUDA support at CMake configure time on conda-forge CI. I tried alot of configurations before I found a combination that works with local vanilla `conda build`, and don't immediately recognise what would be the problem on your CI (I have added `cuda-version >=12.0` to `meta.yml` since then). Knowledge base references `__cuda`, but I think that's related to run constraints, so shouldn't interfere with build.

The build runners do not have any GPUs. Ensure that you are using our CMAKE_ARGS environment variable and the compilers that are set to the environment variables CC, CXX, etc.

carterbox:
Here in staged recipes, the package will only be built against CUDA 11.2. Once in the feedstock, it will be rerendered and then it will be built against 11.2, 11.8, and 12.x.

Here are some examples that are similar to your recipes:

https://github.com/conda-forge/staged-recipes/pull/23415 -> https://github.com/conda-forge/carterbox-torch-radon-feedstock

https://github.com/conda-forge/libmagma-feedstock

recipes/pyflamegpu/conda_build_config.yaml
```diff
@@ -0,0 +1,19 @@
...
+  - 12.0                       
+cuda_compiler_version_min:
+  #- 11.2                       # [linux or win64]
+  - 12.0                       # [linux or win64]
+  - 12.0                       
+  
+c_compiler: # [win]
+  - vs2022 # [win]
+cxx_compiler: # [win]
+  - vs2022 # [win]
```
carterbox, line 19: Delete this file. CUDA and compiler variants are managed by our infrastructure.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+  number: 0
+  script_env:
+    - FLAMEGPU_CONDA_BUILD_THREADS
+    - FLAMEGPU_CONDA_CUDA_ARCHITECTURES
+  missing_dso_whitelist:
+    - $RPATH/libcuda.so.1  # [linux] Ignore as this comes from CUDA driver
+    - $RPATH/libnvrtc.so.12  # [linux] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/libcudart.so.12  # [linux] Conda-build can't find this in run reqs cuda-nvrtc
+    - $RPATH/nvrtc64_120_0.dll  # [win64] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/cudart64_12.dll  # [win64] Conda-build can't find this in run reqs cuda-nvrtc
```
carterbox, line 24:
```suggestion
```
These should not be needed because in staged-recipes there is only one CUDA variant available: CUDA 11.2.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+  version: {{ version_py }}
+
+source:
+  # Temporary, will switch to release, when conda recipe is working
+  url: https://github.com/FLAMEGPU/FLAMEGPU2/archive/refs/heads/rob_conda.tar.gz
+  # url: https://github.com/FLAMEGPU/FLAMEGPU2/archive/refs/tags/{{ version_cpp }}.tar.gz
+
+build:
+  number: 0
+  script_env:
```
carterbox, line 16:
```suggestion
build:
  number: 0
  skip: true  # [cuda_compiler_version == "None"]
  skip: true  # [py < 310]
  script_env:
```
Skip building the default non-cuda variant forever. Skip lower python versions for faster debugging for now, then unskip later.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - $RPATH/libcuda.so.1  # [linux] Ignore as this comes from CUDA driver
+    - $RPATH/libnvrtc.so.12  # [linux] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/libcudart.so.12  # [linux] Conda-build can't find this in run reqs cuda-nvrtc
+    - $RPATH/nvrtc64_120_0.dll  # [win64] Conda-build can't find this in run reqs cuda-cudart
+    - $RPATH/cudart64_12.dll  # [win64] Conda-build can't find this in run reqs cuda-nvrtc
+  ignore_run_exports:
+    - python  # [linux] Not clear why conda thinks this isn't used
+    - astpretty  # [linux, win64] Not clear why conda thinks this isn't used
+    - vc14_runtime  # [win64] Not clear why conda thinks this isn't used (we aren't static linking)
+    - ucrt  # [win64] Not clear why conda thinks this isn't used (we aren't static linking)
```
carterbox, line 29:
```suggestion
```
This is not the purpose of ignore_run_exports. You should only add host packages here that you know are NOT being used at runtime. e.g. You are using a C library that has some header-only APIs.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cuda-cudart-dev  # [linux]
+    - cuda-nvrtc-dev  # [linux]
+    - libcurand-dev  # [linux]
```
carterbox, line 37:
```suggestion
  {% if cuda_major == 12 %}
    - cuda-cudart-dev  # [linux]
    - cuda-nvrtc-dev  # [linux]
    - libcurand-dev  # [linux]
  {% endif %}
```
These packages are only for CUDA >= 12. Our package structure changed for CUDA 12. Before that, everything was shipped as one monolithic cudatoolkit package.

Also, these are libraries, so they belong in the HOST requirements section.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+  host:
+    - python
+    - astpretty
+    - python-devtools
+    - pip
+    - setuptools >=40.8.0
+    - wheel
+    - python-build
+  run:
+    - cuda-version >=12.0
```
carterbox, line 51:
```suggestion
```
The cuda version is run_exported by our cuda compiler package.

recipes/pyflamegpu/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+  license_family: MIT
+  summary: GPU accelerated agent based modelling framework for Python
+  description: |
+    FLAME GPU is a GPU accelerated simulator for domain independent complex systems simulations. Version 2 brings a complete re-write of the existing library offering greater flexibility, an improved interface for agent scripting, CUDA C++ & python3 interfaces and better research software engineering.
+    FLAME GPU provides a mapping between a formal agent specifications with C based scripting and optimised CUDA code. This includes a number of key ABM building blocks such as multiple agent types, agent communication and birth and death allocation.
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE.md
+  doc_url: https://docs.flamegpu.com/
+  dev_url: https://github.com/FLAMEGPU/FLAMEGPU2
```
carterbox, line 74: Looks like there are separate CXX and Python APIs? Can you build the CXX libraries first, then link the Python modules to them dynamically?

carterbox (changes requested):

Robadob:
Sorry, it took so long to receive a response that I was moved onto another
project and haven't had capacity to continue moving this forwards.

Unclear/if when we will pick of back up.

PR #24506, opened 2023-11-15
Isolate conda python from site packages, like venv does by default

itcarroll:
The package ~~(which is a one liner)~~ addresses an issue repeatedly brought up in conda/conda#394, conda/conda#448, conda/conda#7707, conda/conda#7173, and conda/conda#13337. The solution is different from the solution introduced in the [conda-ecosystem-user-package-isolation](https://anaconda.org/conda-forge/conda-ecosystem-user-package-isolation) package; it only effects Python in the environment. All this package does it does during ~~build~~`conda create`, which is to  ~~create/append the line "include-system-site-packages = false" to a "pyvenv.cfg" file read by the `site` module from the Python standard library.~~ create a "pyvenv.cfg" file, if one does not exist, with the line "include-system-site-packages = false" via a conda plugin implementing a post-command extension to `conda create`. As a result this package, if installed in the base environment,

- affects all new environments created by conda that include Python
- does not affect Python in the base environment
- will cease to have any effect if there is a change to conda that adds a pyvenv.cfg file during python installation.

Checklist
- [x] Source is from official source. **Not Applicable**
- [x] Build number is 0.
- [x] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details). **Not Applicable**


itcarroll:
I am willing to be a listed maintainer.

itcarroll:
@mfansler Thank you for suggesting a new package, and I would really appreciate your help. I am new to conda-forge.

There are two outstanding things that I know I don't know how to do:
- [ ] Indicate that the package requires python >= 3.3
- [ ] Write a bld.bat script for Windows that mirrors build.sh

chrisburr:
Is this editing a file in-place? If so, you can't rely on the installation order so people could get either the fixed or unfixed version.

jaimergp:
AFAIK this is shipping the file, which is subject to clobbering. Having no deps, maybe it's installed last or first, not sure how the toposort treats lone leaves.

An alternative is to use `post-link.sh` to edit things at env creation time, but it might suffer from the same clobbering problems.

Realistically speaking I don't know if there are many packages clobbering `pyenv.cfg`, so it's not that bad, but I'm not sure about how good a solution this is, given this would be opt-in for every environment created and tied to the conda-forge channel when it's more of a ecosystem-wide configuration aspect.

A `conda` specific alternative would be to have:

1. A `post-solve` plugin hook
2. A plugin implementing such hook, watching `LINK` transactions. If `python` is included, it ensures `pyvenv.cfg` is configured properly via `python -mvenv`.

Another option is to have this depend on `python` and use `python -mvenv` to do all the configuration. But again, nothing guarantees those changes are undone by other packages.

At any rate, if this goes through, we need to be careful about where in the prefix tree this is placed, because according to [PEP 405](https://peps.python.org/pep-0405/), it might change the values of `sys.prefix` and friends.

itcarroll:
If there is interest here in a conda solution, please comment on conda/conda#13337.

itcarroll:
I think I see two concerns about clobbering a `pyvenv.cfg` file.

1. If this package is installed before Python, it might have no effect.
2. If another package is installed after this one, it might undo this package's effect.

I don't think 1 is a problem, because the $PREFIX directory exists for any conda environment and I don't think installation of Python would remove an existing pyvenv.cfg file.

I guess for 2 I'm not sure that we want to prevent users from altering a configuration file. The point of this package is to make it easy to create. If a user decides later they want a different configuration, that's okay.

recipes/conda-ecosystem-site-packages-isolation/meta.yaml
```diff
@@ -0,0 +1,28 @@
+package:
+  name: conda-ecosystem-site-packages-isolation
+  version: 0.1.0
+
+build:
+  number: 0
+
```
mfansler, line 7: Constrain any Python installations with

```suggestion

requirements:
  run_constrained:
    - python >=3.3

```

recipes/conda-ecosystem-site-packages-isolation/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+    default behavior of virtual environments created by Python's built-in "venv" module.
+    Python in the conda environment gets isolated from system site-packages, including user
+    site-packages.
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file: LICENSE.txt
+
+extra:
+  recipe-maintainers:
+    - itcarroll
```
mfansler, line 28: I'm happy to be added as a maintainer (I also maintain the related [`conda-ecosystem-user-package-isolation-feedstock`](https://github.com/conda-forge/conda-ecosystem-user-package-isolation-feedstock)).

```suggestion
    - itcarroll
    - mfansler
```

recipes/conda-ecosystem-site-packages-isolation/meta.yaml
```diff
@@ -0,0 +1,28 @@
...
+    - python -c "import sys, site; sys.exit(0 if site.ENABLE_USER_SITE==False else 1)"
+
+about:
+  home: http://conda-forge.org/
+  summary: Prevent the environment's Python from searching for packages outside the environment
+  description: >
+    With respect to the list of locations searched by Python for packages, mimic the
+    default behavior of virtual environments created by Python's built-in "venv" module.
+    Python in the conda environment gets isolated from system site-packages, including user
+    site-packages.
```
mfansler, line 25: I'd suggest leading with *what* it does, then mentioning this aligns with `venv` behavior.

```suggestion
    This package uses the mechanism provided by Python's `site` module to ignore 
    system-level site-packages via setting `include-system-site-packages=false` in a 
    prefix-level "pyvenv.cfg". This aligns Conda environments with the default behavior 
    of `venv` virtual environments.
```

mfansler:
I agree @jaimergp that it makes more sense to implement the changes in a hook rather than ship as a static file. If one really wanted to enforce the status, one could check it (and potentially correct it) via an activation script.

itcarroll:
I'm not sure I understand what the suggestions to use a hook imply regarding a conda-forge package, but I believe the two alternatives proposed above (the `post-solve` plugin hook, or the ... well, other one) are ideas for a solution on conda itself. I fully support a fix over there, an encourage closing this PR if anyone can work up a conda solution. It's beyond me.

mfansler:
@itcarroll apologies - I was perhaps a bit loose in referring to package scripts as "hooks". I'll clarify, since indeed I was still talking about this package.

Currently, this recipe ships a static `pyvenv.cfg`file. So if other recipes do something similar, the ultimate installed version in an environment will depend on (possibly arbitrary) installation order.

Jaime suggested to instead use a `post-link.sh` script (after the package has been technically "installed") to generate/edit a `pyvenv.cfg` on-the-fly. That avoids this package overwriting others, but does nothing to prevent our setting from being wiped. (Jaime also had suggestions for an alternate approach via a plugin instead of a package.)

I then suggested this package could also use an `activate.sh` script to keep the setting enforced. That is, the setting value could be checked and repaired, rather than setting it only once and hoping nothing messes it up. To me this seems on par with setting environment variables during environment activation, but I'd like to know if others see it differently.

Anyway, I think of these scripts as using "hooks" (invoked on events), but realize the Conda documentation has a precise meaning of "hook" in the context of plugin API. _Mea culpa!_

itcarroll:
Thanks all for the suggestions and help. I have refactored the package as a post-command plugin to conda, but continue to "ship" the pyvenv.cfg file. I don't see any alternative to using a pyvenv.cfg to configure the search paths; Python's mechanism is based on this file. By changing the package to a post-command hook that only acts during environment creation, I believe it is now also clear that the package will not try to prevent clobbering of the pyvenv.cfg. This behavior is how `python -m venv` works too, there are no guard rails on the configuration file.

I considered the advantages and disadvantages of using `venv` to create the configuration file, and opted to create the configuration file manually. The `venv` module is inflexible about including lines that do not apply to conda environment creation, so the manual approach is cleaner than trying to prune a file created by `venv`.

itcarroll:

recipes/conda-create-no-user-site/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - python >=3.3
+
+test:
+  requires:
+    - python
+  commands:
+    - python -c "import sys, site; sys.exit(0 if site.ENABLE_USER_SITE==False else 1)"
+
+about:
+  home: http://conda-forge.org/
```
jaimergp, line 28:
```suggestion
  home: http://github.com/conda-forge/conda-create-no-user-site-feedstock
```

recipes/conda-create-no-user-site/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  summary: conda create will configure Python to ignore packages outside the environment
+  description: >
+    This package extends `conda create` to configure Python so it will ignore packages
+    installed with `pip install --user`. Envrionment creation will now add a
```
jaimergp, line 32:
```suggestion
    installed with `pip install --user`. Environnment creation will now add a
```

recipes/conda-create-no-user-site/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    prefix-level "pyvenv.cfg" file with the line `include-system-site-packages = false`,
+    which gets read by Python's `site` module whenver Python is run. This aligns Python
```
jaimergp, line 34:
```suggestion
    which gets read by Python's `site` module whenever Python is run. This aligns Python
```

recipes/conda-create-no-user-site/conda_create_no_user_site/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+        with pyvenv_cfg.open("x", encoding="utf-8") as f:
+            f.write("include-system-site-packages = false\n")
+
+
+@plugins.hookimpl
+def conda_post_commands():
+    yield plugins.CondaPostCommand(
+        name="no-user-site",
+        action=action,
+        run_for={"create"},
```
jaimergp, line 22: I would also consider adding `install`, `remove` and maybe even `update`. The reasons:

- A user had created an environment without Python, but then installs something that brings Python in.
- The user removes `python` from the environment, and there's a lingering `pyvenv.cfg` file, which should be removed if possible.

recipes/conda-create-no-user-site/meta.yaml
```diff
@@ -0,0 +1,44 @@
+package:
+  name: conda-create-no-user-site
+  version: 0.1.0
+
+source:
+  path: .
```
jaimergp, line 6: What's the maintenance model for this package? Feedstocks are not usually equipped to deal with release processes and PRs, so maybe it's a better idea to commit this to a separate repo (maybe under conda-incubator, if you are inclined). I like seeing more conda plugins in the world and it would help with visibility.

jezdez:
Hi there, conda maintainer here (and former virtualenv developer):

I believe this proposed package is solving the problem on the wrong level as it's using the plugin system to change a fundamental feature of conda and in effect makes _you_ the maintainer of this pattern in the conda ecosystem going forward. That's not sustainable and would be better solved by an config option in conda instead. It would also make this available for user of conda that aren't using conda-forge, e.g. users of Anaconda Distribution.

I would encourage you instead to work on a patch to conda to fix https://github.com/conda/conda/issues/13337 via a condarc setting that would enable users to switch to `include-system-site-packages=false` if they choose to do so. We can document it in the official conda docs and point out pros and cons there, when making that switch. 

Maybe, when we get feedback from users, we can even change the default for new conda installations in the future. Happy to discuss what we would need to know to make that decision.

Hope that helps!

itcarroll:
Thank you @jaimergp and @jezdez for the review and comment! I would also much prefer to see a solution at the conda level, or perhaps on the anaconda/python "feedstock" (if there is such a thing for anaconda packages outside conda-forge) since it is specific to environments with Python. I started this package because I got zero support on conda/conda#13337. If you can generate some interest on the conda issue, we can continue finding the right solution over there and I'll just close this PR.

PR #24522, opened 2023-11-17
Add lsst-ts-xml recipe

couger01:


rbovill:
I have been, and ever shall be, a maintainer of the ts_xml package.

tribeiro:
Agree to be a maintainer.

beckermr (changes requested):
Please remove the comments.

recipes/lsst-ts-xml/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/lsst-ts/ts_xml/archive/v{{ version }}.tar.gz
+  sha256: 04a8c8053c1c26180b689c61ab470de8117f6ac243158f17b17aa1a288dfc5fe
+
+build:
+  skip: True  # [win, py<310]
```
beckermr, line 13: This recipe should be noarch python with a constraint on python in the recipe directly.

beckermr (changes requested):

recipes/lsst-ts-xml/meta.yaml
couger01: Except it didn't build successfully with Windows and therefore when I set the skip: true [win], the linter said that having the skip meant that noarch:python couldn't be used.

beckermr: You can ignore the windows failure. Noarch python only builds on Linux.

recipes/lsst-ts-xml/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+    - setuptools_scm
+  run:
+    - python
+    - astropy
```
beckermr, line 24:
```suggestion
    - astropy
    - _unix
```
This will ensure it only installs on osx and Linux.

beckermr (commented):
This change will ensure people cannot install this on windows.

recipes/lsst-ts-xml/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  sha256: 04a8c8053c1c26180b689c61ab470de8117f6ac243158f17b17aa1a288dfc5fe
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python
```
beckermr, line 19:
```suggestion
    - python >=3.10
```

recipes/lsst-ts-xml/meta.yaml
couger01: It has to be with a double underscore.

PR #24539, opened 2023-11-20
Adding CppInterOp package

alexander-penev:
Checklist
- [x] Build number is 0.


vgvassilev:
@SylvainCorlay, @JohanMabille, @anutosh491, here is our CppInterOp as we discussed. It probably has rough edges but we'd be glad if you could help landing it.

recipes/CppInterOp/build.sh
```diff
@@ -0,0 +1,147 @@
...
+# LLVM/Clang settings
+
+export LLVM_DIR="${SRC_DIR}/llvm_project"
+export LLVM_BUILD_DIR="${LLVM_DIR}/build"
+
+if [[ "${backend}" == "repl" ]]; then
+  export LD_LIBRARY_PATH="${LD_LIBRARY_PATH}:${LLVM_BUILD_DIR}/lib:${LLVM_DIR}/lib"
+  #export CPLUS_INCLUDE_PATH="${CPLUS_INCLUDE_PATH}:${LLVM_BUILD_DIR}/tools/clang/include:${LLVM_BUILD_DIR}/include:${LLVM_DIR}/clang/include:${LLVM_DIR}/llvm/include"
+
+  x=$(conda list -q --json "^clang$")
```
JohanMabille, line 29: We should not fallback to building cling, clang, and LLVM if clang has not been installed. This should be considered as a critical error since a dependency that is guaranteed to be installed by conda/mamba is not.

recipes/CppInterOp/conda_build_config.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - ""
+  - a292a61b
+backend:
+  - repl
+  - cling
+zip_keys:
+  -
+    - clangdev
+    - clingdev
+    - clingdev_tag
```
SylvainCorlay, line 17: There seems to be an issue in the zip keys.

SylvainCorlay:
> That is the intent, yes. The major difference is CppInterOp is all-in-one, embeddable library which can co-exist with other versions of LLVM at the same time.

I thought that the embedding everything was only going to be for the cling use case. Could we envision a mode where we embed nothing, link dynamically with llvm and clang, and depend on them?

vgvassilev:
> > That is the intent, yes. The major difference is CppInterOp is all-in-one, embeddable library which can co-exist with other versions of LLVM at the same time.
> 
> I thought that the embedding everything was only going to be for the cling use case. Could we envision a mode where we embed nothing, link dynamically with llvm and clang, and depend on them?

No, we should not do this because if CppInterOp is embedded into a framework that uses another version of LLVM both will clash. For example, if I `dlopen(CppInterOp)` it will bring the dependency of LLVM16 and if the user code `dlopen`s something that requires another version of LLVM such as libMesa it will crash.

The design of CppInterOp is to encapsulate its LLVM and use it to compile and run without interfering other projects that use LLVM, too.

recipes/CppInterOp/build.sh
anutosh491: + 1

SylvainCorlay:
> The design of CppInterOp is to encapsulate its LLVM and use it to compile and run without interfering other projects that use LLVM, too.

I get that but on conda-forge, if you use shared objects (incuding for LLVM), everybody will use the same flavor of llvm, which presumably would resolve the issue of differing versions.

vgvassilev:
> > The design of CppInterOp is to encapsulate its LLVM and use it to compile and run without interfering other projects that use LLVM, too.
> 
> I get that but on conda-forge, if you use shared objects (incuding for LLVM), everybody will use the same flavor of llvm, which presumably would resolve the issue of differing versions.

I can get CppInterOp via conda on my system, build a tool and deploy it on another system that has libMesa installed using another package manager. In that case we are in the situation I described.

vgvassilev:
@SylvainCorlay ping.

JohanMabille (approved):

SylvainCorlay:
I am merging this here, but I am not sure we should use it in Xeus-cpp.

anutosh491:
> I am merging this here, but I am not sure we should use it in Xeus-cpp.

From a wasm build perspective too, I can imagine things getting complicated (every dependency more is a pain)

vgvassilev:
I believe we discussed that and I think that's the way to go in a longer term. We can support wasm in CppInterOp...

PR #24540, opened 2023-11-20
Add meos, pymeos_cffi and pymeos packages

Diviloper:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/meos/build.sh
+++ b/recipes/meos/build.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+mkdir -p build && cd build
+
+cmake ${CMAKE_ARGS} \
+      -D CMAKE_BUILD_TYPE=Release \
+      -D MEOS=on \
+      -D GSL_INCLUDE_DIR=$PREFIX/lib \
+      -D GSL_LIBRARY=$PREFIX/lib \
+      -D GSL_CBLAS_LIBRARY=$PREFIX/lib \
+      -D PROJ_INCLUDE_DIRS=$PREFIX/include \
+      -D PROJ_LIBRARIES=$PREFIX/lib \
+      ${SRC_DIR}
+
+make -j
+
+make install
--- a/recipes/meos/build.sh
+++ b/recipes/meos/build.sh
@@ -1,15 +1,25 @@
 #!/bin/bash
 
+echo "PREFIX variable: ${PREFIX}"
+
+sed -i "s,/usr/local,${PREFIX},g" ./meos/CMakeLists.txt
+
 mkdir -p build && cd build
 
 cmake ${CMAKE_ARGS} \
       -D CMAKE_BUILD_TYPE=Release \
+      -D CMAKE_INSTALL_PREFIX=${PREFIX} \
+      -D CMAKE_INSTALL_LIBDIR=lib \
       -D MEOS=on \
-      -D GSL_INCLUDE_DIR=$PREFIX/lib \
-      -D GSL_LIBRARY=$PREFIX/lib \
-      -D GSL_CBLAS_LIBRARY=$PREFIX/lib \
-      -D PROJ_INCLUDE_DIRS=$PREFIX/include \
-      -D PROJ_LIBRARIES=$PREFIX/lib \
+      -D GSL_INCLUDE_DIR=${BUILD_PREFIX}/include \
+      -D GSL_LIBRARY=${BUILD_PREFIX}/lib/libgsl.so \
+      -D GSL_CBLAS_LIBRARY=${BUILD_PREFIX}/lib/libgslcblas.so \
+      -D PROJ_INCLUDE_DIRS=${BUILD_PREFIX}/include \
+      -D PROJ_LIBRARIES=${BUILD_PREFIX}/lib/libproj.so \
+      -D GEOS_INCLUDE_DIR=${BUILD_PREFIX}/include \
+      -D GEOS_LIBRARY=${BUILD_PREFIX}/lib/libgeos.so \
+      -D JSON-C_INCLUDE_DIRS=${BUILD_PREFIX}/include/json-c \
+      -D JSON-C_LIBRARIES=${BUILD_PREFIX}/lib/libjson-c.so \
       ${SRC_DIR}
 
 make -j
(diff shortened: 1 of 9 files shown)


estebanzimanyi:
I would like to collaborate with Victor in the role of maintainer
Many thanks 
Esteban

Diviloper:
Hi @conda-forge/help-python-c , ready for review!

Diviloper:

recipes/meos/build.sh
```diff
@@ -0,0 +1,25 @@
...
+      -D PROJ_INCLUDE_DIRS=${PREFIX}/include \
+      -D PROJ_LIBRARIES=${PREFIX}/lib/libproj.so \
+      -D GEOS_INCLUDE_DIR=${PREFIX}/include \
+      -D GEOS_LIBRARY=${PREFIX}/lib/libgeos_c.so \
+      -D JSON-C_INCLUDE_DIRS=${PREFIX}/include/json-c \
+      -D JSON-C_LIBRARIES=${PREFIX}/lib/libjson-c.so \
+      ${SRC_DIR}
+
+make -j
+make install
```
ocefpaf, line 25: Is there a `make check` or `ctest` that we can run here?

recipes/meos/meta.yaml
```diff
@@ -0,0 +1,52 @@
+{% set name = "meos" %}
+{% set version = "1.1.0b1" %}
```
ocefpaf, line 2: This one falls under CFEP-05. See https://github.com/conda-forge/cfep/blob/main/cfep-05.md

recipes/pymeos/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 363b5fe10314f481f71361ed9e6f21443e4bf94c2546e4bdc1d31ffa70c85fce
+
+build:
+  skip: true  # [not linux]
+  skip: true  # [py < 37]
```
ocefpaf, line 14: This one can probably be noarch.

Diviloper: I think so too, but I set it like this since it's the same as the `pymeos_cffi` package which this depends on. I can change it if you think it's better

recipes/meos/build.sh
Diviloper: I don't think there is now, but I need @estebanzimanyi to confirm or deny this.

recipes/pymeos/meta.yaml
ocefpaf: It is fine to keep this noarch and let Windows and mac-OS fail with a missing dependency , or at least OS-noarch. That way we avoid the Python build matrix and save some CI resources.

recipes/meos/meta.yaml
Diviloper: Should I just change the "b" for "rc" (in the three recipes)? I see more instructions in the spec, but only for existing feedstrocks.

recipes/pymeos/meta.yaml
Diviloper: I've made it noarch and moved the python constraints into the host and run sections following the docs at https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-builds

recipes/meos/meta.yaml
ocefpaf: All development releases fall under CFEP-05. You can either add the `conda_build_config.yaml` and publish it in a development label, or wait/bug upstream for a stable release.

Diviloper: I see. I'll talk with the rest of the team to decide whether it's worth it to publish it now or just wait a bit. Thank you!

recipes/meos/build.sh
estebanzimanyi: `ctest` is used for all the CI tests in the MobilityDB (PostgreSQL) binding, e.g.,
https://github.com/estebanzimanyi/MobilityDB/actions/runs/7252347113/job/19756608599
We didn't used `ctest` for the PyMEOS binding.

Diviloper: Indeed, we don't have a check for MEOS, since the library is tested through MobilityDB.
Will this be an issue? If so, we could activate the testing in the `pymeos` recipe, which would report any issue with the `meos` library.

recipes/meos/meta.yaml
Diviloper: @ocefpaf I updated the version numbers to fix this, since the stable release of all three packages has just been published.

recipes/pymeos_cffi/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name | replace("-", "_") }}/{{ name | replace("-", "_") }}-{{ version }}.tar.gz
+  sha256: 08d0b50293f2af458188aa3a41af0d6ecd2d2db91130128ddf33ffb47343d57b
+
+build:
+  skip: true  # [not linux]
+  skip: true  # [py < 37]
```
ocefpaf, line 14:
```suggestion
  skip: true  # [py<37]
```

recipes/pymeos_cffi/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name | replace("-", "_") }}/{{ name | replace("-", "_") }}-{{ version }}.tar.gz
+  sha256: 08d0b50293f2af458188aa3a41af0d6ecd2d2db91130128ddf33ffb47343d57b
+
+build:
+  skip: true  # [not linux]
```
ocefpaf, line 13: What is happening on Windows and macOS? Can you get at lest macOS to work?

Diviloper: When the PR was created MEOS only supported Linux, but now MacOS is supported too.
I'll work on making the python package available for MacOS too.
Thanks for pointing it out!!

Diviloper: We managed to make the compilation work for macOS too, although in the way we found a couple issues that are fixed in the

PR #24610, opened 2023-11-29
Add hostess to conda-forge

Sierra-MC:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/hostess/meta.yaml
+++ b/recipes/hostess/meta.yaml
@@ -13,6 +13,7 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
+  skip: true  # [win]
 
 requirements:
   host:
--- a/recipes/hostess/meta.yaml
+++ b/recipes/hostess/meta.yaml
@@ -10,7 +10,6 @@ source:
   sha256: 319916189bc0cb8cf2c85cc2479f5e90cffcc858774a0734856c1365e3cdc912
 
 build:
-  noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
   skip: true  # [win]
--- a/recipes/hostess/meta.yaml
+++ b/recipes/hostess/meta.yaml
@@ -47,11 +47,12 @@ test:
     - pip
 
 about:
+  home: https://github.com/MillionConcepts/hostess
   summary: intuitive admin library
   license: BSD-3-Clause
   license_file: LICENSE
 
 extra:
   recipe-maintainers:
-    - sierrabrown
+    - Sierra-MC
     - m-stclair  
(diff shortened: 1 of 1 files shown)


m-stclair:
I am willing to be a maintainer of this package on conda-forge.

Sierra-MC:
@conda-forge/staged-recipes please review.

Sierra-MC:
@conda-forge/help-python ready for review (sorry, I didn't realize there was a language specific tag so using that now so this goes to the right team)

recipes/hostess/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/hostess-{{ version }}.tar.gz
+  sha256: 319916189bc0cb8cf2c85cc2479f5e90cffcc858774a0734856c1365e3cdc912
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
+  skip: true  # [py<39]
```
xhochy, line 16: This looks like it can be
```suggestion
  noarch: python
```

Any reason not to?

xhochy (changes requested):

recipes/hostess/meta.yaml
Sierra-MC: We need a python version specification (3.9+) and when adding that while using `noarch: python` the bot replied:
"noarch packages can't have skips with selectors. If the selectors are necessary, please remove noarch: python."

Prior to that I had added the windows skip because otherwise the cicd does not pass as windows is not compatible with this package due to the python-magic dependency (which is not compatible with windows).

recipes/hostess/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
```
ocefpaf, line 23:
```suggestion
    - python >=3.9
    - pip
  run:
    - python >=3.9
```

ocefpaf: That fixes the noarch lint you had. You should make it noarch and add the min python here.

ocefpaf:
@xhochy you were the first reviewer here. I'll leave the merging to you.

Sierra-MC:
@ocefpaf @xhochy after adding in these changes the build is failing on both windows (expected since the windows skip was removed by adding noarch python as described above) and osx. Can I change the recipe back and still be eligible for merge?

Sierra-MC:
@ocefpaf @xhochy after adding in these changes the build is failing on windows (expected since the windows skip was removed by adding noarch python as described above). Please let me know if this is acceptable and ready for merge or if you'd like the commits reverted so the cicd checks skip windows as it was previously set up to do.

xhochy (approved):

PR #24669, opened 2023-12-04
Add recipe for baybe

AdrianSosic:
Checklist
- [x] Build number is 0.


AdrianSosic:
@conda-forge/help-python, this PR is ready for review.

- We would highly appreciate your help in setting up the dependencies correctly. Our packages has several optional dependency groups and it is not clear to us what is the right way to specify those in the recipe. Here is our current [pyproject.toml](https://github.com/emdgroup/baybe/blob/main/pyproject.toml) file.
- Also, we are unsure if the dependency version requirements are to be copied 1:1 from the toml.

Scienfitz:
Happy to be maintainer.

recipes/baybe/meta.yaml
```diff
@@ -0,0 +1,66 @@
+{% set name = "baybe" %}
+{% set version = "0.7.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```
ocefpaf, line 9: Upstream published wheels only. You should request them to publish the source distribution (.tar.gz)  or see if they have releases on GitHub or something similar.

AdrianSosic: Hi @ocefpaf, thanks for your help. For clarity: we are the creators of the package and released it's first public version just yesterday. So in principle, we could publish the source distribution as well, if it helps!

However, we also have a first release in Github, and I tried using that with `grayskull`. I've just updated the recipe to what `grayskull` produced, but this brings me back to my original question how to properly account for our optional dependencies. In fact, `grayskull` immediately run into a problem since we are referencing one of the dependency groups in the main deps:
image

Also, none of the other groups are included...

AVHopp:
Happy to be a maintainer.

recipes/baybe/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - ngboost >=0.3.12
+    - numpy >=1.24.1
+    - pandas >=1.4.2
+    - protobuf <=3.20.3
+    - scikit-learn >=1.1.1
+    - scikit-learn-extra >=0.3.0
+    - scipy >=1.10.1
+    - setuptools-scm >=7.1.0
+    - pytorch >=1.11.0
+    - baybe
```
ocefpaf, line 39: The package cannot depend on itself.

```suggestion
```

AdrianSosic: Noticing missing conda deps in my recipe. I'm on it...

AdrianSosic:
Hi @ocefpaf, just FYI: since there are several missing conda dependencies for our package, we decided not add conda support for now. Hence, I'll close this PR. Thanks for your support, though, we might reconsider adding conda at a later point 👍🏼

PR #24726, opened 2023-12-08
Adding the USNISTGOV SCTK (word error rate scoring toolkit) as a conda package

egaznep:
Checklist
- [x] Build number is 0.


egaznep:
@conda-forge-admin, please ping team @conda-forge/help-c-cpp

I need help regarding the license file. Even though SPDX lists [NIST-Software](https://spdx.org/licenses/NIST-Software.html) as a valid identifier, the linter complains about it. 
(I plan to resolve rest of the issues myself)

recipes/sctk/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/usnistgov/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 43bd2bfe60e0ad04a563ddef025e84fca4c073f8f7c02f056b59fb8e2ec4f022
+
+build:
+  number: 3
```
ocefpaf, line 13: First published package of the same version should start as build 0.

```suggestion
  number: 0
```

recipes/sctk/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - cmake
+    - make
+    - m2-base  # [win]
+
+test:
+  source_files:
+    - makefile
+    - src/**
+  commands:
+    - make check -C $SRC_DIR  # [not win]
```
carterbox, line 29: If you want to run make check, please do it as part of the build script instead of including a copy of the entire source tree in the package.

My understanding is that this package is a collection of CLI tools, so the tests section should be calling "$ tool --help" or like "$ tool --version", to check that they return without error.

carterbox (changes requested):

recipes/sctk/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+build:
+  number: 0
+  skip: True  # [win]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
```
carterbox, line 20:
```suggestion
```
There appears to be no CMakeLists.txt in the upstream project, so CMake is probably not needed.

carterbox (changes requested):

recipes/sctk/meta.yaml
egaznep: I understand. Unfortunately the developer did not add either option. What would be the recommendation in this case, simply checking for the existence of the files?

recipes/sctk/meta.yaml
egaznep: I did this change myself, should've "accepted your suggestion" instead. I would be glad if you could re-check.

egaznep:
@conda-forge-admin, please ping team @conda-forge/help-c-cpp, ready for review!

PR #24761, opened 2023-12-12
Add APtools

ax3l:
A collection of tools for accelerator physics.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/aptools/meta.yaml
+++ b/recipes/aptools/meta.yaml
@@ -5,7 +5,6 @@ package:
   version: {{ version }}
 
 source:
-  fn: {{ version }}.tar.gz
   url: https://github.com/AngelFP/APtools/archive/refs/tags/v{{ version }}.tar.gz
   sha256: d382b6700668194d099c014b63c8e6d34cf0cb26cccc0a85daeef8a5883de8d2
 
--- a/recipes/aptools/meta.yaml
+++ b/recipes/aptools/meta.yaml
@@ -27,6 +27,10 @@ requirements:
     - scipy
 
 test:
+  requires:
+    - pip
+  commands:
+    - pip check
   imports:
     - aptools
 
--- a/recipes/aptools/meta.yaml
+++ b/recipes/aptools/meta.yaml
@@ -27,10 +27,6 @@ requirements:
     - scipy
 
 test:
-  requires:
-    - pip
-  commands:
-    - pip check
   imports:
     - aptools
 
--- a/recipes/aptools/meta.yaml
+++ b/recipes/aptools/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   number: 0
   noarch: python
-  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  script: {{ PYTHON }} -m pip install . -vv
 
 requirements:
   host:
--- a/recipes/aptools/meta.yaml
+++ b/recipes/aptools/meta.yaml
@@ -1,0 +1,54 @@
+{% set version = "0.2.4" %}
+
+package:
+  name: aptools
+  version: {{ version }}
+
+source:
+  fn: {{ version }}.tar.gz
+  url: https://github.com/AngelFP/APtools/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d382b6700668194d099c014b63c8e6d34cf0cb26cccc0a85daeef8a5883de8d2
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
+
+requirements:
+  host:
+    - pip
+    - python >=3.6
+    - setuptools
+  run:
+    - deprecated
+    - h5py
+    - numpy
+    - openpmd-api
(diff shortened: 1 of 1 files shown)


ax3l:
Needed for https://github.com/conda-forge/staged-recipes/pull/24760

ax3l:
@conda-forge/help-python ready for review :)

AngelFP:
I agree to be listed in this packages as a co-maintainer.

recipes/aptools/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set version = "0.2.4" %}
+
+package:
+  name: aptools
+  version: {{ version }}
+
+source:
+  fn: {{ version }}.tar.gz
```
ocefpaf, line 8:
```suggestion
```

recipes/aptools/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  url: https://github.com/AngelFP/APtools/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: d382b6700668194d099c014b63c8e6d34cf0cb26cccc0a85daeef8a5883de8d2
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
ocefpaf, line 15:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

dhirschfeld: Hi @ocefpaf :wave:
I'm curious about this one as I've previously recommended to drop the `--no-deps --no-build-isolation` flags with the understanding that they were already configured on `conda-forge`.

My go-to reference which I usually recommend is to follow the example recipe:
https://github.com/conda-forge/staged-recipes/blob/b9fe52599c957d19c9cdbf926d19a66be7460cdb/recipes/example/meta.yaml#L38


Is that still good advice or should we be recommending to add those flags?

ax3l: The KB seems to also follow the simple version implicitly in some examples:
https://conda-forge.org/docs/maintainer/knowledge_base.html

Will stand by for advise :)

dhirschfeld: From gitter (thanks to `jakirkham`) it seems you only need to use `-vv` here as the rest are set by `conda-forge` and this recipe doesn't use `outputs`.

The underlying issue with `outputs` - https://github.com/conda/conda-build/issues/3993

ax3l: Confirmed via chat:
https://matrix.to/#/!SOyumkgPRWoXfQYIFH:matrix.org/$17024301973176NcLkK:matrix.org?via=matrix.org&via=gitter.im&via=cadair.com

Conda adds automatically
```
... -vv --no-deps --no-index --no-build-isolation --ignore-installed --cache-dir pip_cache
```
unless recipe `outputs` are used.
https://github.com/conda-forge/implicit-feedstock/blob/3832f659196b1aae468470f1bda12b59e268ced4/recipe/meta.yaml#L42

ax3l:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

ocefpaf: In theory conda-build has those flags but it doesn't hurt to add them in fact latest version of grayskull is adding them by default. The reason is to avoid bugs like in the multiple output recipe where the flags were not initialized.

recipes/aptools/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - setuptools
+  run:
+    - deprecated
+    - h5py
+    - numpy
+    - openpmd-api
+    - python >=3.6
+    - scipy
+
+test:
```
xhochy, line 29:
```suggestion
test:
  requires:
    - pip
  commands:
    - pip check
```

xhochy (changes requested):

ax3l:
@xhochy thanks for the test hint. I do not understand why it fails with
```
+ pip check
aptools 0.2.4 requires openpmd-api, which is not installed.
```
because `openpmd-api` is installed as part of the `run` section.

recipes/aptools/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  requires:
+    - pip
+  commands:
+    - pip check
```
ax3l, line 33: Skipping for now:
```suggestion
```

dhirschfeld:
> I do not understand why it fails

`pip check` is a tool designed for the `pip` ecosystem and as such, IME, it doesn't always work with `conda` packages. When it does, it's nice to have as it adds an extra layer of protection against publishing broken packages, but when it doesn't I just skip it.

There is a feature request for `pip` to allow ignoring certain packages so we can still get the benefits for the rest:
  - https://github.com/pypa/pip/issues/11157
  - https://github.com/pypa/pip/pull/11159

xhochy (approved):

PR #24825, opened 2023-12-19
add python-ghostscript package

akrherz:
Checklist
-  If static libraries are linked in, the license of the static library is packaged.
- [x] Build number is 0.

See conda-forge/camelot-py-feedstock#25 for the need for this package.  There is some ambiguity though, as [pypi has python3-ghostscript](https://pypi.org/project/python3-ghostscript/) version 0.5, but points to [gitlab](https://gitlab.com/pdftools/python-ghostscript), which has a version 0.6, but is named `python-ghostscript`


akrherz:
I've now found a 0.7 release within the gitlab repo and updated the PR accordingly.

akrherz:
Eh, I am unsure how important it is that the windows CI fails on a noarch package?

akrherz:
Yeah, I had better figure out a way to fix this prior to merging

> RuntimeError: Can not find Ghostscript DLL in registry

akrherz:
oOo, green :)  a gentle pin to @ocefpaf to see if he has thoughts on what I did here :)

akrherz:
Now I realize that having OS specific selectors on the patch won't work for a noarch feedstock.  So carrying both patches shouldn't be of too much concern. 

- `windll_path.patch` only touches a code path used on windows
- `win_version.patch` hard codes the version number, which is brittle/dangerous for the next release, but will engage upstream on this.

recipes/python-ghostscript/windll_path.patch
```diff
@@ -0,0 +1,11 @@
+--- python-ghostscript-v0.7.orig/ghostscript/_gsprint.py	2023-12-20 13:18:30.453497088 -0600
++++ python-ghostscript-v0.7/ghostscript/_gsprint.py	2023-12-20 13:45:31.513348016 -0600
+@@ -487,6 +487,8 @@
+     from distutils.version import LooseVersion
+     import os
+ 
++    return os.environ.get("GS_DLL", os.environ["CONDA_PREFIX"] + "\\library\\bin\\gsdll64.dll")
```
ocefpaf, line 7: You may need to pin ghostscript to ensure a new version, with a different DLL version number in it, won't break the package, right? I'm not an expert on Windows to know how that name can change.

recipes/python-ghostscript/win_version.patch
```diff
@@ -0,0 +1,11 @@
+--- python-ghostscript-v0.7.orig/setup.cfg	2023-12-20 13:18:30.452497083 -0600
++++ python-ghostscript-v0.7/setup.cfg	2023-12-20 13:18:40.747546844 -0600
+@@ -22,7 +22,7 @@
+ 
+ [metadata]
+ name = ghostscript
+-version = attr:ghostscript.__version__
++version = 0.7
```
ocefpaf, line 8: Just curious, why do you need this patch? What went wrong here?

recipes/python-ghostscript/windll_path.patch
akrherz: Thanks for the review!  On my windows test system, there is no versioning with the DLL that I can find.
Screenshot from 2023-12-21 08-00-26

recipes/python-ghostscript/win_version.patch
akrherz: On windows, setuptools is failing to evaluate the version in this manner.  I needed it to make CI green here, but perhaps it would not be necessary within a feedstock building on linux64 for noarch

ocefpaf: Yep, noarch with a failing Windows would be ok. I also would like to see the error, I'm curious what is going on there.

akrherz: Well, I must have been dreaming that I tested this without the hardcoded path patch.  Onward and upward!

PR #24841, opened 2023-12-20
Add `google-cloud-cpp-compute` feedstock

coryan:
`compute` is a very large feature. It compiles nearly a thousand `.cc` files and creates about 100 `.so` files. Moving the feature to a separate feedstock makes it available for application developers, without impacting the build time for the `google-cloud-cpp-core` or `google-cloud-cpp` feedstocks.

Part of the work for https://github.com/conda-forge/google-cloud-cpp-feedstock/issues/141 

Checklist
- [x] Build number is 0.


coryan:
@h-vetinari I think this is ready for a review.

dbolduc:
I will support this feedstock

recipes/google-cloud-cpp-compute/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+  number: 0
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+    - libgrpc
+    - libprotobuf
+    - libgoogle-cloud-devel =={{ version }}
```
h-vetinari, line 24: Why is this in build?

recipes/google-cloud-cpp-compute/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+  - name: libgoogle-cloud-compute
+    script: install-libgoogle-cloud.sh   # [unix]
+    script: install-libgoogle-cloud.bat  # [win]
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - cmake
+        - ninja
+        - libgoogle-cloud-devel =={{ version }}
```
h-vetinari, line 45: Same here

recipes/google-cloud-cpp-compute/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - libabseil
+        - libprotobuf
+        - openssl
+        - libcxx  # [osx]
+      run:
+        - libgoogle-cloud =={{ version }}
+        - libabseil
+        - libprotobuf
+        - openssl
+        - libcxx  # [osx]
```
h-vetinari, line 57: All this should be covered by run-exports

recipes/google-cloud-cpp-compute/recipe-scripts-license.txt
h-vetinari, line 1: This file is superfluous I think?

h-vetinari (commented):
Basically LGTM, some details remain

recipes/google-cloud-cpp-compute/meta.yaml
coryan: Because I still don't have a firm grasp of what goes in `build` vs. `host` vs. `run`.  Removed.

recipes/google-cloud-cpp-compute/recipe-scripts-license.txt
coryan: Removed

recipes/google-cloud-cpp-compute/meta.yaml
coryan: Without it, the local build fails with:

```
CMake Error at CMakeLists.txt:254 (find_package):
  Could not find a package configuration file provided by
  "google_cloud_cpp_common" with any of the following names:

    google_cloud_cpp_commonConfig.cmake
    google_cloud_cpp_common-config.cmake

  Add the installation prefix of "google_cloud_cpp_common" to
  CMAKE_PREFIX_PATH or set "google_cloud_cpp_common_DIR" to a directory
  containing one of the above files.  If "google_cloud_cpp_common" provides a
  separate development package or SDK, be sure it has been installed.
```

`google-cloud-cpp-compute` depends on some of the libraries created in the `google-cloud-cpp-core-feedstock`.

I still don't have a firm grasp of what goes in `build` vs. `host` vs. `run`. So if it really belongs elsewhere I am happy to change it.  Or if you can suggest some documentation / tutorial that explains how each ought to be used, and would be thankful for that too.

recipes/google-cloud-cpp-compute/meta.yaml
coryan: Removed

recipes/google-cloud-cpp-compute/meta.yaml
coryan: Fixed.

recipes/google-cloud-cpp-compute/meta.yaml
h-vetinari: > I still don't have a firm grasp of what goes in `build` vs. `host` vs. `run`. So if it really belongs elsewhere I am happy to change it. Or if you can suggest some documentation / tutorial that explains how each ought to be used, and would be thankful for that too.

It's a thing that's hard to grasp indeed. I think the cross-compilation-case is actually the most instructive. Imagine a x64 agent, but we want to compile a library for aarch64. That means the stuff we need in the build environment (compiler, linker, etc. - everything that's actually executed during the build) needs to match the x64 architecture.

On the other hand, the host environment needs to match the target architecture (here aarch64), and any libraries we're compiling against must be in the host-environment, with the right ABI. Compiling against a static or header-only library _only_ requires a host-dependence (everything necessary gets consumed during compilation), whereas compiling against a shared library generates a runtime requirement (a "run-export" from `host:` to `run:`), which ensures that the actual DSO that's necessary to provide the symbols at runtime will be present. That's why we need `libgoogle-cloud-X-devel` in host, but only run-export `libgoogle-cloud-X`.

While there are some exceptions (e.g., we might need `libprotobuf` in the build environment for cross-compilation, because we need to execute `protoc`, which we can only do with the CPU arch of the agent), it's generally a sign of something going wrong to include "libraries I need to compile against" in the `build` environment. What's likely happening is that something somewhere is using a relative lookup (or some other heuristic, e.g. relative to the compilers) that expects the libraries in `$BUILD_PREFIX`, but we really should point that logic to `$PREFIX` (==host). In CMake that's usually something like `CMAKE_PREFIX_PATH=$PREFIX`.

Now the above focussed on the cross-compilation part for illustration purposes, but it holds as well for the native compilation case. The build environment is where the compilers and other build tools live, which are unrelated (including any transitive dependencies they might drag along) from the libraries we're compiling against.

coryan: Thanks, that was useful.   Let me know if it all looks good now.  I will change the other PRs once we have an agreed structure for them.

recipes/google-cloud-cpp-compute/build.sh
```diff
@@ -0,0 +1,28 @@
...
+      -DOPENSSL_ROOT_DIR=$PREFIX \
+      -DCMAKE_BUILD_TYPE=release \
+      -DCMAKE_CXX_STANDARD=17 \
+      -DCMAKE_INSTALL_PREFIX=$PREFIX \
+      -DCMAKE_INSTALL_LIBDIR=lib \
+      -DProtobuf_PROTOC_EXECUTABLE=$BUILD_PREFIX/bin/protoc \
+      -DGOOGLE_CLOUD_CPP_GRPC_PLUGIN_EXECUTABLE=$BUILD_PREFIX/bin/grpc_cpp_plugin \
+      -DGOOGLE_CLOUD_CPP_ENABLE_WERROR=OFF
+  cmake --build .build/compute
+  echo "$(date -u '+%Y-%m-%dT%H:%M:%SZ'): DONE - Building compute"
```
h-vetinari, line 28: Indentation here is off

recipes/google-cloud-cpp-compute/build.sh
```diff
@@ -0,0 +1,28 @@
...
+export OPENSSL_ROOT_DIR=$PREFIX
+
+if [[ "${target_platform}" == osx-* ]]; then
+  # See https://conda-forge.org/docs/maintainer/knowledge_base.html#newer-c-features-with-old-sdk
+  CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+  echo "$(date -u '+%Y-%m-%dT%H:%M:%SZ'): Building compute"
+  cmake ${CMAKE_ARGS} \
+      -GNinja -S . -B .build/compute \
```
h-vetinari, line 14: Can we be more explicit about which folders exist already, and avoid names with a leading dot? And also use the same names consistently between building and testing?

This might all seem a bit unnecessary, but it's possible (though not recommended) to pipe some file from the build stage through to the testing stage (and similarly, it's possible that the sources already contain a `./build` folder).

By consistently creating the the respective folders in the scripts, we demonstrate that we're not relying on either of that. What I'm thinking is something along the lines of:

```
mkdir -p build/compute
cd build/compute  # or pushd, if you prefer

cmake -G Ninja \
    ${CMAKE_ARGS} \
    -D... \
    -D... \
    ../..  # or $SRC_DIR

cmake --build .
```

recipes/google-cloud-cpp-compute/bld.bat
```diff
@@ -0,0 +1,25 @@
+@echo on
+setlocal EnableDelayedExpansion
+
+:: CMake does not like paths with \ characters
+set LIBRARY_PREFIX="%LIBRARY_PREFIX:\=/%"
+set BUILD_PREFIX="%BUILD_PREFIX:\=/%"
+set SRC_DIR="%SRC_DIR:\=/%"
+
+cmake -G "Ninja" ^
+    -S . -B .b ^
```
h-vetinari, line 10: Would prefer to avoid names with leading dots, especially on windows.

recipes/google-cloud-cpp-compute/bld.bat
```diff
@@ -0,0 +1,25 @@
+@echo on
+setlocal EnableDelayedExpansion
```
h-vetinari, line 2: I see no variables defined in if-branches, so I don't think we actually need this line?

recipes/google-cloud-cpp-compute/install-libgoogle-cloud.bat
```diff
@@ -0,0 +1,22 @@
...
+set BUILD_PREFIX="%BUILD_PREFIX:\=/%"
+set SRC_DIR="%SRC_DIR:\=/%"
+
+:: Once DLLs are working, we should install the non-devel packages using
+::     cmake --install .b --component google_cloud_cpp_runtime
+:: and the devel packages using
+::     cmake --install .b --component google_cloud_cpp_development
+
+if [%PKG_NAME%] == [libgoogle-cloud-compute-devel] (
+  cmake --install .b
```
h-vetinari, line 15: here we do need to refer back to the folder defined in `bld.bat`, and it's good if it's self-explanatory.

recipes/google-cloud-cpp-compute/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+    - nlohmann_json
+    - zlib # Needed by gRPC
+
+outputs:
+  - name: libgoogle-cloud-compute
+    script: install-libgoogle-cloud.sh   # [unix]
+    script: install-libgoogle-cloud.bat  # [win]
+    build:
+      run_exports:
+        - libgoogle-cloud =={{ version }}
```
h-vetinari, line 40: This should probably just run-export itself, otherwise we probably get a circularity problem when trying to build `libgoogle-cloud` that includes `libgoogle-cloud-compute`

recipes/google-cloud-cpp-compute/run_test_feature.bat
```diff
@@ -0,0 +1,19 @@
+@echo on
+setlocal EnableDelayedExpansion
+
+:: CMake does not like paths with \ characters
+set LIBRARY_PREFIX="%LIBRARY_PREFIX:\=/%"
+
+set FEATURE=%PKG_NAME:libgoogle-cloud-=%
+set FEATURE=%FEATURE:-devel=%
```
h-vetinari, line 8: Windows batch variable syntax is not very self-explanatory; I'd prefer a comment à la
```suggestion
:: strip off prefix and (possible) suffix around the main name of the library (==feature)
set FEATURE=%PKG_NAME:libgoogle-cloud-=%
set FEATURE=%FEATURE:-devel=%
```

recipes/google-cloud-cpp-compute/run_test_feature.sh
```diff
@@ -0,0 +1,19 @@
...
+# Stop on first error
+set -euo pipefail
+
+if [[ "${target_platform}" == osx-* ]]; then
+  # as in build.sh
+  CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+fi
+
+feature=${PKG_NAME/#libgoogle-cloud-/}
+feature=${feature/%-devel/}
```
h-vetinari, line 12: Same here actually - I wouldn't even know how that works. Are `#`/`%` a different spelling of regex(-like) markers?

h-vetinari (commented):
We're down to minor details now. :)

recipes/google-cloud-cpp-compute/run_test_feature.sh
coryan: Added a comment.

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

recipes/google-cloud-cpp-compute/run_test_feature.bat
coryan: Removed for this feedstock.  Will fix on the feedstocks where it is relevant.

recipes/google-cloud-cpp-compute/meta.yaml
coryan: I apologize, but that went over my head.  The thing that will depend on `libgoogle-cloud-compute` is `libgoogle-cloud-all`?

recipes/google-cloud-cpp-compute/install-libgoogle-cloud.bat
coryan: Changed. FWIW, I used a very short name to workaround the problems outlined in  https://github.com/googleapis/google-cloud-cpp/pull/13351

recipes/google-cloud-cpp-compute/bld.bat
coryan: Removed.

recipes/google-cloud-cpp-compute/bld.bat
coryan: Are you concerned about folks building the code on FAT filesystems with only 8.3 filename support?  I am not sure anything would work in that case, but no matter, changed.

recipes/google-cloud-cpp-compute/build.sh
coryan: > Can we be more explicit about which folders exist already,

I am not sure what you mean by this.

> and avoid names with a leading dot?

Happy to do that, though they are harmless.  AFAIK none of the Windows filesystems have limitations around a leading dot.

> And also use the same names consistently between building and testing?

Happy to do that too.

> This might all seem a bit unnecessary, but it's possible (though not recommended) to pipe some file from the build stage through to the testing stage (and similarly, it's possible that the sources already contain a `./build` folder).

It seems more likely that the sources would contain a `./build` directory than a `./.build` folder, but meh.

> By consistently creating the the respective folders in the scripts, we demonstrate that we're not relying on either of that. What I'm thinking is something along the lines of:
> 
> ```
> mkdir -p build/compute
> cd build/compute  # or pushd, if you prefer
> 
> cmake -G Ninja \
>     ${CMAKE_ARGS} \
>     -D... \
>     -D... \
>     ../..  # or $SRC_DIR
> 
> cmake --build .
> ```

`cmake -S <path to source> -B <path to build>` is a portable way to create the build directory and define the source directory. It is documented since CMake 3.13, and works since at least CMake 3.5.  Why use more commands to do the same thing that CMake already does?  It is not more explicit, or less explicit, just maybe less familiar thank explicit `mkdir` calls?

recipes/google-cloud-cpp-compute/build.sh
coryan: Fixed

coryan:
I think this is ready to be merged.

coryan:
@h-vetinari is this ready now?

recipes/google-cloud-cpp-compute/meta.yaml
h-vetinari: Run-exports are for capturing the relationship that compiling against a given shared library generates a run-time requirement for the same (or an ABI-compatible) library, in order to be able to resolve the symbols.

In other words, compiling against `libgoogle-cloud-compute[-devel]` should generate a run-export on `libgoogle-cloud-compute`. In a way, it's harmless though, because either way, a package compiled against `libgoogle-cloud-compute` will end up with a `libgoogle-cloud` runtime requirement, as `*-compute` itself needs that (and it gets generated as a run-export from the `libgoogle-cloud-devel` host-dep).

recipes/google-cloud-cpp-compute/run_test_feature.sh
h-vetinari: OK, it _is_ a pattern-matching syntax. It's rare enough that I would have just put the link into the comment, but fine.

h-vetinari (approved):
Sorry for the wait, let's get this in.

recipes/google-cloud-cpp-compute/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+about:
+  home: https://github.com/googleapis/google-cloud-cpp
+  license: Apache-2.0
+  license_family: Apache
+  license_file: LICENSE
+  summary: Google Cloud Client Library for C++
+
+extra:
+  feedstock-name: google-cloud-cpp-core
```
coryan, line 116: The PR is merged and the staged feedstock got removed, but still no feedstock repository.  I wonder if using the wrong name here caused some problems. 
@h-vetinari FYI.

h-vetinari: Yes, this will cause things to go wrong. Sorry for overlooking it

h-vetinari: @beckermr, it's weird that the recipe got removed "after conversion into feedstock", if the name exists already and it couldn't actually be created. https://github.com/conda-forge/staged-recipes/commit/85480c3bf0859ed9c0cbef6f43cbb3c13c5931ae

h-vetinari: It's fixed now.

PR #24845, opened 2023-12-21
Add recipe for llvm-bazel-cache

xhochy:
Checklist
- [ ] Build number is 0.


xhochy:
@h-vetinari @hmaarrfk would you mind reviewing this? I would like to use it in `jaxlib` to ensure that we stay inside the CI limit. It can also be useful for Tensorflow but it would there mainly reduce the time in the local builds but won't help us fit the builds into the 6h window.

recipes/llvm-bazel-cache/compile_bundle.py
```diff
@@ -0,0 +1,66 @@
...
+    current_src = astor.to_source(kw.value)[:-1]
+    kw.value = ast.parse(f"{current_src} + ['-l{name}']").body[0].value
+
+
+def new_kwarg(name, code):
+    return ast.keyword(arg="linkopts", value=ast.parse(code).body[0].value)
+
+
+def rewrite_cc_library(symbol, node):
+    sym = symbol.lower()
```
h-vetinari, line 24: Can you write a short explanation here? AFAICT this modifies both the AST of `BUILD.bazel` _and_ copies libraries into `$PREFIX`. I don't under stand the role of the `llvm_for_tf` cache here, because the libraries get copied to `$PREFIX/lib` directly, which means that we might get conflicts/clobbering with other of our llvm-builds.

Also, I'd like to see in the logs which libs get copied into `$PREFIX` like that.

recipes/llvm-bazel-cache/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+requirements:
+  build:
+    - python *
+    - astor
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - bazel
+    - bazel-toolchain
+    - rsync
```
h-vetinari, line 38: Should we constrain how this can be combined with `llvmdev` / `mlir`?

Or should we build those feedstocks for the same commits into the same label (`llvm-bazel-cache`), so that we can host-depend on them here and auto-deduplicate `$PREFIX/lib/*.a`?

recipes/llvm-bazel-cache/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - test -f "${PREFIX}/lib/libLLVMTFSupport.a"
+    - test -f "${PREFIX}/share/llvm_for_tf/llvm/BUILD.bazel"
+    - test -f "${PREFIX}/share/llvm_for_tf/llvm/include/llvm/Pass.h"
+    - test -f "${PREFIX}/lib/libMLIRTFSupport.a"
+    - test -f "${PREFIX}/share/llvm_for_tf/mlir/BUILD.bazel"
+    - test -f "${PREFIX}/share/llvm_for_tf/mlir/include/mlir/InitAllPasses.h"
+
+about:
+  home: http://llvm.org/
+  dev_url: https://github.com/llvm-mirror/llvm
```
h-vetinari, line 51: Let's use the proper address here: https://github.com/llvm/llvm-project/

h-vetinari (commented):
Still not very good with bazel, but I've got some questions. Otherwise LGTM in principle.

recipes/llvm-bazel-cache/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - astor
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - bazel
+    - bazel-toolchain
+    - rsync
+
+test:
+  commands:
+    - test -f "${PREFIX}/lib/libLLVMTFSupport.a"
```
hmaarrfk, line 42: it is strange to me that a static file is needed. can linkers take static + dynamic and make a dynamic library out of it?

recipes/llvm-bazel-cache/patches/build.patch
```diff
@@ -0,0 +1,46 @@
...
+index 7770284e5543..0b45127495dc 100644
+--- a/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
++++ b/utils/bazel/llvm-project-overlay/llvm/BUILD.bazel
+@@ -218,13 +218,15 @@ cc_library(
+         "lib/Support/BLAKE3/llvm_blake3_prefix.h",
+     ] + select({
+         "@platforms//cpu:aarch64": [
+-            "lib/Support/BLAKE3/blake3_neon.c",
++            # TODO(b/234415414): temporary disabled
++            #            "lib/Support/BLAKE3/blake3_neon.c",
```
hmaarrfk, line 11: do you intend to ever enable support for these?

recipes/llvm-bazel-cache/meta.yaml
```diff
@@ -0,0 +1,59 @@
+{% set version = "18.0.0.jax0420_b3e353" %}
```
hmaarrfk, line 1:
```suggestion
# This feedstock is used in jaxlib to ensure that we stay inside the CI limit.
# To find the correct version to build, browse ________
# It can also be useful for Tensorflow but it would there mainly reduce the 
# time in the local builds but won't help us fit the builds into the 6h window.
{% set version = "18.0.0.jax0420_b3e353" %}
```

Maybe add a small description on how to find the right version?

recipes/llvm-bazel-cache/meta.yaml
xhochy:
```suggestion
  dev_url: https://github.com/llvm/llvm-project
```

recipes/llvm-bazel-cache/meta.yaml
xhochy: We should not de-deduplicate anything here. We will probably never build `llvmdev` or `mlir` for these commits. I would add a `run_constraints: - llvmdev <0a0` here to avoid co-installation.

recipes/llvm-bazel-cache/meta.yaml
```diff
@@ -0,0 +1,63 @@
+# This feedstock is used in jaxlib to ensure that we stay inside the CI limit.
+# To find the correct version to build, browse ________
```
hmaarrfk, line 2: Btw, there is a "blank" here that I would kindly request your help to fill in.

Thanks for your other response!

recipes/llvm-bazel-cache/compile_bundle.py
xhochy: I wrote some comments and added a logging statement.

recipes/llvm-bazel-cache/meta.yaml
xhochy: Done in https://github.com/conda-forge/staged-recipes/pull/24845/commits/be228452be3cc72bbfb6e6cee0cf1790d0c9d392

h-vetinari (approved):
Basically LGTM (modulo passing CI; we're cutting it very close with the 6h on osx, but should hopefully be fixable by restarting).

xhochy:
I will do some slight tuning to ensure it stays in the limit and then merge. Thanks for the review!

PR #24864, opened 2023-12-27
Added audio-separator recipe

beveradb:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/audio-separator/conda-forge.yml
+++ b/recipes/audio-separator/conda-forge.yml
@@ -1,8 +1,4 @@
 channel_priority: strict
-azure:
-  settings_win:
-    pool:
-      vmImage: windows-2019
 channels:
   sources:
     - pytorch
--- a/recipes/audio-separator/conda-forge.yml
+++ b/recipes/audio-separator/conda-forge.yml
@@ -1,0 +1,6 @@
+channels:
+  sources:
+    - pytorch
+    - conda-forge
+  targets:
+    - ["conda-forge", "main"]
--- a/recipes/audio-separator/conda-forge.yml
+++ b/recipes/audio-separator/conda-forge.yml
@@ -1,5 +1,4 @@
 channel_priority: strict
 channels:
   sources:
-    - pytorch
     - conda-forge
--- a/recipes/audio-separator/conda-forge.yml
+++ b/recipes/audio-separator/conda-forge.yml
@@ -1,6 +1,9 @@
+channel_priority: strict
+azure:
+  settings_win:
+    pool:
+      vmImage: windows-2019
 channels:
   sources:
     - pytorch
     - conda-forge
-  targets:
-    - ["conda-forge", "main"]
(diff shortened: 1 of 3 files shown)


beveradb:

I confirm I'm willing to be listed as a maintainer :) 

(also, thank you for your time, happy festive period! etc. 🎄 )

beveradb:
I'd been trying various things to get the windows CI build to work (it fails because `nothing provides requested pytorch` and I'm not sure if it's possible to install pytorch from the `pytorch` channel just for the windows build)

I guess that's not necessary though, as the [guidelines](https://conda-forge.org/docs/orga/guidelines.html) say 
> noarch: python can fail in platforms other than Linux (e.g. missing dependency)

So, I think this is ready to go now even though the win_64 build is failing due to the missing pytorch.
The important thing is that the `cuda` and `cpu` versions of pytorch and onnxruntime are correctly included on linux, and they are! 🎉 

@conda-forge-admin, please ping @conda-forge/help-python

beveradb:

Ready for review, thanks!

recipes/audio-separator/conda-forge.yml
```diff
@@ -0,0 +1,5 @@
+channel_priority: strict
+channels:
+  sources:
+    - pytorch
```
ocefpaf, line 4: This would create some complications for conda-forge users b/c we do not recommend folks to mix channels and yet this package would only work with another channel in their setup.

beveradb: Thanks for taking a look at this @ocefpaf !

I've pushed a commit removing this, but the reason I had this here is that conda-forge is missing a `win-64` pytorch build and seems to be outdated compared to the pytorch channel (plus, the pytorch channel is what pytorch themselves tell people to use: https://pytorch.org/get-started/locally/)

Would you happen to know of a good example feedstock which requires pytorch & torchaudio as dependencies from conda-forge?

I've been really struggling to figure out the _right_ way to get this working, and every example I've found myself (just using github code search with various combinations of filters trying to find feedstock repos) has been frustratingly varied in approach. Any assistance greatly appreciated, I just want audio-separator to be available to people on conda-forge!

Thank you!

ocefpaf: Failing on Windows is fine and folks should be able to add the pytorch channel on their setup at their own discretion. Things should work but we, conda-forge, don't recommend it b/c it can mess up new users setups.

beveradb:
What needs to happen for the feedstock to get converted, @ocefpaf ?
I thought based on the 10 minute cron schedule GHA workflow it would have happened by now but while the repo has been created, the recipe still seems to be in `main`

https://github.com/conda-forge/audio-separator-feedstock 

Apologies if this is an unnecessary ping and I should just be patient, I'm just trying to understand the process as it's my first time publishing to conda-forge!

BastianZim:
Something is indeed wrong, and this needs manual intervention. Normally feedstock creations can take up to 24h. 
@conda-forge/core could someone have a look at why the conversion is failing? Thanks!

ocefpaf:
Should be fixed now. See https://github.com/conda-forge/audio-separator-feedstock

xref.: https://github.com/conda-forge/staged-recipes/pull/24969

beveradb:
Thank you both!

PR #24876, opened 2024-01-01
Add `q` (CLT to mimic SQL in the terminal)

Sann5:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/q/meta.yaml
+++ b/recipes/q/meta.yaml
@@ -1,0 +1,47 @@
+{% set name = "q" %}
+{% set version = "3.1.6" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/harelba/q/archive/v{{ version }}.tar.gz
+  sha256: e63ba4ae49647f764c5255ad7065d2c614fdf03a2f7349a795de69529701fab8
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python
+    - pip
+  run:
+    - python
+
+test:
+  requires:
+    - pip
+  commands:
+    - pip check
+    - q -h
+
+about:
+  home: https://github.com/harelba/q
+  summary: 'Run SQL directly on delimited files and multi-file sqlite databases'
+  description: |
+    q's purpose is to bring SQL expressive power to the Linux command line and to provide easy access to text as actual data. It allows 1)
+    performing SQL-like statements directly on tabular text data, auto-caching the data in order to accelerate additional querying on the same file, and 2)
+    performing SQL statements directly on multi-file sqlite3 databases, without having to merge them or load them into memory.
+  license: GPL-3.0-only
+  license_family: GPL
+  license_file: LICENSE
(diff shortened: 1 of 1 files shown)


dbast:
can this be `noarch: generic` and depend on `__unix` as runtime dependency, so it is only installable on unix?

Sann5:
hey @dbast. thanks for your attention. I'm still working on passing a few mysterious tests that fail. I will try out your suggestion once this is done :)

Sann5:
Hey @conda-forge/staged-recipes @conda-forge/help-python , haven't heard from you in a while.

Sann5:
@conda-forge/staged-recipes @conda-forge/help-python

Sann5:
@dbast

BastianZim:
From what I can see your package looks fine but I am slightly worried about merging a package that is only called q. @conda-forge/core would that cause any problems or any other considerations?

jaimergp:
Yep, we don't have a big precedent for single-letter packages... I think only `r`? There are a few with two letters. But if `r` exists then it must work so I guess it's ok proceed.

recipes/q/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+source:
+  url: https://github.com/Sann5/q/archive/pack_issue_326.zip
+  sha256: 420e92546a4ba42989be609768c67d636097bc2d5054756c7381d7c766a30533
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
+  skip: true  # [py>=311]
```
jaimergp, line 16: I know this was brought up before, but this could be replaced with the following run requirements:

```yaml
run:
  - python >=3.8,<3.11.0dev0
  - __unix
  - six
```

Sann5: Im not sure exactly what you mean. Can you make the change suggestion with the review tool so I can see exactly what you propose? Also then I can quickly accept it if it looks reasonable.

jaimergp: Yes. There's no reason to not make this package `noarch`. We let it fail on Windows and then constrain `python` so 3.11 is not supported. Let me add a couple comments.

recipes/q/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+requirements:
+  host:
+    - python
```
jaimergp, line 20:
```suggestion
    - python >=3.8,<3.11.0dev0
```

recipes/q/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - pip
+  run:
+    - python
```
jaimergp, line 23:
```suggestion
    - python >=3.8,<3.11.0dev0
```

recipes/q/meta.yaml
```diff
@@ -0,0 +1,56 @@
+{% set name = "q" %}
+{% set version = "3.1.6" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/Sann5/q/archive/pack_issue_326.zip
```
jaimergp, line 9: Is this the correct URL? We usually get the archive from the tagged version of the official release. This looks more like a fork?

PR #24898, opened 2024-01-02
Add fixit

apmorton:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/fixit/meta.yaml
+++ b/recipes/fixit/meta.yaml
@@ -10,25 +10,25 @@ source:
   sha256: b31665cb6491d659d8dfef5a6078a7e9f786e299826636d03d6bd91b6f71e95b
 
 build:
-  skip: true  # [py<38]
   entry_points:
     - fixit = fixit.cli:main
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  noarch: python
   number: 0
 
 requirements:
   host:
-    - python
+    - python >= 3.8
     - hatchling
     - hatch-vcs
     - pip
   run:
-    - python
+    - python >= 3.8
     - click >=8.0
     - libcst >=0.3.18
     - moreorless >=0.4.0
     - packaging >=21
-    - tomli >=2.0  # [py<311]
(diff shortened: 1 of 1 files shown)


apmorton:

(windows failure appears systemic)

recipes/fixit/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - hatchling
+    - hatch-vcs
+    - pip
+  run:
+    - python
+    - click >=8.0
+    - libcst >=0.3.18
+    - moreorless >=0.4.0
+    - packaging >=21
+    - tomli >=2.0  # [py<311]
```
ocefpaf, line 31: It can probably be noarch if we are OK installing `tomli` in >=311. What do you think?

apmorton: I originally did that, but decided against it because it is a genuine unused dependency.

(as opposed to "try using tomli if there, otherwise use tomllib").

ocefpaf: We are between an unused dependency vs several artifacts that could be one and save CI time. In conda-forge we are usually biased to avoid the latter.

apmorton: If you insist, I will change this - but this policy does seem at-odds with the defaults in grayskull, which states:

> The main goal of this project is to generate concise recipes for conda-forge

ocefpaf: Sure. Grayskull is one thing, conda-forge is another. Also, note that grayskull has a conda-forge strict option that tries to make it closer to our latest policies, it may be outdated though.

apmorton: I generated this recipe using `mamba greyskull pypi --strict-conda-forge fixit`

ocefpaf: Yep, that is the option but, as I mentioned above, it is probably a bit outdated. It is hard to capture the latest policies. Well, we don't even write everything down b/c they change to quickly :-)

TL;DR ideally we should have a semi-arch for Python like we have for OS-noarch now.

PR #25026, opened 2024-01-13
Add CUDA 11.8

jakirkham:
Now that CUDA 11.8 is included in the global pinnings. Add it to staged-recipes to ensure it is tested as well.

xref: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/5340

PR #25027, opened 2024-01-13
[NO MRG] Test CUDA 11.8

jakirkham:
Test out the CUDA 11.8 CI job addition in PR ( https://github.com/conda-forge/staged-recipes/pull/25026 ) with a recipe that builds on CUDA 11.8


jakirkham:
The linter error is expected as this is reusing an existing conda-forge recipe for testing

Also on CI only the CUDA 11.8 build will meaningfully run for this recipe. All other CI jobs will fast finish

jakirkham:
Seems to be working as expected. Will let it run to completion

jakirkham:
It passed. Closing as testing is complete

jakirkham:
From a `staged-recipes` perspective this is doing the right thing

Though the recipe itself appears to have an issue. Namely the source comes from a GH archive and the checksum of that archive has changed (due to the fact GH archives are dynamically generated). Will debug that separately in PR ( https://github.com/conda-forge/cuda-python-feedstock/pull/79 ) and update here after

That said, not expecting any issues with CUDA 11.8. So think we can go ahead while this is sorted out

PR #25040, opened 2024-01-15
Add pybind11-numpy-example recipe

lkeegan:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pybind11-numpy-example/meta.yaml
+++ b/recipes/pybind11-numpy-example/meta.yaml
@@ -22,11 +22,11 @@ requirements:
     - ninja
     - pip
     - pybind11
-    - python >=3.7
+    - python
     - scikit-build-core
   run:
     - numpy
-    - python >=3.7
+    - python
 
 test:
   imports:
--- a/recipes/pybind11-numpy-example/meta.yaml
+++ b/recipes/pybind11-numpy-example/meta.yaml
@@ -1,0 +1,53 @@
+{% set name = "pybind11-numpy-example" %}
+{% set version = "0.0.11" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pybind11_numpy_example-{{ version }}.tar.gz
+  sha256: 9bab0607c9697673fd1bd6dc3b09689a10cbd8c908aa062298b4ba3801c224b4
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+  host:
+    - cmake
+    - make
+    - ninja
+    - pip
+    - pybind11
+    - python >=3.7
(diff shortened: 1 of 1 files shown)


lkeegan:

carterbox:
@lkeegan, why do you want to publish a conda package of a source code demo? Conda packages are prebuilt binaries. Wouldn't the target audience of this project want to compile the project themselves?

recipes/pybind11-numpy-example/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - make
+    - ninja
+  host:
+    - numpy
+    - pip
+    - pybind11
+    - python
+    - scikit-build-core
+  run:
+    - python
```
carterbox, line 30: The build section is for build tools and the stdlib template.

The host section is for things that must match the target platform when cross-compiling.

I have moved numpy to the host section because you probably use the numpy headers during compiling and need the ABI to match at build and runtime.

carterbox (changes requested):
It looks like you are shipping static libs. We don't allow this because they waste space.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=939539&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=904

recipes/pybind11-numpy-example/meta.yaml
lkeegan: thanks!

lkeegan:
> It looks like you are shipping static libs. We don't allow this because they waste space.
> 
> https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=939539&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=904
> 
> Also, please explain why you need to ship headers and cmake config files for a python extension module? These file should not be needed for python extensions whose purpose is to be loaded by the python intepreter. Additionally, you should not be installing "include", "lib" folders to the site-packages directory.

You're right the headers, cmake config, static lib etc shouldn't be part of the python module installation - thanks I'll fix this upstream.

lkeegan:
@carterbox the issue with shipping a static lib, headers, cmake config, etc should now be fixed

lkeegan:
@carterbox as far as I can tell the warnings are harmless - scikit-build-core doesn't allow `CMAKE_BUILD_TYPE` or `CMAKE_INSTALL_PREFIX` to be overridden from the `CMAKE_ARGS` environment variable:

https://github.com/scikit-build/scikit-build-core/commit/b38f67687e9932badf46f3a18b6b39641a5856c4

but these variables are set elsewhere, and it does respect all the other defines in `CMAKE_ARGS` (and from looking through the issues it seems this behaviour was added specifically to support building on conda-forge: https://github.com/scikit-build/scikit-build/issues/588, https://github.com/scikit-build/scikit-build/issues/722)

Thanks again for the careful review!

PR #25061, opened 2024-01-18
Add bip-utils

MementoRC:
Checklist
- [x] Build number is 0.


MementoRC:
@carterbox This package has issues with the available packages on conda-forge. I requested a re-populate of `pycryptodome`, but `nacl` has the same issue, so I ask the author to consider bumping his reqs, no response yet  - though he already updated `coincurve` per my previous request

MementoRC:
@carterbox I discussed with the original author and he made changes to his requirements, however, after a few iterations he actually made the requirements impossible to be reproduced within conda-forge (for ex, he requested older versions of coincurve for py<3.11). So, it felt that the simpler solution was to fork and maintain a conda-friendly version of the package

carterbox:
I haven't looked into details of how your fork changes from upstream, but if you're only changing the requirements file, you don't need a fork, use the upstream source just carry the patch in this recipe. Carrying patches in the recipe makes it easier for downstream users to see exactly how this build of bip is different from the canonical one.

I don't think you have provided any reason why you would need to make significant changes to the bip API or implementation, so if your fork makes any changes beyond just making things buildable or the deps solvable, I will need to take a closer look.

In summary, it's fine to carry patches in conda recipes which make minor modifications to the package metadata or build scripts. Anything more than that should definitely be merged upstream.

MementoRC:
@carterbox That is mostly the case, however, while building and testing the recipe locally, I kept running into a odd issue that (as far as I understand it) was caused by the `ecdsa` class being picked instead ot the internal bip-utils class. So I refactored the package into a src-layout, so this creates a huge amount of apparent changes

It is a bit hard to work with upstream maintainers (or at least, I don't do a good job at it) who do not care or even know about conda. In this particular case, although I tried to clearly indicate that `coincurve` <19 does not exists, he ended-up modifying the requirements in basically an opposite way by even requiring coincurve 15 for some version of python. I understand his point of view, it's just that it goes contrary to when I need for conda package

carterbox:
> So I refactored the package into a src-layout, so this creates a huge amount of apparent changes

This shouldn't be necessary because (as far as I understand) `ecdsa` is a separate package. Because it is a separate package, the source layout should have no effect on the installed package.

MementoRC:
The issue I was seeing were fails of `bip-utils` tests with `PointJacobi` has no member `from_bytes` as I was changing the version of `ecdsa` in the recipe in order to tell the author which version combination would work for conda-forge. Looking at `bip-utils` classes, the `PointJacobi` has a `fomr_bytes` definition, so it seemed to indicate that when running the tests things got mixed-up.

So, since I thought there werre little chances to align with the author on the requirements that he prefers for his pip-based package, I would fork, then why not implement the src-layout that seems to be cleaner way of packaging - though I am flexible, ofc

carterbox:
> So, since I thought there werre little chances to align with the author on the requirements that he prefers for his pip-based package, I would fork, then why not implement the src-layout that seems to be cleaner way of packaging - though I am flexible, ofc

STOP making personal forks and think about what type of packages you are bringing to conda-forge. It is an ecosystem of *cyptography* packages. These packages need to be as close to the canonical package as possible with minimal, transparent, easily-understandable, and tested changes. Users of these packages value security and stability.

If a new person comes along and wants to download bip-utils from conda forge, but instead of getting the canonical package, they get something built from your personal fork with a large git diff? It looks like you're trying to obfuscate adding security vulnerabilities to their dependency tree.

MementoRC:
@carterbox The errors no longer appear in this minimally patched requirements. It may be ready for review

For reference the failed tests: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=870460&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=3352

recipes/ed25519-blake2b/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - ed25519_blake2b
+  requires:
+    - pip
+    - python
+  source_files:
+    - test_ed25519_kat.py
+    - kat-ed25519-blake2b.txt
+  commands:
+    - pip check
+    - python test_ed25519_kat.py
```
carterbox, line 36:
```suggestion
  commands:
    - pip check
    - edsig
    - python test_ed25519_kat.py
```

carterbox (approved):
LGTM if tests pass.

recipes/ed25519-blake2b/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - ed25519_blake2b
+  requires:
+    - pip
+    - python
+  source_files:
+    - test_ed25519_kat.py
+    - kat-ed25519-blake2b.txt
+  commands:
+    - pip check
+    - edsig
```
carterbox, line 36: This command line script is installed with the package, but it is broken because it relies on the ed25519 package. If ed25519-blake2b is supposed to be co-installable with ed25519, then this script should be removed so there are not clobberings with the ed25519 package. If they are not to be con-installable, then this recipe needs to be run_constrained on `ed25519 ==999999`.

carterbox (changes requested):
Fix broken script or remove it?

recipes/ed25519-blake2b/meta.yaml
MementoRC: I was just looking into it. This is a fork of `python-ed25519` and it does not update the `bin` dir. It seems that it should be removed or updated.
When you say 'should be removed', you're referring to upstream, correct?
I will look into it.

I did not think to test scripts, this is a good lesson on recipe crafting, hope I'll remember and be more thorough

MementoRC: Is it ok to add `ed25519` recipe to this PR?

carterbox: You can remove the edsig file from the install list with a small patch to the setup.py. I believe that /bin/edsig exists in both ed25519-blake2b and ed25519, so they would clobber if you install both.

carterbox: You could also add a small patch which fixes the imports in /bin/edsig. I'm not sure which is more appropriate.

MementoRC: Right, I am testing a patch that changes the import and the name

MementoRC: Looks like it is viable. I went for the patching solution that maintain the script from upstream:
   -  Renamed edsig to edsig_blake2b
   - Patched imports in edsig_blake2b
   - Updated setup.py
   - Updated the README.md

recipes/ed25519-blake2b/meta.yaml
```diff
@@ -33,10 +37,8 @@ test:
     - kat-ed25519-blake2b.txt
   commands:
     - pip check
-    - edsig
+    - edsig_blake2b
     - python test_ed25519_kat.py
-  downstreams:
-    - bip-utils
```
carterbox, line 42: Why is bip-utils no longer downstream?

recipes/ed25519-blake2b/1_rename-edsig-edsig_blake2b.patch
```diff
@@ -0,0 +1,188 @@
+--- bin/edsig      2024-02-16 16:57:39.437391589 -0600
```
carterbox, line 1: Is the diff smaller if you combine patches 1 and 2? (the rename and the replacement of edsig with edsig_blake2b)?

carterbox (changes requested):

recipes/ed25519-blake2b/1_rename-edsig-edsig_blake2b.patch
MementoRC: I tried but it did not work, it complained that I was deleting a file that did not exist, but I could have mistyped something. That was my preferred route s well, let me take a look again

recipes/ed25519-blake2b/meta.yaml
MementoRC: It looks like a bad copy-paste

MementoRC: Related Q: Is it a good practice to add the downsstreams? meaning, when I see a dependence is it a good practice to go into the dependent feedstock and add the new downstream?

recipes/ed25519-blake2b/1_rename-edsig-edsig_blake2b.patch
MementoRC:
```
Applying patch: /home/conda/staged-recipes-copy/recipes/ed25519-blake2b/2_fix-edsig_blake2b.patch
Applying patch: /home/conda/staged-recipes-copy/recipes/ed25519-blake2b/2_fix-edsig_blake2b.patch with args:
['-Np2', '-i', '/tmp/tmppivo3ljo/2_fix-edsig_blake2b.patch.native', '--binary']
checking file bin/edsig
patching file bin/edsig
```
then
`error: [Errno 2] No such file or directory: 'bin/edsig_blake2b'`

Oh, maybe the rename should be empty?

MementoRC: Ok, got it, you're point is that since I have a big patch anyway, I should skip `2_...`. I thought it'd be clearer to verify that `1_` does not change anything but the file name, then `2_...` shows the changes

PR #25078, opened 2024-01-19
add perl-compress-bgzf

jvolkening:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/perl-compress-bgzf/meta.yaml
+++ b/recipes/perl-compress-bgzf/meta.yaml
@@ -22,9 +22,11 @@ build:
 requirements:
   build:
     - perl
+    - perl-module-build
 
   host:
     - perl
+    - perl-module-build
 
   run:
     - perl
--- a/recipes/perl-compress-bgzf/meta.yaml
+++ b/recipes/perl-compress-bgzf/meta.yaml
@@ -12,17 +12,14 @@ source:
 
 build:
   number: 0
-  noarch: generic
   script:
-    - perl Build.PL --installdirs vendor --create_packlist 0
-    - ./Build
-    - ./Build test
-    - ./Build install --verbose
+    - perl Build.PL 
+    - perl ./Build
+    - perl ./Build test
+    - perl ./Build install --installdirs vendor --create_packlist 0
 
 requirements:
   build:
-    - perl
-    - perl-module-build
 
   host:
     - perl
--- a/recipes/perl-compress-bgzf/meta.yaml
+++ b/recipes/perl-compress-bgzf/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "perl-compress-bgzf" %}
+{% set version = "0.006" %}
+{% set sha256 = "fb9716ba560a19d2cf29f385f5bf76d9c362fb71ba9df1ada244450230b1da67" %}
+
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://cpan.metacpan.org/authors/id/V/VO/VOLKENING/Compress-BGZF-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  noarch: generic
+  script:
+    - perl Build.PL --installdirs vendor --create_packlist 0
+    - ./Build
+    - ./Build test
+    - ./Build install --verbose
+
+requirements:
+  build:
+    - perl
+
+  host:
+    - perl
+
+  run:
+    - perl
+
+test:
+  imports:
+    - Compress::BGZF
+    - Compress::BGZF::Reader
+    - Compress::BGZF::Writer
+
+about:
+  home: http://metacpan.org/pod/Compress-BGZF
+  summary: 'Read/write blocked GZIP (BGZF) files'
+  license: GPL-3.0-or-later
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - jvolkening
--- a/recipes/perl-compress-bgzf/meta.yaml
+++ b/recipes/perl-compress-bgzf/meta.yaml
@@ -12,6 +12,8 @@ source:
 
 build:
   number: 0
+  noarch: generic
+  skip: true  # [win]
   script:
     - perl Build.PL 
     - perl ./Build
--- a/recipes/perl-compress-bgzf/meta.yaml
+++ b/recipes/perl-compress-bgzf/meta.yaml
@@ -21,12 +21,9 @@ build:
     - perl ./Build install --installdirs vendor --create_packlist 0
 
 requirements:
-  build:
-
   host:
     - perl
     - perl-module-build
-
   run:
     - perl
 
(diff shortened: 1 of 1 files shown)


jvolkening:
@conda-forge/help-perl Trying to migrate the first of my packages from bioconda, and can't get past this build failure with this package using `Module::Build`. As far as I can tell, the  build goes fine but no files are actually found/packaged. Relevant lines from the build log appear to be:
```
...
Result: PASS
Building Compress-BGZF                                                                           
Installing $BUILD_PREFIX/lib/perl5/vendor_perl/Compress/BGZF.pm                                  
Installing $BUILD_PREFIX/lib/perl5/vendor_perl/Compress/BGZF/Reader.pm                           
Installing $BUILD_PREFIX/lib/perl5/vendor_perl/Compress/BGZF/Writer.pm                           
Installing $BUILD_PREFIX/man/man3/Compress::BGZF::Writer.3                                       
Installing $BUILD_PREFIX/man/man3/Compress::BGZF::Reader.3                                       
Installing $BUILD_PREFIX/man/man3/Compress::BGZF.3                                               
Installing $BUILD_PREFIX/bin/bgzip.pl                                                            
                                                                                                 
Resource usage statistics from building perl-compress-bgzf:                                      
   Process count: 3                                                                              
   CPU time: Sys=0:00:00.0, User=0:00:00.2                                                       
   Memory: 28.0M                                                                                 
   Disk usage: 1.0K                                                                              
   Time elapsed: 0:00:04.0                                                                       
                                                                                                 
                                                                                                 
Packaging perl-compress-bgzf                                                                     
Packaging perl-compress-bgzf-0.006-pl5321hd8ed1ab_0                                              
WARNING: No files or script found for output perl-compress-bgzf                                  
number of files: 0
...
```
Any advice? I can't find any other `perl-*` packages on conda-forge which use Module::Build, but it's not a particularly obscure packager.

jvolkening:
Looks like @conda-forge/help-perl team is empty, so pinging @conda-forge/perl for any advice on above.

mfansler:
Not a Perl expert, but looks like it's installing into `BUILD_PREFIX` instead of `PREFIX`. Probably shouldn't have `perl` as a `build` dependency, but just `host` (and `run`). This also would be consistent with what `conda skeleton cpan Compress::BGZF` generates.

Have you tried just the `conda skeleton cpan Compress::BGZF` recipe?

jvolkening:
Thanks @mfansler, good idea and the recipe built by `conda skeleton cpan` works. There are some parts of it that I don't think are in line with the latest recommendations, but I was able to compare the differences and narrow the issue down to the inclusion of `noarch: generic` in the recipe. The `conda skeleton cpan` does not add this, although my understanding is that it is the correct thing to do for pure-perl packages. If I remove it from my own recipe, the build works. If I add it to the skeleton recipe, it fails. When `noarch: generic` is included, the actual build succeeds but the tests fail (in this case, only trying to import the modules which can't be found).

What makes this more confusing is that for the existing conda-forge perl recipes, all of which are for packages using the `ExtUtils::MakeMaker` build system, at least some are marked `noarch: generic` and it doesn't seem to cause a problem (e.g. [perl-text-csv](https://github.com/conda-forge/perl-text-csv-feedstock/)). So, it seems to be specific to the use of `Module::Build`. In theory, both systems should be installing the files to the same locations.

I just saw that the win64 build failed because `perl-module-build` is not build for win64. This is another issue that I will ask about on that feedstock (is there any reason to skip Windows build).

jvolkening:

mfansler:
> *"...narrow the issue down to the inclusion of `noarch: generic` in the recipe. The `conda skeleton cpan` does not add this, although my understanding is that it is the correct thing to do for pure-perl packages. If I remove it from my own recipe, the build works. If I add it to the skeleton recipe, it fails. When `noarch: generic` is included, the actual build succeeds but the tests fail (in this case, only trying to import the modules which can't be found).*"

I can't replicate this problem. Running the `linux-64` build locally on the current recipe with `noarch: generic` works just fine for me. I'd suggest adding the `noarch` back here and seeing where it goes. We only need `linux-64` to pass.

jvolkening:
> I can't replicate this problem. Running the `linux-64` build locally on the current recipe with `noarch: generic` works just fine for me. I'd suggest adding the `noarch` back here and seeing where it goes. We only need `linux-64` to pass.

Yes, you beat me to it. Maybe my issues were a result of cached build files, but when I removed the build artifacts and re-tried with `noarch: generic` it worked. More testing suggests that your original comment points to the problem, which is including `perl` as a `build` dependency. I've read the "Defining metadata" docs and wish I understood better why this was, but at least it fixes the immediate issue.

recipes/perl-compress-bgzf/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+requirements:
+  build:
+
+  host:
+    - perl
+    - perl-module-build
+
+  run:
+    - perl
```
mfansler, line 31: I suggest we tidy this up:

```suggestion
  host:
    - perl
    - perl-module-build
  run:
    - perl
```

recipes/perl-compress-bgzf/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+test:
+  imports:
+    - Compress::BGZF
+    - Compress::BGZF::Reader
+    - Compress::BGZF::Writer
+
+about:
+  home: http://metacpan.org/pod/Compress-BGZF
```
mfansler, line 40: That link gave "not found". Perhaps it should be...

```suggestion
  home: https://metacpan.org/dist/Compress-BGZF
```

mfansler (changes requested):
A couple suggestions, otherwise looks ready!

recipes/perl-compress-bgzf/meta.yaml
jvolkening: Thanks, pushed a commit with a fixed URL (slightly different than suggested). I will be more careful to check these in future.

recipes/perl-compress-bgzf/meta.yaml
jvolkening: Thanks, just pushed a commit with the suggested changes.

mfansler (approved):

mfansler:
Thanks for taking the time to transfer this! ❤️ 

Feel free to tag me if any other Bioconda transfers need a review or troubleshooting.

jvolkening:
@mfansler Thanks again for your guidance. This was one of a few "test runs" I'm doing to help me understand the nuances of conda-forge, but I have plans to try to do more of a "bulk" migration as I become more comfortable. I have opened up an issue [here](https://github.com/conda-forge/staged-recipes/issues/25110) to collect notes related to issues/questions I run into, so that I can use best practices out of the gate. If you have any guidance on the points in that post, hopefully I can understand/fix some of these things locally before opening up more PRs.

PR #25131, opened 2024-01-24
add litestar 2.8.0

bollwyvl:
Checklist
- [x] Build number is 0.

References:
- fixes https://github.com/conda-forge/starlite-feedstock/issues/47
- unblocks https://github.com/conda-forge/strawberry-graphql-feedstock/pull/162

Notes:
- `win-64` builds will fail on `-with-cli` and `-standard` due to lack of upstream `uvloop` support for that platform 
  - we could code around it with `noarch_platforms` or something, but doesn't really seem worth it


bollwyvl:
@conda-forge/starlite please comment here if you'd like to co-maintain.

thewchan:
@bollwyl sure happy to help

bollwyvl:
@thewchan thanks, will update when i resolve `piccolo -> targ -> ???`

bollwyvl:
yeah, `piccolo` is a mess, not dealing with it now

bollwyvl:
Also the `pydantic-extra-types` stack is missing a bunch of stuff:

```
-   - nothing provides requested pycountry >=23,<24
-   - nothing provides requested python-ulid >=1,<2
-   - nothing provides pycountry >=23,<24 needed by pydantic-extra-types-with-all-2.5.0-py_0
```

bollwyvl:
@conda-forge/staged-recipes @conda-forge/help-python ready for review

ocefpaf:
@bollwyvl I understand that you want to bring your own tooling for maintaining the recipe and I know you will probably have good arguments to do so. However, we cannot handle that and the long lasting effects are unpredictable.  Even though you use a different name for the template recipe, that can be confusing for existing (and future) parsing tools in conda-forge.

Do you mind adapting that recipe to the custom format we support and remove the extra stuff?

bollwyvl:
> the custom format we support 

Happy to name the stuff whatever, put it wherever: is there some docs or prior art for examples?

ocefpaf:
> Happy to name the stuff whatever, put it wherever: is there some docs or prior art for examples?

We don't have prior art in what you are trying to achieve. The core group did find two more recipes that ended up passing to the feedstock with those scripts and extra yaml file. See https://github.com/search?q=org%3Aconda-forge%20path%3A*.j2.yaml&type=code

What we ask is that you use just the meta.yaml and remove the extra stuff. We are afraid that extra files that are not supported by conda-forge will cause confusing and can even break future iterations of our infrastructure.

I know that this is frustrating but we are few and it is hard to accommodate custom stuff, specially if all maintainers start building their own!

With all that said, if you believe what you are doing there could benefit the larger community, please get in touch with the bot team and work with them to see if we can assimilate those ideas.

bollwyvl:
What if I just call it `test_recipe.py`, which _happens_ to contain a jinja template and _happens_ to also have a meaning when invoked directly with `python test_recipe.py --update`?

ocefpaf:
> What if I just call it `test_recipe.py`, which _happens_ to contain a jinja template and _happens_ to also have a meaning when invoked directly with `python test_recipe.py --update`?

We would need some agreement with the @conda-forge/core group. I'm inclined to reduce the complexity if possible.

recipes/litestar/meta.yaml
```diff
@@ -0,0 +1,536 @@
...
+      home: https://litestar.dev
+      dev_url: https://github.com/litestar-org/litestar
+      doc_url: https://docs.litestar.dev
+      summary: Light-weight and flexible ASGI API Framework
+      license: MIT
+      license_file: LICENSE
+
+  - name: litestar-with-annotated-types
+    build:
+      noarch: python
```
carterbox, line 73: I believe that all of these metapackages can be `noarch: generic` instead of `noarch: python` because they are empty packages which do not contain any python modules (they only depend on other packages). Thus, they also do not need a host section or a runtime dependency on python since the litestar base package already constrains the python version.

Please try this, as it will shorten the length of this recipe and probably shorten build times.

carterbox (changes requested):

recipes/litestar/meta.yaml
bollwyvl:
```
Attempting to finalize metadata for litestar-with-annotated-types
Error: Failed to render jinja template in /home/conda/staged-recipes-copy/recipes/litestar/meta.yaml:
'PYTHON' is undefined
```

bollwyvl: So the above _could_ perhaps be avoided by splitting into `litestar-feedstock` and a sibling `litestar-extras-feedstock`, both cut from the same version of the upstream sdist. We ended up going this way when splitting `dagster-(libs-)feedstock`.

carterbox: This error is because the recipe is rerendered for every output, but PYTHON is only defined when the current output uses python. So when rerendering for the generic outputs, jinja doesn't know what to do when rendering the part of the for the python parts. However, it doesn't need to know the value of PYTHON becuase that part of the recipe will be discarded anyways.

We can either use Jinja to define PYTHON = 'some-placeholder-so-jinja-does-not-error' when PYTHON is undefined, or we can just use the environment variable syntax in the script. Typically we use `script: {{ PYTHON }}` because it works for both unix and Windows, but here the recipe will be built once on Linux, so we can use `script: $PYTHON`.

carterbox:
The Windows build will fail, and that's probably OK.

carterbox (approved):

recipes/litestar/meta.yaml
bollwyvl: Sure, even though it's irrelevant whether it builds on windows, I've added the `if PYTHON is not defined` bit, with a note about `noarch: generic`.

bollwyvl:
@conda-forge/staged-recipes ready for re-review

thewchan:
Some other package now depends on this https://github.com/conda-forge/evidently-feedstock/pull/33

ocefpaf:
@conda-forge/core what do you think about this way for adding the recipe generator?

bollwyvl:
Updated. _Still_ anticipating fails on `win-64`, and can take the recipe generator out entirely, just using it in my local repo, if this is really a big sticking point.

thewchan:
Any update on this? @bollwyvl @conda-forge/core

bollwyvl:
Thanks all.

PR #25132, opened 2024-01-24
Add recipe for Pink

stephane-caron:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pink/meta.yaml
+++ b/recipes/pink/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/pin-{{ name }}/pin_{{name}}-{{ version }}.tar.gz
+  url: https://pypi.io/packages/source/{{ name[0] }}/pin-{{ name }}/pin_{{ name }}-{{ version }}.tar.gz
   sha256: daa245925c739d69a14ce2c4b80f9f3758648b7c6f4987b392a78692e6bc67f1
 
 build:
--- a/recipes/pink/meta.yaml
+++ b/recipes/pink/meta.yaml
@@ -32,9 +32,6 @@ requirements:
 test:
   imports:
     - pink
-  commands:
-    - python -c "import pinocchio as pin; print(pin)"
-    - python -c "import pink; print(pink)"
 
 about:
   home: https://pypi.org/project/pin-pink
--- a/recipes/pink/meta.yaml
+++ b/recipes/pink/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/pin_pink-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


stephane-caron:
I'm willing to be listed as a maintainer.

stephane-caron:

recipes/pink/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - proxsuite
+    - qpsolvers >=2.5.0
+    - quadprog >=0.1.11
+
+test:
+  imports:
+    - pink
+  commands:
+    - python -c "import pinocchio as pin; print(pin)"
+    - python -c "import pink; print(pink)"
```
ocefpaf, line 37: You probably don't need these. One is testing a dependency and the other one is redundant with the test import above.

```suggestion
```

stephane-caron: Got it, thank you!

recipes/pink/meta.yaml
```diff
@@ -0,0 +1,45 @@
+{% set name = "pink" %}
+{% set version = "1.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/pin-{{ name }}/pin_{{ name }}-{{ version }}.tar.gz
```
ocefpaf, line 9: The PyPI name is pin-pink, right? What is the reason to rename it to just pink?

stephane-caron: This is related to the main dependency of this project, which is called Pinocchio but whose PyPI package is just `pin` (because `pinocchio` was already taken I guess).

The same happened with this project: it is called Pink, but there was already a package named `pink` on PyPI (which seems inactive). We then named the PyPI package `pin-pink`, adding the `pin` prefix in reference to Pinocchio.

ocefpaf:
We prefer to keep the names as close to PyPI as possible to avoid confusion. If you are the upstream developer and you prefer otherwise we can make an exception but it will be confusing for users who have to swtich from PyPI to conda in their day jobs (which is quite common BTW).

stephane-caron:
I understand the rationale for trying to keep PyPI and conda-forge in sync where possible. In this project, our users are rather students and researchers, and the `pin-pink` name on PyPI in itself already creates some confusion. The typical example is: "I installed pink but it doesn't work!" to which we check "did you `pip install pin-pink` and not `pink`?", and that's usually the issue...

I am the upstream maintainer, and for this reason I would indeed prefer we make an exception :smiley: The plan is that we will incent users to use Conda rather than PyPI (there are also performance reasons to do so), so I'd rather not we carry over the historical naming from PyPI.

ocefpaf:
> I understand the rationale for trying to keep PyPI and conda-forge in sync where possible. In this project, our users are rather students and researchers, and the `pin-pink` name on PyPI in itself already creates some confusion. The typical example is: "I installed pink but it doesn't work!" to which we check "did you `pip install pin-pink` and not `pink`?", and that's usually the issue...
> 
> I am the upstream maintainer, and for this reason I would indeed prefer we make an exception 😃 The plan is that we will incent users to use Conda rather than PyPI (there are also performance reasons to do so), and I'd rather we don't carry over the historical naming from PyPI.

That is all I need to know 👍🏽

PR #25137, opened 2024-01-25
perl-number-format migration from bioconda

jvolkening:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -13,16 +13,16 @@ source:
 build:
   number: 0
   noarch: generic
+  skip: true  # [win]
   script:
-    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1  # [not win]
-    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1 MAKE=nmake  # [win]
-    - make && make test && make install VERBINST=1  # [not win]
-    - jom && jom test && jom install VERBINST=1  # [win]
+    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
+    - make
+    - make test
+    - make install VERBINST=1
 
 requirements:
   build:
-    - make  # [not win]
-    - jom  # [win]
+    - make
   host:
     - perl
   run:
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -13,11 +13,14 @@ source:
 build:
   number: 0
   noarch: generic
+  script_env:
+    - MAKE=make  # [not win]
+    - MAKE=gmake  # [win]
   script:
     - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
-    - make
-    - make test
-    - make install VERBINST=1
+    - $MAKE
+    - $MAKE test
+    - $MAKE install VERBINST=1
 
 requirements:
   build:
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -13,16 +13,16 @@ source:
 build:
   number: 0
   noarch: generic
-  script_env:
-    - PERL_MM_OPT='MAKE=nmake'  # [win]
   script:
     - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
-    - make && make test && make install VERBINST=1  # [not win]
-    - nmake && nmake test && nmake install VERBINST=1  # [win]
+    - make
+    - make test
+    - make install VERBINST=1
 
 requirements:
   build:
     - make  # [not win]
+    - m2w64-make  # [win]
   host:
     - perl
   run:
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -1,0 +1,42 @@
+{% set cpan_name = "Number-Format" %}
+{% set version = "1.76" %}
+{% set sha256 = "0e0060eb363635a885706c6a26f5fcaafeae759f7b2acae49dda70e195dd44d6" %}
+
+package:
+  name: perl-{{ cpan_name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://cpan.metacpan.org/authors/id/R/RJ/RJBS/Number-Format-{{ version }}.tar.gz
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  noarch: generic
+  script:
+    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
+    - make
+    - make test
+    - make install VERBINST=1
+
+requirements:
+  build:
+    - make
+  host:
+    - perl
+  run:
+    - perl
+
+test:
+  imports:
+    - Number::Format
+
+about:
+  home: https://metacpan.org/pod/Number::Format
+  summary: Perl extension for formatting numbers
+  license: Artistic-1.0-Perl OR GPL-1.0-or-later
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - jvolkening
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -13,7 +13,6 @@ source:
 build:
   number: 0
   noarch: generic
-  skip: true  # [win]
   script:
     - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
     - make
--- a/recipes/perl-number-format/meta.yaml
+++ b/recipes/perl-number-format/meta.yaml
@@ -13,14 +13,16 @@ source:
 build:
   number: 0
   noarch: generic
+  script_env:
+    - PERL_MM_OPT='MAKE=nmake'  # [win]
   script:
     - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
     - make && make test && make install VERBINST=1  # [not win]
-    - gmake && gmake test && gmake install VERBINST=1  # [win]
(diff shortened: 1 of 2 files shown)


jvolkening:
@mfansler Since you offered, any idea why the Windows builds are failing for this (and also #25136 -- a slightly different error but in almost the same place).

Also, the knowledge base says 

    By default, conda-forge will only build noarch packages on a linux_64 CI runner

Any idea why these are being built on all three platforms in the first place? Many thanks.

jvolkening:
Windows build is failing during the TEST stage, after creating the conda environment, with:
```
(%PREFIX%) %SRC_DIR%>"%PREFIX%\bin/perl" "%SRC_DIR%\run_test.pl" 
The system cannot find the path specified.
```

recipes/perl-number-format/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+  number: 0
+  noarch: generic
+  script:
+    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
+    - make && make test && make install VERBINST=1  # [not win]
+    - gmake && gmake test && gmake install VERBINST=1  # [win]
+
+requirements:
+  build:
+    - make
```
mfansler, line 23: Have you tried the m2sys make?

```suggestion
  script:
    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
    - make && make test && make install VERBINST=1

requirements:
  build:
    - make     # [not win]
    - m2-make  # [win]
```

mfansler:
In the end, it isn't so crucial that the**win-64** build passes. On the feedstock, only **linux-64** will run. Here, the additional builds are useful in that they indirectly give some evidence that the generic package will be functional on other systems.

recipes/perl-number-format/meta.yaml
jvolkening: Yes, actually that was what the first commit used...the CI just hung at the first `make` call, for several hours until I eventually pushed another commit and it canceled.

jvolkening:
> In the end, it isn't so crucial that the **win-64** build passes. On the feedstock, only **linux-64** will run. Here, the additional builds are useful in that they indirectly give some evidence that the `noarch` package will be functional on other systems.

Okay, I was wondering why the other platform builds were running, but it makes sense that this is just for staging.

There are a number of packages that involve compilation, and it would be nice to figure out the Windows build on something simple like this pure-Perl package so those don't all need to be marked `skip: true  # [win]`. In theory they should build on Windows.

jvolkening:
Do you recommend just ignoring the failed Windows builds on `noarch: generic` recipes then, assuming I can't figure this out?

jvolkening:
To explain the last few attempts, `gmake` is supposed to be part of Strawberry Perl, which it appears is what the conda-forge `perl` package uses for Windows, but apparently it is not included in the conda package. The [knowledge base](https://conda-forge.org/docs/maintainer/knowledge_base.html#using-cmake) suggests that `nmake` should also be available:

    On Windows, you can also use nmake to build, but that does not need to be explicitly included.

But the last failure suggests that is also not the case. I wish I had some way to test the Windows builds locally using the same environment that the CI does.

recipes/perl-number-format/meta.yaml
mfansler: There are no CI logs for a run with `m2-make` for Windows. Can you point me to this? The first commit simply shows `make` without using the Cygwin (POSIX safe) version for Windows.

recipes/perl-number-format/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+  script:
+    - perl Makefile.PL INSTALLDIRS=vendor NO_PERLLOCAL=1 NO_PACKLIST=1
+    - make
+    - make test
+    - make install VERBINST=1
+
+requirements:
+  build:
+    - make  # [not win]
+    - m2w64-make  # [win]
```
mfansler, line 25: `m2w64` is only for `run`

```suggestion
    - m2-make  # [win]
```

recipes/perl-number-format/meta.yaml
jvolkening: > There are no CI logs for a run with `m2-make` for Windows. Can you point me to this? The first commit simply shows `make` without using the Cygwin (POSIX safe) version for Windows.

My apologies, that was actually for the original batch of recipes I tried for PR #25136. The first couple of commits all used `m2-make` like your suggestion, but the Windows build for the third commit just hung. I then decided, rather than trying to trouble-shoot a multi-recipe PR which I assume would eat additional CI time, to open a new PR with a pure-Perl package that in theory should be simple to build and use that to troubleshoot the Windows build.

recipes/perl-number-format/meta.yaml
jvolkening: Okay...in any case the build failed with the same error as using `make`. Do you think I should try `m2-make` again, in case the previous issue was a CI problem?

recipes/perl-number-format/meta.yaml
jvolkening: I just pushed a commit using `m2-make` and it has hung at the same point it did previously in the other PR.

jvolkening: Since I'm not sure if canceled CI jobs can be seen (I couldn't see the last one), the tail end of the log prior to the job freezing was:
```
(base) %SRC_DIR%>set "VS_YEAR=2015" 

(base) %SRC_DIR%>set "CMAKE_GENERATOR=Visual Studio 14 2015 Win64" 

(base) %SRC_DIR%>set "MSYS2_ARG_CONV_EXCL=/AI;/AL;/OUT;/out" 

(base) %SRC_DIR%>set "MSYS2_ENV_CONV_EXCL=CL" 

(base) %SRC_DIR%>call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64 
The system cannot find the path specified.

(base) %SRC_DIR%>set "INCLUDE=%PREFIX%\Library\include;" 

(base) %SRC_DIR%>set "LIB=%PREFIX%\Library\lib;" 

(base) %SRC_DIR%>call "C:\Miniforge\Scripts\..\condabin\conda_hook.bat" 

(base) %SRC_DIR%>call "C:\Miniforge\Scripts\..\condabin\conda.bat" activate "%PREFIX%" 
Checking if your kit is complete...
Looks good
Generating a gmake-style Makefile
Writing Makefile for Number::Format
Writing MYMETA.yml and MYMETA.json
Microsoft Windows [Version 10.0.20348.2227]
(c) Microsoft Corporation. All rights reserved.
```
These lines:
```
Checking if your kit is complete...
Looks good
Generating a gmake-style Makefile
Writing Makefile for Number::Format
Writing MYMETA.yml and MYMETA.json
```
I recognize as coming from the `perl Makefile.PL` command. After that should come the call to `make`. Not sure what the last few Microsoft copyright statements are, as I thought `m2-make` installed an MSYS version of GNU make.

jvolkening:
> Perhaps an advantage ...

Possibly...but I think in this case the CI is not failing during the build stage; the logs indicate that the build completes successfully (including `make test` which is where the actual unit testing is done for Perl E::MM packages). I can try pushing the direct output from `cpan skeleton` to see what happens, although in some cases I have found it produces incorrect recipes as well.

I may have identified the issue. I looked again at the actual error message during the TEST stage:
```shell
(%PREFIX%) %SRC_DIR%>"%PREFIX%\bin/perl" "%SRC_DIR%\run_test.pl" 
The system cannot find the path specified.
```
I went ahead and installed Miniforge on a Windows laptop, created an environment with only `perl`, and tested out the path to `perl` above. What I found was that for me it was installed to:
```
%PREFIX%\Library\bin\perl
```
rather than
```
%PREFIX%\bin\perl
```
(The mixed forward and backward slashes didn't make a difference)

Is this possibly a bug in conda-build?

jvolkening:
I believe it is a conda-build bug. In conda-build `config.py`, the path above is set [here](https://github.com/conda/conda-build/blob/465d386318cca048d660c3a5aa7088b076473ab5/conda_build/config.py#L567C3-L572C19)
```python
    def _get_perl(self, prefix, platform):
        if platform.startswith("win"):
            res = join(prefix, "Library", "bin", "perl.exe")
        else:
            res = join(prefix, "bin/perl")
        return res
```
and [here](https://github.com/conda/conda-build/blob/465d386318cca048d660c3a5aa7088b076473ab5/conda_build/config.py#L742-L743)
```python
    def perl_bin(self, prefix, platform):
        return self._get_perl(prefix, platform)
```
and finally called in `build.py` [here](https://github.com/conda/conda-build/blob/465d386318cca048d660c3a5aa7088b076473ab5/conda_build/build.py#L3234-L3240)
```python
        if pl_files:
            tf.write(
                '"{perl}" "{test_file}"\n'.format(
                    perl=metadata.config.perl_bin(
                        metadata.config.test_prefix, metadata.config.host_platform
                    ),
                    test_file=join(metadata.config.test_dir, "run_test.pl"),
```
I'm guessing that `host_platform` for this recipe evaluates to `noarch` (not sure though), so the path is not set correctly even though the actual test is running on Windows. Does this seem right?

jvolkening:
> I believe it is a conda-build bug

Actually, maybe *bug*is not the right word. If the above is correct, then the problem is the interaction of conda-build with the conda-forge CI.

jvolkening:
@mfansler If you think the above logic looks right, I'm considering submitting a PR to `conda-build` proposing to replace `host_platform` with `sys.platform`, i.e.
```python
        if pl_files:
            tf.write(
                '"{perl}" "{test_file}"\n'.format(
                    perl=metadata.config.perl_bin(
                        metadata.config.test_prefix, sys.platform
                    ),
                    test_file=join(metadata.config.test_dir, "run_test.pl"),
```

jvolkening:
@mfansler Do you think we can just ignore the Windows failure for now, for this and similar noarch recipes? If so, does this recipe look okay (I more or less reverted back to where I started with it, since it didn't seem to be a`make` issue after all). I did open a PR in conda-build for the above issue, but I don't know how that will go.

mfansler (approved):

mfansler:
> @mfansler Do you think we can just ignore the Windows failure for now, for this and similar noarch recipes?

Yes, sounds good. I'll go ahead and merge this.

Thanks for your work on this! ❤️

PR #25165, opened 2024-01-30
Create conda package for sqlalchemy-access

jepegit:
Checklist
- [x] Build number is 0.


recipes/sqlalchemy-access/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - sqlalchemy >=2.0.0
+    - pyodbc >=4.0.27
+    - pywin32-on-windows
```
xhochy, line 25: If this package is `win` only, you could add the following
```suggestion
    - pywin32-on-windows
    - __win
```

xhochy: Why didn't you end up using this?

jepegit: Hi @xhochy,
I tried it, but the check failed. So then I tried a lot of other things, and failed, and failed, and failed. Until it did not (allowing only for installation on windows, since the `sqlalchemy-access` plugin is for reading Microsoft Access and it is only sensible to do on windows).
Afterthought: I probably did something wrong when merging your suggested change, and it might also have worked. 
Question: what does the `__win` do?

xhochy: > I tried it, but the check failed.

It failed with which result? In some cases a failing CI on other platforms is OK.

`__win` adds a dependency on Windows as the runtime environment, i.e. the resulting package, even if marked as `noarch`, is only installable on Windows

jepegit: I see. It failed only on the non-windows environments. I will try again with your version.

xhochy:
Please add a file `conda-forge.yml` with the following content:

```
noarch_platforms: win_64
```

This ensures that the CI in the generated feedstock runs on the correct platform. Then we can merge..

xhochy:
Don't edit the top-level but create a new file in recipes/sqlalchemy-access ;)

jepegit:
Hi @xhochy  
Reverted to the original `conda-forge.yml` on the top folder and added `conda-forge.yml` inside recipes/sqlalchemy-access :-)

xhochy (approved):

PR #25195, opened 2024-01-31
Add recipe for OpenMM-XTB

peastman:
Checklist
- [x] Build number is 0.

@RaulPPelaez I included you as a maintainer.  I hope that's ok?

### DIFF
--- a/recipes/openmm-xtb/build.sh
+++ b/recipes/openmm-xtb/build.sh
@@ -23,6 +23,7 @@ mkdir -p "${TEST_DIR}"
 cp -a ../python/tests/* tests/Test* "${TEST_DIR}"
 if [[ "$target_platform" == osx* ]]; then
     find "${TEST_DIR}" -name "Test*" -perm +0111 -type f \
+        -exec python $RECIPE_DIR/patch_osx_tests.py "{}" \; \
         -exec cp "{}" $PREFIX/share/openmm-xtb/tests/ \;
 else
     find "${TEST_DIR}" -name "Test*" -executable -type f -exec cp "{}" $PREFIX/share/openmm-xtb/tests/ \;


RaulPPelaez:
thats ok

RaulPPelaez:
I can help with the recipe but I do not have push access into this PR.
The build script should be simple, like this one:
```shell
#!/bin/bash

set -euxo pipefail

rm -rf build || true


CMAKE_FLAGS="${CMAKE_ARGS} -DCMAKE_INSTALL_PREFIX=${PREFIX} -DCMAKE_BUILD_TYPE=Release"
CMAKE_FLAGS+=" -DPython3_EXECUTABLE=${PYTHON}"
CMAKE_FLAGS+=" -DOPENMM_DIR=${PREFIX}"


mkdir build && cd build
cmake ${CMAKE_FLAGS} ${SRC_DIR}
make -j$CPU_COUNT
make -j$CPU_COUNT install
```

peastman:
If you want to make changes yourself, you can create a PR to my branch and I'll merge them.

RaulPPelaez:
The contents of the recipe folder translate directly t the final feedstock AFAIK. So that should go to recipe/build.sh.
Let me PR to your fork then.

peastman:
I think the script also needs to `make PythonInstall`?  In addition, I see the [OpenMM-PLUMED build script](https://github.com/conda-forge/openmm-plumed-feedstock/blob/main/recipe/build.sh) packages up the tests.  Should this one do the same?

xhochy:
Please don't request a review before this is ready for review.

RaulPPelaez:
xtb is not built for windows. So that should be skipped in this package too
https://github.com/conda-forge/xtb-feedstock/blob/98a2f6893733e62bb411be616fd46503f3d424a9/recipe/meta.yaml#L14

RaulPPelaez:
In OSX the lib is successfully built and installed, but then the test does not find the dynamic library:
`['  ERROR (openmm-xtb,share/openmm-xtb/tests/TestXtbForce): /Users/runner/Miniforge3/conda-bld/openmm-xtb_1707754971797/work/build/libOpenMMXTB.dylib not found in packages, sysroot(s) nor the missing_dso_whitelist.\n.. is this binary repackaging?']`

I do not know what to do about that.

peastman:
Earlier in the log it says,

> 2024-02-12T16:25:23.2315870Z -- Installing: $PREFIX/lib/libOpenMMXTB.dylib

So it clearly is being built.  Perhaps it's looking in the wrong location?  A few lines before the error, there's the line

> 2024-02-12T16:25:45.4007130Z    INFO (openmm-xtb,lib/python3.10/site-packages/_openmmxtb.cpython-310-darwin.so): lib/libOpenMMXTB.dylib found in this package

I think that means the Python package successfully linked to it.  But when it tries to run `TestXtbForce`, it looks in a different location.

RaulPPelaez:
Maybe some rpath play is required here?

```yaml
  rpaths:
    - lib/

```

peastman:
The documentation says that's a Linux feature, and it's ignored on other platforms.

peastman:
Possibly that error message is misleading and the real problem is something different.  Earlier in the log we get the warning

> WARNING (openmm-xtb): interpreter (Python) package conda-forge/osx-64::python==3.10.13=h00d2728_1_cpython in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)

And then just before the error about TextXtbForce there's an exception

> conda_build.exceptions.OverLinkingError: overlinking check failed

[The documentation](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#ignoring-files-in-overlinking-overdepending-checks) briefly mentions options related to overlinking and overdepending, but I can't find any definition of what those terms mean.

It also seems like it could be an issue related to relocatable binaries: it's looking for the library in the environment where it was built instead of where it's being tested.  But the documentation says binaries are made relocatable by default.

RaulPPelaez:
I see a million of those warnings in all feedstocks. I always assumed it was an overprotecting test. As in, it is not correctly understanding what "used" is.
Perhaps one has to use ${PYTHON} instead of python when testing?

RaulPPelaez:
What about silencing the error with:
```yaml
build:
   missing_dso_whitelist:
      - libOpenMMXTB.dylib
```

peastman:
Success!

peastman:
This is now working.  It's ready to be reviewed and merged.

RaulPPelaez:
cc @conda-forge/help-python-c could you please take a look? Thanks.

recipes/openmm-xtb/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - swig                                   # [build_platform != target_platform]
+    - openmm >=8.1.1                         # [build_platform != target_platform]
+    - cmake
+    - make
+    - conda-build  # [osx]
```
xhochy, line 26: This should never be listed as a dependency.
```suggestion
```

recipes/openmm-xtb/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  host:
+    - python
+    - pip
+    - openmm >=8.1.1
+    - xtb
+    - swig                                   # [not arm64]
+  run:
+    - python
+    - {{ pin_compatible('openmm', min_pin='x.x', max_pin='x.x') }}
```
xhochy, line 35: Aren't the libraries using a run_exports? Normally you shouldn't need to do this for native libraries.

recipes/openmm-xtb/build.sh
```diff
@@ -0,0 +1,33 @@
...
+
+
+mkdir build && cd build
+cmake ${CMAKE_FLAGS} ${SRC_DIR}
+make -j$CPU_COUNT
+make -j$CPU_COUNT install PythonInstall
+
+# Copy tests to the share folder
+# Include test executables too
+TEST_DIR="${PREFIX}/share/${PKG_NAME}/tests/"
```
xhochy, line 23: Typically, we don't install tests. Why are you doing it here?

recipes/openmm-xtb/build.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+CMAKE_FLAGS="${CMAKE_ARGS} -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON"
```
xhochy, line 5: This line is not actually used.

xhochy (changes requested):

recipes/openmm-xtb/meta.yaml
peastman: Apparently not.  There's no `run_exports` in https://github.com/conda-forge/openmm-feedstock/blob/main/recipe/meta.yaml.

recipes/openmm-xtb/build.sh
peastman: @jaimergp this is copied from the build script you wrote for openmm-plumed.  Do you remember what the reason was?

peastman: @xhochy is there a way to run the tests without copying them to the package?  I looked up a few other packages to see how they handled tests, but they did exactly the same thing as here.

According to the [conda-build documentation](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#test-section) you can add a `source_files` section to the recipe to make it copy files to the directory where the tests will be run.  I tried adding

```yaml
test:
  source_files:
    - python/tests/TestXtbForce.py
```

But when I then try to run it with `python TestXtbForce.py`, it says the file isn't there.

jaimergp: This was handled this way in Plumed because they were only four files and it was easier that way. IIRC, the test files also had the prefix hardcoded and tests would fail otherwise. It was also the only way to test reliably whether the package was correctly built.

`files` and `source_files` will end up "packaging" the files too in the sense that they will be published as part of the `info/` directory, although maybe not part of the PREFIX, if that's what we are concerned with.

> But when I then try to run it with python TestXtbForce.py, it says the file isn't there.

Try `python python/tests/TestXtbForce.py`. They are copied with the relative paths untouched.

peastman: The situation is the same here.  There are only two files (one C++ test and one Python test), both very small.  If `source_files` also adds it to the package, I don't see any advantage to doing it that way?

recipes/openmm-xtb/meta.yaml
peastman: Fixed.

recipes/openmm-xtb/build.sh
peastman: Fixed.

recipes/openmm-xtb/meta.yaml
peastman: Actually, removing it broke the Mac build.

```
2024-03-06T01:30:42.7708550Z Traceback (most recent call last):
2024-03-06T01:30:42.7710240Z   File "/Users/runner/work/1/s/recipes/openmm-xtb/patch_osx_tests.py", line 4, in <module>
2024-03-06T01:30:42.7710700Z     from conda_build.os_utils import macho
2024-03-06T01:30:42.7711150Z ModuleNotFoundError: No module named 'conda_build'
```

peastman:
@conda-forge/help-python-c this is ready for another review.

peastman:
Can someone review this?

peastman:
If no one else has any comments, can someone merge this?

recipes/openmm-xtb/meta.yaml
jaimergp: That error comes from the test helper script, where we had to patch some test binaries so they don't contain the hardcoded prefix. I'm not sure this is needed in this project (again, that was a plumed thing). I'd start simple with more standard approaches first, and then add hacks as needed using the existing openmm-* recipes as hints.

recipes/openmm-xtb/build.sh
```diff
@@ -0,0 +1,30 @@
...
+
+# Copy tests to the share folder
+# Include test executables too
+TEST_DIR="${PREFIX}/share/${PKG_NAME}/tests/"
+
+mkdir -p "${TEST_DIR}"
+cp -a ../python/tests/* tests/Test* "${TEST_DIR}"
+if [[ "$target_platform" == osx* ]]; then
+    find "${TEST_DIR}" -name "Test*" -perm +0111 -type f \
+        -exec python $RECIPE_DIR/patch_osx_tests.py "{}" \; \
```
jaimergp, line 26: Can we check if we still need this?

peastman: It's still needed.  WIthout it we get the `conda_build.exceptions.OverLinkingError: overlinking check failed` error.

recipes/openmm-xtb/meta.yaml
peastman: As noted in my other comment, it really is needed.  I did exactly what you described: started with the simple approach, then copied over the fixes for problems from other recipes.

recipes/openmm-xtb/build.sh
jaimergp: For future reference, this is the full error:

```
  ERROR (openmm-xtb,share/openmm-xtb/tests/TestXtbForce): /Users/runner/Miniforge3/conda-bld/openmm-xtb_1712618910781/work/build/libOpenMMXTB.dylib not found in packages, sysroot(s) nor the missing_dso_whitelist.
```

Where we can see that the `build_dir` path is hardcoded so it doesn't end up being replaced by the PREFIX machinery in conda-build. To me this sounds like something that could be fixed in the CMake config, preferrably.

That said I don't know exactly where that'd go so I'd support this workaround for now.

@xhochy, what do you think?

xhochy: Sounds like a good workaround. We have tried to fix  this in (upstream) Arrow, too, but we never found a reliable solution :(

peastman: Thanks!  Is it ok to merge this?

xhochy: From my PoV yes. I'll let @jaimergp have a final look.

jaimergp (approved):

PR #25196, opened 2024-01-31
Add ast-grep recipe.

asford:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/ast-grep/0001-Downgrade-tree-sitter-cpp-for-statis_assert-build-er.patch
+++ b/recipes/ast-grep/0001-Downgrade-tree-sitter-cpp-for-statis_assert-build-er.patch
@@ -1,7 +1,7 @@
 From 8c566f0e4ae6533d73e300233814b8214a347819 Mon Sep 17 00:00:00 2001
 From: Alex Ford <alex.ford@abcellera.com>
 Date: Wed, 31 Jan 2024 16:03:24 -0800
-Subject: [PATCH] Downgrade tree-sitter-cpp for statis_assert build error
+Subject: [PATCH] Downgrade tree-sitter-cpp for static_assert build error
 
 ---
  Cargo.lock                 | 4 ++--
--- a/recipes/ast-grep/0001-Downgrade-tree-sitter-cpp-for-statis_assert-build-er.patch
+++ b/recipes/ast-grep/0001-Downgrade-tree-sitter-cpp-for-statis_assert-build-er.patch
@@ -1,0 +1,49 @@
+From 8c566f0e4ae6533d73e300233814b8214a347819 Mon Sep 17 00:00:00 2001
+From: Alex Ford <alex.ford@abcellera.com>
+Date: Wed, 31 Jan 2024 16:03:24 -0800
+Subject: [PATCH] Downgrade tree-sitter-cpp for statis_assert build error
+
+---
+ Cargo.lock                 | 4 ++--
+ crates/language/Cargo.toml | 4 ++--
+ 2 files changed, 4 insertions(+), 4 deletions(-)
+
+diff --git a/Cargo.lock b/Cargo.lock
+index 30420efd..ad9ad46a 100644
+--- a/Cargo.lock
++++ b/Cargo.lock
+@@ -1902,9 +1902,9 @@ dependencies = [
+ 
+ [[package]]
+ name = "tree-sitter-cpp"
+-version = "0.20.3"
++version = "0.20.2"
+ source = "registry+https://github.com/rust-lang/crates.io-index"
+-checksum = "23b4b625f46a7370544b9cf0545532c26712ae49bfc02eb09825db358b9f79e1"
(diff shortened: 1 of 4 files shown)


recipes/ast-grep/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/{{ name }}/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
+  sha256: a1021a96fe778fcce7e600727e862b1b37f51031bad9f85d5d99c7f1aef2ffda
+
+build:
+  number: 0
+  skip: true # [win]
```
asford, line 14:
```suggestion
  skip: true  # [win]
```

recipes/ast-grep/0001-Downgrade-tree-sitter-cpp-for-statis_assert-build-er.patch
```diff
@@ -0,0 +1,49 @@
+From 8c566f0e4ae6533d73e300233814b8214a347819 Mon Sep 17 00:00:00 2001
+From: Alex Ford <alex.ford@abcellera.com>
+Date: Wed, 31 Jan 2024 16:03:24 -0800
+Subject: [PATCH] Downgrade tree-sitter-cpp for statis_assert build error
```
asford, line 4:
```suggestion
Subject: [PATCH] Downgrade tree-sitter-cpp for static_assert build error
```

asford:
@conda-forge/help-rust 

This feedstock is ready for review.

I need to make a changed to a pinned dependency due to a failure in static linking.
This is because the dependency has an improper use of `static_assert` for the C standard used in compilation of the feedstock.

This builds properly outside of the build container, so I suspect there may be some config needed to pass appropriate CFLAGS need to define the C standard version used during compilation? 
I think downgrading this dependency is acceptable for the time being.

recipes/ast-grep/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  run:
+      # for details see conda_build_config.yaml
+    - __osx >=10.12     # [osx and x86_64]
```
xhochy, line 30:
```suggestion
    - __osx >= {{ MACOSX_DEPLOYMENT_TARGET|default('10.12') }}     # [osx and x86_64]
```

recipes/ast-grep/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    write patterns as if you are writing ordinary code. It will match all code
+    that has the same syntactical structure. You can use $ sign + upper case
+    letters as wildcard, e.g. $MATCH, to match any single AST node. Think it as
+    REGEX dot ., except it is not textual.
+  license: MIT
+  license_family: MIT
+  # It is required to include a license file in the package,
+  # (even if the license doesn't require it) using the license_file entry.
+  # Please also note that some projects have multiple license files which all need to be added using a valid yaml list.
+  # See https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#license-file
```
xhochy, line 52:
```suggestion
```

recipes/ast-grep/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+      # for details see conda_build_config.yaml
+    - __osx >=10.12     # [osx and x86_64]
+
+
+test:
+  commands:
+    - sg --version
+
+about:
+  home: https://github.com/simplejson/simplejson
```
xhochy, line 38: This needs to be updated.

xhochy (changes requested):
You could probably set a newer C std by appending it to the CFLAGS

recipes/ast-grep/meta.yaml
asford:
```suggestion
  home: https://github.com/ast-grep/ast-grep
```

recipes/ast-grep/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+  run:
+      # for details see conda_build_config.yaml
+    - __osx >= {{ MACOSX_DEPLOYMENT_TARGET|default('10.12') }}     # [osx and x86_64]
```
asford, line 30:
```suggestion
    - __osx >={{ MACOSX_DEPLOYMENT_TARGET | default('10.12') }}     # [osx and x86_64]
```

asford:
> You could probably set a newer C std by appending it to the CFLAGS

Thanks, I wasn't able to fully diagnose this yet but think we can roll with a build now with the downgraded dep.

xhochy (approved):

PR #25240, opened 2024-02-05
Adding ttp package

cousinbr:
Checklist
- [ ] Build number is 0.


cousinbr:

recipes/ttp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7,<4.0
+    - poetry-core >=1.0.0
+    - pip
+  run:
+    - python >=3.7,<4.0
+  run_constrained:
```
xhochy, line 26: This list of run_constraint will make it very hard to be installed. These are probably optional dependencies for e.g. docs. You should drop them from the recipe.

xhochy (changes requested):

cousinbr:
Hello @xhochy, 

I'm not familiar with conda packaging, so sorry if i make mistake. I have removed sphinx related package in "run_constrained". I have read the documentation about "run_constrained" parameter but it's not clear for me. Does theses dependancies will be installed when i install the "ttp" package or we have a system to install optional dependencies like pip have, eg: "pip install ttp" vs "pip install ttp[full]"

Regards

recipes/ttp/meta.yaml
cousinbr: Hello @xhochy,

I'm not familiar with conda packaging, so sorry if i make mistake. I have removed sphinx related package in "run_constrained". I have read the documentation about "run_constrained" parameter but it's not clear for me. Does theses dependancies will be installed when i install the "ttp" package or we have a system to install optional dependencies like pip have, eg: "pip install ttp" vs "pip install ttp[full]"

Regards

xhochy: run_constrainted is only for additional constraints that limit the version ranges of these packages if tgey were to be installed because od another package. They don't get auto-installed. For optional dependencies, you can have a look at the black-feedstock. You need to use multiple outputs for that. I would not recommend that here as it's probably unlikely that one wants the conda package to work on the documents and not work on that in the actual source tree.

xhochy (approved):

PR #25289, opened 2024-02-07
Add `httpx-ntlm`, `httpx-negotiate-sspi`

dhirschfeld:
Checklist
- [x] Build number is 0.


dhirschfeld:
ping @conda-forge/help-python!

recipes/httpx-negotiate-sspi/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/httpx_negotiate_sspi-{{ version }}.tar.gz
+  sha256: cbf5ae118f33c92986eedffe18913e46252788ae97cd43f75f8efbad5725d0f4
+
+build:
+  skip: True  # [not win]
```
ocefpaf, line 13: We can make this an OS-noarch. See https://conda-forge.org/docs/maintainer/knowledge_base.html#noarch-packages-with-os-specific-dependencies

dhirschfeld: Ah, cool. Thanks will make it `noarch`...

dhirschfeld: Actually @ocefpaf, this package only makes sense to install on Windows - it uses Windows specific apis that are not available on Linux.

IIUC the Linux equivalent would be `GSSAPI` which can work with both platforms but requires a workign Kerberos configuration. Windows users tend to just use SSPI because if works out of the box on that platform.

I'm publishing both packages here so that on Windows users can use `httpx-negotiate-sspi` and on Linux (where that doesn't work) they can instead fallback to `httpx-ntlm`.

Given that it doesn't work on Linux should it still be `noarch`?

dhirschfeld: > Given that it doesn't work on Linux should it still be `noarch`?

If it makes the Linux package effectively a dummy package that might be a good thing as it then won't force all packages which depend on it to also be arch-specific 🤔 

The import would fail, but likely for a `ModuleNotFound` for `pywin32` rather than `httpx_negotiate_sspi`:
```python
from httpx_negotiate_sspi import HttpNegotiateAuth
```

dhirschfeld:
I might need some help makeing this `noarch` @ocefpaf! 😬 

I assume that if I add a [`noarch_platforms`](https://conda-forge.org/docs/maintainer/conda_forge_yml.html#noarch-platforms) key to `conda-forge.yml` it will Just Work? Given that this is `staged-recipes` I'm not supposed to change `conda-forge.yml` here right?

recipes/httpx-negotiate-sspi/meta.yaml
ocefpaf: You can make it windows only. The reason for os-niarch is to avoid the Python versions matrix.

ocefpaf:
> I'm not supposed to change `conda-forge.yml` here right?

You can add one in the recipe directory and things should "just work."

dhirschfeld:
Unsurprisingly, building a non-functional package on Linux fails to import:
```
import: 'httpx_negotiate_sspi'
  File "/home/conda/staged-recipes/build_artifacts/httpx-negotiate-sspi_1707571383325/test_tmp/run_test.py", line 2, in <module>
    import httpx_negotiate_sspi
  File "/home/conda/staged-recipes/build_artifacts/httpx-negotiate-sspi_1707571383325/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/python3.12/site-packages/httpx_negotiate_sspi/__init__.py", line 3, in <module>
    from .sspi import HttpSspiAuth
  File "/home/conda/staged-recipes/build_artifacts/httpx-negotiate-sspi_1707571383325/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_place/lib/python3.12/site-packages/httpx_negotiate_sspi/sspi.py", line 10, in <module>
    import pywintypes
ModuleNotFoundError: No module named 'pywintypes'
```

But I suppose that's no worse than `ModuleNotFoundError: No module named 'httpx_negotiate_sspi'` whilst letting the package (and everything which depends on it) still be `noarch`

dhirschfeld:
> > I'm not supposed to change `conda-forge.yml` here right?
> 
> You can add one in the recipe directory and things should "just work."

That seemed to do the job :+1:

PR #25290, opened 2024-02-07
Add `lineax`

dhirschfeld:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/lineax/meta.yaml
+++ b/recipes/lineax/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "lineax" %}
+{% set version = "0.0.4" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/lineax-{{ version }}.tar.gz
+  sha256: e68f1eba2f352122fdce9adc0556684f31eb8364b1a00acee484dd6e44a34e5e
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Windows is skipped upstream:
https://github.com/conda-forge/jaxlib-feedstock/blob/main/recipe/meta.yaml#L23

dhirschfeld:
@conda-forge/core I think this is good to merge. Windows fails as it can't be installed as the upstream dep isn't being built for the Windows platform:
```
nothing provides jaxlib >=0.4.11 needed by jax-0.4.13-pyhd8ed1ab_0
```

I think we still want to build a `noarch` recipe here as it will Just Work as soon as upstream publishes a build of `jaxlib` for Windows.

Until that time the `noarch` package wouldn't be able to be installed on Window (due to the missing dep) but would work fine on all other platforms where `jaxlib >=0.4.11` is available.

recipes/lineax/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python >=3.9,<4.dev0
+    - hatchling
+    - pip
+  run:
+    - python >=3.9
```
ocefpaf, line 23: While I don't think that is needed, it should be the same as above.

```suggestion
    - python >=3.9,<4.dev0
```

ocefpaf: It doesn't hurt b/c I believe the upper pin is unnecessary. I've seen this bug in grayskull before BTW, but I was too lazy to report ;-p

PR #25298, opened 2024-02-07
added musclex recipe

tristanleduc:
recipe of musclex software for diffraction pattern analysis

### DIFF
--- a/recipes/musclex/conda_build_config.yaml
+++ b/recipes/musclex/conda_build_config.yaml
@@ -1,0 +1,2 @@
+channel_sources:
+  - conda-forge,fastai,defaults
--- a/recipes/musclex/meta.yaml
+++ b/recipes/musclex/meta.yaml
@@ -1,0 +1,73 @@
+{% set name = "musclex" %}
+{% set version = "1.23.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/musclex-{{ version }}.tar.gz
+  sha256: 239bcb1accefae4d1bafb8837fec79996d0d7cdeb56bac3590c44a43b948ad00
+
+build:
+  entry_points:
+    - musclex=musclex.main:main
+    - musclex-launcher=musclex.launcher:LauncherForm.main
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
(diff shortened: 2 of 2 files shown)


tristanleduc:
approved

recipes/musclex/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+channel_sources:
+  - conda-forge,fastai,defaults
```
xhochy, line 2: We only want packages that depend on conda-forge in conda-forge. Please add any missing dependencies to conda-forge first.

recipes/musclex/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - docs
+    - hooks
+    - musclex
+  requires:
+    - pip
+
+about:
+  home: https://github.com/biocatiit/musclex
+  summary: Muscle X The Muscle X Application Suite is a collection of programs intended to assist with analyzing diffraction X-ray images.
+  license: LicenseRef-Proprietary
```
xhochy, line 68: The license file looks like MIT license.

recipes/musclex/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+test:
+  imports:
+    - AppImageBuilder
+    - dev_docs
+    - docs
+    - hooks
+    - musclex
+  requires:
+    - pip
```
xhochy, line 63:
```suggestion
    - pip
  commands:
    - pip check
```

recipes/musclex/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - fisx <=1.2.1
+    - future <=0.18.3
+    - pymca 
+
+test:
+  imports:
+    - AppImageBuilder
+    - dev_docs
+    - docs
+    - hooks
```
xhochy, line 60: These should not be installed. This sounds like there is an issue in upstream's call to find_packages.

xhochy (changes requested):

recipes/musclex/conda_build_config.yaml
tristanleduc: Hi, 

 i need to use opencv-python-headless, which is not available on conda-forge, which is why I used an alternative channel. Is there a solution to address that? 

Regards, 
Tristan Leduc

xhochy: Yes, we need a variant of that in conda-forge. See the opencv-feedstock for discussions/progress on it.

PR #25340, opened 2024-02-10
Add kanidm-clients

freundTech:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/kanidm-clients/conda_build_config.yaml
+++ b/recipes/kanidm-clients/conda_build_config.yaml
@@ -1,0 +1,3 @@
+# https://conda-forge.org/docs/maintainer/knowledge_base.html#requiring-newer-macos-sdks
+MACOSX_SDK_VERSION:            # [osx and x86_64]
+  - '10.12'                    # [osx and x86_64]
--- a/recipes/kanidm-clients/meta.yaml
+++ b/recipes/kanidm-clients/meta.yaml
@@ -1,0 +1,53 @@
+{% set name = "kanidm-clients" %}
+{% set version = "v1.1.0rc16" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/kanidm/kanidm/archive/refs/tags/{{ version.replace("rc", "-rc.") }}.tar.gz
+  sha256: 26e275bf6cf19dba34a2cc0d382c22e3b3c887f15e9fe64a1c011098173a1de9
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+    - pkg-config
+    - openssl
+    - libudev  # [linux]
+    - sysroot_linux-64 2.17  # [linux64 and cdt_name!='cos6']
+    - {{ cdt('libudev-devel') }}  # [linux and cdt_name=='cos6']
+  host:
+    - libudev  # [linux and cdt_name!='cos6']
+    - openssl
+  run:
+    - openssl
+
+test:
+  commands:
+    - kanidm version
+
+about:
+  home: https://kanidm.com/
+  summary: 'A simple, secure and fast identity management platform'
+  description: |
+    Kanidm is a simple and secure identity management platform, allowing other applications and services to offload the challenge of authenticating and storing identities to Kanidm.
+    The goal of this project is to be a complete identity provider, covering the broadest possible set of requirements and integrations. You should not need any other components (like Keycloak) when you use Kanidm - we already have everything you need!
+    To achieve this we rely heavily on strict defaults, simple configuration, and self-healing components. This allows Kanidm to run from small home labs, families, small business, and all the way to the largest enterprise needs.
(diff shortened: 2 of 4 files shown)


recipes/kanidm-clients/conda_build_config.yaml
```diff
@@ -0,0 +1,3 @@
+# https://conda-forge.org/docs/maintainer/knowledge_base.html#requiring-newer-macos-sdks
+MACOSX_SDK_VERSION:            # [osx and x86_64]
+  - '10.12'                    # [osx and x86_64]
```
freundTech, line 3: This is needed as otherwise `_getentropy` can't be linked against.

See https://github.com/conda-forge/py-sr25519-bindings-feedstock/issues/5 and https://github.com/conda-forge/ruff-feedstock/issues/165

freundTech:
@conda-forge/help-rust I could use some help fixing the OS X build here.

I get the following during the cargo build, suggesting that the build script wants to use openssl, but can't dynamically link to it.
```
dyld: Library not loaded: @rpath/libssl.3.dylib
    Referenced from: /Users/runner/Miniforge3/conda-bld/kanidm-clients_1707651829074/work/target/release/build/webauthn-authenticator-rs-2608b7a98b23ac5b/build-script-build
    Reason: image not found
```
openssl is installed as build, host and run dependency. 
The recipe works fine locally on an M1 Mac. I don't have a x86 Mac for local testing.

freundTech:
Looks like the auto label workflow wrongly labels this as `R`, because I mentioned `rust`. 
I created a PR to fix that: #25342

xhochy:
Ideas:

1. Set `OPENSSL_DIR=$PREFIX`
2. Set `OPENSSL_LIB_DIR` and `OPENSSL_INCLUDE_DIR`
3. Install `pkg-config` in build.

freundTech:
`pkg-config` is already installed and the env varibales only affect `openssl-sys` at build-time.

The problem is with `build.rs` of `webauthn-authenticator-rs`. It needs to be built and run before the rest of `webauthn-authenticator-rs` can be compiled.
`build.rs` builds fine against openssl, but when trying to run it it can't dynamically link against openssl for some reason.

Setting `DYLD_FALLBACK_LIBRARY_PATH=$BUILD_PREFIX/lib` tells it where to find it (`$PREFIX/lib` also worked, but would probably break when cross-compiling)

This is now ready for review @conda-forge/help-rust.

recipes/kanidm-clients/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "kanidm-clients" %}
+{% set version = "v1.1.0rc16" %}
```
xhochy, line 2: Either set this to a release or set the channel_target to a dev/rc channel.

recipes/kanidm-clients/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - pkg-config
+    - openssl
+    - libudev  # [linux]
+    - sysroot_linux-64 2.17  # [linux64 and cdt_name!='cos6']
+    - {{ cdt('libudev-devel') }}  # [linux and cdt_name=='cos6']
+  host:
+    - libudev  # [linux and cdt_name!='cos6']
+    - openssl
+  run:
+    - openssl
```
xhochy, line 29: Should come via `run_exports`
```suggestion
```

recipes/kanidm-clients/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+    - pkg-config
+    - openssl
+    - libudev  # [linux]
+    - sysroot_linux-64 2.17  # [linux64 and cdt_name!='cos6']
+    - {{ cdt('libudev-devel') }}  # [linux and cdt_name=='cos6']
+  host:
+    - libudev  # [linux and cdt_name!='cos6']
```
xhochy, line 26: Why not use `libudev` always?

xhochy (changes requested):

recipes/kanidm-clients/meta.yaml
freundTech: Despite this being a 1.1.0 rc there is no non-rc release yet.
Kanidm went from 1.0.0rc3 to 1.1.0-alpha without a 1.0.0 release

recipes/kanidm-clients/meta.yaml
freundTech: I had problems linking against `libudev` (I don't remember the exact error message), so I took this solution from [librealsense-feedstock](https://github.com/conda-forge/librealsense-feedstock/blob/45ef8a2bda4df7f949bc52ab45351afb3039a7fa/recipe/meta.yaml).

libudev is linux only, so that selector needs to stay. As far as I understand the cos6 selector is because libudev is only packaged for cos7, while for cos6 the cdt must be used.

xhochy: Then we should build for cos7 only. There is little value in having this workaround for little value.

xhochy (approved):

PR #25425, opened 2024-02-15
Add dry-attr package from PyPi

iamthebot:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/dry-attr/meta.yaml
+++ b/recipes/dry-attr/meta.yaml
@@ -1,0 +1,41 @@
+{% set version = "0.3.2" %}
+{% set name = "dry-attr" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/a/attr/attr-{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/dry-attr/meta.yaml
```diff
@@ -0,0 +1,41 @@
+{% set version = "0.3.2" %}
+{% set name = "dry-attr" %}
```
xhochy, line 2: Why have you prefixed this with dry?

xhochy (changes requested):

recipes/dry-attr/meta.yaml
iamthebot: @xhochy `attr` [is already a feedstock](https://github.com/conda-forge/attr-feedstock) for a different library and the author of this library has noted that `attr` as an import is shadowed by a more popular library and hence to use `dry-attr`.

I'm using this new name for the library here as a result.

xhochy: 👍 Normally, we would use `python-attr` as a way to avoid this conflict. It probably makes sense to add that alias long-term.

xhochy (approved):

PR #25438, opened 2024-02-17
Sortmerna conda recipe

biocodz:
The current recipe will be maintained parallel to the Sortmerna [bioconda-recipe](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/sortmerna). The bioconda recipe currently uses pre-built binaries. The current recipe performs the whole build including building 3rd party libraries. This recipe is also going to support Windows builds (not supported on bioconda).


bgruening:
I don't think we should maintain the bioconda recipe in parallel. If this one is ready here we should remove the bioconda recipe.

biocodz:
I agree. Mentioned the parallel just for suggestions. How do we proceed with removing the bioconda recipe? Usual pull request?

martin-g:
I was not aware of this PR.
I also worked on the Bioconda recipe to build packages for more platforms - https://github.com/bioconda/bioconda-recipes/pull/46644
It is a duplicated work for now! Once this PR is merged I will create an Arch Migrator PR to add builds for linux-aarch64 and linux-ppc64le and another PR for dropping the Bioconda recipe!

biocodz:
Created [bioconda-recipes PR 48607](https://github.com/bioconda/bioconda-recipes/pull/48607) to remove sortmerna recipe due relocation to conda-forge

bgruening (approved):

biocodz:
filed [issue 5391](https://github.com/conda/conda-build/issues/5391) on conda-build

dslarm:
Anyone still working on this?
It looks like the recipe is deleted from bioconda, but this recipe has never been merged and landed as its own feedstock - possibly due to windows build failing?   Would it make sense to kill the windows build for now, and get this working for the linux side

martin-g:
> Would it make sense to kill the windows build for now, and get this working for the linux side

+1

biocodz:
yeah, I guess disabling Win build is an option. Do I have to prepare another PR?

dslarm:
> yeah, I guess disabling Win build is an option. Do I have to prepare another PR?

No, I doubt it: you should be able to add the commit in your existing branch, and it will automatically pull.

recipes/sortmerna/conda_build_config.yaml
mfansler, line 1: This should probably be removed. MacOS SDK is redundant with defaults. Don't linux default compilers work?

recipes/sortmerna/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/{{ name }}/{{ name }}/archive/v{{ version }}.tar.gz
+  sha256: 6b62def30704ea956e1de060b602f9bebc8f2ba68107c59329b332500997b1d2
+
+build:
+  number: 0
+  run_exports: []
+  #skip: True  # [osx]
```
mfansler, line 15: If you want `run_exports` to keep parity with Bioconda standards, then use them; otherwise, tidy up.

```suggestion
```

recipes/sortmerna/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
mfansler, line 19: Needs `stdlib`:

```suggestion
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
```

recipes/sortmerna/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - jinja2
+    - requests
+    - ninja
+  run:
+    - python
+
+test:
+  commands:
+#  - test -f ${PREFIX}/bin/{{ name }}  # [not win]
+#  - if not exist %PREFIX%\bin\{{ name }}.exe exit 1  # [win]
```
mfansler, line 35: Tidy up

```suggestion
```

recipes/sortmerna/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  - {{ name }} -h  # [not win]
+  - {{ name }}.exe -h  # [win]
+
+about:
+  home: http://bioinfo.lifl.fr/RNA/sortmerna
```
mfansler, line 40: Additional URLs:

```suggestion
  home: http://bioinfo.lifl.fr/RNA/sortmerna
  dev_url: https://github.com/sortmerna/sortmerna
  doc_url: https://sortmerna.readthedocs.io/en/latest/
```

recipes/sortmerna/bld.bat
mfansler, line 1: Missing trailing newline

recipes/sortmerna/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+echo "running python setup.py -n all"
+env
```
mfansler, line 3: Tidy up

```suggestion
```

recipes/sortmerna/build.sh
mfansler, line 1: Missing trailing newline

mfansler (changes requested):

mfansler (changes requested):
Another point for discussion is that this is vendoring `zlib` and `rocksdb`, both of which are available through Conda Forge. It would be preferable that the recipe be reworked to use the Conda Forge versions.

recipes/sortmerna/build.sh
biocodz: fixed

biocodz:
> Another point for discussion is that this is vendoring `zlib` and `rocksdb`, both of which are available through Conda Forge. It would be preferable that the recipe be reworked to use the Conda Forge versions.

There are problems with this:
- static rocksdb.lib is built with all possible zip libs (lz4-c, snappy, zstd, bzip2, zlib) -> need to include all of them into sortmerna build, and create the cmake modules for all (except zlib). Sortmerna only uses zlib.
- not all versions of rocksdb are available.

recipes/sortmerna/conda_build_config.yaml
biocodz: I was not able to build without configuration as specified in conda_build_config.yaml

recipes/sortmerna/meta.yaml
biocodz: Adding `{{ stdlib('c') }}` throws
```
ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform linux-64: {MatchSpec("c_linux-64")}
  - nothing provides requested c_linux-64

└─ c_linux-64 does not exist (perhaps a typo or a missing channel).
```
works fine without it

recipes/sortmerna/conda_build_config.yaml
mfansler: Was this done together with the `stdlib` spec? How did it fail?

Perhaps push a version with these changes so we can see the logs from any failing CI.

biocodz:
I'll disable Win build after checking one more time the reason for failure. Conda-build has no problems on my local machine

recipes/sortmerna/conda_build_config.yaml
biocodz: Linux build fails under compiler >11.4 due incompatibilities with old 3rd party c libraries used by Sortmerna. Have to stick with max 11.4 until this is resolved. Same goes for MacOS.

mfansler:
> > Another point for discussion is that this is vendoring `zlib` and `rocksdb`, both of which are available through Conda Forge. It would be preferable that the recipe be reworked to use the Conda Forge versions.
> 
> There are problems with this:
> 
> * static rocksdb.lib is built with all possible zip libs (lz4-c, snappy, zstd, bzip2, zlib) -> need to include all of them into sortmerna build, and create the cmake modules for all (except zlib). Sortmerna only uses zlib.
> * not all versions of rocksdb are available.

The `rocksdb` feedstock appears amenable to supporting specific versions as branches (see [PR history](https://github.com/conda-forge/rocksdb-feedstock/pulls?q=+is%3Apr+)). If a v7.10.x branch is needed, it could possibly be added there.

As for zip libs, I admit I don't follow the implication. So, maybe I'm not the one to discuss this with (CC: @conda-forge/help-python-c). Naively, I don't understand why anything needs to change outside of switching from static to dynamically-linked libraries.

Sorry to be a bother about this, but it just does not look great to be running 100+ minutes per platform and 90% of that time is rebuilding libraries Conda Forge nominally has available for dynamic linking.

---

Otherwise, if one wants to merge with the vendored libraries, then the licenses of those need to also be explicitly bundled through the `about.license_file` list. E.g., [`zlib` license](https://github.com/madler/zlib/blob/v1.3.1/LICENSE) and [`rocksdb` license](https://github.com/facebook/rocksdb/blob/v7.10.2/LICENSE.Apache).

biocodz:
Yeah, we are well aware of the described problem, and tried to solve it, including the dynamic linking, and the pre-built libraries, and each time we ran into some kind of problem. For the time being, the existing solution is the only one, and fixing it requires a considerable effort/time. Considering this project is run at 100% voluntarily dedicated free developer's time, it's not going to happen fast.

mfansler:
Just checking back in. Thanks for the clarifications.

Since reworking the dependencies appears tedious, let's rather ensure the zlib and rocksdb licenses are added. One could do this dynamically, but simply including copies (linked in https://github.com/conda-forge/staged-recipes/pull/25438#issuecomment-2511219960) would suffice. Either adjust the names/paths to reflect which license corresponds to which vendored software, or include an additional file that indicates what is what.

mfansler (approved):
LGTM!

I'm going to leave this open a day or two, in case anyone from @conda-forge/help-python proper has comments. Otherwise, I think it's good to merge.

PR #25517, opened 2024-02-21
Adding recipe for iotaa

maddenp:
Checklist
- [x] If static libraries are linked in, the license of the static library is packaged. **NA**
- [x] Build number is 0.


maddenp:
I am willing to be a maintainer.

maddenp:

maddenp:
Not sure if I'm phrasing this right, but:

@conda-forge-admin, please ping team conda-forge/help-python

recipes/iotaa/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set version = "0.7.2" %}
+{% set build = "0" %}
+{% set python = "python >=3.8" %}
```
ocefpaf, line 3: Please don't template these. They obfuscate the recipe and don't help much.

ocefpaf (commented):
The error is b/c you have a conda recipe in the tarball and conda-build is trying to build it too. You should delete it.

recipes/iotaa/meta.yaml
```diff
@@ -1,28 +1,24 @@
-{% set version = "0.7.2" %}
-{% set build = "0" %}
-{% set python = "python >=3.8" %}
-
 package:
   name: iotaa
-  version: {{ version }}
```
ocefpaf: The version should be in the jinja template! My comment was pointing to the python version and build number only. Sorry for the confusion but when in doubt, check our docs and examples.

maddenp: Understood, will fix, thanks.

The Linux/OSX builds fail trying to run the `.exe`, so I will revert to the separate tests with selectors, unless you have another idea.

I tried adding an `rm -rf recipe`, which is fine on Linux/OSX, but which of course does not work on Windows, which I am also no expert on. Would it be bad practice to omit a Windows build for this package? If so I will find a solution.

ocefpaf: The idea is to not use the `.exe` and see if Windows can call the executable without it. As far as I know it is possible to run, on Windows, with both `cli` or `cli.exe`. The former will save you an extra line in the recipe and the need for the OS selector.

recipes/iotaa/meta.yaml
maddenp-noaa: Fixed.

PR #25532, opened 2024-02-23
Add CAD_to_OpenMC surface mesh creation package.

ebknudsen:
Checklist
- [x] Build number is 0.


ebknudsen:
@conda-forge/help-python ready for review I believe

recipes/cad-to-openmc/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+{% set version = "0.3.3a2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  sha256: 15689aff401b196a8b2a95b7571dab9dc328c4b70fd79f38fa812e806fe5c986
+  url: https://github.com/openmsr/CAD_to_OpenMC/archive/refs/tags/v{{ version }}.tar.gz
+build:
```
ocefpaf, line 11:
```suggestion

build:
```

recipes/cad-to-openmc/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
```
ocefpaf, line 12:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

recipes/cad-to-openmc/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - {{ compiler('c') }}
+  host:
+    - python
+    - setuptools >=61.0
+    - pip
+  run:
+    - python >=3.9
+    - gmsh
+    - python-gmsh
+    - moab >=5.3.0
```
ocefpaf, line 27: You can skip Windows b/c this package is not available there.
BTW, do you want to check that macOS error?

ebknudsen: I do want to check the MacOS error but not know as such. I think it is a bigger upstream issue though, so I think I'll have to punt a present. I.e. I should perhaps skip it. I tried only building on linux (f280da2) but that failed. should that have been [linux64] perhaps?

ocefpaf: Can you add skips for Windows and macOS then?

recipes/cad-to-openmc/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  version: {{ version }}
+
+source:
+  sha256: 15689aff401b196a8b2a95b7571dab9dc328c4b70fd79f38fa812e806fe5c986
+  url: https://github.com/openmsr/CAD_to_OpenMC/archive/refs/tags/v{{ version }}.tar.gz
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: True  # [not linux64]
```
ocefpaf, line 15:
```suggestion
  skip: True  # [not linux]
```

recipes/cad-to-openmc/meta.yaml
ebknudsen: Sure - can do. I have tried two things: skipping by means of "not linux" and "not linux_64". The former fails, whereas the latter works. AFAICT the fail is on python 3.12rc which is not done for the latter.
I'll try skipping win and osx explicitly, but I don't think that's the issue here. Not quite sure what the sane way of doing this is now - I shall have to run a few more tests.

ebknudsen: Have added skips for win and osx as well as skips for python < 3.8 & >3.12. osx have to come later as I don't have access to infrastructure for that right now. Python version problems are upstream problems which I can't do much about I'm afraid

PR #25586, opened 2024-02-28
Add pybip39

MementoRC:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pybip39/meta.yaml
+++ b/recipes/pybip39/meta.yaml
@@ -24,6 +24,7 @@ requirements:
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
     - cargo-bundle-licenses
+    - python
   host:
     - maturin >=0.14,<0.15
     - python
--- a/recipes/pybip39/meta.yaml
+++ b/recipes/pybip39/meta.yaml
@@ -1,0 +1,59 @@
+{% set name = "pybip39" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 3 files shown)


MementoRC:
Could adding python in the build section resolve the issue? It looks like python is found in order:
   1 PYO3_PYTHON
   2 venv or conda PREFIX
   3 search in bin directories

There is also an [env] in Cargo.toml, but it is likely not dynamic - I don't have a windows system, so I may go debug-commit mode :\

recipes/pybip39/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
+    - {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
+    - cargo-bundle-licenses
+    - python
```
xhochy, line 27: This should not be here and will likely lead to break, rather try setting `PYO3_PYTHON` for windows.
```suggestion
```

xhochy (changes requested):

recipes/pybip39/meta.yaml
MementoRC: It passes though :'(

Do you think that this should work:
```
  script_env:
    - PYO3_PYTHON: %PREFIX%\bin\python.exe  # [win]
```
or:
```
  script_env:
    - PYO3_PYTHON: {{ PYTHON }}  # [win]
```
Well, the latter should work

xhochy: Yes, try that. The other one might lead to an empty package at one point.

xhochy (approved):

PR #25608, opened 2024-03-01
Add `logfire` recipe

Kludex:


recipes/logfire/meta.yaml
```diff
@@ -0,0 +1,88 @@
...
+  imports:
+    - logfire
+  commands:
+    - pip check
+    - logfire --help
+  requires:
+    - pip
+
+about:
+  summary: Coming soon...
```
Kludex, line 81:
```suggestion
  home: https://pypi.org/project/logfire/
  summary: Coming soon...
```

recipes/logfire/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - python >=3.8
+    - opentelemetry-sdk >=1.21.0
+    - opentelemetry-exporter-otlp-proto-http >=1.21.0
+    - opentelemetry-instrumentation >=0.41b0
+    - rich >=13.4.2
+    - protobuf >=4.23.4
+    - typing-extensions >=4.0.0
+    - tomli >=2.0.1
+    - psycopg >=2.9.1
+  run_constrained:
```
xhochy, line 34: What's the reason for the long list here?

Kludex: I don't understand the question. Can't a package have dependency constraints?

Kludex: We have many optional dependencies.

xhochy: Yes, a package can have constraints but this seems like an unreasonably large number. This will make the installation of this package for people that don't use **all the optional** features very hard. I would propose to not handle them this way.

samuelcolvin: @Kludex it looks like some of these requirements can be removed completely, and some moved to optional groups.

If condaforge doesn't support optional groups, just remove them.

xhochy: conda(-forge) supports optional dependencies but this way it isn't user-friendly. You're better off doing something like black does: https://github.com/conda-forge/black-feedstock/blob/main/recipe/meta.yaml

Kludex: I've simplified this. 👍

Kludex:
I'm ready when you are. 🙏

PR #25683, opened 2024-03-10
Added polars-ds as a package

thewchan:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/polars-ds/meta.yaml
+++ b/recipes/polars-ds/meta.yaml
@@ -17,7 +17,7 @@ build:
 requirements:
   build:
     - {{ compiler('c') }}
-    - {{ compiler('rust') }}
+    - {{ compiler('rust') }} 1.77
     - make
   host:
     - python
--- a/recipes/polars-ds/meta.yaml
+++ b/recipes/polars-ds/meta.yaml
@@ -19,14 +19,14 @@ requirements:
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
   host:
-    - python >=3.8
+    - python
     - maturin >=1.2,<2.0
     - pip
 
     # from maturin[patchelf]
     - patchelf
   run:
-    - python >=3.8
+    - python
     - polars >=0.20.6
 
     - __osx >={{ MACOSX_DEPLOYMENT_TARGET|default("10.9") }}  # [osx and x86_64]
--- a/recipes/polars-ds/meta.yaml
+++ b/recipes/polars-ds/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 167f29db7b498083fc8c3ac4855bd1bef7711843cd921cb048026bf5d603e729
 
 build:
-  skip: True  # [py<38]
+  skip: True  # [py<38 or win]
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
 
@@ -47,3 +47,4 @@ about:
 extra:
   recipe-maintainers:
     - thewchan
+    - xhochy
--- a/recipes/polars-ds/meta.yaml
+++ b/recipes/polars-ds/meta.yaml
@@ -26,7 +26,6 @@ requirements:
   run:
     - python
     - polars >=0.20.6
-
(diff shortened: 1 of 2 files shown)


thewchan:
Gosh this took forever to bulid... but win build is having some trouble:

```
  fatal runtime error: Rust cannot catch foreign exceptions
  error: could not compile `polars-arrow` (lib)
```

@conda-forge/help-rust @conda-forge/help-python Any pointers here?

recipes/polars-ds/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - python
+    - maturin >=1.2,<2.0
+    - pip
+
+    # from maturin[patchelf]
+    - patchelf  # [not-win]
+  run:
+    - python
+    - polars >=0.20.6
+
```
xhochy, line 32:
```suggestion
```

recipes/polars-ds/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
+    - make
+  host:
+    - python
+    - maturin >=1.2,<2.0
+    - pip
+
+    # from maturin[patchelf]
+    - patchelf  # [not-win]
```
xhochy, line 28: Do you need this at all? The selector is invalid, so I guess not. If you would need it, it should go into build.
```suggestion
    - patchelf  # [not win]
```

xhochy (changes requested):

recipes/polars-ds/meta.yaml
thewchan: I can fix the selector; it might not be needed since it was built fine. Upstream just specified maturin[patchelf] that's why I included it. https://github.com/abstractqqq/polars_ds_extension/blob/v0.3.3/requirements.txt

I can remove it if you think best @xhochy?

xhochy: Better remove it.

xhochy:
The windows error looks like out-of-memory. It might be that we need to apply the same workaround there as we do on the polars-feedstock.

thewchan:
@xhochy Do you know what kind of work around for win build is needed?

pavelzw:
You need to do cross-compilation from linux to windows. Regular cross-compilation with python in the host env from linux to windows is unfortunately not possible as windows envs cannot be installed on linux in conda. Thus, you need to do manual maturin build, see https://github.com/conda-forge/polars-feedstock/blob/bb361469100b5e86c7f582c46c57fcc67cc219d0/recipe/build.sh#L16-L20

pavelzw:
But this requires a conda-forge.yml to be present so you can only start doing this once the feedstock exists: https://github.com/conda-forge/polars-feedstock/blob/bb361469100b5e86c7f582c46c57fcc67cc219d0/conda-forge.yml#L10

thewchan:
@xhochy @pavelzw this seems pretty tricky but I'll do my best to take a stab... But at least for now it seems like the feedstock first needs to be created by merging first without a valid win build before we can do anythin

thewchan:
@pavelzw I can do that, @xhochy do you agree with this approach?

xhochy:
Yes, let's do that and add me as a maintainer so I can assist you on the feedstock.

thewchan:
Did I mess something up? Now all the builds are failing... @xhochy

recipes/polars-ds/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+build:
+  skip: True  # [py<38 or win]
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
```
xhochy, line 20:
```suggestion
    - {{ compiler('rust') }} 1.77
```

xhochy:
I have updated the rust compiler in the last days. That could be related.

thewchan:
oops this was dropped off my queue, is this good to be merged? @xhochy ?

xhochy (approved):

PR #25688, opened 2024-03-10
Add a noqt package and variants noqt5, noqt6

hmaarrfk:
See https://github.com/conda-forge/opencv-feedstock/issues/293 and more recent interest: https://github.com/conda-forge/staged-recipes/pull/25669

@conda-forge/qt-main requests for comments and thoughts is greatly appreciated

@ocefpaf one of my concerns was about "maintainability" of this package. It seems that the only way to update such a package is through repodata patches. As such, I decided to create out of an abundance of caution a noqt4 package. Let me know what you think.

cc: @traversaro @jschueller 

Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -1,7 +1,7 @@
 {% set version = "1.0" %}
 
 package:
-  name: noqt
+  name: noqt-helpers
   version: {{ version }}
 
 
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -55,10 +55,10 @@ outputs:
         - echo "no tests for this package"
 
 about:
-  home: https://github.com/conda-forge/nomkl-feedstock
+  home: https://github.com/conda-forge/noqt-feedstock
   license: BSD-3-Clause
   license_file: LICENSE
-  summary: 'prevents qt installation'
+  summary: 'Prevents Qt installation'
 
 extra:
   recipe-maintainers:
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -13,19 +13,8 @@ outputs:
   - name: noqt
     requirements:
       run:
-        - {{ pin_subpackage('noqt4', exact=True) }}
         - {{ pin_subpackage('noqt5', exact=True) }}
         - {{ pin_subpackage('noqt6', exact=True) }}
-    test:
-      commands:
-        - echo "no tests for this package"
-  - name: noqt4
-    build:
-      noarch: generic
-    requirements:
-      run_constrained:
-        # qt4 only ever had a "mega qt" package called qt
-        - qt <0.0a0|>=5.0a0
     test:
       commands:
         - echo "no tests for this package"
@@ -35,7 +24,7 @@ outputs:
     requirements:
       run_constrained:
         # qt-main only exists as of 5.x
-        - qt-main <0.0a0
+        - qt-main <5.0a0|>=6.0a0
         # before the creation of qt-main, a megapackage called qt existed
         - qt <0.0a0
     test:
@@ -47,8 +36,10 @@ outputs:
       noarch: generic
     requirements:
       run_constrained:
+        # While qt-main still doesn't point to qt6, it may in the future...
+        - qt-main <6.0a0|>=7.0a0
         # For qt6, we always had qt6-main
-        - qt6-main <0.0a0
+        - qt6-main <6.0a0
     test:
       commands:
         - echo "no tests for this package"
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -64,3 +64,4 @@ extra:
   recipe-maintainers:
     - hmaarrfk
     - conda-forge/qt-main
+  feedstock-name: noqt
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -24,7 +24,7 @@ outputs:
     requirements:
       run_constrained:
         # qt-main only exists as of 5.x
-        - qt-main <5.0a0|>=6.0a0
+        - qt-main <5.0a0|>=6.0
         # before the creation of qt-main, a megapackage called qt existed
         - qt <0.0a0
     test:
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -37,7 +37,7 @@ outputs:
     requirements:
       run_constrained:
         # While qt-main still doesn't point to qt6, it may in the future...
-        - qt-main <6.0a0|>=7.0a0
+        - qt-main <6.0a0
         # For qt6, we always had qt6-main
         - qt6-main <6.0a0
     test:
--- a/recipes/noqt/meta.yaml
+++ b/recipes/noqt/meta.yaml
@@ -9,24 +9,23 @@ build:
   number: 0
   noarch: generic
 
-requirements:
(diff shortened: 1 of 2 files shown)


recipes/noqt/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+        - echo "no tests for this package"
+  - name: noqt5
+    build:
+      noarch: generic
+    requirements:
+      run_constrained:
+        # qt-main is essentially qt5
+        - qt-main <0.a0
+        # before the creation of qt-main, a megapackage called qt existed
+        - qt <0.a0
```
hmaarrfk, line 41: should this be loosened to `qt <5.a0` to allow for qt4 to be installed? i feel like that will cause a whole set of other problems.

ocefpaf:
> @ocefpaf one of my concerns was about "maintainability" of this package. It seems that the only way to update such a package is through repodata patches. As such, I decided to create out of an abundance of caution a noqt4 package. Let me know what you think.

For some odd reason I missed this ping. I'm OK with it. I don't see any obvious  downsides.

hmaarrfk:
I've reopened this because it seems like it may help at compilation to resolve environments.
https://github.com/conda-forge/freecad-feedstock/pull/122

recipes/noqt/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - name: noqt4
+    build:
+      noarch: generic
+    requirements:
+      run_constrained:
+        # qt4 only ever had a "mega qt" package called qt
+        - qt <0.a0|>=5.a0
+    test:
+      commands:
+        - echo "no tests for this package"
```
h-vetinari, line 32: I would skip `noqt4` completely. qt4 is so old (the last [build](https://anaconda.org/conda-forge/qt/files?version=4.8.7) was >5 years ago) that it's essentially uninstallable anyway. 

With `noqt5` requiring both `qt <0.0a0` and `qt-main <0.0a0`, that already encompasses all remnants of qt4 that might exist.

recipes/noqt/meta.yaml
h-vetinari, line 41: We need a minor version before the "a0". Also, I would only do `qt-main <5.0a0`, so we don't close the door that `qt6-main` eventually becomes `qt-main`. That the output got renamed was to enable co-installation; once qt5 fades away, we won't need that anymore. Not saying that's a good enough reason to then change it, but it might happen.
```suggestion
        # qt-main is essentially qt5
        - qt-main <5.0a0
        # before the creation of qt-main, a megapackage called qt existed
        - qt <0.0a0
```

recipes/noqt/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+  - name: noqt6
+    build:
+      noarch: generic
+    requirements:
+      run_constrained:
+        # For qt6, we always had qt6-main
+        - qt6-main <0.a0
```
h-vetinari, line 52:
```suggestion
      run_constrained:
        # For qt6, we always had qt6-main
        - qt6-main <0.0a0
```

recipes/noqt/meta.yaml
h-vetinari: Exactly, hence why I think the distinction between `noqt5` and `noqt4` isn't worth it.

h-vetinari (commented):
Thanks for coming up with this! I think the idea is great, and it would be great to have for certain build scenarios! Some comments below.

h-vetinari:
> one of my concerns was about "maintainability" of this package. It seems that the only way to update such a package is through repodata patches.

What do you mean here? Like if the structure of the qt packages changes, we need to modify the repodata? I think that's an acceptable price if we end up doing big changes (which would have to be well-justified in and of themselves anyway).

recipes/noqt/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - {{ pin_subpackage('noqt4', exact=True) }}
+    - {{ pin_subpackage('noqt5', exact=True) }}
+    - {{ pin_subpackage('noqt6', exact=True) }}
+test:
+  commands:
+    - echo "no tests for this package"
+
+outputs:
+  - name: noqt
+  - name: noqt4
```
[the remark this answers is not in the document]
hmaarrfk: ok

recipes/noqt/meta.yaml
hmaarrfk: ok. i removed `noqt4`.

recipes/noqt/meta.yaml
hmaarrfk: Thanks o hope i addressed the 2 points.

recipes/noqt/meta.yaml
hmaarrfk: thnaks.

hmaarrfk:
I mostly mean helping to reduce the scope of this.

For now, that means removing the `noqt4` package.

recipes/noqt/meta.yaml
```diff
@@ -13,29 +13,18 @@ outputs:
...
-      commands:
-        - echo "no tests for this package"
   - name: noqt5
     build:
       noarch: generic
     requirements:
       run_constrained:
         # qt-main only exists as of 5.x
-        - qt-main <0.0a0
+        - qt-main <5.0a0|>=6.0a0
```
h-vetinari, line 27: I remember some discussion in mamba recently about these kind of "or" specifiers. Can we get confirmation that they're fully supported in mamba (@jjerphan @JohanMabille) & pixi (@wolfv @baszalmstra)?

recipes/noqt/meta.yaml
```diff
@@ -47,8 +36,10 @@ outputs:
       noarch: generic
     requirements:
       run_constrained:
+        # While qt-main still doesn't point to qt6, it may in the future...
+        - qt-main <6.0a0|>=7.0a0
```
h-vetinari, line 40: If you allow `qt7` here, then `noqt` needs a separate run-constraint (probably a better idea anyway than depending on the subpackages IMO), because just depending on `noqt{5,6}` would then still allow `noqt` to be installed with `qt7`.

hmaarrfk:
```suggestion
        # While qt-main still doesn't point to qt6, it may in the future...
        - qt-main <6.0a0
```

recipes/noqt/meta.yaml
wolfv: Fully supported by rattler/py-rattler/pixi/rattler-build as well.

recipes/noqt/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    test:
+      commands:
+        - echo "no tests for this package"
+  - name: noqt5
+    build:
+      noarch: generic
+    requirements:
+      run_constrained:
+        # qt-main only exists as of 5.x
+        - qt-main <5.0a0|>=6.0a0
```
h-vetinari, line 27: Don't really need the `a0` for the `>=` condition.
```suggestion
        - qt-main <5.0a0|>=6.0
```

h-vetinari (approved):
LGTM, thanks!

h-vetinari:
I think we're good to merge here?

hmaarrfk:
yes, feel free to press merge, but i don't think it resolves your main concern. (more info on the original feedstock that brought us here)

hmaarrfk:
I think this can at least help CIs avoid a qt download. users can add "noqt" to help the solving from finding qt.

PR #25689, opened 2024-03-11
Adding `eval_type_backport` recipe

tasansal:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/eval-type-backport/meta.yaml
+++ b/recipes/eval-type-backport/meta.yaml
@@ -1,0 +1,44 @@
+{% set name = "eval_type_backport" %}
+{% set version = "0.1.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/eval_type_backport-{{ version }}.tar.gz
+  sha256: d83ee225331dfa009493cec1f3608a71550b515ee4749abe78da14e3c5e314f5
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.7
+    - setuptools
+    - wheel
+    - setuptools-scm
+    - pip
+  run:
+    - python >=3.7
+
+test:
+  imports:
+    - eval_type_backport
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/alexmojaki/eval_type_backport
+  summary: Like `typing._eval_type`, but lets older Python versions use newer typing features.
+  license: MIT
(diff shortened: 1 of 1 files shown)


tasansal:
@alexmojaki I also added you as a maintainer here, can you please leave a comment to confirm you are willing to be listed as a maintainer.

alexmojaki:
:+1: I am willing to be listed as a maintainer.

tasansal:

tasansal:
@ocefpaf quick question. Just realized this. The package on pypi is `eval-type-backport` and the Python module is `eval_type_backport`. 

In here, should we rename the `pagkage::name` section to the one with hyphens for consistency? or does conda not care if its hyphen or underscore?

TIA.

ocefpaf:
> In here, should we rename the `pagkage::name` section to the one with hyphens for consistency? or does conda not care if its hyphen or underscore?

A package name doesn't necessarily have to be the same as its import name. With that said, it is nice to use the name that the upstream devs used to publish it. However, in Python land, specifically PyPI, both `-` and `_` versions are OK for installing a package. That is not true for conda b/c conda is not a Python package manager and name-something is different from name_something. What we can do to mimic PyPI's name handling is to publish another package, using multiple outputs, with the name variant. See https://github.com/conda-forge/importlib_resources-feedstock/blob/main/recipe/meta.yaml for an example.

tasansal:
> > In here, should we rename the `pagkage::name` section to the one with hyphens for consistency? or does conda not care if its hyphen or underscore?
> 
> A package name doesn't necessarily have to be the same as its import name. With that said, it is nice to use the name that the upstream devs used to publish it. However, in Python land, specifically PyPI, both `-` and `_` versions are OK for installing a package. That is not true for conda b/c conda is not a Python package manager and name-something is different from name_something. What we can do to mimic PyPI's name handling is to publish another package, using multiple outputs, with the name variant. See https://github.com/conda-forge/importlib_resources-feedstock/blob/main/recipe/meta.yaml for an example.

ok I see. Is it ok to reuse the feedstock that was created from this PR? Or should we make a new one and archive the old one?

ocefpaf:
> ok I see. Is it ok to reuse the feedstock that was created from this PR?

Yep. That is preferable. No need to create a new one.

alexmojaki:
Why do both of these work already?

- https://anaconda.org/conda-forge/eval-type-backport
- https://anaconda.org/conda-forge/eval_type_backport

tasansal:
> Why do both of these work already?
> 
> * https://anaconda.org/conda-forge/eval-type-backport
> * https://anaconda.org/conda-forge/eval_type_backport

I have modified the package name in https://github.com/conda-forge/eval_type_backport-feedstock to make it work with hyphens. We now just need to modify the recipe `meta.yaml` there per @ocefpaf's suggestion to make sure it updates both. Right now it will only update the one with hyphens.

PR #25766, opened 2024-03-17
Adding fluoriclogppka package

PetroPavliukk:
Checklist
- [x] Build number is 0.


zklaus:
It's `pip check` that is failing, not detecting `rdkit`. Since the conda-forge `rdkit` package is installed, that suggests that the `rdkit` package is missing the required pip installation information.

zklaus:
Indeed, the latest version (2023.09.6) has the required `lib/python3.11/site-packages/rdkit-2023.9.6.dist-info` directory that is missing from the old version that you are pinning here. Generally speaking, pinning everything to specific versions is a bit frowned upon, since then we cannot update packages. Can you relax the pins? Which of them are actually really needed?

PetroPavliukk:
> Indeed, the latest version (2023.09.6) has the required `lib/python3.11/site-packages/rdkit-2023.9.6.dist-info` directory that is missing from the old version that you are pinning here. Generally speaking, pinning everything to specific versions is a bit frowned upon, since then we cannot update packages. Can you relax the pins? Which of them are actually really needed?

In theory, I can install a new version of rdkit, but there is a problem that for the dataset on which we generated features and trained the models, the single feature calculation function started to work differently than in the version we used and our reproduced prediction results will a little different. But from what I understand, in that case, I'd better write my own function similar to the one in the old version of rdkit and use the latest version of rdkit?

zklaus:
I am not familiar enough with `rdkit` to comment on that. It is fine to skip the `pip check`. In that case, I would not completely remove it, but comment it out and leave a comment in the `meta.yaml` why this is not feasible at the moment.

PetroPavliukk:
i'll try, thanks a lot for your help

ogurbych:
I would like to be the reviewer for this package

DkraAI:
I would like to be the reviewer for this package.

PetroPavliukk:

recipes/fluoriclogppka/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - dgl
+    - dgllife ==0.3.2
+
+test:
+  imports:
+    - fluoriclogppka
+  commands:
+    - pip check || true  # We allow the check to fail because of a problem with rdkit, because the rdkit package is missing the required pip installation information.
+  requires:
+    - pip
```
BastianZim, line 52:
```suggestion
```

BastianZim (approved):

recipes/fluoriclogppka/meta.yaml
BastianZim: Either one works but this is a bit more clean.

PR #25791, opened 2024-03-20
Add USalign

SSchott:
Checklist
- [ x] Build number is 0.

First recipe I am testing by myself for a package we use in our projects. Source code is plain cpp, no tarball. License included as stated in source code, and assigned as BSD-2-Clause as disclosed by the author in scripts under https://zhanggroup.org/US-align/bin/module/USalignLinux64.zip.

### DIFF
--- a/recipes/usalign/meta.yaml
+++ b/recipes/usalign/meta.yaml
@@ -1,0 +1,42 @@
+{% set name = "USalign" %}
+{% set version = "20240303" %}
+{% set sha256 = "10408172007c6c99abe69b0ba5b7848cbf980dd5701c0e27c5b7e57db9dc9cc4" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://zhanggroup.org/US-align/bin/module/USalign.cpp  
+  sha256: {{ sha256 }}
+
+build:
+  skip: True # [win]
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+
+test:
+  commands:
+    - which USalign
+
+about:
+  home: https://zhanggroup.org/US-align
+  summary: 'US-align: universal structure alignment of monomeric and complex proteins and nucleic acids'
+  description: |
+    US-align (Universal Structural alignment) is a unified protocol
+    to compare 3D structures of different macromolecules (proteins, RNAs and DNAs)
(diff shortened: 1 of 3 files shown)


SSchott:
Changed to use the zip file instead.

I think this should be ready for feedback? @conda-forge/help-c-cpp

recipes/usalign/meta.yaml
```diff
@@ -0,0 +1,42 @@
+{% set name = "USalign" %}
+{% set version = "20240303" %}
+{% set sha256 = "b4da9fc58449f96a98eabf4137082cfc161787a5df739df351186e70b9c87950" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://zhanggroup.org/US-align/bin/module/USalignLinux64.zip
```
carterbox, line 10: https://zhanggroup.org/US-align/bin/module/USalign.cpp

It should be this address, so we don't download the precompiled binary?

recipes/usalign/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+about:
+  home: https://zhanggroup.org/US-align
+  summary: 'US-align: universal structure alignment of monomeric and complex proteins and nucleic acids'
+  description: |
+    US-align (Universal Structural alignment) is a unified protocol
+    to compare 3D structures of different macromolecules (proteins, RNAs and DNAs)
+    in different forms (monomers, oligomers and heterocomplexes) for
+    both pairwise and multiple structure alignments.
+  license: BSD-2-Clause
+  license_family: BSD
```
carterbox, line 35: [This is not BSD-2-Clause](https://opensource.org/license/bsd-2-clause). Use "OTHER" for the license description or figure out what the license actually is.

recipes/usalign/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - sysroot_linux-64  # [linux]
+
+test:
+  commands:
+    - which USalign
```
carterbox, line 24:
```suggestion
    - USalign --help
```

```suggestion
    - USalign --version
```
Do something like this instead because the binaries is actually executed. If you want to check that it's the correct USalign, you should use an existence check to `$PREFIX/bin/USalign`.

carterbox (changes requested):

PR #25857, opened 2024-03-25
Add package scikit-fdiff

thealanjason:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/scikit-fdiff/meta.yaml
+++ b/recipes/scikit-fdiff/meta.yaml
@@ -23,7 +23,7 @@ requirements:
     - numpy
     - numba
     - scipy
-    - sympy <1.6
+    - sympy
     - streamz
     - more-itertools
     - xarray
@@ -45,10 +45,7 @@ requirements:
 test:
   imports:
     - skfdiff
-  commands:
-    - pip check
-  requires:
-    - pip
+  
 
 about:
   home: https://gitlab.com/celliern/scikit-fdiff/
--- a/recipes/scikit-fdiff/meta.yaml
+++ b/recipes/scikit-fdiff/meta.yaml
@@ -1,0 +1,62 @@
+{% set name = "scikit-fdiff" %}
+{% set version = "0.7.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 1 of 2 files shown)


thealanjason:
@conda-forge/help-python

recipes/scikit-fdiff/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - numba
+    - scipy
+    - sympy <1.6
+    - streamz
+    - more-itertools
+    - xarray
+    - dask-core
+    - pyyaml
+    - joblib
+    - path
```
synapticarbors, line 33: It looks like this should be `path.py` rather than `path` according to the upstream `setup.cfg`. I'm not sure if you'll need quotation marks around it due to the dot in the name. Could be fine without.

recipes/scikit-fdiff/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+about:
+  home: https://gitlab.com/celliern/scikit-fdiff/
+  summary: Automatic finite difference discretization for 1D PDE with fast temporal solvers.
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - thealanjason
+    - conda-forge/python
```
synapticarbors, line 64: You shouldn't list `conda-forge/python` as a co-maintainer.

synapticarbors (changes requested):

recipes/scikit-fdiff/meta.yaml
thealanjason: Fixed

recipes/scikit-fdiff/meta.yaml
thealanjason: I shall now try with quotation marks. Previously it did not work without them.

thealanjason (commented):
Updated feedback from Review

synapticarbors:
It looks like `sympy <1.6` is causing issues. sympy 1.5.1 (which I think is the last version that satisfies the constraint) was only built for python <3.9. I'm wondering if we can relax the sympy version requirement? You'd probably have to remove the `pip check` then.

thealanjason (commented):
Building passes now. 
Thanks for the hint on sympy version.

thealanjason:
@synapticarbors Please have a look :)

PR #25869, opened 2024-03-26
Adding AreTomo2 recipe

apeck12:
Checklist
- [ x ] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [ x ] Source is from official source.
- [ x ] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).
- [ x ] If static libraries are linked in, the license of the static library is packaged.
- [ x ] Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md).
- [ x ] Build number is 0.
- [ x ] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details).

### DIFF
--- a/recipes/AreTomo2/build.sh
+++ b/recipes/AreTomo2/build.sh
@@ -1,0 +1,3 @@
+which nvcc
+make exe -f makefile11 [CUDAHOME=path/cuda-xx.x]
+cp AreTomo2 ${PREFIX}/bin/
--- a/recipes/AreTomo2/meta.yaml
+++ b/recipes/AreTomo2/meta.yaml
@@ -1,0 +1,90 @@
+# Note: there are many handy hints in comments in this example -- remove them when you've finalized your recipe
+# If your package is python based, we recommend using Grayskull to generate it instead:
+# https://github.com/conda-incubator/grayskull
+
+# Jinja variables help maintain the recipe as you'll update the version only here.
+# Using the name variable with the URL in line 16 is convenient
+# when copying and pasting from another recipe, but not really needed.
+{% set name = "AreTomo2" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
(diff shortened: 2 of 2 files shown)


szhengczii:
I am willing to be a maintainer of this conda feedstock.

apeck12:
@alisterburt, thanks for showing me the conda-forge ropes earlier! It looks like we've correctly tracked down the CUDA path and are on to the next bug:
```
make: g++: No such file or directory
make: *** [makefile11:192: Util/CNextItem.o] Error 127
```
so it seems like it fails with the first C++ file that it tries to compile. Unfortunately I'm at a loss for how to debug this. 

I'm not sure if it's relevant, but I forgot that there actually are two pre-compiled libraries that other files are linked against during compilation. The source code for those is also included, so they could be regenerated by running their respective makefiles if needed. Do they also need to be referenced somewhere in the meta.yaml file?

alisterburt:
@apeck12 I'm also not great at building/debugging these things but it looks like this SO answer might be on the right track 

https://stackoverflow.com/questions/12919081/gcc-g-no-such-file-or-directory

If that's right, some files which are included in the file being compiled are not found by the compiler...

@shawnzhengczii how are include directories handled in your build/makefile?

alisterburt:
@apeck12 I missed your last comment about the two libraries

Ideally those would be built prior to AreTomo2 if they are required 🙂

alisterburt:
If they are shared libraries you will need to copy them into $PREFIX/lib similar to how we copy the aretomo binary into the bin directory

apeck12:
Thanks for that tip! Testing whether copying the prebuilt libraries to ${PREFIX}/lib resolves the issue. The makefile specifies the path to those libraries using `-I` flag.

PR #25877, opened 2024-03-27
add hatch-docstring-description

flying-sheep:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/hatch-docstring-description/meta.yaml
+++ b/recipes/hatch-docstring-description/meta.yaml
@@ -35,7 +35,7 @@ test:
 about:
   home: https://github.com/flying-sheep/hatch-docstring-description
   summary: Hatch plugin to derive package description from Python package docstring
-  license: GPL-3.0
+  license: GPL-3.0-or-later
   license_file: LICENSE.md
 
 extra:
--- a/recipes/hatch-docstring-description/meta.yaml
+++ b/recipes/hatch-docstring-description/meta.yaml
@@ -1,0 +1,42 @@
+{% set name = "hatch-docstring-description" %}
+{% set version = "1.0.2" %}
+
+package:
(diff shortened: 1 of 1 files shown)


recipes/hatch-docstring-description/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - hatch_docstring_description
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/flying-sheep/hatch-docstring-description
+  summary: Hatch plugin to derive package description from Python package docstring
+  license: GPL-3.0
```
xhochy, line 38: The linter is correct. This must either be `GPL-3.0-or-later` or `GPL-3.0-only`.

xhochy (changes requested):

recipes/hatch-docstring-description/meta.yaml
flying-sheep: well, that’s an actionable thing, since I’m the maintainer and can say what it meant.

flying-sheep:
Hm, I did this PR for https://github.com/conda-forge/legacy-api-wrap-feedstock/pull/5, but it seems like that builds without having this build dependency installed? Is it enough to have build dependencies on PyPI?

xhochy:
> Is it enough to have build dependencies on PyPI?

No, build dependencies also must be in conda-forge. Maybe in this case it worked because you are building from the sdist and the hatch plugin is only needed when generating the sdist and not during the actual installation process.

xhochy (approved):

flying-sheep:
Ah, leaving out the build dep just happened to work. There’s not description in the installed metadata, but that’s not important.

PR #25934, opened 2024-04-02
Add tmuxp and libtmux (noarch python)

AntoinePrv:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/libtmux/conda-forge.yml
+++ b/recipes/libtmux/conda-forge.yml
@@ -1,0 +1,3 @@
+noarch_platforms:
+  - linux_64
+  - win_64
--- a/recipes/libtmux/conda-forge.yml
+++ b/recipes/libtmux/conda-forge.yml
@@ -1,3 +1,2 @@
 noarch_platforms:
   - linux_64
-  - win_64
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -16,7 +16,7 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - pip
     - poetry-core >=1.0
   run:
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -13,7 +13,7 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
   number: 0
-  skip: true  # [win]
+  skip: true
 
 requirements:
   host:
@@ -22,7 +22,7 @@ requirements:
     - poetry-core >=1.0
   run:
     - python >=3.8
-    - tmux
+    - tmux  # [not win]
 
 test:
   imports:
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -16,11 +16,11 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - pip
     - poetry
   run:
-    - python
+    - python >=3.8
 
 test:
   imports:
@@ -48,4 +48,3 @@ about:
 extra:
   recipe-maintainers:
     - AntoinePrv
-
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -12,7 +12,7 @@ source:
 build:
   script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
   number: 0
-  skip: true  # [win]
+  noarch: python
 
 requirements:
   host:
@@ -20,8 +20,10 @@ requirements:
     - pip
     - poetry-core >=1.0
   run:
-    - python
-    - tmux
+    - python >=3.8
+    - tmux    # [unix]
+    - __unix  # [unix]
+    - __win   # [win]
 
 test:
   imports:
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -50,3 +50,4 @@ about:
 extra:
   recipe-maintainers:
     - AntoinePrv
+    - tony
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -1,0 +1,51 @@
+{% set name = "libtmux" %}
+{% set version = "0.36.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 12b5554b3a19d663d2a04f30b87fb063bd6456463a3ef6c6445a721fd7f7569a
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+    - poetry
+  run:
+    - python
+
+test:
+  imports:
+    - libtmux
+    - libtmux.session
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://libtmux.git-pull.com/
+  doc_url: https://libtmux.git-pull.com/
+  dev_url: https://github.com/tmux-python/libtmux
+  summary: Python API / wrapper for tmux
+  description: |
+    libtmux is a typed Python library that provides a wrapper for interacting programmatically
+    with tmux, a terminal multiplexer. You can use it to manage tmux servers, sessions, windows,
+    and panes.
+    Additionally, libtmux powers tmuxp, a tmux workspace manager.
+  license: MIT
+  license_family: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - AntoinePrv
+
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -21,9 +21,7 @@ requirements:
     - poetry-core >=1.0
   run:
     - python >=3.8
-    - tmux    # [unix]
-    - __unix  # [unix]
-    - __win   # [win]
+    - tmux
 
 test:
   imports:
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -13,14 +13,16 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
   number: 0
+  skip: true  # [win]
 
 requirements:
   host:
     - python >=3.8
     - pip
-    - poetry
+    - poetry-core >=1.0
   run:
     - python >=3.8
+    - tmux
 
 test:
   imports:
--- a/recipes/libtmux/meta.yaml
+++ b/recipes/libtmux/meta.yaml
@@ -16,11 +16,11 @@ build:
 
 requirements:
   host:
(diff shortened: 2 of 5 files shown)


AntoinePrv:
Welcome @tony, thanks for joining. Once this is merged, it will get moved to its own repository.
A lot of things are automated, and all we'll likely be doing is merge PR and ensure dependencies are in sync with the project. Full do is available on [conda-forge.org.](https://conda-forge.org/).

AntoinePrv:
@conda-forge/help-python I think this is ready. I had to make the package an arch one because of a dependency that does not exists on Windows, and because I believe conda-build mistook a Python file for a binary one (and so was making an error).
I think we can merge this and make it noarch in the feedstock once there is no Windows CI.

recipes/libtmux/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 12b5554b3a19d663d2a04f30b87fb063bd6456463a3ef6c6445a721fd7f7569a
+
+build:
+  script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
+  number: 0
+  skip: true  # [win]
```
ocefpaf, line 15: My guess is that these packages can be OS-noarch. See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

ocefpaf: >  is it OK to change that file in the PR on staged-recipe?

You have to add one in the recipe directory, that one will be ported to the feedstock.

AntoinePrv:
@ocefpaf the error on Windows (something to do with [this](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/c-runtime-error-r6033?view=msvc-170)?) was really convoluted (especially for a noarch package) so the `tmuxp` package on Windows in simply empty. I think it's a good compromise since anyways the package is of no use without `tmux`. Let me know what you think.

ocefpaf:
> @ocefpaf the error on Windows (something to do with [this](https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/c-runtime-error-r6033?view=msvc-170)?) was really convoluted (especially for a noarch package) so the `tmuxp` package on Windows in simply empty. I think it's a good compromise since anyways the package is of no use without `tmux`. Let me know what you think.

I though you did not want Windows there, no? That is why I suggested a OS-noarch, so you could do linux and macOS noarch. Does that make sense?

recipes/libtmux/conda-forge.yml
```diff
@@ -0,0 +1,3 @@
+noarch_platforms:
+  - linux_64
+  - win_64
```
ocefpaf, line 3: My guess is that you want it on macOS but not Windows, right?

```suggestion
  - osx_64
```

recipes/libtmux/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+requirements:
+  host:
+    - python
+    - pip
+    - poetry-core >=1.0
+  run:
+    - python >=3.8
+    - tmux    # [unix]
+    - __unix  # [unix]
+    - __win   # [win]
```
ocefpaf, line 26:
```suggestion
```

recipes/tmuxp/conda-forge.yml
```diff
@@ -0,0 +1,3 @@
+noarch_platforms:
+  - linux_64
+  - win_64
```
ocefpaf, line 3: My guess is that you want it on macOS but not Windows, right?

```suggestion
  - osx_64
```

recipes/tmuxp/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: 23b3ff0a8862a44c2bc687a553f78f4afd8f1e03371d775555a7699ed54572b4
+
+build:
+  # Seeing weird errors like the following on Windows, along with weird encoding.
+  # https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/c-runtime-error-r6033?view=msvc-170
+  script:
+    - {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv  # [not win]
```
ocefpaf, line 16:
```suggestion
    - {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv 
```

recipes/tmuxp/meta.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - python
+    - pip
+    - poetry-core >=1.0
+  run:
+    - python >=3.8
+    - colorama >=0.3.9
+    - pyyaml >=6.0
+    - libtmux >=0.36  # [unix]
+    - __unix          # [unix]
+    - __win           # [win]
```
ocefpaf, line 31:
```suggestion
```

recipes/libtmux/conda-forge.yml
AntoinePrv: I don't think we need `osx_64` as a platform since the `linux_64` and `osx_64` builds are the same (and are only constrained by `__unix`).
I would remove `win64` platform and selectors altogether but it seems in this CI the Windows builds will run nonetheless and fail, hence the need to make a dummy windows package.

ocefpaf: I believe we do need it here when makin os-noarch, or the binary will be installable only on Linux.

AntoinePrv: My understanding of `noarch_platforms` is that to run mutually exclusive `noarch` packages, one may still need different  platforms, for instance if dependencies are different. However this is not the case of `macos` and `linux` here which should have the same `noarch` package, in the same way that a truly `noarch` package can be built on `linux-64` only.
What do you think? Am I missing something ?

Running a clean build locally with `run_locally.py linux64`,  `cat build_artifacts/linux-64/current_repodata.json`  is empty and `cat build_artifacts/noarch/current_repodata.json`  outputs
```json
{
  "info": {
    "subdir": "noarch"
  },
  "packages": {},
  "packages.conda": {
    "libtmux-0.36.0-pyhca7485f_0.conda": {
      "build": "pyhca7485f_0",
      "build_number": 0,
      "depends": [
        "python >=3.8",
        "tmux"
      ],
      "legacy_bz2_md5": null,
      "license": "MIT",
      "license_family": "MIT",
      "md5": "b7591b3a7a24a5e3ec81231865222012",
      "name": "libtmux",
      "noarch": "python",
      "sha256": "502e150ef97a3ee747a91b6338bb3be76cb535cd8fd6f78cb959a39686c71048",
      "size": 48595,
      "subdir": "noarch",
      "timestamp": 1712670081414,
      "version": "0.36.0"
    },
    "tmuxp-1.45.0-pyhca7485f_0.conda": {
      "build": "pyhca7485f_0",
      "build_number": 0,
      "depends": [
        "colorama >=0.3.9",
        "libtmux >=0.36",
        "python >=3.8",
        "pyyaml >=6.0"
      ],
      "legacy_bz2_md5": null,
      "license": "MIT",
      "license_family": "MIT",
      "md5": "b67b0c8dbc4c43c234294a21891ace9f",
      "name": "tmuxp",
      "noarch": "python",
      "sha256": "49f76bc22975ab9dc18f653c7e94d255fe80f81f963c9d7e8c64d9ae04195bfd",
      "size": 47418,
      "subdir": "noarch",
      "timestamp": 1712670121330,
      "version": "1.45.0"
    }
  },
  "removed": [],
  "repodata_version": 1
}
```

ocefpaf: > What do you think? Am I missing something ?

Nope. I was probably missing that you did not had any specific dependency for macOS. Sorry for the red-herring.

recipes/libtmux/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  sha256: 12b5554b3a19d663d2a04f30b87fb063bd6456463a3ef6c6445a721fd7f7569a
+
+build:
+  script: {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
+  number: 0
+  noarch: python
+
+requirements:
+  host:
+    - python
```
ocefpaf, line 19:
```suggestion
    - python >=3.8
```

recipes/libtmux/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - pip
+    - poetry-core >=1.0
+  run:
+    - python >=3.8
+    - tmux
```
ocefpaf, line 24:
```suggestion
    - tmux
   - __unix  # [unix]
```

recipes/tmuxp/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  # Seeing weird errors like the following on Windows, along with weird encoding.
+  # https://learn.microsoft.com/en-us/cpp/error-messages/tool-errors/c-runtime-error-r6033?view=msvc-170
+  script:
+    - {{ PYTHON }} -m pip install --no-deps --no-build-isolation . -vv
+  number: 0
+  noarch: python
+
+requirements:
+  host:
+    - python
```
ocefpaf, line 22:
```suggestion
    - python >=3.8
```

AntoinePrv:
@ocefpaf thank you again for the feedback. I feel like this went full circle, how do you want to move forward?
This is what I see so far, unless you have other suggestions.

#### A - Regular `noarch: python`
No selectors. The Windows CI will be failing here but we merge anyways since there will not be Windows CI in the feedstock.
Optionally, we can add `__unix` dependency. This is is only change the error on Windows from "_`tmuxp` requires `libtmux` which requires `tmux` which is not available on Windows_" to "_`tmuxp` is not available on Windows_" (in idea).

#### B - `noarch: python` + `win_64`  in `noarch_platforms` with a dummy windows package
Same as before, but we use a Windows `noarch_platforms` to add selectors and make the CI pass here. This is only to please the Windows CI here. On the downsides it creates a `tmuxp` package on Windows that is empty (the linter won't accept `skip: True # [win]` with a `__unix` constraint).

#### C- Arch packages with `skip: True #[win]`
Also works with the CI here, but it needlessly creates many packages. We can reverse it in the feedstock (since again there won't be any Windows CI). On the plus side, it never creates a Windows package.

recipes/tmuxp/meta.yaml
ocefpaf: We need run and host with the same python constraint.

recipes/tmuxp/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - pip
+    - poetry-core >=1.0
+  run:
+    - python >=3.8
+    - colorama >=0.3.9
+    - pyyaml >=6.0
+    - libtmux >=0.36
```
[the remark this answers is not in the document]
ocefpaf: This will ensure the package is installable only on Linux and macOS.

recipes/tmuxp/meta.yaml
AntoinePrv: Sorry I forgot to push that one.

recipes/tmuxp/meta.yaml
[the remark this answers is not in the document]
AntoinePrv: I added it without the selector otherwise the linter complains.

PR #25983, opened 2024-04-09
Add Scipp to conda-forge

MridulS:
Checklist
- [x] Build number is 0.

(Will keep as draft till CI is happy.)


tkittel:
Nice, looking forward to this on conda-forge :-)

I might be wrong, but I believe you are supposed to add a `{{ stdlib("c") }}` to your build requirements (or wherever you depend on any compiler package): https://conda-forge.org/news/2024/03/24/stdlib-migration/

tkittel:
Regarding the linter feedback:

About the python version, check out https://github.com/conda-forge/ncrystal-feedstock/blob/main/recipe/meta.yaml for how you can use a skip statement instead. I.e. add something like 

```
  skip: true  # [py<38]
```

and then just list `python` in the actual requirements sections.

The other two lints are pretty self-explanatory I hope :-)

tkittel:
If you build with pip, make sure to add a `--no-deps` flag to the pip command (otherwise you might inadvertently pull in some PyPI packages). And then it is a good idea to add a `pip check` command to the tests. That way you will discover if scipp one day added a new dependency but you forgot to update the conda-forge meta.yaml accordingly.

tkittel:
So my experiences with SCM-based versioning is that you can't use github tarballs, but need to use the PyPI source dists as your sources. I think that is the cause if the latest build error. You can see an example here of how to sources from PyPI:

https://github.com/conda-forge/simple-build-system-feedstock/blob/main/recipe/meta.yaml

PR #25986, opened 2024-04-09
Add recipe for metatensor

Luthaf:
Checklist
- [ ] ~If static libraries are linked in, the license of the static library is packaged.~
- [x] Build number is 0.

----

This is adding a recipe for the core library in https://github.com/lab-cosmo/metatensor. A future PR will add another recipe for the torch bindings which live in the same repo (as well as the Python bindings to both the -core and -torch versions of the code).

This package is implemented in Rust, but builds with CMake and generally behaves like a C++ package from an outside point of view.


recipes/metatensor/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=x86_64-apple-darwin" %} #[osx and x86_64]
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=aarch64-apple-darwin" %} #[osx and arm64]
+
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=x86_64-pc-windows-msvc" %} #[win and x86_64]
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=aarch64-pc-windows-msvc" %} #[win and arm64]
+
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=x86_64-unknown-linux-gnu" %} #[linux and x86_64]
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=aarch64-unknown-linux-gnu" %} #[linux and aarch64]
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=powerpc64le-unknown-linux-gnu" %} #[linux and ppc64le]
```
Luthaf, line 15: @conda-forge/help-rust is there an easier way to do this?

Luthaf:
The windows error is related to our CMake script, I'll check it. The osx error is a bit stranger, and seems related to MACOSX_DEPLOYMENT_TARGET=10.9? But it should automatically be set to 10.12 according to https://conda-forge.org/news/2024/01/29/rust-175-osx-sdk/. Is there anything else I should be doing here?

Luthaf:
@conda-forge/help-rust and @conda-forge/help-c-cpp, ready for review!

recipes/metatensor/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  url: {{ url_base }}/metatensor-core-v{{ version }}/metatensor-core-cxx-{{ version }}.tar.gz
+  sha256: 8ecac73604d79d7f3e46880cd14370d85b26ade3c9aa8ab9594678e8f3083a4f
+  patches:
+    - mktemp-mkdir.patch
+
+build:
+  number: 0
+  script: |
+    cmake -G Ninja {{ CMAKE_CONFIG }} {{ METATENSOR_CONFIG }} {{ RUST_TARGET }} {{ CMAKE_INSTALL_PREFIX }} .
+    cmake --build . --config Release --target install
```
xhochy, line 34: Please put this into a `build.sh` script. Otherwise, this way too much Jinja2 variables in here and makes it quite unreadable.

recipes/metatensor/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set version = "0.1.5" %}
+{% set url_base = "https://github.com/lab-cosmo/metatensor/releases/download" %}
+
+{% set CMAKE_INSTALL_PREFIX = "-DCMAKE_INSTALL_PREFIX=$PREFIX" %}  # [not win]
+{% set CMAKE_INSTALL_PREFIX = "-DCMAKE_INSTALL_PREFIX=%PREFIX%/Library" %}  # [win]
```
xhochy, line 5: You can get this by passing `$CMAKE_ARGS` to the cmake call.

recipes/metatensor/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+{% set RUST_TARGET = "-DRUST_BUILD_TARGET=x86_64-apple-darwin" %}  # [osx and x86_64]
```
xhochy, line 7: Please use `-DRUST_BUILD_TARGET=$CONDA_RUST_TARGET` instead of manually listing them.

xhochy (changes requested):
Please also bundle third-party rust licenses using cargo-bundle-licenses (see other Rust PRs for that).

recipes/metatensor/meta.yaml
Luthaf: Do I need to add both build.sh and build.bat then? My justification for putting it here is that otherwise I need to duplicate the exact same code for unix & windows

recipes/metatensor/meta.yaml
Luthaf: thanks, I was looking for something like this!

Luthaf: Hmm, it looks like `$CONDA_RUST_TARGET` is not using the right format: it is `X86_64_UNKNOWN_LINUX_GNU` instead of `x86_64-unknown-linux-gnu` on linux for example. Is there another variable I can use instead?

xhochy: `CARGO_BUILD_TARGET` is the one then. See https://github.com/conda-forge/rust-activation-feedstock/blob/77337e3832b712a6de383c0a96dbe1a33d5f864f/recipe/activate.sh#L9-L16 and https://github.com/conda-forge/rust-activation-feedstock/blob/77337e3832b712a6de383c0a96dbe1a33d5f864f/recipe/conda_build_config.yaml for the source of these values.

Luthaf:
@conda-forge/help-rust and @conda-forge/help-c-cpp, this is ready for a second round of review =)

xhochy (approved):

PR #26068, opened 2024-04-17
Add recipe for ckzg

step21:
Checklist
- [x] Build number is 0.


zklaus:
Note that upstream has in the meantime added a patch for the Windows issue at https://github.com/ethereum/c-kzg-4844/pull/419. You can turn that into a patch here, or skip Windows until a new release with the patch comes out.

step21:
> Note that upstream has in the meantime added a patch for the Windows issue at [ethereum/c-kzg-4844#419](https://github.com/ethereum/c-kzg-4844/pull/419). You can turn that into a patch here, or skip Windows until a new release with the patch comes out.

I saw this, because as you can see I was in that discussion with upstream, but there is no indication that that patch solves this particular windows issue. I had wanted to test it, but as I wrote had/have trouble getting a system able to test it.

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - make
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+    - clang
```
xhochy, line 21: Why is `clang` listed here? You probably would either list it as a compiler or use `clangdev`.

step21: thanks. Most likely because it worked, and I didn't know that I could add it with compiler(). Interestingly, doing that meant it failed on Windows. So now I changed it to that but only use that line for `[lin]`.

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    - make
+    - {{ compiler('c') }}
+    - {{ compiler('clang') }}  # [linux]
```
xhochy, line 20: A cleaner approach for this is described in https://conda-forge.org/news/2024/04/30/clang-everywhere/ Havig both `compiler('c')` and `compiler('clang')` can lead to issues. Putting clang into cbc.yaml should prevent this.

step21: I tried that but it fails unfortunately. This might be because the build script explicitly uses gcc for windows. https://github.com/step21/c-kzg-4844/blob/main/src/Makefile

xhochy: Best is to limit it with `  # [linux]` on all lines of the cbc.yaml

step21: Somehow this is not working. Error from OSX `  /var/folders/n8/w901w1rn4ldg6yskstndsj200000gn/T/assembly-f95753.s:6555:1: error: invalid CFI advance_loc expression
  .cfi_adjust_cfa_offset 8`
Windows build just tries to apply clang options to gcc, which fails ofc.

step21: @xhochy  Please advise if I should fix/remove the build script from upstream to enable clang for all (assuming it works) or revert to the version that worked using the built-in build script, but uses gcc for windows.

step21:
@conda-forge/core please review/approve with the original build script or tell me how I can force/replace the build script to use only clang. The way @xhochy suggested did not work, thus I reverted the commits to a working build.

hmaarrfk:
what if you removed clang alltogether?

step21:
> what if you removed clang alltogether?

I assume it would still clash with this Makefile https://github.com/ethereum/c-kzg-4844/blob/main/src/Makefile which has 
```
# Cross-platform compilation settings.
ifeq ($(PLATFORM),Windows)
	CC = gcc
	CFLAGS += -D_CRT_SECURE_NO_WARNINGS
else
	CC = clang
	CFLAGS += -fPIC -Werror
endif
```
Windows also probably does not really use this file, as in `setup.py` there is

```
class CustomBuild(build_ext):
    def run(self):
        if system() == "Windows":
            try:
                check_call(["blst\\build.bat"])
            except Exception:
                pass
        check_call(["make", "-C", "src", "blst"])
        super().run()
```

I'm not saying these are great, but I assume I would have to patch them only use either clang or gcc. (and I do not know if there was a reason to only use either)

isuruf:
you'll have to patch `setup.py` to use something like
```
check_call(["make", "-C", "src", "blst", f"CC={os.environ['CC']}"])
```

step21:
This unfortunately fails on Win, which still uses cl.exe. Any ideas? (I also patched out the use of the windows bat file, as it didn't use either gcc or clang and I'm not sure what kind of magic it is doing.
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=961378&view=logs&jobId=240f1fee-52bc-5498-a14a-8361bde76ba0&j=240f1fee-52bc-5498-a14a-8361bde76ba0&t=7c0f8eae-6d6f-51bf-636f-73a1a7fb1bc4

jakirkham:
Maybe disable Windows and then we can review/merge?

The Windows portion could be revisited in the feedstock

step21:
> Maybe disable Windows and then we can review/merge?
> 
> The Windows portion could be revisited in the feedstock

Ok good idea. Windows build is skipped now.

step21:
This is ready for review, @conda-forge/help-python

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ckzg-{{ version }}.tar.gz
+  sha256: 4295acc380f8d42ebea4a4a0a68c424a322bb335a33bad05c72ead8cbb28d118
+  patches:
+    - build.patch
```
jakirkham, line 12: Let's drop this patch if unused. It can be attempted again in the feedstock

```suggestion
```

recipes/ckzg/build.patch
```diff
@@ -0,0 +1,18 @@
...
+-        if system() == "Windows":
+-            try:
+-                check_call(["blst\\build.bat"])
+-            except Exception:
+-                pass
+-        check_call(["make", "-C", "src", "blst"])
++        check_call(["make", "-C", "src", "blst", f"CC={os.environ['CC']}"])
+         super().run()
+ 
+
```
jakirkham, line 18:
```suggestion
```

recipes/ckzg/meta.yaml
jakirkham: This should be unneeded as we have `{{ compiler('c') }}` already. Let's drop it

```suggestion
```

jakirkham (commented):
Thanks Florian! 🙏

Added a couple minor suggestions below

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  build:
+    - make
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+    - clang
```
jakirkham, line 24: This should be unneeded as we have `{{ compiler('c') }}` already. Let's drop it

```suggestion
```

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,48 @@
+{% set name = "ckzg" %}
+{% set version = "1.0.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ckzg-{{ version }}.tar.gz
```
jakirkham, line 9:
```suggestion
  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```

step21:
Hey, @jakirkham without any patch, we are now again at the juncture of this Makefile I referred to here: https://github.com/conda-forge/staged-recipes/pull/26068#issuecomment-2181023138
This forces clang for linux and osx. On OSX that works, because it is installed by default. The main question is - how to force it to use only one compiler? In the pursuit of first merging, then fixing the feedstock, should I add clang back, so that only clang is used?

MementoRC:
@step21 After a few days looking into it, I proposed this: https://github.com/conda-forge/staged-recipes/pull/27412, which should build Windows using gcc 5, separate `blst` as its own package (with the py-binding being worked in a separate recipe that still has a few snags on linux)
Note that it is still work in progress as I need to fine-tune windows/osx by going through CI.

MementoRC:
Cool. I am still working on it. I want to enforce linking to the versioned dynamic library and it's a bit trickier. It feels more reliable, but maybe overkill. As for windows, it's my Achilles heel, I had it working but must have changed something during the many refactoring. I'll ping you when I get it to pass

MementoRC:
Similar functionality with separated `blst`/`ckzg` build to address the upstream package assumptions on compiler: https://github.com/conda-forge/staged-recipes/pull/27490

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  requires:
+    - pip
+
+about:
+  home: https://github.com/ethereum/c-kzg-4844
+  summary: Python bindings for C-KZG-4844
+  license: Apache-2.0
+  license_file:
+    - LICENSE
+    - blst/LICENSE
```
MementoRC, line 51: This might not be needed

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ckzg-{{ version }}.tar.gz
+  sha256: 4295acc380f8d42ebea4a4a0a68c424a322bb335a33bad05c72ead8cbb28d118
+  patches:
+    - setup.py.patch
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
```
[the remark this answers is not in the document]
step21: +1

recipes/ckzg/meta.yaml
step21: true

step21:
thanks for your comments. TBH I don't really care which PR we continue. Also added you as maintainer here. Though your has a lot more changes, I only patched out the build.bat so far, so I assume I would also have to carry those over.

step21:
builds! all pkgs are using clang. Only thing is that there is a deprecation warning for setup.py, but I would ignore it for now.

wolfv:
Looks good to me!

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  number: 0
+
+requirements:
+  build:
+    - make
+    - {{ compiler('clang') }}
+    - {{ stdlib('c') }}
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - libblst
```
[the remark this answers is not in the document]
step21: I wasn't quite sure, but this matches what is recommended under cross-compilation in the knowledge base. See https://conda-forge.org/docs/maintainer/knowledge_base/#cross-compilation (you have to scroll a bit)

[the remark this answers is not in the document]
jakirkham: Usually we only need things that we execute (like `python`, compilers, etc.)

So we can probably drop `libblst` and `libblst-headers`

recipes/ckzg/conda_build_config.yaml
```diff
@@ -0,0 +1,8 @@
+c_compiler:
+  - clang
+c_compiler_version:
+  - 18
+cxx_compiler:
+  - clangxx
+cxx_compiler_version:
+  - 18
```
jakirkham, line 8: As we don't appear to use C++ here, we can drop this part

```suggestion
```

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - libblst-headers
```
jakirkham, line 26: Suggesting dropping these per the discussion in this thread: https://github.com/conda-forge/staged-recipes/pull/26068#discussion_r1810520733

```suggestion
```

step21:
Thanks @jakirkham for the suggestions, should all be applied now.

recipes/ckzg/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+  host:
+    - python
+    - pip
+    - libblst
+    - libblst-headers
+  run:
+    - python
+    - libblst
```
jakirkham, line 32: Appears `libblst` is missing `run_exports`, have raised this discussion in issue: https://github.com/conda-forge/libblst-feedstock/issues/2

Also it appears we don't have `libblst` in the global pinnings. So raised issue: https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6623

Would be good to sort these out so we can ensure this is being pinned appropriately

step21: ﻿Should be resolved once https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6655 is merged.

jakirkham (approved):

step21:
Are we waiting for one more review now?

jakirkham:
Right there were a bunch of folks looking at this. So was waiting to see if anyone else still had more thoughts

step21:
Are we still waiting?

xhochy (approved):

step21:
thanks!

PR #26084, opened 2024-04-18
bump conda-forge-ci-setup version & conda-build lower bound

h-vetinari:
Discussed in core call today; the conda-build used here is constrained by the old ci-setup; update to what the rest of our infra is using as well.

First time I'm doing an upgrade of staged-recipes, please let me know if I missed something.

### DIFF
--- a/.ci_support/requirements.txt
+++ b/.ci_support/requirements.txt
@@ -1,8 +1,8 @@
 conda>=23.7.3
 conda-libmamba-solver>=23.7.0
-conda-build>=3.16,!=3.28.3
+conda-build>=24.3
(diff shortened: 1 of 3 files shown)


beckermr:
Isn't there a conda mambabuild command we need to change too? I thought we couldn't use boa anymore now?

h-vetinari:
> Isn't there a conda mambabuild command we need to change too? I thought we couldn't use boa anymore now?

I looked around, but all I could find was that `build_all.py` calls the conda-build API directly; there was no direct callsite to `conda mambabuild` or `conda-build` that I found.

PR #26087, opened 2024-04-18
[NOARCH] Add a recipe for helios-menuinst

dokempf:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/helios-menuinst/meta.yaml
+++ b/recipes/helios-menuinst/meta.yaml
@@ -16,6 +16,10 @@ requirements:
   run:
     - jupyterlab
 
+test:
+  commands:
+    - test -f $PREFIX/Menu/helios-menuinst.json
+
 about:
   home: https://github.com/3dgeo-heidelberg/helios
   summary: 'Menu items for Helios++'
--- a/recipes/helios-menuinst/meta.yaml
+++ b/recipes/helios-menuinst/meta.yaml
@@ -10,7 +10,7 @@ build:
   number: 0
   script:
     - mkdir -p {{ PREFIX }}/Menu
-    - cp {{ SRC_DIR }}/menu.json {{ PREFIX }}/Menu/helios.json
+    - cp {{ SRC_DIR }}/menu.json {{ PREFIX }}/Menu/helios-menuinst.json
 
 requirements:
   run:
--- a/recipes/helios-menuinst/meta.yaml
+++ b/recipes/helios-menuinst/meta.yaml
@@ -1,0 +1,32 @@
+package:
+  name: helios-menuinst
+  version: "0.1"
+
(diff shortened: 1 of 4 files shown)


dokempf:
@conda-forge/help-python This is ready for review. Sorry for hijacking the Python team here, there is no "Generic" team to ping for review.

Maybe a note on what this package does is in order: I am in the process of creating `constructor`-based installers for the package `helios`. For that I would like to add `menuinst` menus, but restrict them to `constructor` installers and ignore them in regular terminal installations. I consulted jaimergp on the constructor Matrix chat and he confirmed that putting the `menuinst` JSONs into a separate package is the way to go.

recipes/helios-menuinst/run_test.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+if ! test -f $PREFIX/Menu/helios.json; then
```
jaimergp, line 3: You can probably get by by adding this directly in meta.yaml as:

```
test -f $PREFIX/Menu/helios.json
```

recipes/helios-menuinst/meta.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+source:
+  path: src
+
+build:
+  noarch: generic
+  number: 0
+  script:
+    - mkdir -p {{ PREFIX }}/Menu
+    - cp {{ SRC_DIR }}/menu.json {{ PREFIX }}/Menu/helios.json
```
jaimergp, line 13: The target json file must match or start with the package name.

```suggestion
    - cp {{ SRC_DIR }}/menu.json {{ PREFIX }}/Menu/helios-menuinst.json
```

recipes/helios-menuinst/run_test.sh
dokempf: I did something like that and it executed with `sh` instead of `bash` and produced an error.

jaimergp: Hm, weird. That's standard posix 🤔 Can you try again so we can see the traceback?

jaimergp:
The important bit for menuinst is the conda-build validation (which doesn't error out if it fails, only a warning; to be addressed in the future).

In this case, the logs do say:

```
Validating Menu/*.json files
'Menu/helios.json' is a valid menuinst JSON document
```

So we are fine.

Note: given how the build script is Unix-y, Windows builds are failing, but given how simple the JSON document is, it should work ok.

dokempf:
@jaimergp It does indeed pass now with the tests in `meta.yaml`. I am not sure what exactly I had done wrong before (and force-pushed it away).

jaimergp (approved):

PR #26116, opened 2024-04-21
Add package to allow noarch python packages to depend on cupy

hmaarrfk:
At this stage, i'm curious how it gets through the build process.

xref: https://github.com/conda-forge/cupy-feedstock/issues/193

Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/cupyifcudaavailable/LICENSE
+++ b/recipes/cupyifcudaavailable/LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2014, conda-forge
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+* Redistributions in binary form must reproduce the above copyright notice,
+  this list of conditions and the following disclaimer in the documentation
+  and/or other materials provided with the distribution.
+
+* Neither the name of staged-recipes nor the names of its
+  contributors may be used to endorse or promote products derived from
+  this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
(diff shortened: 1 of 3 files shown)


recipes/cupyifcudaavailable/LICENSE
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2014, conda-forge
```
hmaarrfk, line 1:
```suggestion
Copyright (c) 2024, conda-forge
```

hmaarrfk:
Hmm well this fails if you use mamba quite catastrophically due to:
https://github.com/mamba-org/mamba/issues/3222

recipes/cupyifcudaavailable/meta.yaml
```diff
@@ -0,0 +1,30 @@
...
+  noarch: generic
+  string: cuda_{{ PKG_BUILDNUM }}  # [with_cuda]
+  string: cpu_{{ PKG_BUILDNUM }}   # [not with_cuda]
+
+requirements:    # [with_cuda]
+  run:           # [with_cuda]
+    - __cuda     # [with_cuda]
+    - cupy       # [wtih_cuda]
+
+# CIs don't really have cuda, and thus  would fail at resolving __cuda
```
hmaarrfk, line 18:
```suggestion
# CIs don't really have cuda, and thus would fail at resolving __cuda
```

recipes/cupyifcudaavailable/meta.yaml
```diff
@@ -0,0 +1,31 @@
...
+  number: 0            # [not with_cuda]
+  number: 300          # [    with_cuda]
+  noarch: generic
+  string: cuda_{{ PKG_BUILDNUM }}  # [with_cuda]
+  string: cpu_{{ PKG_BUILDNUM }}   # [not with_cuda]
+
+requirements:    # [with_cuda]
+  run:           # [with_cuda]
+    - __cuda     # [with_cuda]
+    - cupy       # [wtih_cuda]
```
hmaarrfk, line 16:
```suggestion
    - cupy       # [with_cuda]
```

hmaarrfk:
@jakirkham @leofang this is the package I have in mind to help me depend on cupy.

It allows downstream packages to just depend on "one thing" without creating `cpu/cuda` variants, or without creating `osx/linux/windows` variants.

leofang:
Thanks @hmaarrfk for your relentless push. I finally understand what's missing in your use case, sorry it took me a while. I suggest
1. We close this PR. This is simply not a scalable approach if everyone wants a GPU package to be optionally installable on a CPU only env (assuming the package supports it).
2. We move `cuda-version` [here](https://github.com/conda-forge/cupy-feedstock/blob/ee6580d553b9f8a1efb450cdde79de002c540c3b/recipe/meta.yaml#L191-L192) from `cupy-core`'s `run` to `run_constrained` (but leave `cupy` alone). This will ensure that `cuda-version` is used to distinguish CPU/GPU envs [as per its design](https://github.com/conda-forge/cuda-version-feedstock/blob/main/recipe/README.md), and also ensure we don't install extra CUDA deps in a CPU-only env (because `cupy-core` should really be used here).

hmaarrfk:
I agree that this is not a scalable approach. cupy is quite a unique package in that it targets a very specific hardware platform.

I'm happy to close this PR.

As i understand cupy does not target osx. so `cupy-core` would not be OK to have a dependency on noarch packages.

We can continue this discussion elsewhere more on the cupy feedstock.

leofang:
Let's continue the discussion in https://github.com/conda-forge/cupy-feedstock/pull/264 :)

PR #26148, opened 2024-04-26
Add NCEPLIBS-bufr

darothen:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/NCEPLIBS-bufr/meta.yaml
+++ b/recipes/NCEPLIBS-bufr/meta.yaml
@@ -20,6 +20,7 @@ requirements:
   build:
     - {{ compiler('fortran') }}
     - {{ compiler('c') }}
+    - {{ stdlib('c') }}
     - make  # [linux]
     - cmake
     - numpy
--- a/recipes/NCEPLIBS-bufr/meta.yaml
+++ b/recipes/NCEPLIBS-bufr/meta.yaml
@@ -14,7 +14,7 @@ build:
   skip: True  # [win]
   # NOTE: For simplicity, only support recent Python versions
   skip: True  # [py<=39]
-  number: 1
+  number: 0
 
 requirements:
   build:
@@ -47,4 +47,3 @@ about:
 extra:
   recipe-maintainers:
     - darothen
-    - jbathegit
--- a/recipes/NCEPLIBS-bufr/meta.yaml
+++ b/recipes/NCEPLIBS-bufr/meta.yaml
@@ -12,8 +12,11 @@ source:
 build:
   # NOTE: No Windows support in initial feedstock (known use cases are only osx and linux)
   skip: True  # [win]
-  # NOTE: For simplicity, only support recent Python versions
-  skip: True  # [py<=39]
+  # NOTE: For simplicity, only support recent Python versions - but not Python==3.12, 
+  # because the stable NCEPLIBS-bufr release hard-codes a Python setup that uses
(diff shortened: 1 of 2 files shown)


darothen:
@jbathegit I'm developing a tool which would greatly benefit from [NOAA-EMC/NCEPLIBS-bufr](https://github.com/NOAA-EMC/NCEPLIBS-bufr) being available on `conda-forge`, so I'm taking the liberty of creating a feedstock it. Since you're listed as the code manager on that repo, I thought it prudent that you were listed as a maintainer here.

Note that I need to work through tweaking the build to create a dynamic library before completing this (e.g. https://github.com/conda-forge/cfep/blob/main/cfep-18.md). So this isn't immediately ready to ping for review, but it would be great if you could comment here confirming that you can help as a maintainer.

darothen:
The current osx64 build fails quite reproducibly for a very specific reason, and I was able to debug it locally on my MacBook.

Take the test **intest7_4** as an example. In this test, `subroutine errwrt` is locally overridden within the actual test program, and `module Share_errstr_intest7` is used as a shim to allow the this subroutine to be called in the actual NCEPLIBS-bufr library routines. However, on platform osx64 something is going awry and the local override of the subroutine isn't being called, so we never populate the variable errstr (defined in `module Share_errstr_intest7`). Consequently, we fail the error check on L66 after we call `isetprm ( 'DUMMY', 20 )`; errstr is never set with the proper error message so the `index` function doesn't return 0, and we stop the program.

Not entirely sure how to fix this; it works totally fine on linux_64 (also tested locally, on a different machine). We _could_ just skip the handful of test cases that use this pattern and not even worry about the problem in the first place - all we care about is that the Python bindings work, and those tests pass just fine. I'm inclined to do that to avoid making any upstream changes to the NCEPLIBS-bufr code, unless @jbathegit has strong feelings either way.

darothen:
Following the discussion at NOAA-EMC/NCEPLIBS-bufr#589, I'm making the following changes:

1. Removing @jbathegit as a proposed maintainer - @darothen will be the sole maintainer.
2. Modifying tests to suppress the handful that fail on osx64 due to the dynamic library patching issue as documented previously.

recipes/NCEPLIBS-bufr/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+
+source:
+  url: https://github.com/NOAA-EMC/NCEPLIBS-bufr/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 525f26238dba6511a453fc71cecc05f59e4800a603de2abbbbfb8cbb5adf5708
+
+build:
+  # NOTE: No Windows support in initial feedstock (known use cases are only osx and linux)
+  skip: True  # [win]
+  # NOTE: Only support recent Python releases. Cannot support Python==3.12 because build
+  # system in NCEPLIBS-bufr_v12.0.1 uses a hard-coded dependency on distutils, which was
```
ocefpaf, line 16: Fortran extensions are in a pickle with py312 :-/
They need to move to cmake or meson and that is not an easy change. Let me know if you figure out a path with least friction to solve this!

darothen: The good news is that this particular package does have an improvement to its cmake-based build already merged into trunk, so the next time they cut a release we'll be able to support Py312. But yeah, wish there was a generic way (maybe a back-port of distutils?)

PR #26316, opened 2024-05-10
Add recipe for libmetatensor-torch

Luthaf:
This is the continuation of https://github.com/conda-forge/metatensor-feedstock/issues/2, now packaging the second native library. This one is pure C++, and links to libtorch.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/libmetatensor-torch/meta.yaml
+++ b/recipes/libmetatensor-torch/meta.yaml
@@ -1,0 +1,51 @@
+{% set version = "0.5.0" %}
+{% set url_base = "https://github.com/lab-cosmo/metatensor/releases/download" %}
+
+{% set cmake_args = "$CMAKE_ARGS" %}  # [not win]
+{% set cmake_args = "%CMAKE_ARGS%" %}  # [win]
+
+package:
+  name: libmetatensor-torch
+  version: {{ version }}
+
+source:
+  url: {{ url_base }}/metatensor-torch-v{{ version }}/metatensor-torch-cxx-{{ version }}.tar.gz
+  sha256: 904cf858d8f98b67b948e8a453d8a6da56111e022050d6c8c3d32a9a2cc83464
+
+build:
+  number: 0
+  script: |
+    cmake -G Ninja {{ cmake_args }} -DBUILD_SHARED_LIBS=ON  .
+    cmake --build . --config Release --target install
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - cmake
+    - ninja
+  host:
+    - libmetatensor >=0.1.0,<0.2
+    - libtorch
+  run:
+    - libmetatensor >=0.1.0,<0.2
+    - libtorch
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libmetatensor_torch$SHLIB_EXT  # [unix]
+    - if not exist %PREFIX%\\Library\\bin\\metatensor_torch.dll exit 1  # [win]
+
+about:
+  home: https://github.com/lab-cosmo/metatensor
+  summary: TorchScript/C++ bindings to metatensor
+  license: BSD-3-Clause
+  license_family: BSD
+  license_file:
+    - LICENSE
+    - _deps/nlohmann_json-src/LICENSE.MIT
+  doc_url: https://lab-cosmo.github.io/metatensor/
+
+extra:
+  recipe-maintainers:
+    - Luthaf
+    - PicoCentauri
--- a/recipes/libmetatensor-torch/meta.yaml
+++ b/recipes/libmetatensor-torch/meta.yaml
@@ -28,11 +28,11 @@ requirements:
     - ninja
   host:
     - libmetatensor >=0.1.8,<0.2
-    # we only need the CPU version of libtorch to build the code, but we can run
-    # with either the CPU or CUDA version
+    # We build against the CPU version of torch to not have to deal with CUDA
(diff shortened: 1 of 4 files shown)


Luthaf:
Just to check, the version of libtorch is pinned across the ecosystem, right? And any bump in the libtorch version will trigger a rebuild of all feedstocks depending on it?

Luthaf:
@conda-forge/help-c-cpp ready for review!

recipes/libmetatensor-torch/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+test:
+  requires:
+    - cmake
+    - ninja
+    - {{ compiler('cxx') }}
+    - libtorch  =*=cpu*
+  files:
+    - test-project/
+  commands:
```
xhochy, line 43: Please add a simple file existence test for checking that the shared library is part of the package and at the correct path.

recipes/libmetatensor-torch/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+package:
+  name: libmetatensor-torch
+  version: {{ version }}
+
+source:
+  url: {{ url_base }}/metatensor-torch-v{{ version }}/metatensor-torch-cxx-{{ version }}.tar.gz
+  sha256: 904cf858d8f98b67b948e8a453d8a6da56111e022050d6c8c3d32a9a2cc83464
+
+build:
```
xhochy, line 12: Shared libraries also need a run_exports.

xhochy (changes requested):

PicoCentauri:
Thanks for the work @Luthaf. Of course I am willing to help maintaining this recipe.

recipes/libmetatensor-torch/meta.yaml
Luthaf: It's not very clear what run_exports does. Does it only add a `run` dependency to all packages that built against the library and had it as a `host` dependency? Or is there something else I need to be aware of here?

carterbox: https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#export-runtime-requirements

run_exports export runtime requirements to downstream packages which dynamically link to your package. It's a statement which says "hey, if you link to my package, then this is the version that you will need at runtime."

Luthaf: Thanks! I've added `run_exports` to [`libmetatensor`](https://github.com/conda-forge/libmetatensor-feedstock) and removed it from the `run` requirements here. 

I'd guess `libtorch` also has a run_exports, but is there a way to express "build against the cpu variant but run with any variant, cpu or cuda"? For now I do two separate host & run requirements.

Luthaf:
@conda-forge/help-c-cpp ready for another round of review!

recipes/libmetatensor-torch/meta.yaml
carterbox: > I'd guess libtorch also has a run_exports

[Yes.
](https://github.com/conda-forge/pytorch-cpu-feedstock/blob/2953d76cb7cfa5fd5b8049c1df9baf0758bab4af/recipe/meta.yaml#L43)

Looks like the run_exports for `libtorch` are variant agnostic, so if you build against any variant of `libtorch` you can get any variant at runtime. It looks like this package doesn't link against `libtorch_cuda`, so you don't need to worry about contraining the `libtorch` build variant.

recipes/libmetatensor-torch/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - {{ stdlib("c") }}
+    - cmake
+    - ninja
+  host:
+    - libmetatensor >=0.1.8,<0.2
+    # we only need the CPU version of libtorch to build the code, but we can run
+    # with either the CPU or CUDA version
+    - libtorch  =*=cpu*
+  run:
+    - libtorch
```
carterbox, line 35:
```suggestion
    # We don't link to any gpu libraries
    - libtorch
```

carterbox:
The logs show that the build system is also looking for zlib? Is that a missing host dependency?

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=935468&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=808

Luthaf:
> The logs show that the build system is also looking for zlib? Is that a missing host dependency?

We are not trying to load ZLIB ourself. Digging around the CMake files, this is what seems to happen:

We load torch with `find_package(Torch)`, which then does `find_package(Caffe2)`

https://github.com/pytorch/pytorch/blob/6c503f1dbbf9ef1bf99f19f0048c287f419df600/cmake/TorchConfig.cmake.in#L68

Which then does `find_package(Protobuf)`: 

https://github.com/pytorch/pytorch/blob/6c503f1dbbf9ef1bf99f19f0048c287f419df600/cmake/public/protobuf.cmake#L4

which finally tries to find ZLIB:

https://github.com/protocolbuffers/protobuf/blob/2f3242c576504459621fd7d78bbccf39bfaa49c5/cmake/protobuf-config.cmake.in#L5
https://github.com/protocolbuffers/protobuf/blob/2f3242c576504459621fd7d78bbccf39bfaa49c5/CMakeLists.txt#L172

PR #26342, opened 2024-05-13
add ml4xcube package

Julia310:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/ml4xcube/meta.yaml
+++ b/recipes/ml4xcube/meta.yaml
@@ -1,0 +1,56 @@
+{% set name = "ml4xcube" %}
+{% set version = "0.0.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/ml4xcube-{{ version }}.tar.gz
+  sha256: 5629a87459aeb6d3aa4df4689fc1352561a66ebed8aee599836cd7d0f5549ec5
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - setuptools >=42
+    - wheel
+    - pip
+  run:
+    - python >=3.8
+    - conda-forge::xcube
+    - dask-core >=2023.2.0
+    - numpy >=1.24
+    - pandas >=2.2
+    - scikit-learn >1.3.1
+    - xarray >2023.8.0
+    - zarr >2.11
+    - rechunker >=0.5.1
+    - sentinelhub
+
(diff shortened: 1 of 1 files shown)


Julia310:

recipes/ml4xcube/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - pip
+  run:
+    - python >=3.8
+    - conda-forge::xcube
```
synapticarbors, line 23: I'm not familiar with the `::` in a package name preceded by the the channel (?). I think this could just be `xcube`. Could you clarify why the dependency is listed the way that it is?

Julia310: Hi @synapticarbors 

Thank you for reviewing my Python package!

I wanted to install `xcube` specifically from the conda-forge channel because the latest versions of this package are only available there. 

As mentioned in [this GitHub issue](https://github.com/conda/conda/issues/7202), the syntax `channel::package` can be used in an `environment.yml` file to force the installation from a specific channel. I assumed this would work in the `meta.yaml` as well, so I used this approach to ensure that the package is installed from conda-forge.

Since I am not shure if this is the right approach for `meta.yaml` I simply added a package version to the definition.

synapticarbors: On conda-forge we only pull from the conda-forge channel for requirements, and for end-users it's probably better to specify the correct version rather than trying to force a particular channel.

Julia310: Hi @synapticarbors,

Thank you for merging my code. I have been trying to find the ml4xcube package in the available conda-forge packages, but it seems it is not yet available for download. Could you please let me know when it will be possible to download it?

As I am new to this procedure, I would appreciate your feedback in this regard.

Best,
Julia

Julia310: In the commit history I saw that the receipe was removed. Do I need to make another PR?

grafik

synapticarbors: The feedstock was created here https://github.com/conda-forge/ml4xcube-feedstock but it didn't actual build the packages for some reason. Let me look into it.

Julia310: Hi @synapticarbors,

I hope you're doing well. I wanted to follow up on the status of the ml4xcube package. Is there any update on why the package build didn't succeed. How could I proceed from here?

synapticarbors: See https://github.com/conda-forge/ml4xcube-feedstock/pull/2#issuecomment-2118389200 -- one way to proceed would be to remove the `pip check` there which I can do

PR #26343, opened 2024-05-13
add libtinyspline

mantkiew:
Checklist
- [X] Build number is 0.


mantkiew:

Questions:

1. there's [tinyspline](https://pypi.org/project/tinyspline/) on PyPi that can be added separately.
2. this one is C/C++ only. I'm confused about the naming conventions. Should there be `tinyspline` just for Python and then `tinysplinec` and `tinysplinecxx` for C and C++, respectively?
3. I tried using `ctest` to run the tests but the command is not available in the build environment. I have the `cmake` in the requirements and that should bring in the `ctest` command as well.

xhochy (changes requested):
You should add a lib prefix to distinguish it from the python package.

Also, as this ships shared libraries, it should have a run_exports statement.

mantkiew:
@xhochy is that ok?

Finally, can we use `ctest` to execute tests? That would be great but so far my builds were failing because `ctest` command is not available. Looks like it's excluded from the `cmake` installation in the build image.

recipes/libtinyspline/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - strip --strip-unneeded lib/libtinyspline*  # [linux]
+    - cmake --build . --target install
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('libtinyspline', max_pin='x') }}
+
+requirements:
+  build:
+    - cmake
+    - binutils  # [linux]
```
xhochy, line 28: This should not be required.
```suggestion
```

recipes/libtinyspline/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+source:
+  url: https://github.com/msteinbeck/tinyspline/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 3ea31b610dd279266f26fd7ad5b5fca7a20c0bbe05c7c32430ed6aa54d57097a
+
+build:
+  script: 
+    - mkdir -p build
+    - cd build
+    - cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DTINYSPLINE_BUILD_EXAMPLES=OFF -DTINYSPLINE_BUILD_DOCS=OFF -DTINYSPLINE_INSTALL_LIBRARY_DIR=lib -DCMAKE_INSTALL_PREFIX=${PREFIX}  # [unix]
+    - cmake .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DTINYSPLINE_BUILD_EXAMPLES=OFF -DTINYSPLINE_BUILD_DOCS=OFF -DTINYSPLINE_INSTALL_LIBRARY_DIR=lib -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX%  # [win]
```
xhochy, line 17:
```suggestion
    - cmake ${CMAKE_ARGS} .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DTINYSPLINE_BUILD_EXAMPLES=OFF -DTINYSPLINE_BUILD_DOCS=OFF -DTINYSPLINE_INSTALL_LIBRARY_DIR=lib -DCMAKE_INSTALL_PREFIX=${PREFIX}  # [unix]
    - cmake %CMAKE_ARGS% .. -G Ninja -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON -DTINYSPLINE_BUILD_EXAMPLES=OFF -DTINYSPLINE_BUILD_DOCS=OFF -DTINYSPLINE_INSTALL_LIBRARY_DIR=lib -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX%  # [win]
```

xhochy (changes requested):

xhochy:
> Finally, can we use `ctest` to execute tests? That would be great but so far my builds were failing because `ctest` command is not available. Looks like it's excluded from the `cmake` installation in the build image.

This should work 🤔 I have been doing this in other locations. You need to run these tests though during the build phase of conda-build.

recipes/libtinyspline/meta.yaml
mantkiew: I added `binutils` to have the  command `strip`. Otherwise, the build was failing with "command not found". Also, stripping only works on Linux (and all binaries should be stripped).

mantkiew:
> > Finally, can we use `ctest` to execute tests? That would be great but so far my builds were failing because `ctest` command is not available. Looks like it's excluded from the `cmake` installation in the build image.
> 
> This should work 🤔 I have been doing this in other locations. You need to run these tests though during the build phase of conda-build.

Ah, that's the problem! I was running `ctest` in the `test` section not in the `build`. Also, I couldn't find any mention of `ctest` in the documentation at all.

recipes/libtinyspline/meta.yaml
xhochy: There should be a `$STRIP` command set that comes in via the compiler. Please use that one.

mantkiew: In the end, I switched to using `cmake --install . --strip` instead of manually using the command `$STRIP` because cmake knows how to do it across all platforms.

xhochy (approved):

PR #26357, opened 2024-05-14
Add recipe for `mrmr-selection`

svengiegerich:
Checklist
- [x] Build number is 0.


recipes/mrmr_selection/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set name = "mrmr_selection" %}
+{% set version = "0.2.8" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  # no sdist provided, no tags on github
+  # see issue: https://github.com/smazzanti/mrmr/issues/45
```
ocefpaf, line 10: If you believe that the wheel was created after this commit,

https://github.com/smazzanti/mrmr/commit/3ea01aac88628b678aa88f734524594b0edd63d7

we could use that instead. At leas it would give us a traceable source for it.

PS: Philosophical questions.  If a package doesn't publish an sdist or tag to a released wheel, is it really open?

recipes/mrmr_selection/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  url: https://files.pythonhosted.org/packages/a9/f4/b09243bbcd133952f393b54924553c4a21ab2f4e4a7a083050f25afe5fcb/{{ name }}-{{ version }}-py3-none-any.whl
```
ocefpaf, line 11:
```suggestion
  url: https://github.com/smazzanti/mrmr/archive/{{ commit }}.tar.gz
```

recipes/mrmr_selection/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set name = "mrmr_selection" %}
+{% set version = "0.2.8" %}
```
ocefpaf, line 2:
```suggestion
{% set version = "0.2.8" %}
{% set commit = "3ea01aac88628b678aa88f734524594b0edd63d7" %}
```

recipes/mrmr_selection/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+source:
+  # no sdist provided, no tags on github
+  # see issue: https://github.com/smazzanti/mrmr/issues/45
+  url: https://files.pythonhosted.org/packages/a9/f4/b09243bbcd133952f393b54924553c4a21ab2f4e4a7a083050f25afe5fcb/{{ name }}-{{ version }}-py3-none-any.whl
+  sha256: 70723070d516802397e467075366dc3be07117de6378deee6821d5aba2686aa4
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install {{ name }}-{{ version }}-py3-none-any.whl -vv --no-deps
```
ocefpaf, line 17:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

ocefpaf (commented):
> How I would use it exactly? Do you mean mentioning it as a comment, or smth else?

@svengiegerich something like this should work.

recipes/mrmr_selection/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  # no sdist provided, no tags on github
+  # see issue: https://github.com/smazzanti/mrmr/issues/45
+  url: https://github.com/smazzanti/mrmr/archive/{{ commit }}.tar.gz
+  sha256: 70723070d516802397e467075366dc3be07117de6378deee6821d5aba2686aa4
```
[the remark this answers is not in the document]
ocefpaf: Yep. You should update that. You can use the one in the error message.

recipes/mrmr_selection/meta.yaml
ocefpaf: > do you know what I need to adjust? 'jinja2' is already listed in the requires.

Yes, you need all the dependencies listed here: https://github.com/smazzanti/mrmr/blob/dc8a810bc9d01b6fe20e8767de0098d4671b50b6/setup.py#L19-L26

ocefpaf: Give me a sec to check it locally...

ocefpaf: OK. It looks like the project imports itself at build time and need to be importable before being built. That is an old practice but buggy. If you can, please open an issue upstream. Our workaround is to specify all the run dependencies at build time too. I did that here for you. Let's see if that works.

PR #26370, opened 2024-05-15
Add python-cudnn-frontend

weiji14:
The cuDNN FrontEnd(FE) API is a C++ header-only library that wraps the cuDNN C backend API. Repo at https://github.com/NVIDIA/cudnn-frontend

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/cudnn-frontend/meta.yaml
+++ b/recipes/cudnn-frontend/meta.yaml
@@ -22,6 +22,7 @@ requirements:
     - {{ compiler('c') }}
     - {{ compiler('cxx') }}
     - {{ compiler('cuda') }}
+    - {{ stdlib("c") }}
     - cmake >=3.17
     - cudnn
     - make
@@ -29,10 +30,11 @@ requirements:
   host:
     - pip
     - pybind11
-    - python >=3.7
+    - python
     - setuptools >=64
   run:
     - cudnn
+    - python
 
 test:
   imports:
@@ -43,6 +45,7 @@ test:
     - pip
 
 about:
+  home: https://github.com/NVIDIA/cudnn-frontend
   summary: cuDNN FrontEnd API
   license: MIT
   license_file:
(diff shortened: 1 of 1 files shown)


weiji14:

xhochy:
If this is a header-only library, why does it need Python?

weiji14:
This is built with the Python bindings (https://github.com/NVIDIA/cudnn-frontend?tab=readme-ov-file#python-api). I could try to remove `python` from the host and runtime deps and see if it works?

xhochy:
It depends on what you want to achieve with this package. If it should be one for the headers, you should remove python. If you want to have the Python bindings, then this looks good.

weiji14:
Yes, I'll be using this with the python bindings. Should I add a `-python` suffix perhaps?

xhochy:
> Yes, I'll be using this with the python bindings. Should I add a `-python` suffix perhaps?

For these cases, we normally add a python- prefix.

recipes/cudnn-frontend/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  skip: true           # [not linux]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - {{ stdlib("c") }}
+    - cmake >=3.17
+    - cudnn
```
xhochy, line 27: This should be in host and then you probably don't need to explicitly mention it in run anymore.

weiji14: Ok, done! Thanks again for checking :smiley:

weiji14:
> In your conda_build_config.yaml, please change the name of `MACOSX_DEPLOYMENT_TARGET`, to `c_stdlib_version`!

Guessing that this is a false alarm since there's no `conda_build_config.yaml` file added? Can probably just ignore also since this recipe won't be built on macOS.

xhochy (approved):

PR #26371, opened 2024-05-15
Adding conda-recipe-manager as a conda-forge feedstock

schuylermartin45:
- This will allow us to start publishing versions that others can consume
- The most immediate need is for an [ask in Grayskull](https://github.com/conda/grayskull/issues/486)

Checklist
- [ x] If static libraries are linked in, the license of the static library is packaged.
- [ x] Build number is 0.


schuylermartin45:
@conda-forge/help-python I believe this is ready for review. Thanks!

recipes/conda-recipe-manager/LICENSE
```diff
@@ -0,0 +1,32 @@
+BSD 3-Clause License
```
synapticarbors, line 1: It looks like this file is already in the source distribution you're using so you shouldn't include an additional copy of it here.

synapticarbors (changes requested):

recipes/conda-recipe-manager/LICENSE
schuylermartin45: Ok I'm confused then because every piece of documentation I saw says to include it with the recipe file. I thought that was standard practice.

https://github.com/conda-forge/staged-recipes/blob/a504af81c05491bf7b0b018b2fa1efe64767985c/recipes/example/meta.yaml#L52-L55
```yaml
  # It is strongly encouraged to include a license file in the package,
  # (even if the license doesn't require it) using the license_file entry.
  # See http://conda.pydata.org/docs/building/meta-yaml.html#license-file
  license_file: LICENSE.txt
```
  
https://conda-forge.org/docs/maintainer/adding_pkgs/#the-staging-process
> With the help of [the staging process](https://conda-forge.org/docs/maintainer/adding_pkgs/#creating-recipes), add a package's recipe and license to the [staged-recipes repository](https://github.com/conda-forge/staged-recipes) and create a PR.

https://conda-forge.org/docs/maintainer/adding_pkgs/#checklist
> Ensure that you have included a license file if your license requires one – most do. (see [here](https://github.com/conda-forge/staged-recipes/blob/a504af81c05491bf7b0b018b2fa1efe64767985c/recipes/example/meta.yaml#L52-L55))

BastianZim: Just to clarify, you are correct that this means that it should be included. Most packages package the license file in the source distribution, the file then just needs to list it in the meta.yaml and it is included. You only need to manually include the license file if it isn't included.
So the difference comes from including the actual file (which is often not necessary) and including the path to the file (which is always necessary).

schuylermartin45: Thanks for the clarification! Maybe the docs or linter could be improved to reflect that nuance?

Either way, I just removed the file in the latest commit.

schuylermartin45:
I'm not sure how to get the bot to change the label, but I, as the author, made a contribution last week and I don't think the label is accurate anymore.

recipes/conda-recipe-manager/LICENSE
BastianZim: Yes, I think there is an issue for that, we just haven't gotten around to it... 😄

PR #26449, opened 2024-05-24
Correctly set the MACOSX_DEPLOYMENT_TARGET

xhochy:
We need to set `MACOSX_DEPLOYMENT_TARGET` so that `conda-forge-ci-setup` set the correct value if the recipe doesn't set anything. This value is parsed before the global pinning is taken into account.

This also removes the top-level `conda_build_config.yaml` which should never been placed there.


Tobias-Fischer:
Hi @xhochy - is this one good to go?

xhochy:
It is probably not needed anymore.

PR #26544, opened 2024-06-03
First recipe for eo_tools

odhondt:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/eo_tools/meta.yaml
+++ b/recipes/eo_tools/meta.yaml
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/odhondt/eo_tools/archive/{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


recipes/eo_tools/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python
+    - setuptools
+    - wheel
+    - pip
+  run:
+    - python ==3.12.3
```
synapticarbors, line 24: I would highly advise against such a strict pinning of `python`, unless this is meant to be a stand-alone application and not a library that can be used with other dependencies. I can see that upstream the dependencies are tightly pinned, but I'm guessing that `python` itself could just specify something like `python >=3.12,<3.13`.

odhondt: Thanks for the advice! I have made the change.

PR #26556, opened 2024-06-04
Add burstH2MM

harripd:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/bursth2mm/meta.yaml
+++ b/recipes/bursth2mm/meta.yaml
@@ -25,7 +25,7 @@ requirements:
   run:
     - python >=3.8
     - numpy
-    - matplotlib-base
+    - matplotlib-base >=3.0.1
     - seaborn >=0.11.1
     - fretbursts
     - h2mm-c >=1.0.3
--- a/recipes/bursth2mm/meta.yaml
+++ b/recipes/bursth2mm/meta.yaml
@@ -19,12 +19,12 @@ build:
 
 requirements:
   host:
-    - python
+    - python >=3.8
     - setuptools >=64
     - setuptools-scm >=8.0
     - pip
   run:
-    - python
+    - python >=3.8
     - numpy
     - matplotlib-base
     - seaborn >=0.11.1
--- a/recipes/bursth2mm/meta.yaml
+++ b/recipes/bursth2mm/meta.yaml
@@ -13,7 +13,7 @@ build:
   noarch: python
   script_env:
     - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
-  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
 
 requirements:
--- a/recipes/bursth2mm/meta.yaml
+++ b/recipes/bursth2mm/meta.yaml
@@ -1,0 +1,60 @@
+{% set name = "burstH2MM" %}
+{% set version = "0.1.8" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
+  url: https://github.com/harripd/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: d06654d8e37a1d30ce39c97f861bcc1542b25333fecf43ae5c1a07a2e1892a8f
+
+build:
+  noarch: python
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  number: 0
+
+requirements:
+  host:
+    - python
+    - setuptools >=64
+    - setuptools-scm >=8.0
+    - pip
+  run:
+    - python
+    - numpy
+    - matplotlib-base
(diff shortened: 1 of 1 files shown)


harripd:
I am willing to be a maintainer.

harripd:
@conda-forge/help-python ready for review/

recipes/bursth2mm/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+requirements:
+  host:
+    - python >=3.8
+    - setuptools >=64
+    - setuptools-scm >=8.0
+    - pip
+  run:
+    - python >=3.8
+    - numpy
+    - matplotlib-base
```
ocefpaf, line 29:
```suggestion
    - matplotlib-base >=3.0.1
```

recipes/bursth2mm/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - setuptools >=64
+    - setuptools-scm >=8.0
+    - pip
```
ocefpaf, line 25: While I could not find any numpy compiled extension upstream, it does declare a build dependency on numpy in https://github.com/harripd/burstH2MM/blob/670a88b4ae033fa6c2a4bc499be003f5ca62e026/pyproject.toml#L6

Maybe that is a mistake b/c the recipe seems to work fine as noarch here.

recipes/bursth2mm/meta.yaml
```diff
@@ -0,0 +1,60 @@
+{% set name = "burstH2MM" %}
+{% set version = "0.1.8" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  fn: {{ name }}-{{ version }}.tar.gz
```
ocefpaf, line 9:
```suggestion
```

recipes/bursth2mm/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  url: https://github.com/harripd/{{ name }}/archive/{{ version }}.tar.gz
+  sha256: d06654d8e37a1d30ce39c97f861bcc1542b25333fecf43ae5c1a07a2e1892a8f
+
+build:
+  noarch: python
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script: "{{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir -vvv"
```
ocefpaf, line 17:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

recipes/bursth2mm/meta.yaml
harripd: Both FRETBursts and H2MM_C are dependent on numpy, in pip it uses oldest-supported-numpy, which is then replaced with pin-compatible-numpy in their conda-forge recipes

recipes/bursth2mm/meta.yaml
harripd: Fine with me

recipes/bursth2mm/meta.yaml
harripd: Fine with me

recipes/bursth2mm/meta.yaml
harripd: As long as it still builds

recipes/bursth2mm/meta.yaml
ocefpaf: > Both FRETBursts and H2MM_C are dependent on numpy

To run or build? B/c you don't have it here in host, which is used to build and, if that is true, it should fail as noarch, but it didn't. oldest-supported-numpy/pin-compatible-numpy are used when the package has a compiled extension against numpy, which doesn't seem to be the case, even though it is declared upstream.

recipes/bursth2mm/meta.yaml
ocefpaf: @harripd, I'm a conda-forge core member and staged-recipes reviewer. Why do you think I would send you something that would not work?

recipes/bursth2mm/meta.yaml
harripd: Not sure why. I included numpy in the build requirements long ago, just assuming it was needed. 

I checked on my local machine, I can remove numpy from the build requirements and it makes no difference. So it does not actually require numpy to build, just to run.

Not sure why it doesn't give an error though given that numpy is stated in the build requirements of the current repo...

recipes/bursth2mm/meta.yaml
harripd: I didn't see any indication that it couldn't build, I wasn't saying I did trust you, sorry if I implied that.

recipes/bursth2mm/meta.yaml
harripd: Would you like me to make an update (also on pip where I remove the numpy *build* dependency? Just let me know and I'll make the appropriate updates (new version, update the version number and sha256 sum or add pin-compatible-numpy to build) so I can resolve the converstaion

ocefpaf: This is not a problem for this package now, but auto-updating tools will be confused. It is a bug upstream b/c it declares something that it doesn't need.

PR #26581, opened 2024-06-06
add taskgroup 0.0.0a4

bollwyvl:
Checklist
- [x] Build number is 0.

References:
- blocks https://github.com/conda-forge/hypercorn-feedstock/pull/44

### DIFF
--- a/recipes/taskgroup/meta.yaml
+++ b/recipes/taskgroup/meta.yaml
@@ -1,0 +1,41 @@
+{% set version = "0.0.0a4" %}
+
+package:
+  name: taskgroup
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/t/taskgroup/taskgroup-{{ version }}.tar.gz
+  sha256: eb08902d221e27661950f2a0320ddf3f939f579279996f81fe30779bca3a159c
+
+build:
+  number: 0
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
+  host:
+    - flit-core >=3.2,<4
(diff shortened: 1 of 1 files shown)


bollwyvl:

recipes/taskgroup/meta.yaml
```diff
@@ -0,0 +1,41 @@
+{% set version = "0.0.0a4" %}
```
ocefpaf, line 1: This one falls under https://github.com/conda-forge/cfep/blob/main/cfep-05.md

bollwyvl: Hm, [no stable release](https://pypi.org/project/taskgroup/#history) exists, and this is [depended upon](https://github.com/conda-forge/hypercorn-feedstock/pull/44) by an existing feedstock.

ocefpaf: You can place this one is a dev channel and still use it there by adding an extra channel in the conda-forge yaml file. However, that is too cumbersome and, IMO, people should just stop with the dev releases that should not have been a dev release to start with. PyPI set a bad precedent with that "pre" policy IMO.

bollwyvl: I heartily agree, and really `black` did the worst damage, leaving a lot of people with `--pre` on in various setups. Indeed, having a second `dev` registry, and the _ability_ to specify it per dep would have likely have been more useful, as quietly allowing a `==` pin a pre-release package feels super bad.

ocefpaf: And black did not need that BTW! It was the only pre release that could be installed as a stable one. Go figure why...

PR #26651, opened 2024-06-13
PyGnome initial recipe.

JamesMakela-NOAA:
Checklist
- [x] Build number is 0.


ChrisBarker-NOAA:
I agree to be a maintainer

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/NOAA-ORR-ERD/PyGnome/archive/v{{ version }}.tar.gz
+  sha256: f4263ea7e373414ea66ca0f630b705b07e0131258b2fafad03d369b0f5ef0ee6
+
+build:
+  skip: true  # [linux]
```
ChrisBarker-NOAA, line 13: why is linux skipped?

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+requirements:
+  build:
+    - cython
+    - numpy
+    - netCDF4
+    - scikit-build-core
+    - clang 15.*
+    - cmake
+  host:
+    - python >=3.7
```
ChrisBarker-NOAA, line 26: let's do >= 3.9

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+  run:
+    - netCDF4
+    - openssl
+    - libzlib
+    - bzip2
+
+test:
+  imports:
+    - gnome
+  requires:
```
ChrisBarker-NOAA, line 39: I don't think you need all the requirements here -- just the ones that are extra for the tests.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - pyogrio
+    - pyproj
+    - shapely
+    - gridded
+    - pynucos
+    - cell_tree2d
+    - py_gd
+    - adios_db
+  source_files:
+    - py_gnome/tests/unit_tests
```
ChrisBarker-NOAA, line 60: we don't need that if we're only doing an import test.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - netCDF4
+    - scikit-build-core
+    - clang 15.*
+    - cmake
+  host:
+    - python >=3.7
+    - pip
+    - numpy
+    - scikit-build-core
+  run:
```
ChrisBarker-NOAA, line 30: this is where the full list of requirements should go -- not in the tests stanza.

ChrisBarker-NOAA (changes requested):
getting there -- putting netcdf et al in host may solve the build issue

recipes/pygnome/meta.yaml
JamesMakela-NOAA: It is skipped because we didn't verify it on linux.  Don't have an environment for it.  Although our Gitlab runners are on CentOS.  I will enable it just to see if it works.

recipes/pygnome/meta.yaml
JamesMakela-NOAA: will do...

recipes/pygnome/meta.yaml
JamesMakela-NOAA: I was failing the import test without the presence of those python packages.

recipes/pygnome/meta.yaml
JamesMakela-NOAA: Ok, I was testing if we could run the pytests, and I left that in there.  I will remove it.

recipes/pygnome/meta.yaml
JamesMakela-NOAA: So they should go in the requirements/run?

ChrisBarker-NOAA: Exactly -- `requirements/run` is what is need to, well, run the package -- that's what will get installed by conda when you conda install this package.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+
+requirements:
+  build:
+    - cython
+    - numpy
+    - netCDF4
+    - scikit-build-core
+    - clang 15.*
+    - cmake
+  host:
```
ChrisBarker-NOAA, line 37: and I think this is the problem -- all those libs it can't find should be in the host requirements, so they'll be there.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/NOAA-ORR-ERD/PyGnome/archive/v{{ version }}.tar.gz
+  sha256: f4263ea7e373414ea66ca0f630b705b07e0131258b2fafad03d369b0f5ef0ee6
+
+build:
+  skip: True  # [py<38]
+  missing_dso_whitelist:
```
ChrisBarker-NOAA, line 14: There's got to be something wrong here -- missing_dso_whitelist  should be only rarely needed in special circumstances !?!?!

JamesMakela-NOAA: Yeah well, this entire configuration file was working just fine locally on my mac.  Warnings?  Yeah a few, but nothing that should have broken the entire thing.

Whatever is causing this has to do with the Azure instances they are running this on.  Tough situation huh?  Something that doesn't even fail locally, but fails catastrophically in the GitHub pipeline.

ChrisBarker-NOAA: I suspect it's about the cross-compiling infrastructure -- we're not doing that at all locally, and even if it's not being done here, the system is in place.

Which means we need to figure out "build" vs "host".

ChrisBarker-NOAA: you can ping the right group on the PR for help (can't find the list right now) -- and/or ask in the conda community forum:

https://conda.discourse.group/

JamesMakela-NOAA: > I suspect it's about the cross-compiling infrastructure -- we're not doing that at all locally, and even if it's not being done here, the system is in place.
> 
> Which means we need to figure out "build" vs "host".

I am sure that is an important distinction.  However not even the osx64 target is building.  I would like us to first and foremost get that target working since it is what we are familiar with.

recipes/pygnome/meta.yaml
```diff
@@ -18,13 +18,14 @@ requirements:
...
     - cross-python_{{ target_platform }}  # [build_platform != target_platform]
-    - numpy                               # [build_platform != target_platform]
+    - numpy
     - cython
     - netCDF4
     - scikit-build-core
     - clang 15.*             # [osx]
     - {{ compiler('c') }}    # [not osx]
     - {{ compiler('cxx') }}  # [not osx]
+    - make
```
ChrisBarker-NOAA, line 28: Should this be [now win] ?

JamesMakela-NOAA: Maybe.  I don't think it is bothering the windows build step though, so I was keeping it simple.

recipes/pygnome/LICENSE.txt
MementoRC, line 1: This file may not be needed as it should be in your source tarball

JamesMakela-NOAA: That is correct, but how do I get the linter to stop complaining about it?

MementoRC: Oh, that is odd, it should not complain. A more knowledgeable reviewer should be able to help you further (I am just another recipe maintainer who worked with sckit-build-core)

JamesMakela-NOAA: Ok, well it isn't hurting anything so I will leave it in to satisfy the linter.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+  number: 0
+  script: cd py_gnome && {{ PYTHON }} -m pip install ./ -vv
+
+requirements:
+  build:
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+    - numpy
+    - cython
+    - netCDF4
```
xhochy, line 26: These should also be in `host`. Maybe you need them here for the `build_platform != target_platform` but really only for that.
```suggestion
    - numpy
    - cython
    - netCDF4
```

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - scikit-build-core
+    - clang 15.*             # [osx]
+    - {{ compiler('c') }}    # [not osx]
+    - {{ stdlib('c') }}      # [not osx]
+    - {{ compiler('cxx') }}  # [not osx]
```
xhochy, line 31: Set `c(xx)_compiler_version` in `conda_build_config` instead. Otherwise, you won't get the right `C(XX)FLAGS) environment variables.
```suggestion
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
    - {{ compiler('cxx') }}
```

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - libaec
+    - libxml2
+    - libzip
+    - openssl
+    - zlib
+    - zstd
+  run:
+    # Even though we have python defined in reqs->host, the windows build is
+    # still complaining about it not being in reqs->run.
+    - python  # [win]
```
xhochy, line 55: Python packages should always have an explicit `python` in  `run` 
```suggestion
    - python
```

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,103 @@
...
+    - pynucos
+    - cell_tree2d
+    - py_gd >=2.2
+    - adios_db
+
+test:
+  imports:
+    - gnome
+  requires:
+    - pytest
```
xhochy, line 78: You're not using it.
```suggestion
```

xhochy (changes requested):

recipes/pygnome/meta.yaml
JamesMakela-NOAA: ok, I made the change.  I put these three items in *both* the build and host sections, and nothing seems to have been broken.

ChrisBarker-NOAA:
@xhochy:

I moved the clang version (15) to conda_build_config -- but now it won't work because apparently somewhere else it's requiring 16 -- ideas?

MementoRC:
Could the recipe need to re re-rendered?

recipes/pygnome/meta.yaml
JamesMakela-NOAA: Ok, the clang 15 requirement is now in conda_build_config.yaml.

recipes/pygnome/meta.yaml
JamesMakela-NOAA: It is now removed.

ChrisBarker-NOAA:
@conda-forge/help-python-c r, ready for review!

We've cleaned it up quite a bit from the last one -- hopefully ready to go now!

Thanks!

recipes/pygnome/meta.yaml
xhochy: In build, they should have a `# [build_platform != target_platform]` comment.

xhochy (changes requested):

JamesMakela-NOAA:
@xhochy:  I added the build_platform comment for numpy, cython, & netCDF4, and I was able to build locally without any failures, but now the checks are failing for all platforms.  The failure in the azure pipeline is:

`2024-08-20T18:13:55.1688580Z   make[2]: CYTHON-NOTFOUND: No such file or directory`

Can you tell me what the purpose was for adding the build_platform comments to these lines.

xhochy:
> Can you tell me what the purpose was for adding the build_platform comments to these lines.

The purpose is to ensure that these dependencies are picked always from the host environment. Only in the cross-compiling case, we will also need to have them present to ensure we can run some binaries natively during the build.

Please re-add it to Cython. If it only works without the comment, there is a deeper issue in dependency discovery with the setup and something else needs to be fixed.

ChrisBarker-NOAA:
>Please re-add it to Cython. If it only works without the comment, there is a deeper issue in dependency discovery with the setup and something else needs to be fixed.

Done -- however, I'm confused -- adding that selector means that Cython will not be a build requirement unless cross-compiling -- so it will never be there at all.

In contrast to the numpy and netCDF4, which are also host and run requirements, so will be there.

Should Cython be a host requirement?

But Cython is not a run-time binary, and needs to be there on the machine running the build, not the target machine -- more like cmake than numpy, which is why we put it in build, without the selector.

so???

ChrisBarker-NOAA:
@xhochy: I put the selector back for cython, and indeed it failed again, 'cause we told it din't need cython unless cross compiling -- so where should we be specifying cython??

It sure seem to me to be a build requirement.

ChrisBarker-NOAA:
cython now in host, but still failures, but only on linux_64. So now I'm really confused.

 I *really* think that cython is a straightforward build requirement:

"""
When both build and host sections are defined, the build section can be thought of as "build tools" - things that run on the native platform, but output results for the target platform.
"""

Cython runs on the native platform, and, indeed, outputs results (C code) that's platform independent. Some projects actually put the generated C code in the sdist, and it can be compiled anywhere. so there's not reason to do anything different when cross compiling.

So why the selector? and it doesn't need to be available at run time (or link time) for the target platform -- so why in host?

Anyway, closer, but not there yet -- we really need this!

xhochy:
> I really think that cython is a straightforward build requirement

No, because cython actually has a Python dependency and should also be using the host Python to produce the correct output. That differentiates it from the normal build/host split. For anything based on Python (in Python packages), it should go primarily in host.

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+  script: cd py_gnome && {{ PYTHON }} -m pip install ./ -vv
+
+requirements:
+  build:
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+    - numpy                               # [build_platform != target_platform]
+    - netCDF4                             # [build_platform != target_platform]
+    - cython                              # [build_platform != target_platform]
+    - scikit-build-core
```
xhochy, line 27: Same argument as with Cython
```suggestion
    - cython                              # [build_platform != target_platform]
    - scikit-build-core             # [build_platform != target_platform]
```

xhochy (changes requested):

ChrisBarker-NOAA:
we'll see if this works now ....

xhochy:
Can you switch the build to `ninja` so that we can the failing compiler command?

ChrisBarker-NOAA:
> Can you switch the build to ninja

I have no idea what that means, or how to do it. 

I *think* ninja is being used by scikit-build-core -- but that's under the hood as far as I can tell.

Cross-compiling is HARD.

>  Cython ... should also be using the host Python to produce the correct output

That's not how Cython used to work -- has something changed? I thought Cython produced platform-independent C (C++) code. There is *some* dependence on the Python version (i.e. the Python C API), but I don't think there's anything platform dependent in what it generates. In fact, it used to be common practice to include the generated code in an sdist, so it could be built elsewhere without running Cython at all. Has something changed?

Anyway, it all seemed to work with Cython as a regular build dependency -- and now it doesn't :-(

But if ninja can get it work, great, but I have no idea how to do that / what it means?

recipes/pygnome/meta.yaml
xhochy:
```suggestion
    - ninja
```

xhochy:
> That's not how Cython used to work -- has something changed? I thought Cython produced platform-independent C (C++) code. There is some dependence on the Python version (i.e. the Python C API), but I don't think there's anything platform dependent in what it generates. In fact, it used to be common practice to include the generated code in an sdist, so it could be built elsewhere without running Cython at all. Has something changed?

We basically want to avoid having a Python installation in `build` as long as we aren't cross-compiling.

> But if ninja can get it work, great, but I have no idea how to do that / what it means?

See the suggestion.

ChrisBarker-NOAA (approved):
maybe there?

ChrisBarker-NOAA:
```
CMake was unable to find a build program corresponding to "Unix Makefiles".  
```
do we need both make and ninja? Or do we need to make out build scripts ninja-based (again, no idea how...)

ChrisBarker-NOAA:
Darn, now getting:

` /Users/runner/Miniforge3/conda-bld/pygnome_1727896937254/work/py_gnome/lib_gnome/Basics.h:15:10: fatal error: 'cstdlib' file not found
  #include <cstdlib>`

That seems pretty basic -- what's up???

ChrisBarker-NOAA:
@conda-forge/help-python-c -- now I"m really confused -- we had it building opn all abut one cross-platform build (I forgot which one) -- but now It's failing on OS-X  with:

```
/Users/runner/Miniforge3/conda-bld/pygnome_1727903416636/work/py_gnome/lib_gnome/Basics.h:15:10: fatal error: 'cstdlib' file not found
  #include <cstdlib>
           ^~~~~~~~~
```
which seems pretty darn basic. Any ideas?

note: we had a lot of struggles with how to get Cython working on the platform builds, but this seems to be something different.

xhochy:
We're running into https://github.com/conda-forge/clangdev-feedstock/issues/310 Seems like `clang=15` is not well supported anymore.

h-vetinari:
> We're running into [conda-forge/clangdev-feedstock#310](https://github.com/conda-forge/clangdev-feedstock/issues/310) Seems like `clang=15` is not well supported anymore.

If you need clang 15, you can just add `libcxx-devel` to your requirements.

ChrisBarker-NOAA:
@h-vetinari : thanks, but `libcxx-devel` is there in the host requirements arlready 

So that's not it :-(

h-vetinari:
The compiler is in the `build:` environment, so `libcxx-devel` should be too.

ChrisBarker-NOAA:
cross your fingers ....

recipes/pygnome/meta.yaml
```diff
@@ -0,0 +1,105 @@
...
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+    - numpy                               # [build_platform != target_platform]
+    - netCDF4                             # [build_platform != target_platform]
+    - scikit-build-core                   # [build_platform != target_platform]
+    - cython                              # [build_platform != target_platform]
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+    - {{ compiler('cxx') }}
+    - libcxx-devel  # [osx]
```
h-vetinari, line 33: You'll need to limit this to `<19`, because libcxx 19 cannot be parsed by clang 15 anymore.

ChrisBarker-NOAA:
OMG! it's working!   

Thanks all!!!!

don't merge just yet -- I want to make sure we've got this pointed to the latest. and greatest.

ChrisBarker-NOAA:
One question:

This package doesn't work on py < 3.9 and hasn't been tested, may not work, with py3.13, so we have:

```
build:
  skip: True  # [py<39 or py>312 or python_impl == 'pypy']
```

in there.

Should this be handled by pinning python in the requirements instead? If so, in run? or everywhere?

NOTE: I have no idea why we did it that way in the first place -- but I don't want to mess with it unless someone advises me to -- this is building for the first time in months!

h-vetinari:
> Should this be handled by pinning python in the requirements instead?

The skip is fine, though you can drop the pypy part; it's obsolete.

ChrisBarker-NOAA:
OK then -- ready for review / merge!

xhochy (approved):

ChrisBarker-NOAA:
Thanks all for the support on a long slog!

PR #26810, opened 2024-06-30
Add tad-libcint package

marvinfriede:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/tad-libcint/meta.yaml
+++ b/recipes/tad-libcint/meta.yaml
@@ -1,0 +1,56 @@
+{% set name = "tad-libcint" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/tad_libcint-{{ version }}.tar.gz
+  sha256: c47332ec4c1b8a177b2d3c8f093a3ae04a507719efda5feb6c91991a23933444
+
+build:
+  skip: true  # [py>=313 or not linux]
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+    - cmake >=3.10
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+  host:
+    - libblas
+    - python
+    - pip
+  run:
+    - python
+    - numpy <2
+    - tad-mctc
+    - pytorch >=1.11,<2.4
+
+test:
+  imports:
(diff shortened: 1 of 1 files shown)


recipes/tad-libcint/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - {{ stdlib("c") }}
+    - {{ compiler('cxx') }}
+    - cmake >=3.10
+    - python                              # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+  host:
+    - libblas
+    - libgomp  # [linux]
+    - openmp  # [osx]
+    - make
```
xhochy, line 29: This should only be in build
```suggestion
```

recipes/tad-libcint/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/tad_libcint-{{ version }}.tar.gz
+  sha256: c47332ec4c1b8a177b2d3c8f093a3ae04a507719efda5feb6c91991a23933444
+
+build:
+  skip: true  # [py>=313 or not linux]
```
xhochy, line 13: Don't skip py313 yet or otherwise the migration might not come around later
```suggestion
  skip: true  # [not linux]
```

xhochy (changes requested):

marvinfriede:
The build fails for the second Python version. The workflow tries to compile for Python 3.8 - 3.12. I am not sure if this is the correct way or if isolation is required. But clearly, the second build mixes in paths from the first build ([error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=970750&view=logs&j=6f142865-96c3-535c-b7ea-873d86b887bd&t=22b0682d-ab9e-55d7-9c79-49f3c3ba4823&l=1889)).
I tried to draw inspiration from [PySCF](https://github.com/conda-forge/pyscf-feedstock/tree/main/.azure-pipelines), because they use the same C library. But there are a lot more modifications (e.g. in the .azure_pipelines directory). Not sure, how to proceed.

xhochy:
The problem seems to be that the CMake build picks up the system Python instead of the conda python. You probably need to pass -DPython_Executable or similar to the cmake command.

marvinfriede:
@xhochy I couldn't figure out a fix. That's why I restricted the recipe to only build for Python 3.11. I plan on updating the feedstock (with changes outside the recipe directory as in [PySCF](https://github.com/conda-forge/pyscf-feedstock/tree/main)) once the package is registered. Is this an acceptable way forward?

xhochy:
At the moment, this is skipping CI completely. You should figure out how to select the right Python or we cannot merge this as the packages are prone to be faulty (the build python can change at any time).

PR #26812, opened 2024-07-01
Fix UX papercuts while debugging locally

jaimergp:
The build_steps.sh contained a couple papercuts that made it difficult to debug locally (at least on macOS for Linux). This PR fixes these:

- Directory considered unsafe by git
- Permissions issue with conda notices
- Unnecessary build_artifacts copy

### DIFF
--- a/.scripts/build_steps.sh
+++ b/.scripts/build_steps.sh
@@ -36,6 +36,10 @@ show_channel_urls: true
 solver: libmamba
 CONDARC
 
+# Workaround for errors related to "unsafe" directories:
+# https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765 
+git config --global --add safe.directory "${FEEDSTOCK_ROOT}"
+
 # Copy the host recipes folder so we don't ever muck with it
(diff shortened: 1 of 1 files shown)


jaimergp:
@conda-forge/core - this is ready. WDYT?

beckermr:
Let's try it! Thank you!

.scripts/build_steps.sh
```diff
@@ -36,8 +36,16 @@ show_channel_urls: true
 solver: libmamba
 CONDARC
 
+# Workaround for errors related to "unsafe" directories:
+# https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765 
+git config --global --add safe.directory "${FEEDSTOCK_ROOT}"
```
jakirkham, line 41: Understand this on Linux in the Docker image. Though on macOS (where it is run on the user's system) should this be guarded?

jakirkham:
Thanks Jaime! 🙏

Had a question above

.scripts/build_steps.sh
jaimergp: MacOS has its own script separate from this, it shouldn't be affected? Or do you mean when macOS is the host?

PR #26890, opened 2024-07-08
add lima

danielnachun:
Checklist
- [ ] Build number is 0.


danielnachun:
@xhochy @pavelzw This will be of interest to you as it should resolve whatever was blocking https://github.com/conda-forge/staged-recipes/pull/23930 from working.  One thing I struggled with is that there is hard dependency on macOS 11.0 for the C standard library, but the error was extremely cryptic.

recipes/lima/meta.yaml
```diff
@@ -0,0 +1,40 @@
+{% set name = "lima" %}
```
mfansler, line 1: I believe a prohibitive concern here is that the name `lima` is used for [a different recipe in Bioconda](https://github.com/bioconda/bioconda-recipes/tree/master/recipes/lima) that builds [a `lima` package from PacificBiosciences](https://github.com/PacificBiosciences/barcoding). Such a convergence is problematic for Bioconda because they prioritize the `conda-forge` channel, such that this package would mask theirs.

Can we consider shipping this as `lima-vm` instead?

mfansler (changes requested):

recipes/lima/meta.yaml
danielnachun: Yes, I actually meant to specifically ask if `lima-vm` should be used instead but forgot to put that in my comment.  I will change the recipe name as you have suggested.  Thanks!

xhochy (approved):

recipes/lima-vm/build.sh
```diff
@@ -0,0 +1,22 @@
...
+sed -i "s/codesign --entitlements/codesign --force --entitlements/" Makefile
+
+make VERSION=${PKG_VERSION}
+cp -r _output/bin/* ${PREFIX}/bin
+cp -r _output/share/* ${PREFIX}/share
+
+mkdir -p ${PREFIX}/etc/bash_completion.d
+mkdir -p ${PREFIX}/share/zsh/site-functions
+mkdir -p ${PREFIX}/share/fish/vendor_completions.d
+limactl completion bash > ${PREFIX}/etc/bash_completion.d/lima
```
pavelzw, line 18: Does this actually work? 🤯

danielnachun: It definitely generates the completion files - I just tested that now locally.  Whether the users's shell properly adds them will probably depend on their specific configuration.  Did you run into issues with this in the previous PR?  Unfortunately the logs were expired so I wasn't able to see where it was failing.  If there are any tweaks to be made, please feel free to make an issue or PR on the feedstock!

PR #26986, opened 2024-07-15
Add input4MIPs validation

znichollscr:
Add input4mips-validation, a package for validating input data for climate model runs. See https://github.com/climate-resource/input4mips_validation

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -20,22 +20,21 @@ requirements:
   host:
     - python >=3.9
     - poetry-core
-    - pip
   run:
     - python >=3.9.0
     - loguru >=0.7
     - cf_xarray >=0.8
     - xarray >=2022.0
     - pint-xarray >=0.3
-    - numpy >=1.23,<2.0
+    - numpy >=1.23
     - cftime >=1.6
     - typing-extensions >=4.0
     - pooch >=1.0
     - netcdf4 >=1.0
     - attrs >=22.0
-    - typer >=0.9.0,<0.10.0
-    - validators >=0.28.3,<0.29.0
-    - cattrs >=23.2.3,<24.0.0
+    - typer >=0.9.0
+    - validators >=0.28.3
+    - cattrs >=23.2.3
   run_constrained:
     - matplotlib-base >=3.0
     - notebook >=6.0
@@ -57,9 +56,8 @@ outputs:
   - name: input4mips-validation-locked
     requirements:
       host:
-        - python >=3.9,<4.0
+        - python >=3.9
         - poetry-core
-        - pip
       run:
         - python 3.9.*
         - {{ pin_subpackage('input4mips-validation', exact=True) }}
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -18,7 +18,7 @@ build:
 
 requirements:
   host:
-    - python >=3.9
+    - python >=3.9.0
     - poetry-core
   run:
     - python >=3.9.0
@@ -55,14 +55,12 @@ outputs:
   # (e.g. dependency updates with breaking changes)
   # for those who just want the CLI.
   - name: input4mips-validation-locked
-    build:
-      skip: true  # [py != 39]
     requirements:
       host:
-        - python
+        - python >=3.9.0,<3.10
         - poetry-core
       run:
-        - python
+        - python >=3.9.0,<3.10
         - {{ pin_subpackage('input4mips-validation', exact=True) }}
         - {{ pin_compatible('attrs', min_pin='23.2.0', max_pin='23.2.1') }}
         - {{ pin_compatible('cattrs', min_pin='23.2.3', max_pin='23.2.4') }}
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -62,6 +62,7 @@ outputs:
         - pip
       run:
         - python ==3.9
+        - {{ pin_subpackage('input4mips-validation', exact=True) }}
         - {{ pin_compatible('attrs', min_pin='23.2.0', max_pin='23.2.1') }}
         - {{ pin_compatible('cattrs', min_pin='23.2.3', max_pin='23.2.4') }}
         - {{ pin_compatible('cftime', min_pin='1.6.4', max_pin='1.6.5') }}
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -18,11 +18,11 @@ build:
 
 requirements:
   host:
-    - python >=3.9,<4.0
+    - python >=3.9
     - poetry-core
     - pip
   run:
-    - python >=3.9.0,<4.0.0
+    - python >=3.9.0
     - loguru >=0.7
     - cf_xarray >=0.8
     - xarray >=2022.0
@@ -61,7 +61,7 @@ outputs:
         - poetry-core
         - pip
       run:
-        - python ==3.9
+        - python 3.9.*
         - {{ pin_subpackage('input4mips-validation', exact=True) }}
         - {{ pin_compatible('attrs', min_pin='23.2.0', max_pin='23.2.1') }}
         - {{ pin_compatible('cattrs', min_pin='23.2.3', max_pin='23.2.4') }}
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -61,7 +61,7 @@ outputs:
         - poetry-core
         - pip
       run:
-        - python==3.9
+        - python ==3.9
         - {{ pin_compatible('attrs', min_pin='23.2.0', max_pin='23.2.1') }}
         - {{ pin_compatible('cattrs', min_pin='23.2.3', max_pin='23.2.4') }}
         - {{ pin_compatible('cftime', min_pin='1.6.4', max_pin='1.6.5') }}
@@ -85,6 +85,7 @@ outputs:
         - pip
 
 about:
+  home: https://github.com/climate-resource/input4mips_validation
   summary: Validation of input4MIPs data (checking file formats, metadata etc.).
   license: BSD-3-Clause
   license_file: LICENCE
--- a/recipes/input4mips-validation/meta.yaml
+++ b/recipes/input4mips-validation/meta.yaml
@@ -57,10 +57,10 @@ outputs:
   - name: input4mips-validation-locked
     requirements:
       host:
-        - python >=3.9.0,<3.10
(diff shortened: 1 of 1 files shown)


znichollscr:

znichollscr:

znichollscr:

znichollscr:
@conda-forge-admin, please ping conda-forge/core 

As one other note, I'd be happy to help out as a reviewer of conda recipes. I'm no expert, but I have done this for a few packages and would be happy to give back. (As a note, my contributions have been all under my other account: @znicholls)

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - pint-xarray >=0.3
+    - numpy >=1.23,<2.0
+    - cftime >=1.6
+    - typing-extensions >=4.0
+    - pooch >=1.0
+    - netcdf4 >=1.0
+    - attrs >=22.0
+    - typer >=0.9.0,<0.10.0
+    - validators >=0.28.3,<0.29.0
+    - cattrs >=23.2.3,<24.0.0
```
jaimergp, line 38: Are these strict upper bounds necessary or just "out of caution". If the latter, they are not usually needed and we prefer being optimistic in the honor of flexibility. If something does break, we can repodata patch and the constraints after releasing it without rebuilding anything.

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+  - name: input4mips-validation
+
+  # A locked version, which will help avoid installation headaches
+  # for those who just want the CLI.
+  - name: input4mips-validation-locked
+    requirements:
+      host:
+        - python >=3.9,<4.0
+        - poetry-core
+        - pip
```
jaimergp, line 62: I think you don't need this here.

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+      run:
+        - python ==3.9
```
jaimergp, line 64:
```suggestion
        - python 3.9.*
```

I doubt you only want `3.9.0` but not `3.9.1` etc.

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+build:
+  entry_points:
+    - input4mips-validation = input4mips_validation.cli:app
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.9,<4.0
```
jaimergp, line 21:
```suggestion
    - python >=3.9
```

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - poetry-core
+    - pip
+  run:
+    - python >=3.9.0,<4.0.0
```
jaimergp, line 25:
```suggestion
    - python >=3.9.0
```

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,96 @@
...
+    - pip check
+    - input4mips-validation --help
+  requires:
+    - pip
+
+outputs:
+  - name: input4mips-validation
+
+  # A locked version, which will help avoid installation headaches
+  # for those who just want the CLI.
```
jaimergp, line 56: Smart! First time I see this type of output, not sure how I feel about it 🤔 How is this helping with "installation headaches". Fearing a breaking dependency update or something?

recipes/input4mips-validation/meta.yaml
znichollscr: Thank you, very helpful!

recipes/input4mips-validation/meta.yaml
znichollscr: Ah yes these can all be removed. I think they're the result of using Grayskull to build the recipe based on a repository which uses poetry (which automatically adds these upper bounds). I'll remove them, the flexibility is definitely the goal (particular given there is also a 'locked' version).

recipes/input4mips-validation/meta.yaml
znichollscr: So the idea is that you have a locked version of the package, which is insensitive to dependency updates. For people who want to use the package's CLI as an application, this locked version provides a stable interface that is insensitive to dependency updates. We'll have lots of users who will just run away if they get installation issues, so I want to offer them a path that is as likely to work as possible (and avoid breaking due to dependency updates). For people who want to use the package as a library, they can still use the non-locked version with the full flexibility it provides.

As a concrete example, the locked version would mean that, even if numpy updates to 3.0, if a user installs input4mips-validation-locked, the package would still work (the latest numpy update to 2.0 has caused chaos throughout our stack, and having these locked versions would have helped a lot). 

As another reference for this kind of behaviour, there is a plugin for this in the Python world (https://pdm-build-locked.readthedocs.io/en/stable/), I am basically implementing this behaviour by hand here because the stability is key for this project.

recipes/input4mips-validation/meta.yaml
znichollscr: Thanks, will try removing

recipes/input4mips-validation/meta.yaml
znichollscr: Removed upper bounds

znichollscr:
Thanks for the review!

znichollscr:
I think I've fixed everything now and CI is green. The thread about the locked version is the only remaining thing. Thanks again @jaimergp

recipes/input4mips-validation/meta.yaml
znichollscr: > They thing is that they'll probably need to install it in a new environment to maximize their chances of not running into conflicts with other projects due to the hard pinning

Correct.

> At that point, isn't something like PyInstaller or nuitka more suited?

Perhaps. For me, there's a few issues with these solutions and I would really appreciate merging with the locked version as is. Even if it isn't the perfect solution, it is a solution. A bit more reasoning:

1. Our community knows and uses conda. I had never heard of PyInstaller and nuitka and I would appreciate the ability to avoid asking our community to become familiar with another tool.
  - You are completely correct that this solution is basically using conda a bit like "pipx". I know that's not the intended use case, but I think it does work as a fine solution and I don't see any great harm (but please correct me if I'm wrong)?  (Arguably, a pre-built docker image would also be a great solution, as I said, I'm trying to avoid Docker because our community just isn't ready for it yet.)
3. Distributing with PyInstaller/nuitka looks like a case of "build somewhere, send files to others to use". Building for multiple platforms requires doing multiple builds on different platforms I assume. I don't have my head around this for those tools (I assume it is not so hard, but spending more hours working out the CI etc. for that is a step I would love to be able to not do right now). On the other hand, I do know how to make a multi-platform compatible tool with conda.
4. It's not obvious to me that PyInstaller and nuitka work with conda specific dependencies. Their docs all focus on PyPI dependencies. So, while they might work for conda, if they don't work out of the box, I'm going to have quite a bit of learning to do. As I said, I would love to avoid that.

That's obviously a fairly long answer. As I said, I completely agree that there are probably other ways to do this. I am happy to accept that those ways might be better. I hope you can see that this way should also achieve the desired goals and hence we can merge this as is.

jaimergp: No one else opposed so let's see. In the future if we see that it's not as useful as intended it can be removed once in the feedstock.

If you want an example of PyInstaller / nuitka, for the sake of curiosity, here are two examples:

- PyInstaller: https://github.com/conda-forge/conda-standalone-feedstock
- nuitka: https://github.com/conda/conda-standalone/pull/80

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+outputs:
+  - name: input4mips-validation
+
+  # A locked version, which will help avoid installation headaches
+  # (e.g. dependency updates with breaking changes)
+  # for those who just want the CLI.
+  - name: input4mips-validation-locked
+    requirements:
+      host:
+        - python >=3.9
```
jaimergp, line 60:
```suggestion
        - python 3.9.*
```

We probably want this here too.

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+  requires:
+    - pip
+
+outputs:
+  - name: input4mips-validation
+
+  # A locked version, which will help avoid installation headaches
+  # (e.g. dependency updates with breaking changes)
+  # for those who just want the CLI.
+  - name: input4mips-validation-locked
```
jaimergp, line 57:
```suggestion
  - name: input4mips-validation-locked
    build:
      skip: true  # [py != 39]
```

recipes/input4mips-validation/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    requirements:
+      host:
+        - python >=3.9
+        - poetry-core
+      run:
+        - python 3.9.*
```
jaimergp, line 63:
```suggestion
        - python
```

Since this is non-noarch, it will get a build per Python version and OS (hence the skip above). So no need to repeat ourselves here.

recipes/input4mips-validation/meta.yaml
znichollscr: Very cool, thank you!

recipes/input4mips-validation/meta.yaml
znichollscr: Hmm this didn't work as expected. I'll try some different routes now.

znichollscr:
@jaimergp I went with a more basic solution than [your suggestion](https://github.com/conda-forge/staged-recipes/pull/26986#discussion_r1684691201) (because the suggestions didn't work for reasons I couldn't figure out). Does this now look ok to merge to you?

jaimergp:
I think that might be a bug in the linter 🤔 What if you list the outputs [like this](https://github.com/conda-forge/napari-feedstock/blob/main/recipe/meta.yaml), without a top-level package. You might need to adjust the `package.name` field so it doesn't match any of the outputs (again take a look at the example I linked). Let me know if it's not clear.

recipes/input4mips-validation/meta.yaml
znichollscr: Thanks for the tip, seems to be behaving now

recipes/input4mips-validation/meta.yaml
znichollscr: Thanks for the tip, seems to be behaving now

znichollscr:
Thanks for the pointer to napari, seems to be behaving now

jaimergp (approved):

PR #27048, opened 2024-07-21
Add mesonlsp, checkmake, neocmakelsp

danielnachun:
Checklist
- [ ] Build number is 0.


danielnachun:
@conda-forge/staged-recipes This is now ready for review.  

I had to skip the Windows build for `mesonlsp` because it cannot find `libcurl` with `pkg-config`, and the build system does not provide another way I could find to manually specify the paths to the `libcurl` headers and shared libraries.  I believe the `libcurl` package on Windows is missing the `libcurl.pc` file `pkg-config`, so this may be fixable in the `libcurl` feedstock eventually, but I'd like to get feedstock for `mesonlsp` set up for the other OSes first.

recipes/mesonlsp/build.sh
```diff
@@ -0,0 +1,23 @@
...
+sed -i "s/'-Wl,-ld_classic'//" meson.build
+sed -i 's/and !defined(__x86_64__)//g' src/polyfill/polyfill.hpp
+
+export CXXFLAGS="${CXXFLAGS} -D_LIBCPP_DISABLE_AVAILABILITY"
+if [[ ${target_platform} =~ .*linux.* ]]; then
+    export CFLAGS="${CFLAGS} -pthread"
+    export CXXFLAGS="${CXXFLAGS} -pthread"
+fi
+
+meson --prefix=${PREFIX} \
```
xhochy, line 18: Probably this should allow you to omit the `--libdir` and will make sure that cross-compiling for the ARM arches works smoothly.
```suggestion
meson $MESON_ARGS --prefix=${PREFIX} \
```

recipes/mesonlsp/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - pkgconf
+    - pygls
+    - libuuid
+  run:
+    - fmt  # [osx]
+    - libarchive
+    - libcurl
+    - nlohmann_json
+    - pkgconf
+    - libuuid
```
xhochy, line 40: Most of of these should come in via `run_exports` or are header-only.
```suggestion
  run:
    - pkgconf
```

recipes/pkgconf/build.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
+meson --prefix=${PREFIX} \
```
xhochy, line 5:
```suggestion
meson $MESON_ARGS --prefix=${PREFIX} \
```

xhochy (changes requested):

recipes/mesonlsp/build.sh
danielnachun: It turns out `${MESON_ARGS}` also provides `--prefix=${PREFIX}` and `--buildtype=release` so I was able to remove those as well.

recipes/mesonlsp/meta.yaml
danielnachun: Upon further examination of my local installation of `mesonlsp`, it is dynamically linking to `libpkgconf`.  So I added a `run_export` to `pkgconf` as well.  I believe this is okay to do here because if the user is using `pkgconf` only to find dependencies, they should put it in `build`, which as far as I know does not propagate `run_exports`.  On the other hand if they want to link to `libpkgconf`, they would need to put `pkgconf` in `host`, which would propagate `run_exports`.  If my assumption here is wrong I will revert that change!

xhochy (approved):
Sounds right!

PR #27129, opened 2024-07-31
Add dydxprotocol v4 client py

MementoRC:
Checklist
- [x] Build number is 0.


MementoRC:
@xhochy How to resolve this issue? I don't think I deleted anything. I'll adda `recipe.yaml` in example, but it does not make much sense to me

recipes/dydxprotocol-v4-client-py/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+  noarch: python
+  script:
+    - cd v4-client-py-v2 && poetry build -f wheel
+
+requirements:
+  host:
+    - dydxprotocol-v4-proto-py ==5.0.6
+    - poetry
+    - python >=3.9
+    - typing-extensions ==4.12
```
xhochy, line 26: What's the reason to have this block?
```suggestion
requirements:
  host:
    - dydxprotocol-v4-proto-py ==5.0.6
    - poetry
    - python >=3.9
    - typing-extensions ==4.12
```

xhochy:
> @xhochy How to resolve this issue? I don't think I deleted anything. I'll adda `recipe.yaml` in example, but it does not make much sense to me

This is resolved and other failures depend on missing dependencies, right?

MementoRC:
> > @xhochy How to resolve this issue? I don't think I deleted anything. I'll adda `recipe.yaml` in example, but it does not make much sense to me
> 
> This is resolved and other failures depend on missing dependencies, right?

Yes, it is, thx and welcome back. It's pending v4-proto

MementoRC:
proto is live

MementoRC:
v4-proto is live

MementoRC:
@conda-forge/help-python , hello, ready for review

recipes/dydxprotocol-v4-client-py/meta.yaml
```diff
@@ -0,0 +1,106 @@
...
+    test:
+      imports:
+        - dydx_v4_client
+      requires:
+        - pip
+      commands:
+        - pip check
+
+  # Package name required by v4-client upstream pyproject.toml
+  - name: v4-client
```
xhochy, line 79: This won't be visible in any way to the `pyproject.toml`. This is a conda package and not showing up in `pip`/`python` metadata. Please remove this entry.

xhochy (changes requested):

recipes/dydxprotocol-v4-client-py/meta.yaml
MementoRC: I don't understand, would it not be installable, like `conda install v4-client`? My comment is likely misleading, I menat to say that if a package has a `v4-client` requirement in its pyprokect.toml (or else), then a user would be able to `conda search v4-client` and add it to its `environment.yml`

xhochy: With that comment, it makes more sense. I would stick to PyPi names though and there it isn't available either, right?

xhochy: With that comment, it makes more sense. I would stick to PyPi names though and there it isn't available either, right?

MementoRC: I mixed-up the names, there is a `v4-client-py`, but it may be the older package. I just saw that they released a `dydx-v4-client` about 15days ago, and that uses the same git as the one I propose.

It make sense to remove it then since they have the pip package that matches`dydx-v4-client`

MementoRC: It is a bit messy, they released a version 1.0.1 in pip, but they refer to it as:
`pip install git+https://github.com/dydxprotocol/v4-clients/blob/main/v4-client-py-v2`, which I translated to their tags `v4-client-js@{{ version }}`. They only tag the JS release

I am going to simplify it and remove the split, simply doing a poetry build and pip install.

I would prefer to follow the github, but I could also use the pypi if you recommend it. I can also contact them, they responded for the JS package

xhochy (approved):

PR #27175, opened 2024-08-05
Add highspy package

adament:
conda-forge already carries a highs package for the C++ library and executable. This is a package for the python library with the same name as the pypi package. Even though the C++ package builds a shared library on non-windows platforms, it seems that upstream only supports linking the python module statically against the library code. So without major surgery to the upstream build system, I think this is the best I can do.

Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/highspy/meta.yaml
+++ b/recipes/highspy/meta.yaml
@@ -1,0 +1,71 @@
+{% set version = "1.7.2" %}
+
+package:
+  name: highspy
+  version: {{ version }}
+
+source:
+  url: https://github.com/ERGO-Code/HiGHS/archive/v{{ version }}.tar.gz
+  sha256: 5ff96c14ae19592d3568e9ae107624cbaf3409d328fb1a586359f0adf9b34bf7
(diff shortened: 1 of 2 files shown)


adament:
@conda-forge/help-python I think this is ready for review.

@julian-belina do you have any comments?

julian-belina:
@adament this looks promising. Does this recipe also compile the original code? The only installation command I can find installs the python package of the library. Unfortunately, I was not able recreate the highs pip package on conda-forge

adament:
It compiles the C++ code of the original library as object files and links them into the finale python module, which is equivalent to a linking statically against the original library. This replicates to the best of my understanding how the pip packages are built. All of this under the assumption that I understand the cmake code correctly.

So the highs and highspy packages should now be separate and independent, with the highspy package being equivalent to the pip package and the highs package with the C++ library, headers and standalone executable.

julian-belina:
I have another library that depends on highs and will test if it works as expected. Thank you for creating the recipe

PR #27188, opened 2024-08-06
Add adacpp

Krande:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/adacpp/conda_build_config.yaml
+++ b/recipes/adacpp/conda_build_config.yaml
@@ -1,0 +1,3 @@
+variant:
+  - novtk
+  - all
--- a/recipes/adacpp/conda_build_config.yaml
+++ b/recipes/adacpp/conda_build_config.yaml
@@ -2,7 +2,3 @@ c_stdlib_version:          # [osx and x86_64]
   - "10.15"                # [osx and x86_64]
 MACOSX_SDK_VERSION:        # [osx and x86_64]
   - "10.15"                # [osx and x86_64]
-
-variant:
-  - novtk
-  - all
--- a/recipes/adacpp/conda_build_config.yaml
+++ b/recipes/adacpp/conda_build_config.yaml
@@ -1,4 +1,3 @@
-# Please consult conda-forge/core before doing this
 c_stdlib_version:          # [osx and x86_64]
   - "10.15"                # [osx and x86_64]
 MACOSX_SDK_VERSION:        # [osx and x86_64]
--- a/recipes/adacpp/conda_build_config.yaml
+++ b/recipes/adacpp/conda_build_config.yaml
@@ -1,5 +1,8 @@
-MACOSX_SDK_VERSION:  # [osx and x86_64]
-  - 10.15            # [osx and x86_64]
+# Please consult conda-forge/core before doing this
+c_stdlib_version:          # [osx and x86_64]
+  - "10.15"                # [osx and x86_64]
+MACOSX_SDK_VERSION:        # [osx and x86_64]
+  - "10.15"                # [osx and x86_64]
 
 variant:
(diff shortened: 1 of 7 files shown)


Krande:

Krande:

recipes/adacpp/conda_build_config.yaml
```diff
@@ -0,0 +1,9 @@
+# Please consult conda-forge/core before doing this
```
xhochy, line 1: No need to add this comment here.
```suggestion
```

recipes/adacpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    sha256: 499fb878bb3c1d1bd2bd219f747f6e1f41883c6f63e81e192af204664e8908f2
+    patches:
+      - patches/remove_hardcoded_osx_sysroot_and_deployment_target.patch
+
+build:
+  skip: True  # [py<310]
+  number: 0
+  string: py{{ CONDA_PY }}_{{ variant }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}
+  run_exports:
+    - {{ pin_subpackage(name, exact=True) }} *{{ variant }}*
```
xhochy, line 19: If you pin exact, you don't need the variant anymore. Did you want to limit to `x.x.x` and the variant?

recipes/adacpp/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - scikit-build-core
+    - nanobind
+    - cli11
+    - tinygltf
+    - gmsh
+    - pip
+
+  run:
+    - python
+    - "{{ pin_compatible('gmsh', exact=True) }}"
```
xhochy, line 45: Do you really need the exact same build of gmsh? Normally this is pinned to some patch/minor level.

recipes/adacpp/meta.yaml
Krande: Ah, I didn't consider that exact=True removes the need for the extra variant pinning. 

However, In hindsight, I believe I can skip building against both variants. 

So the background for the 2 variants is that `occt` offers a "lightweight" variant called `novtk` which doesn't rely on VTK, which significantly reduces the number of dependencies. It's been a while since I last tested, but I faintly recall having issues when switching between the occt variants downstream against packages compiled for just one of the variants. But I am suspecting that my issues back then might have been due to something else.

If my understanding is correct, then although adacpp is compiled against the occt variant with VTK (all) or without (novtk) I it should still be ABI compatible (obviously as long as the downstream package doesn't rely on any occt features depending on the VTK compilation being ON).

I just tested compiling adacpp locally by removing the 2 variants. By default it will compile against the `occt=7.7.2=all`. Then I created a new environment where I forced `occt=*=*novtk*`. And It looks like all my tests are passing.

I maintain a few other packages which currently also build against both occt variants. But it seems like I don't need to as long as I don't rely on any VTK-exclusive features of occt. This can potentially reduce the build matrix significantly for a couple of packages (such as [ifcopenshell](https://github.com/conda-forge/ifcopenshell-feedstock)

I'll make an update shortly with the removal of the variants. I'll pin this to the patch-level as I am not certain I'll be able to maintain a 100% compliant semver versioning in the next couple of updates

recipes/adacpp/meta.yaml
Krande: Yea, I think you're right. I'll pin this to the minor level

xhochy (approved):

Krande:
ping @xhochy. Any chance you have time to merge this? :)

Krande:
@conda-forge/help-python-c, given that this PR is approved, maybe we can merge this now? :)

PR #27201, opened 2024-08-07
Add dydx-v4-proto c++

MementoRC:
Checklist
- [x] Build number is 0.


MementoRC:
@conda-forge/help-c-cpp, Hello, ready for review

MementoRC:
@xhochy Is it acceptable to have both `proto` and `client` into one recipe as in: https://github.com/conda-forge/staged-recipes/pull/27325?

Upstream only provides one source for both, but I can continue working on separating them as in this PR

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
```diff
@@ -0,0 +1,171 @@
...
+  # Reuiqres GCC >= 7
+  skip: True  # [win]
+
+requirements:
+  build:
+    - {{ compiler('c') }}  # [not win]
+    - {{ compiler('cxx') }}  # [not win]
+    - {{ stdlib("c") }}
+    # Using experimental m2w64 toolchain for now
+    - mingw-w64-ucrt-x86_64-gcc  # [win and x86_64]
```
xhochy, line 63: These should be available using normal terms now. At least there is a UCRT migration that uses these compiler but the standard `compiler()` functions

MementoRC: You lost me. I need this for gcc>7, the `blst` recipe that I am working on still show `5.3.0` for` compiler('m2w64_c')`

xhochy: This is the migration I'm referring to https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/migrations/r-base44_and_m2w64-ucrt.yaml

MementoRC: `compiler('m2w64_c')` still installs gcc 5.3

MementoRC: Ok, got it with the `conda_build_config.yaml`

MementoRC:
@xhochy Oddly, I don't seem to know how to use the environment, CC and CXX do not seem to be set and by default cmake picks clang (which I don't install ... confused)

xhochy:
It's green now. Do you still need help or should I review?

MementoRC:
I am tempted to say yes and yes.

Also, just so you know ahead of time, I do modify the upstream to separate proto and client and provide dynamic libraries

There are no tests for this lib, so, I am still debugging whether any changes are needed by using it in the `client` PR, there is something odd with protobuf across platforms.

MementoRC:
@xhochy The missing symbols are solved and tested against the client. I think it is ready for review, I remove the versioned .so but kept the `outputs` since I feel it helps me filter the content of the package. Thank you

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
```diff
@@ -0,0 +1,180 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}
+    - cmake >=3.25
+    - libprotobuf >=4.22  # [build_platform == target_platform]
```
xhochy, line 62: I'm confused. I would have expected the reverse condition.

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
```diff
@@ -0,0 +1,180 @@
...
+    - m2w64-binutils  # [win]
+    - ninja
+    - pkg-config
+    - zlib
+  host:
+    - catch2
+    - libabseil >=20240116
+    - libprotobuf >=4.22
```
xhochy, line 70: Omit the pins to get get global pin
```suggestion
    - libabseil
    - libprotobuf
```

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
```diff
@@ -0,0 +1,180 @@
...
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ stdlib("c") }}
+      host:
+        - {{ pin_subpackage('libdydx_v4_proto-headers') }}
+        - libabseil >=20240116
+        - libprotobuf >=4.22
+      run:
+        - {{ pin_subpackage('libdydx_v4_proto-headers') }}
+        - libprotobuf >=4.22  # [win]
```
xhochy, line 100: This should come in via run_exports
```suggestion
```

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
```diff
@@ -0,0 +1,180 @@
...
+    - zlib
+  host:
+    - catch2
+    - libabseil >=20240116
+    - libprotobuf >=4.22
+
+outputs:
+  - name: libdydx_v4_proto
+    build:
+      ignore_run_exports:
```
xhochy, line 75: Please use `ignore_run_exports_from`

xhochy (changes requested):

recipes/dydxprotocol-v4-proto-cpp/meta.yaml
MementoRC: Right, I forgot why (if any logical reason) I tried that

xhochy (approved):

PR #27243, opened 2024-08-11
Adds rattler-build recipe compatibility (and adds jolt-physics)

baszalmstra:
Checklist
- [x] Build number is 0.

This PR adds support for the new recipe format and adds the first such recipe for https://github.com/jrouwe/JoltPhysics

We already did a previous attempt with https://github.com/conda-forge/staged-recipes/pull/27008 which we reverted because of some issues with conda-smithy. A new release is available which should fix this issue.

### DIFF
--- a/.ci_support/build_all.py
+++ b/.ci_support/build_all.py
@@ -1,23 +1,20 @@
-import argparse
-import atexit
-import os
-import re
-import subprocess
-import sys
-import tempfile
-from collections import OrderedDict
-from itertools import chain
 from shutil import rmtree
-
+import tempfile
 import conda.base.context
 import conda.core.index
 import conda.resolve
 import conda_build.api
-import conda_build.variants
 import conda_index.api
 import networkx as nx
-import yaml
 from compute_build_graph import construct_graph
+import argparse
+import atexit
+import re
+import os
+from collections import OrderedDict
+import sys
+import subprocess
+import yaml
 
 try:
     from ruamel_yaml import BaseLoader, load
@@ -30,7 +27,6 @@ EXAMPLE_RECIPE_FOLDERS = ["example", "example-new-recipe"]
 
 def get_host_platform():
     from sys import platform
-
     if platform == "linux" or platform == "linux2":
         return "linux"
     elif platform == "darwin":
@@ -45,21 +41,14 @@ def get_config_name(arch):
 
 
 def build_all(recipes_dir, arch):
-    folders = list(
-        filter(
-            lambda d: os.path.isdir(os.path.join(recipes_dir, d)),
-            os.listdir(recipes_dir),
-        )
-    )
+    folders = list(filter(lambda d: os.path.isdir(os.path.join(recipes_dir, d)), os.listdir(recipes_dir)))
     if not folders:
         print("Found no recipes to build")
         return
 
     platform = get_host_platform()
     script_dir = os.path.dirname(os.path.realpath(__file__))
-    variant_config_file = os.path.join(
-        script_dir, "{}.yaml".format(get_config_name(arch))
-    )
+    variant_config_file = os.path.join(script_dir, "{}.yaml".format(get_config_name(arch)))
 
     has_meta_yaml = False
     has_recipe_yaml = False
@@ -70,11 +59,11 @@ def build_all(recipes_dir, arch):
         meta_yaml = os.path.join(recipes_dir, folder, "meta.yaml")
         if os.path.exists(meta_yaml):
             has_meta_yaml = True
-            with open(meta_yaml, "r", encoding="utf-8") as f:
-                text = "".join(f.readlines())
-                if "cuda" in text:
+            with(open(meta_yaml, "r", encoding="utf-8")) as f:
+                text = ''.join(f.readlines())
+                if 'cuda' in text:
                     found_cuda = True
-                if "sysroot_linux-64" in text:
+                if 'sysroot_linux-64' in text:
                     found_centos7 = True
 
         recipe_yaml = os.path.join(recipes_dir, folder, "recipe.yaml")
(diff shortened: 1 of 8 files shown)


baszalmstra:
Im unsure what is wrong with the WIndows build? There doesnt seem to be an actual error?

.ci_support/build_all.py
h-vetinari, line 1: Please remove all the formatting changes from this file, as it makes the actual changes unreviewable. Or at least, separate out the formatting into a separate commit before or after.

h-vetinari (commented):
This PR does too many things IMO.
* format `build_all.py`
* add rattler-build support
* add a new rattler example recipe
* add a genuinely new recipe

Proving the functionality would already with actually building the example recipe (perhaps with some additional safeguard that we don't create a feedstock), and then a follow-up can add jolt.

W.r.t. to the error on windows, it seems to be failing while trying to call `inspect_artifacts`.
```
Inspecting artifacts
"inspect_artifacts needs conda-forge-ci-setup >=4.6.0"
##[error]Cmd.exe exited with code '1'.
```

baszalmstra:
Thanks for checking out this PR @h-vetinari 

Note that this was already reviewed in https://github.com/conda-forge/staged-recipes/pull/27008 by both @jaimegp and @wolfv  but happy to make more changes.

The problem with the current setup is that there is no easy way to test this. The creation of the feedstock is performed by another repository so without merging this that code won't run and that code won't actually create a feedstock if there is no recipe, therefor we must add a recipe. 

The `build-all.py` script is tested currently by running the CI, if that turns green it should be good to go. I don't see why this recipe should be different from testing the create feedstock code.

The example recipe is there as an example similar to the existing example recipe. I would be happy to add it in another PR but in the previous round of reviews, it was mentioned that this was missing from the PR.

I undid the formatting.

recipes/example-new-recipe/recipe.yaml
```diff
@@ -0,0 +1,99 @@
...
+  #   - win
+  script: python -m pip install . -vv
+  number: 0
+
+requirements:
+  build:
+    # If your project compiles code (such as a C extension) then add the required compilers as separate entries here.
+    # Compiler names include 'c', 'cxx' and 'fortran', among others.
+    - ${{ compiler('c') }}
+    - ${{ stdlib('c') }}  # If you need a compiler, add the `stdlib` equivalent too
```
h-vetinari, line 56: This makes it sound like other compilers (fortran, rust, cxx) also need an equivalent stdlib, but this is not the case. The C stdlib is unfortunately special in many ways (we've iterated on this with Wolf a couple of times, I don't have a better solution for now). I'd write
```suggestion
    - ${{ stdlib('c') }}  # If you need any compiler, add the C standard library ("stdlib") too
```

recipes/example-new-recipe/recipe.yaml
```diff
@@ -0,0 +1,99 @@
...
+  # If getting the source from GitHub, remove the line above,
+  # uncomment the line below, and modify as needed. Use releases if available:
+  # url: https://github.com/simplejson/simplejson/releases/download/${{ version }}/simplejson-${{ version }}.tar.gz
+  # and otherwise fall back to archive:
+  # url: https://github.com/simplejson/simplejson/archive/v${{ version }}.tar.gz
+  sha256: d58439c548433adcda98e695be53e526ba940a4b9c44fb9a05d92cd495cdd47f
+  # sha256 is the preferred checksum -- you can get it for a file with:
+  #  `openssl sha256 <file name>`.
+  # You may need the openssl package, available on conda-forge:
+  #  `conda install openssl -c conda-forge``
```
h-vetinari, line 33: I think it'd be helpful to suggest a variant that doesn't persist the file
```suggestion
  # sha256 is the preferred checksum -- you can get it for a file with:
  #  `openssl sha256 <file name>`,
  # or, in one go without keeping the file with:
  # `curl -L <url> | sha256
  # You may need the packages or openssl and/or curl, available on conda-forge:
  #  `conda install openssl curl -c conda-forge``
```
However, it should be noted that we're not currently [shipping](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Flinux-64%2Fopenssl-3.3.1-h4bc722e_2.conda) any binaries for `sha256` (or `sha256sum`) in `openssl`, and so whether those get picked up depends entirely what else is on your path. Do you want to open an issue on the openssl feedstock to ship those?

.github/workflows/correct_directory.yml
```diff
@@ -31,6 +31,7 @@ jobs:
                     '\nUnfortunately, the recipe was added directly in the `recipes` folder without its own subfolder.\n' +
                     'Please move the recipe file into a folder with the name of the package you want to submit.\n\n' +
                     'For example: if your recipe is currently under `recipes/<your_package>.yaml`, ' +
-                    'it should be moved to `recipes/<your_package>/meta.yaml`.\n' +
+                    'it should be moved to `recipes/<your_package>/meta.yaml` ' +
+                    'or, if you are using the new yaml based recipe format, `recipes/<your_package>/recipe.yaml`\n' +
```
h-vetinari, line 35: I think we need a better label than "new yaml based recipe format" (this partially applies to `example-new-recipe` as well), especially since the old one is already called `meta.yaml`. I know it's not "pure" yaml, but to a casual user this is just confusing. Why not something like "v2 of the recipe specification" or something along those lines?

.ci_support/build_all.py
```diff
@@ -207,6 +242,50 @@ def build_folders(recipes_dir, folders, arch, channel_urls):
...
+    recipes_dir: str, platform, arch, channel_urls: list[str]
+):
+    config = get_config(arch, channel_urls)
+
+    # Remove the example recipes to ensure that they are not also build.
+    for example_recipe in EXAMPLE_RECIPE_FOLDERS:
+        rmtree(os.path.join(recipes_dir, example_recipe), ignore_errors=True)
+
+    # Determine the locations for the variant config files.
+    specs = OrderedDict()
```
h-vetinari, line 255: This is less a question for this PR and more a topic overall across our infrastructure, but at least I thought I'd note it here: https://github.com/conda-forge/conda-smithy/issues/2033

h-vetinari (commented):
Thanks for removing the formatting changes! Looks pretty good at first glance already. I left some comments/suggestions.

h-vetinari:
> Note that this was already reviewed in #27008 by both @jaimegp and @wolfv but happy to make more changes.

Yeah, fair enough, I don't want to complicate this process for you any further. Jaime would actually be a great person to ask for figuring out what (if anything) is going wrong with the artefact inspection. ;-)

.github/workflows/correct_directory.yml
baszalmstra: There has been a whiole discussion going on an we setteld on recipe v1. Ill change this.

```suggestion
                    'or, if you are using the recipe v1 format, `recipes/<your_package>/recipe.yaml`\n' +
```

baszalmstra:
The issue with `inspect-artifacts` should be fixed with https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/336

.ci_support/build_all.py
baszalmstra:
```suggestion
    specs = {}
```

recipes/jolt-physics/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+            -S $SRC_DIR/Build
+        cmake --build . --target install
+
+about:
+  homepage: https://github.com/jrouwe/JoltPhysics
+  license: MIT
+  license_file: LICENSE
+  summary: A multi core friendly rigid body physics and collision detection library.
+  description: A multi core friendly rigid body physics and collision detection library. Written in C++. Suitable for games and VR applications. Used by Horizon Forbidden West. 
+  documentation: https://xtensor.readthedocs.iohttps://jrouwe.github.io/JoltPhysics/
```
jaimergp, line 58:
```suggestion
  documentation: https://jrouwe.github.io/JoltPhysics/
```

recipes/example-new-recipe/recipe.yaml
```diff
@@ -0,0 +1,101 @@
...
+    - ${{ compiler('c') }}
+    - ${{ stdlib('c') }}  # If you need any compiler, add the C standard library ("stdlib") too
+  host:
+    - python
+    - pip
+  run:
+    - python
+
+tests:
+  # Some packages might need a `test/commands` key to check CLI.
```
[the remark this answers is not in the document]
wolfv: True, that comment looks wrong. It is also `script` not command in the test section

PR #27271, opened 2024-08-14
Move example to `example-v1`

### LINKED ISSUE #27264: put recipe.yaml in another folder
@isuruf would like the recipe.yaml and meta.yaml to be in different folders. We should make a pr or revert the old one to move the new recipes to another folder, maybe example-v1 or similar as before. Sorry for the back and forth here. 

Cc @nichmor

jaimergp:
Closes #27264 

Note the paths here should be synced with whatever is checked at https://github.com/conda-forge/conda-smithy/pull/2036, and only merged when that's merged and released in a new conda-smithy version to avoid more linter false positives.


beckermr (changes requested):
We should make the CI job fail if a comment is made so it shows as a failed check in the status.

beckermr:
I am going to merge this since it is unaffected by smithy and we an just use it until smithy is fixed.

PR #27274, opened 2024-08-14
ffbidx package for the fast feedback indexer library

hcstadler:
Checklist
- [x] Build number is 0.


recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+requirements:
+    build:
+        - {{ compiler('cuda') }}
+        - {{ compiler('cxx') }}
+        - {{ stdlib('c') }}
+        - cuda-cudart-dev {{ cuda_compiler_version }}
+        - conda-forge::pkg-config
+        - conda-forge::meson
+        - conda-forge::eigen  # header only library
```
ndevenish, line 27: You don't need `conda-forge::` here

hcstadler: Done

hcstadler:

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set version = "1.0.1" %}
+
+{% set cuda_major = (cuda_compiler_version|default("12.0")).rpartition('.')[0] %}
+
+package:
```
xhochy, line 5: Please indent with two spaces like we do in all other recipes.

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    skip: True  # [not linux64]
+    skip: True  # [(cuda_compiler_version or "None") == "None"]
+    skip: true  # [cuda_major != "12"]
+
+requirements:
+    build:
+        - {{ compiler('cuda') }}
+        - {{ compiler('cxx') }}
+        - {{ stdlib('c') }}
+        - cuda-cudart-dev {{ cuda_compiler_version }}
```
xhochy, line 24: This should be in host

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - pkg-config
+        - meson
+        - eigen  # header only library
+        - python
+        - numpy
```
xhochy, line 29: These four should also be in host

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+    host:
+        - python
+        - numpy
+    run:
+        - python
+        - {{ pin_compatible('numpy') }}
```
xhochy, line 35: This should no longer be used 
```suggestion
```

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+
+test:
```
xhochy, line 37:
```suggestion
test:
  requires:
    - pip
  commands:
    - pip check
```

recipes/ffbidx/build.sh
```diff
@@ -0,0 +1,9 @@
+PKG_CONFIG_PATH=${BUILD_PREFIX}/share/pkgconfig:${PKG_CONFIG_PATH}
```
xhochy, line 1: If you move the dependencies into host, you should no longer need this
```suggestion
```

recipes/ffbidx/build.sh
```diff
@@ -0,0 +1,9 @@
...
+CUDA_HOME=${BUILD_PREFIX} meson setup --reconfigure --buildtype=release --prefix=${PREFIX} --libdir=lib -Ddefault_library=shared -Dinclude-python-api=enabled meson
```
xhochy, line 2:
```suggestion
CUDA_HOME=${BUILD_PREFIX} meson setup --reconfigure --buildtype=release --prefix=${PREFIX} --libdir=lib -Ddefault_library=shared -Dinclude-python-api=enabled meson $MESON_ARGS
```

xhochy (changes requested):

recipes/ffbidx/meta.yaml
hcstadler: Done

recipes/ffbidx/meta.yaml
hcstadler: Done

recipes/ffbidx/meta.yaml
hcstadler: Done.

recipes/ffbidx/meta.yaml
hcstadler: Done

recipes/ffbidx/build.sh
hcstadler: Done

recipes/ffbidx/build.sh
hcstadler: Done

recipes/ffbidx/meta.yaml
hcstadler: Done

hcstadler:
This fails on cuda_118 (which I thought I had skipped) in "Packaging cbfmt" - something I do not use for my code.

recipes/ffbidx/meta.yaml
ndevenish: Are you sure about this? Don't compilers normally go in build? (because they don't need to be tied to the target architecture, and otherwise would prevent e.g. cross compilation).

Or did you just mean cuda-cudart-dev? (which I agree looks like it should be in host)

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  skip: True  # [(cuda_compiler_version or "None") == "None"]
+  skip: true  # [cuda_major != "12"]
+
+requirements:
+  host:
+    - {{ compiler('cuda') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - meson
+    - pkg-config
```
ndevenish, line 25: Usually I would expect these four should be in `build:`. But I don't know if meson needs that to workaround the environment.

ndevenish: Ah, meson was reviewed as needing to be in host, above, so maybe it does belong. It probably doesn't matter that much, if it doesn't export a run requirement.

hcstadler:
The linter fails if compiler('cxx') and stdlib('c') are not present in build.
cuda_120 fails if compiler('cuda') is not present in build.

hcstadler:

recipes/ffbidx/meta.yaml
ndevenish: I am pretty sure that this was a mobile-editing error, and that it was only intended to say that cuda-cudart-dev should be in host. So the compilers and `stdlib("c")` should be in build, not host.

ndevenish:
- I'm pretty sure compilers should definitely not be in `host:`, otherwise I don't think cross-compilation would work. I think above request to move them to host was using an incorrect line selection.
- pkg-config shouldn't _need_ to be in host, because that's just the CLI tool, and meson should still be able to pick up the other dependencies in host when the actual `pkg-config` executable is on the PATH outside the host path. I am not excluding the possibility that meson does something weird requiring it though, so (unless review disagrees) it's probably fine to have in host, along with meson (I don't think either declare an associated run requirement).
- The top two `cuda_home` lines in build.sh look as though they aren't doing anything?

recipes/ffbidx/meta.yaml
xhochy: It will matter for cross-compilation probably and with having meson here, you avoid having a python installation in build.

hcstadler:
@ndevenish: I agree with you on that.
The cuda_home lines in build.sh compute the cuda home path (for a typical installation) from the location of the nvcc binary, which in a typical installation is under bin/nvcc of the cuda home path. If I don't set CUDA_HOME, the meson build fails as far as I remember.
Much of my and mesons confusion stems from the fact that compiler('cuda') results in fundamentally different cuda installations for cuda_118 and cuda_120. I decided to only support cuda_120.

recipes/ffbidx/meta.yaml
xhochy: I think this got mixed up by the GitHub app. Compilers should only be in build.

dagewa:
This came up in our DIALS core meeting today. I'm a bit confused about the status, are we waiting for re-review by @xhochy?

xhochy:
This us waiting for the removal of the compilers from host

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - {{ stdlib('c') }}
+  host:
+    - {{ compiler('cuda') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - meson
+    - pkg-config
+    - eigen  # header only library
+    - python
+    - numpy
```
carterbox, line 30:
```suggestion
requirements:
  build:
    - {{ compiler('cuda') }}
    - {{ compiler('cxx') }}
    - {{ stdlib('c') }}
    - meson
    - pkg-config
  host:
    - eigen  # header only library
    - python
    - numpy
```

recipes/ffbidx/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  url: https://github.com/paulscherrerinstitute/fast-feedback-indexer/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: 99fdd8be09862ceba38436ffb1bf34c164d796e6f88f185b3185d4a606020172
+
+build:
+  number: 0
+  skip: True  # [not linux64]
+  skip: True  # [(cuda_compiler_version or "None") == "None"]
+  skip: True  # [cuda_compiler_version == "11.8"]
```
carterbox, line 15:
```suggestion
  skip: True  # [cuda_compiler_version in (None, "None", "11.8")]
```

recipes/ffbidx/build.sh
```diff
@@ -0,0 +1,10 @@
+cuda_home=$(which nvcc)
+cuda_home=${cuda_home%/bin/nvcc}
+CUDA_HOME=${cuda_home} meson setup $MESON_ARGS --reconfigure -Ddefault_library=shared -Dinclude-python-api=enabled meson
```
carterbox, line 3: Looks like you neeed to add some args to help meson find NumPy in the host prefix.

carterbox (changes requested):

xhochy (approved):

carterbox (changes requested):
If you look at the logs, you will see that the python versions in the build ($BUILD_PREFIX) and host ($PREFIX) environment are different and that meson detects the wrong one (the one in build instead of host). This is why I asked you to add some extra args to your build script to help meson find the python/numpy from the host prefix ($PREFIX).

hcstadler:
@carterbox: Thanks for the great review and suggestions. I added an option to select the python installation in my meson build and updated the recipe, so the python installation to build against is now set to the one installed under $PREFIX.

hcstadler:

xhochy:
Had to merge main here as the build was building a different recipe than the one in the PR.

xhochy (approved):

PR #27277, opened 2024-08-15
kwarray: add package

vicentebolea:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/kwarray/meta.yaml
+++ b/recipes/kwarray/meta.yaml
@@ -1,0 +1,45 @@
+{% set name = "kwarray" %}
+{% set version = "0.7.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/kwarray-{{ version }}.tar.gz
+  sha256: c7015c20330239ba48260ffab53713a8374b6ac2807e0a42f526c8577cd473f0
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
+  host:
+    - pip
+    - python >=3.8
+    - setuptools >=41.0.1
+  run:
+    - numpy
+    - packaging >=21.3
+    - python >=3.8
+    - ubelt >=1.2.3
+
+test:
+  imports:
+    - kwarray
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://gitlab.kitware.com/computer-vision/kwarray
+  summary: The kwarray module
+  license: Apache-2.0
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - Erotemic
+    - vicentebolea


vicentebolea:
@Erotemic I am adding you as maintainer, please answer in a comment so that the CI passes.

Erotemic:
Yes, I will be a maintainer.

vicentebolea:
@conda-forge/help-python, ready for review! CI passing.

recipes/kwarray/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+requirements:
+  host:
+    - pip
+    - python
+    - setuptools >=41.0.1
+  run:
+    - numpy >=1.19.3  # [py<310 and py>=38]
+    - numpy >=1.21.6  # [py<311 and py>=310]
+    - numpy >=1.23.2  # [py<312 and py>=311]
+    - numpy >=1.26.0  # [py>=312]
```
ocefpaf, line 26: While this is in the metadata I'm not sure this is correct. This is a noarch package, right? Or are compiled extensions? I pretty much doubt that API differences would be caught all over the code to handle those different versions of numpy for different versions of Python.

Erotemic: This is explicit, desired, and important. These are automatically determined by using the `xdev` package, and specifically the call `xdev availpkg numpy` to find the lowest possible version of a package (with a binary distribution for targeted operating systems) for a particular Python version (Currently cp313t seems to break the code, but `xdev availpkg scipy` works).

We want this because in some deployment cases, consumers of this packages are stuck on older versions of Python and older versions of numpy, which this package is perfectly compatible with. If we chose a single numpy for all supported Python versions that would prevent our users from having a painless install.

Our CI runs 2 variants of tests: "loose" and "strict". In "loose" pip is free to choose the latest version of the package that still meets the requirement, although we do include the `--prefer-binary` flag to ensure we don't try to compile a binary package that doesn't have a wheel built for it. In "strict" our setup.py script will replace the ">=" with "==" and force it to use minimum versions of all specified dependencies. This ensures that our test suite covers usage in the oldest and most recent version of the dependencies (and also clearly flags when new package releases are breaking us, which happens a lot).

vicentebolea: @Erotemic thanks for pointing this out. I question, is for instance numpy 1.19 incompatible with python 3.12? If it is not would it make sense to pin the lower version of numpy compatible with kwarray regardless of the python version?

Erotemic: Yes, I just fixed the issue with the new cp313t version, and here is what the output looks like:

image

This table shows a breakdown of which platforms and python versions have a binary wheel available for specific versions of numpy. You can see that 3.10+ lack wheels for 1.19.3.

vicentebolea: Thanks for pointing this out. Indeed for securing maximum this versioning scheme is the best way to deal with this, otherwise we will have to limit this package to python>=3.10 I am not sure how much value is in the extended support of python versions for new packages in conda-forge.

ocefpaf: All that is handled automatically by conda-forge, making it noarch is preferred to avoid overloading the CIs. I really doubt you can get an incompatible Python/numpy combination.

vicentebolea:
@ocefpaf ready for review/merge; review comment addressed; CI passing

recipes/kwarray/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/kwarray-{{ version }}.tar.gz
+  sha256: c7015c20330239ba48260ffab53713a8374b6ac2807e0a42f526c8577cd473f0
+
+build:
+  skip: true  # [py<38]
```
xhochy, line 13:
```suggestion
  noarch: python 
```

recipes/kwarray/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - pip
+    - python
```
xhochy, line 20:
```suggestion
    - python >=3.8
```

recipes/kwarray/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - setuptools >=41.0.1
+  run:
+    - numpy
+    - packaging >=21.3
+    - python
```
xhochy, line 25:
```suggestion
    - python >=3.8
```

xhochy (changes requested):

vicentebolea:
@xhochy  ready for review/merge; review comment addressed; CI passing

recipes/kwarray/meta.yaml
ocefpaf: BTW, even pip has a solver nowadays. I'm pretty sure you no longer need that kind of pins there as well.

vicentebolea:
@ocefpaf ready for review/merge; review comment addressed; CI passing

PR #27348, opened 2024-08-22
Add Spherely

benbovy:
New PR using `meta.yml` instead of `recipe.yml`.

Checklist
- [x] Build number is 0.


benbovy:
@jorisvandenbossche would you agree being included as maintainer of this recipe?

jorisvandenbossche:
Yes I do!

benbovy:
@conda-forge/help-python-c, this is ready for review!

recipes/spherely/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - {{ stdlib('c') }}
+    - cmake
+    - make  # [linux]
+  host:
+    - python
+    - pip
+    - scikit-build-core
+    - pybind11 >=2.10.0
+    - s2geometry >=0.11.1
+    - s2geography >=0.1.2
```
xhochy, line 32: The libraries should get a `run_export` and added for conda-forge-pinning. Otherwise, this will not lead to a sustainable pinning.
```suggestion
    - s2geometry
    - s2geography
```

recipes/spherely/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - numpy
+  run:
+    - python
+    - {{ pin_compatible('s2geometry', max_pin='x.x') }}
+    - {{ pin_compatible('s2geography', max_pin='x.x') }}
```
xhochy, line 37:
```suggestion
```

xhochy (changes requested):

recipes/spherely/meta.yaml
jorisvandenbossche: For s2geograhy, we are the first package that depends on it, is that worth to use a global pinning for?

xhochy: Yes, because this will make sure the rebuilds happen via the bot.

benbovy: I was considering doing that afterwards for s2geometry anyway, since at least one other package (pys2index) depends on it.

jorisvandenbossche: But that requires you to manually start a migration? (or is starting the rebuilds also already automated?)

My main concern is the overhead in managing this on the short term, with s2geography and s2geometry being mostly developed by the same people, and so we can easily do PRs in the respective feedstocks when needed.

xhochy: The bot should open the respective PRs automatically. It will add 2-3h of delay compared to manually doing it but it ensures that the pinnings are always compatible.

recipes/spherely/meta.yaml
benbovy: Hmm shouldn't there still be those run dependencies?

```yaml
  - s2geometry
  - s2geography
```

CI is failing..

xhochy: Once they have run_exports and the respective builds have hit the CDN, then they should be automatically as run dependencies.

recipes/spherely/meta.yaml
jorisvandenbossche: > The bot should open the respective PRs automatically. It will add 2-3h of delay compared to manually doing it but it ensures that the pinnings are always compatible.

Sounds good!

recipes/spherely/meta.yaml
benbovy: OK thanks. I opened https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6379. Not sure if that is the right way to do it, though.

> Once they have run_exports and the respective builds have hit the CDN, then they should be automatically as run dependencies.

How does this work in staged recipes? Do we need to re-run the CI manually at some point in this PR?

benbovy:
It should be good now. Both s2geometry and s2geography are correctly added as run dependencies via run_exports. I think https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/6379 is ready too.

benbovy:
@xhochy is it to good to merge? Otherwise I'm happy to make other changes.

PR #27351, opened 2024-08-22
Add DataLabSimpleClient recipe

PierreRaybaut:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/datalabsimpleclient/meta.yaml
+++ b/recipes/datalabsimpleclient/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "DataLabSimpleClient" %}
+{% set version = "0.10.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/c/cdlclient/cdlclient-{{ version }}.tar.gz
+  sha256: 794fc97dc7784f04b369928912efbd6f4f2ed5b544ef18ff2814066716a71ae0
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  entry_points:
+    - cdlclient-tests = cdlclient.tests:run
+  number: 0
+
+requirements:
+  host:
+    - python >=3.8
+    - setuptools
+    - pip
+  run:
+    - python >=3.8
+    - numpy >=1.21
+    - guidata >=3.5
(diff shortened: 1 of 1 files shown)


PierreRaybaut:
@conda-forge/help-python, the [DataLabSimpleClient](https://github.com/DataLab-Platform/DataLabSimpleClient) recipe is ready for review! (thx!)

recipes/datalabsimpleclient/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "DataLabSimpleClient" %}
```
ocefpaf, line 1: Why are you renaming the PyPI package? It should be `cdlclient`, no?

PierreRaybaut: That is a good question indeed. Actually this is a copy of the [DataLab recipe](https://github.com/conda-forge/datalab-feedstock/blob/main/recipe/meta.yaml). DataLab's Python package was named `cdl` because `datalab` was already used. Similarly, DataLab Simple Client package was named `cdlclient`. As the name "DataLab" was available in Conda, it was an opportunity to name properly the package (as it was done before on Debian for example).

Now, this package does not need to be as visible as DataLab itself because it is a utility package that will always be installed following instructions in DataLab's documentation. So, I'm really open to change the name back to `cdlclient` if needed.

ocefpaf: > DataLab's Python package was named cdl because datalab was already used.

You mean used in PyPI, right? While using the "proper" name in conda-forge is fine you should be aware that you may confuse users that are coming from PyPI to conda-forge. Ideally I would recommend to publish in conda-forge using multiple outputs and both names. Similar to what we do in https://github.com/conda-forge/importlib_metadata-feedstock/blob/main/recipe/meta.yaml

PierreRaybaut: OK, thanks for the tip (I see now that it's well documented).
I've just committed changes for creating an alias (CI is currently working on it).

ocefpaf: > I see now that it's well documented

That is a first. We are always behind our docs 😳

PierreRaybaut: Thanks for your review and for the merge. 👍

PR #27363, opened 2024-08-23
Add antio

mscheltienne:
Checklist
- [x] Build number is 0.


mscheltienne:
@larsoner Would you agree to also be listed as maintainer on this recipe?

Also, I'm seeing the following warnings in the linux build:

```
WARNING: Setting build arch. This is only useful when pretending to be on another arch, such as for rendering necessary dependencies on a non-native arch. I trust that you know what you're doing.
WARNING: Setting build platform. This is only useful when pretending to be on another platform, such as for rendering necessary dependencies on a non-native platform. I trust that you know what you're doing.
```

I wonder if I need something like:

```
requirements:
  build:
    - python                                 # [build_platform != target_platform]
    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
```

Last point, knowing that the `tests` and data files are not packaged,do you know what would be a good approach to test the recipe?

recipes/antio/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  license_file:
+    - LICENSE
+    - src/libeep/LICENSE
+    - src/libeep/LICENSE.addendum
+  dev_url: https://github.com/mscheltienne/antio
+
+extra:
+  recipe-maintainers:
+    - mscheltienne
+    - larsoner
```
larsoner, line 60: I agree to be a maintainer

larsoner:
> @larsoner Would you agree to also be listed as maintainer on this recipe?

Yep!

> Last point, knowing that the tests and data files are not packaged in the sdist, do you know what would be a good approach to test the recipe?

You can do for example

https://github.com/conda-forge/mne-qt-browser-feedstock/blob/520312adb50af1487d0f1833a92084a429a1697c/recipe/meta.yaml#L43-L45

But it would be great if you could call some method that ensures that the compiled module can actually be accessed. Maybe at some point add to the C code a little function that just returns a string of the libeep library version or something if it doesn't exist already?

larsoner:
> I wonder if I need something like: [cross-python]

Yes you almost certainly do, e.g.

https://github.com/conda-forge/openmeeg-feedstock/blob/6d33a431476df68c8d0dbd49e1ca41c3a8989466/recipe/meta.yaml#L22-L23

mscheltienne:
@conda-forge/help-python-c Ready for review!

mscheltienne:
I was not expecting to break it.. turns out `#include <pybuffer.h>` is missing in those build environment.. https://github.com/python/cpython/blob/main/Include/pybuffer.h

mscheltienne:
@larsoner Any idea? I don't get that one as the source for the `python` recipe includes this header..

xhochy (approved):

PR #27468, opened 2024-09-02
Adding magic-cta-pipe

aleberti:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/magic-cta-pipe/meta.yaml
+++ b/recipes/magic-cta-pipe/meta.yaml
@@ -40,7 +40,7 @@ requirements:
     - matplotlib-base >=3.0,<4.dev0
     - numba
     - numpy
-    - protozfits>=2.5,<3.dev0
+    - protozfits >=2.5,<3.dev0
     - scipy <1.12,>=1.8
     - scikit-learn >=1.2,<2.dev0
     - pytables
--- a/recipes/magic-cta-pipe/meta.yaml
+++ b/recipes/magic-cta-pipe/meta.yaml
@@ -40,7 +40,7 @@ requirements:
     - matplotlib-base >=3.0,<4.dev0
     - numba
     - numpy
-    - protobuf >=3.20.0,<3.21.dev0
+    - protozfits>=2.5,<3.dev0
     - scipy <1.12,>=1.8
     - scikit-learn >=1.2,<2.dev0
     - pytables
--- a/recipes/magic-cta-pipe/meta.yaml
+++ b/recipes/magic-cta-pipe/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "magic-cta-pipe" %}
-{% set version = "0.5.0" %}
+{% set version = "0.5.1" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


jsitarek:
> @jsitarek @Elisa-Visentin I added you as maintainer of magic-cta-pipe in conda-forge. Could you confirm it is fine with you by commenting this PR?

yes, thanks!

recipes/magic-cta-pipe/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.9
+    - setuptools >=64
```
jsitarek, line 20: in MCP there is no lower limit, just setuptools<=71 in environment (and nothing in setup file)

Elisa-Visentin:
Fine for me, thanks

aleberti:
I have to create a new version of the package, since some dependencies are creating problems. I will update the recipe after that.

aleberti:
@conda-forge/help-python the build is failing on win_64, for the following:

```bash
  - nothing provides requested protozfits >=2.5,<3.dev0
  - nothing provides protozfits >=2.2,<3.dev0 needed by ctapipe_io_lst-0.22.3-pyhd8ed1ab_0
```

but for example the recipe https://github.com/conda-forge/lstchain-feedstock/blob/main/recipe/meta.yaml has the same requirement and it seems it does not have problems. Also, I see that `protozfits` latest version in conda-forge is 2.5.1 (https://anaconda.org/conda-forge/protozfits). Do you have suggestions?

Elisa-Visentin:
@aleberti as far as I know protozfits does not exist for Windows... Maybe using selectors to exclude it for win_64? But
- probably you could build it, but not use it due to the missing package
- ctapipe_io_lst requires it, too

I just checked ctapipe_io_lst package (and also the one you posted above) and they only have the linux entry in the azure_pipelines... Maybe this is a solution?

aleberti:
> @aleberti as far as I know protozfits does not exist for Windows... Maybe using selectors to exclude it for win_64? But
> 
> * probably you could build it, but not use it due to the missing package
> * ctapipe_io_lst requires it, too
> 
> I just checked ctapipe_io_lst package (and also the one you posted above) and they only have the linux entry in the azure_pipelines... Maybe this is a solution?

Yes I know it is not available on windows, but for some reason I though that for lstchain the conda package was available for windows, but it actually is not.

I just checked the PR when `lstchain` was added (https://github.com/conda-forge/staged-recipes/pull/23429) by Max, and also there it fails on windows. Quoting Max:

> The windows failure is expected, although this is a no-arch package, we have a dependency on a python/c++binding package not available on windows

I cannot access the details of the failure since the build is ~1 year old, but from Max says, it is because of protozfits. Being so, I would simply ask for review now.

aleberti:

The windows failure is expected, although this is a no-arch package, we have a dependency on a python/c++binding package not available on windows. In case we need to do anything on the recipe, let us know!

PR #27476, opened 2024-09-03
Add `libblst`

MementoRC:
Checklist
- [x] Build number is 0.


MementoRC:
@conda-forge/help-c-cpp Hello, ready for review

recipes/libblst/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+
+
+  - name: libblst-{{ so_major_version }}
+    files:
+      - lib/libblst.so.{{ so_major_version }}  # [linux]
+      - lib/libblst.so.{{ version }}  # [linux]
+      - lib/libblst.{{ so_major_version }}.dylib  # [osx]
+      - lib/libblst.{{ version }}.dylib  # [osx]
+      - Library\\bin\\blst-{{ so_major_version }}.dll  # [win]
+      - Library\\lib\\blst-{{ so_major_version }}.lib  # [win]
```
xhochy, line 90: This is only needed at compile time and can/should be omitted from the package.
```suggestion
```

recipes/libblst/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+    test:
+      commands:
+        - test -f ${PREFIX}/include/blst.h  # [unix]
+        - test -f ${PREFIX}/include/blst_aux.h  # [unix]
+        - if not exist %LIBRARY_INC%\\blst.h exit 1  # [win]
+        - if not exist %LIBRARY_INC%\\blst_aux.h exit 1  # [win]
+
+
+
+  - name: libblst-{{ so_major_version }}
```
xhochy, line 83: To be sure: They are co-installable?

recipes/libblst/meta.yaml
```diff
@@ -0,0 +1,129 @@
...
+    - {{ compiler('c') }}  # [not win]
+    - {{ compiler('m2w64_c') }}  # [win]
+    - {{ stdlib('c') }}  # [not win]
+    - {{ stdlib('m2w64_c') }}  # [win]
+    - libxcrypt  # [unix]
+    - llvm-tools  # [win]
+    - sed  # [unix]
+
+outputs:
+  - name: libblst
```
xhochy, line 32: Let's call it `-devel` as that is the common name we use for splitted libraries.
```suggestion
  - name: libblst-devel
```

xhochy (changes requested):

recipes/libblst/meta.yaml
MementoRC: Just to be sure, blst-[{ so_major_version }} is the name of the 'primary' library. You're saying that the package should work well with just `blst.lib` being a copy of and describing `blst-0.lib/dll`

recipes/libblst/meta.yaml
MementoRC: Yes? not sure what you mean/ask, I usually follow that pattern that was taught to me when I make my first c-lib. Also in child packages I try to only install the {{ so }} lib.
Though, do teach  a better way ;P

recipes/libblst/meta.yaml
MementoRC: So the -{{so }} is the linkable library, when I install it, I would expect to be able to link with -lblst-{{ so }},  and the compiler would try to find the associated blst-{{ so }}.lib, or will it also find the blst-{{ so }}.dll directly? My uderstanding was that windows needs both, but I know little about windows as my numerous commits can attest

MementoRC: Committed for now

xhochy: You need both at build time, but only the DLL is needed at runtime.

recipes/libblst/meta.yaml
xhochy: We only use this pattern here if you can install and use `libblst-0` and `libblst-1` at the same time. Otherwise, we would name the package simply `libblst` to avoid accidentally having two incompatible versions in the same process.

MementoRC: Oh ... I did not understand that. I really complicate my life for no reason, it was quite hard bending this package (and others to implement the `major_version` scheme!

At this point, do you want me to remove all of it? or would you take pity and let it go? I have 2 packages that rely upon this scheme in the pipeline, meaning depending upon `libblst-0` - Well, I think I know the answer anyway given the potential burden of maintainability on others, I'll get going with the update

The other recipes are `c-kzg-4844` and `blst-py`

MementoRC: So, kidding aside, I think it may be useful to maintain the library version. The author mentioned that the lib was not originally intended for "public usage", so it may evolve in ways that may not maintain backward compatibility.

step21: @xhochy Could you comment whether this reasoning works for you or if you will only accept this without version number? (I'm interested in it as it is necessary for other package updates)

xhochy: Make it without a version number, otherwise, you would need to do some work on top to ensure that `libblst0` and `libblst1` are not co-installed (which is tricky).

MementoRC: It'd be great if you had a bit of time to teach me the principles, or point me to a few recipes. Is the issue that the `0` and `1` may be linked to incompatible external dynamic libraries? I could imagine `ckzg` needing `0` and `blst-py` `1` and having them specifically linked to the appropriate version in their recipe

I'd just like to learn more, if it's just a matter of more work, I did quite a bit already and rather do more than have the many weeks invested be all for naught

In any case, I imagine your time is not that elastic, so I'll get going (with heavy heart) on the refactoring
Thx

MementoRC:
@xhochy I think that it is ready for review. I kept the `ouputs` since I feel it is more reliable to clearly know what is in the package with the `files` syntax. Since I kept that format, I also defined the headers separate

step21:
🥳 thank you @MementoRC for all your work on this

PR #27486, opened 2024-09-04
Add build123d

ndevenish:
Checklist
- [x] Build number is 0.

### Notes on Implementation
- Patch changes upstream dependency to [lib3mf-python](https://pypi.org/project/lib3mf/) instead of [py-lib3mf](https://pypi.org/project/py-lib3mf/). Both these contain the same wrapper implementation (built by [lib3mf](https://github.com/3MFConsortium/lib3mf/blob/master/Autogenerated/Bindings/Python/Lib3MF.py)), and both are present on pypi, except the former is the "official" bundling.
- Noarch, but has a build script to prepare the parts that would be handled by setuptools_scm (which cannot be used as no repository present). Windows build fails without a `build.bat`, but I believe this is fine for noarch?

### DIFF
--- a/recipes/build123d/build.sh
+++ b/recipes/build123d/build.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+export SETUPTOOLS_SCM_PRETEND_VERSION="$PKG_VERSION"
+echo "version = \"$PKG_VERSION\"" >src/build123d/_version.py
+"$PYTHON" -m pip install . -vv
--- a/recipes/build123d/build.sh
+++ b/recipes/build123d/build.sh
@@ -2,4 +2,4 @@
 
 export SETUPTOOLS_SCM_PRETEND_VERSION="$PKG_VERSION"
 echo "version = \"$PKG_VERSION\"" >src/build123d/_version.py
-"$PYTHON" -m pip install . -vv
+"$PYTHON" -m pip install . -vv --no-deps --no-build-isolation 
--- a/recipes/build123d/0001-Use-lib3mf-python.patch
+++ b/recipes/build123d/0001-Use-lib3mf-python.patch
@@ -1,7 +1,7 @@
 From cd9c30ab1d9abc7cba28aeb2fd313382898d004a Mon Sep 17 00:00:00 2001
 From: Nicholas Devenish <ndevenish@gmail.com>
 Date: Wed, 4 Sep 2024 14:15:36 +0100
-Subject: [PATCH] Use lib3mf-python
+Subject: [PATCH 1/2] Use lib3mf-python
 
 ---
  src/build123d/mesher.py | 2 +-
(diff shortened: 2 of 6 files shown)


ndevenish:

ndevenish:

recipes/build123d/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export SETUPTOOLS_SCM_PRETEND_VERSION="$PKG_VERSION"
+echo "version = \"$PKG_VERSION\"" >src/build123d/_version.py
+"$PYTHON" -m pip install . -vv
```
ocefpaf, line 5:
```suggestion
"$PYTHON" -m pip install . -vv --no-deps --no-build-isolation 
```

ocefpaf:
> Windows build fails without a build.bat, but I believe this is fine for noarch?

Ideally we would like to know that it can be build and imported on Windows. But you are correct, it is not necessary for noarch recipes, as long as the resulting noarch package is not broken on Windows.

ndevenish:
> Ideally we would like to know that it can be build and imported on Windows. But you are correct, it is not necessary for noarch recipes, as long as the resulting noarch package is not broken on Windows.

Because of the setuptools_scm parts, I can write this to work but it'll be a script to duplicate work, that will be tested here but won't be tested on the feedstock because (IIRC) that will only build the noarch on linux. I'm prepared to explicitly test windows 

(There might be a more canonical way to work around setuptools_scm but I believe that because the package doesn't read it's own package metadata that writing this version file is the minimal implementation to make it work)

recipes/build123d/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+export SETUPTOOLS_SCM_PRETEND_VERSION="$PKG_VERSION"
```
ocefpaf, line 3: Instead of this file can you try using:

```yaml
  script: SETUPTOOLS_SCM_PRETEND_VERSION={{ version }} {{ PYTHON }} -m pip install . -vv
``` 

in the meta.yaml file `build` section?

ndevenish: You can actually use [build: script_env:](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#use-environment-variables) and an earlier revision did that, and that worked, but the upstream [build123d/version.py](https://github.com/gumyr/build123d/blob/235260a8000fa9d5174df0efea100d7cff332ec6/src/build123d/version.py) doesn't single-source from package metadata and expects it's own hard-copy, which is why I needed the other script line to write the version file.

I could patch that but was trying to steer away from just patching all my problems away.

ndevenish: That said I'm happy to change if I erred on the wrong side of caution there

ocefpaf: > I could patch that but was trying to steer away from just patching all my problems away.

No need to patch but you can open an issue upstream to avoid the hard-copy of that file. It kinds of defeats the purpose.

ndevenish: Yep, good idea, I have a couple of things to upstream, including allowing either lib3mf. Thanks!

PR #27497, opened 2024-09-05
Add recipe for NRSS

pbeaucage:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/NRSS/meta.yaml
+++ b/recipes/NRSS/meta.yaml
@@ -37,8 +37,8 @@ test:
     - pip
 
 about:
-  home: https://github.com/pdudenas/NRSS
-  license: BSD-3-Clause
+  home: https://github.com/usnistgov/NRSS
+  license: NIST-pd
   license_file: LICENSE
 
 extra:
--- a/recipes/NRSS/meta.yaml
+++ b/recipes/NRSS/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "NRSS" %}
+{% set version = "0.0.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/usnistgov/NRSS/archive/{{ version }}.tar.gz
+  sha256: 5b63f2d7b7331874df74f0eb2340e5541cee339116e82880427f51c4f8ed1562
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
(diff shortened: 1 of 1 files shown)


pbeaucage:

recipes/NRSS/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/usnistgov/NRSS/archive/{{ version }}.tar.gz
+  sha256: c56e8450ab938af516ea4e16389f47b78120df6b2fae0587b694f0e794b4e049
+
+build:
+  skip: True  # [not linux64]
```
ocefpaf: Looks like this one can be noarch.

recipes/NRSS/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - scipy
+    - scikit-image
+    - pandas
+    - matplotlib-base
+    - h5py
+    - xarray
+    - cyrsoxs
+test:
+  requires:
+    - python {{ python_min }}
```
ocefpaf, line 37:
```suggestion
    - python {{ python_min }}
    - pip
```

recipes/NRSS/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  imports:
+    - NRSS
```
ocefpaf, line 40:
```suggestion
  imports:
    - NRSS
  commands:
    - pip check
```

recipes/NRSS/meta.yaml
pbeaucage: thanks!  I switched it back over to noarch.  We have a dependency on `cyrsoxs` which is a Linux/CUDA package, so Mac and Windows builds will fail.  This could change at some point in the future though.

pbeaucage:
CI failure on linux comes from pip check - we aren't picking up nodejs which is an upstream dependency (pyhyperscattering).  I'm the feedstock maintainer for that so looking into it now.

pbeaucage:
@conda-forge-admin, please pre-render

pbeaucage:

pbeaucage:
@ocefpaf, thanks for your help with this!  That was a long swim upstream but we are now building and testing successfully on Linux.  Mac and Windows fail (expected) because one of this package's dependencies is Linux/CUDA only.  Do I need to do anything in the `meta.yaml` to address this, or can it be merged as-is with the failures?

ocefpaf:
> Do I need to do anything in the `meta.yaml` to address this, or can it be merged as-is with the failures?

This is fine. If those dependencies are ever available on mac and Windows, this recipe is ready for it and no change is needed.

PR #27535, opened 2024-09-10
Adding libpyvinyl and mcstasscript

tkittel:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/libpyvinyl/meta.yaml
+++ b/recipes/libpyvinyl/meta.yaml
@@ -13,7 +13,7 @@ source:
 build:
   noarch: python
   number: 0
-  script: {{ PYTHON }} -m pip install src --no-deps -vv
+  script: {{ PYTHON }} -m pip install ./src --no-deps -vv
 
 requirements:
   host:
--- a/recipes/libpyvinyl/meta.yaml
+++ b/recipes/libpyvinyl/meta.yaml
@@ -8,12 +8,11 @@ source:
   url: https://pypi.io/packages/source/l/libpyvinyl/libpyvinyl-{{ version }}.tar.gz
   #NB: get sha256 from https://pypi.org/project/libpyvinyl/#files
   sha256: 5031eacda47519734adf0cea21c684bb368659f08f8df48a698b56fb14c837e3
-  folder: src
 
 build:
   noarch: python
   number: 0
-  script: cd ./src/ && {{ PYTHON }} -m pip install . --no-deps -vv
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   host:
@@ -34,9 +33,9 @@ test:
     - libpyvinyl
   commands:
     - pip check
-    - pytest src
+    - pytest .
   source_files:
-    - src/
+    - ./
   requires:
     - pytest
     - pip
--- a/recipes/libpyvinyl/meta.yaml
+++ b/recipes/libpyvinyl/meta.yaml
@@ -36,7 +36,7 @@ test:
     - libpyvinyl
   commands:
     - pip check
-    - pytest .  # [unix]
+    - pytest .
   source_files:
     - ./
   requires:
--- a/recipes/libpyvinyl/meta.yaml
+++ b/recipes/libpyvinyl/meta.yaml
@@ -13,7 +13,7 @@ source:
 build:
   noarch: python
   number: 0
-  script: cd ./src/ && {{ PYTHON }} -m pip install ./src --no-deps -vv
+  script: cd ./src/ && {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   host:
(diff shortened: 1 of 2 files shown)


mads-bertelsen:
I confirm that I want to be added as recipe-maintainer!

tkittel:

recipes/libpyvinyl/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+  name: libpyvinyl
+  version: {{ version }}
+
+source:
+  url: https://github.com/PaNOSC-ViNYL/libpyvinyl/archive/v{{ version }}.tar.gz
+  sha256: 1e83f89b3b702d6b527bba6a2efd64b4c76cad23ad8a8c305ad356381bcd4e93
+
+  #url: https://pypi.io/packages/source/l/libpyvinyl/libpyvinyl-{{ version }}.tar.gz
+  ##NB: get sha256 from https://pypi.org/project/libpyvinyl/#files
+  #sha256: 5031eacda47519734adf0cea21c684bb368659f08f8df48a698b56fb14c837e3
```
ocefpaf, line 13: Why not the PyPI tarball? Also, can you remove these commented out lines?

recipes/libpyvinyl/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
ocefpaf, line 18:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation 
```

recipes/libpyvinyl/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - jsons
+    - h5py
+    - json_tricks
+
+test:
+  imports:
+    - libpyvinyl
+  commands:
+    - pip check
+    - pytest .  # [unix]
```
ocefpaf, line 39: This is noarch, you don't need this selector. However, if there is a significant failure on Windows we need to know so we don't ship a broken package there. Noarch packages are built on linux and ship to all platforms. 

```suggestion
    - pytest .
```

recipes/mcstasscript/meta.yaml
```diff
@@ -0,0 +1,63 @@
+{% set version = "0.0.70" %}
+
+package:
+  name: mcstasscript
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/m/mcstasscript/mcstasscript-{{ version }}.tar.gz
+  #NB: get sha256 from https://pypi.org/project/mcstasscript/#files
```
ocefpaf, line 9:
```suggestion
```

ocefpaf: PS: The bot will take care of that for you in the feedstock.

recipes/mcstasscript/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  sha256: b5d287e97cb76ced81755e7263e58f419edd9f658b16cfe2a8185546b15b47cd
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
ocefpaf, line 15: /
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation 
```

recipes/libpyvinyl/meta.yaml
tkittel: Because of an issue in the upstream project (we did try the pypi source first), meaning that the pypi source dists are broken. This has been reported upstream, but in the meantime the github tarballs are needed as a workaround.

recipes/libpyvinyl/meta.yaml
tkittel: Yes, this is just to make the windows builds pass since a few of the pytests were giving false positives there (essentially the upstream project needs to fix their unit tests to also run on windows, a few tests were using wrong path separators). But the package does work on windows.

tkittel:
@ocefpaf thanks a lot for the review! I basically just accepted all of your suggestions, and removed one extra comment like the one you wanted cleaned up.

The windows build is now failing, since I removed the selector you requested. But the package actually DOES work on windows, the issue is merely that the upstream project has a bug in their tests which gives a false positive failure on windows (some hardcoded path separator instead of using pathlib, but just in the testing code). Another upstream bug is the reason for using github tar-balls, as I also explained. Both issues have been reported upstream, as has the issue of a pinned dependency.

But in any case, I hope you are happy to accept this for conda-forge now? The upstream libpyvinyl issues are not specific to the conda-forge recipe, do not break the conda-forge package, and will be followed up upon by @mads-bertelsen .

ocefpaf:
> upstream project has a bug in their tests which gives a false positive failure on windows (some hardcoded path separator instead of using pathlib

That is not a blocker for is to merge here. Thanks for checking!

PR #27549, opened 2024-09-11
NEW: Add libcudss recipe

carterbox:
Checklist
- [x] Build number is 0.


carterbox:
@adibbley, could you please take a look at this recipe? I'm my first time writing a recipe for a pre-compiled CUDA library.

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,155 @@
...
+build:
+  number: 0
+  skip: true  # [cuda_compiler_version in (None, "None", "11.8")]
+  skip: true  # [not (linux64 or aarch64 or win64)]
+  script:
+    - cp -rv include $PREFIX  # [linux]
+    - cp -rv lib $PREFIX  # [linux]
+    - cp -rv include %PREFIX%  # [win]
+    - cp -rv lib %PREFIX%  # [win]
+    - cp -rv bin %PREFIX%  # [win]
```
adibbley, line 33: For Windows we've been using the `LIBRARY_LIB` and `LIBRARY_BIN` locations rather than `PREFIX`
```suggestion
    - cp -rv lib %LIBRARY_LIB%  # [win]
    - cp -rv bin %LIBRARY_BIN%  # [win]
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,155 @@
...
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcudss.so.{{ lib_version }}  # [linux]
+        - test -f $PREFIX/lib/libcudss_commlayer_nccl.so.{{ lib_version }}  # [linux]
+        - test -f $PREFIX/lib/libcudss_commlayer_openmpi.so.{{ lib_version }}  # [linux]
+        - test -L $PREFIX/lib/libcudss_commlayer_nccl.so.0  # [linux]
+        - test -L $PREFIX/lib/libcudss_commlayer_openmpi.so.0  # [linux]
+        - test -L $PREFIX/lib/libcudss.so.0  # [linux]
+        - test ! -f $PREFIX/lib/libcudss_static.a  # [linux]
+        - if not exist %PREFIX%\\bin\\cudss64_0.dll exit 1  # [win]
```
adibbley, line 70:
```suggestion
        - if not exist %LIBRARY_BIN%\\cudss64_0.dll exit 1  # [win]
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,155 @@
...
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -f $PREFIX/include/cudss.h
+        - test -f $PREFIX/include/cudss_distributed_interface.h
+        - test -L $PREFIX/lib/libcudss_commlayer_nccl.so  # [linux]
+        - test -L $PREFIX/lib/libcudss_commlayer_openmpi.so  # [linux]
+        - test -L $PREFIX/lib/libcudss.so  # [linux]
+        - test ! -f $PREFIX/lib/libcudss_static.a  # [linux]
+        - if not exist %PREFIX%\\lib\\cudss.lib exit 1  # [win]
```
adibbley, line 115:
```suggestion
        - if not exist %LIBRARY_LIB%\\cudss.lib exit 1  # [win]
        - if not exist %LIBRARY_INC%\\cudss.h exit 1  # [win]
        - if not exist %LIBRARY_INC%\\cudss_distributed_interface.h exit 1  # [win]
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,155 @@
...
+  sha256: ac0979610d3eda9e6ce803380587118c6ec72b7dc580e6e96392ac065010196f  # [win64   and (cuda_compiler_version or "").startswith("12")]
+
+build:
+  number: 0
+  skip: true  # [cuda_compiler_version in (None, "None", "11.8")]
+  skip: true  # [not (linux64 or aarch64 or win64)]
+  script:
+    - cp -rv include $PREFIX  # [linux]
+    - cp -rv lib $PREFIX  # [linux]
+    - cp -rv include %PREFIX%  # [win]
```
adibbley, line 31:
```suggestion
    - cp -rv include %LIBRARY_INC%  # [win]
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,155 @@
...
+      host:
+        - {{ pin_subpackage("libcudss", exact=True) }}
+      run:
+        - {{ pin_subpackage("libcudss", exact=True) }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -f $PREFIX/include/cudss.h
+        - test -f $PREFIX/include/cudss_distributed_interface.h
```
adibbley, line 110:
```suggestion
        - test -f $PREFIX/include/cudss.h  # [linux]
        - test -f $PREFIX/include/cudss_distributed_interface.h  #[linux]
```

adibbley (commented):
Thanks, Daniel! Overall looks good to me, just some comments on Windows locations.

cc @billysuh7 for viz

carterbox:
Thanks, @adibbley for the for those suggestions!

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
+        - nccl
+        - openmpi
```
carterbox, line 58: I also want some feedback from @kvoronin just to check that the dependencies are correct, and that they aren't any missing deps. cublas, for example.

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+    - cp -rv bin %LIBRARY_BIN%  # [win]
+
+outputs:
+
+  - name: libcudss
+    build:
+      run_exports:
+        # Breaking changes every version until 1.0
+        - {{ pin_subpackage("libcudss", max_pin="x.x.x") }}
+    files:
```
kvoronin, line 42: Do we (usually) ship cmake config/version files via conda? Or only in libcudss-dev?
What about include/?

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - lib/libcudss_commlayer_nccl.so.*  # [linux]
+      - lib/libcudss_commlayer_openmpi.so.*  # [linux]
+      - lib/libcudss.so.*  # [linux]
+      - bin/cudss64_0.dll  # [win64]
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
kvoronin, line 53: If we use `arm-variant`  and it all works, can we include the jetson build into the conda distribution? We currently have a special build for jetson/Orin for cudss (starting with cudss 0.3.0)

recipes/libcudss/meta.yaml
kvoronin, line 58: nccl and openmpi are only required if customers want to use MGMN mode of cudss. I don't like imposing this requirement on customers who will only use cudss on a single GPU (and never enable MGMN mode). Or WDYT, is having an extra dependency for them is fine? Like there is no nccl for Windows, will it be a problem for using conda package of cudss on Windows?

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+      license: LicenseRef-NVIDIA-End-User-License-Agreement
+      license_file: LICENSE
+      license_url: https://docs.nvidia.com/cuda/cudss/license.html
+      summary: >-
+        NVIDIA cuDSS is an optimized, first-generation
+        GPU-accelerated Direct Sparse Solver library for solving linear systems
+        with very sparse matrices.
+      description: >-
+        NVIDIA cuDSS is an optimized, first-generation
+        GPU-accelerated Direct Sparse Solver library for solving linear systems
```
kvoronin, line 82: A reminder for myself to clean the wording here, I believe it can be improved.

kvoronin:
Will we have a `run_test.<sh|bat>` like we do for other  CUDA math libraries?

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+    GPU-accelerated Direct Sparse Solver library for solving linear systems
+    with very sparse matrices. Direct Sparse Solvers are an important part
+    of numerical computing for real-time applications like autonomous
+    driving and process simulation, where increasing complexity and high
+    throughput demands a robust direct solver.
+  doc_url: https://docs.nvidia.com/cuda/cudss/
+
+extra:
+  recipe-maintainers:
+    - conda-forge/cuda
```
kvoronin, line 156: I also wonder if it makes sense to get one of the cudss developers (me) into the maintainers list.

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+    build:
+      run_exports:
+        # Breaking changes every version until 1.0
+        - {{ pin_subpackage("libcudss", max_pin="x.x.x") }}
+    files:
+      - lib/libcudss.so.*  # [linux]
+      - lib/libcudss_commlayer_nccl.so.*  # [linux]
+      - lib/libcudss_commlayer_openmpi.so.*  # [linux]
+      - lib/libcudss.so.*  # [linux]
+      - bin/cudss64_0.dll  # [win64]
```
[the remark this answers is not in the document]
carterbox: The Windows platform requires the import library at compile time in order to create dynamic links to the shared runtime library (`*.dll`), so without it, no one will be able to link to this package.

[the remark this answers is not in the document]
carterbox: Import libraries are not required at runtime which is why it is being shipped in the `-dev` package.

recipes/libcudss/meta.yaml
carterbox: Good catch! We need to guard `nccl` with a platform selector. We can move the files which link to openmpi to a separate output, so users don't get them by default. Probably we will try to have a nompi build variant in order to match the rest of the channel.

recipes/libcudss/meta.yaml
carterbox:
```suggestion
    - conda-forge/cuda
    - kvoronin
```

recipes/libcudss/meta.yaml
carterbox: I think we can, but I'm not sure if Conda can select for the user the tegra variant automatically yet. @jakirkham would know more about that.

recipes/libcudss/meta.yaml
carterbox: This recipe creates two outputs: `libcudss`, and `libcudss-dev`. The non-dev package contains only artifacts needed at runtime. The dev package contains only things needed at compile time.

We typically ship everything via conda except for documentation, tests, and examples because end users are mostly not developers and people use the internet for documentation and examples.

carterbox:
> Will we have a `run_test.<sh|bat>` like we do for other CUDA math libraries?

There are no GPUs on the build nodes available to run tests. Did you have something besides existence that you wanted to test? We can do a test build with CMAKE to check that the config is valid.

kvoronin:
> > Will we have a `run_test.<sh|bat>` like we do for other CUDA math libraries?
> 
> There are no GPUs on the build nodes available to run tests. Did you have something besides existence that you wanted to test? We can do a test build with CMAKE to check that the config is valid.

I am thinking about a test like for cusparselt:https://github.com/conda-forge/cusparselt-feedstock/blob/main/recipe/run_test.sh or similar. I know there was a general question raised as to how we should write such simple tests across the CUDA math libraries. So maybe there is a different idea. @jakirkham should know more about this.

recipes/libcudss/meta.yaml
carterbox: There is no support for tegra variant yet, but this pattern is a template that should allow it in the future.

recipes/libcudss/meta.yaml
carterbox: @kvoronin, do downstream packages which link against the nccl or openmpi comm layers need to also link against the same nccl/openmpi directly (they are public linkages)? Or are these private linkages which are abstracted away by the cudss API?

Either way, I need to know what versions of openmpi and nccl you link against, so that the recipe can have minimum versions of nccl and openmpi.

recipes/libcudss/test/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+cmake_minimum_required(VERSION 3.12)
+project(all_headers VERSION 0.1 LANGUAGES C CXX CUDA)
+
+find_package(cudss)
+
+set(ALL_TARGETS
+    cudss
```
carterbox, line 7: @kvoronin, I added a compiler test, but it fails because `libcudss_static.a` is missing (as per channel policy we don't ship static libraries). This needs to be fixed upstream. cudss should allow cmake to find the cudss package without finding the static library (the static library should be an optional component).

recipes/libcudss/meta.yaml
carterbox: I was looking at the headers because of making the compile test and it seems that the API does not mention nccl and openmpi, so whether to use either of these libraries is decided at runtime?

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+  home: https://developer.nvidia.com/cudss
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_file: LICENSE
+  license_url: https://docs.nvidia.com/cuda/cudss/license.html
+  summary: >-
+    NVIDIA cuDSS is an optimized, first-generation
+    GPU-accelerated Direct Sparse Solver library for solving linear systems
+    with very sparse matrices.
+  description: >-
+    NVIDIA cuDSS is an optimized, first-generation
```
kvoronin, line 186: Suggestion is to replace this description with:
```
    NVIDIA cuDSS is an optimized, first-generation
    GPU-accelerated Direct Sparse Solver library for solving linear systems
    with sparse matrices. Direct Sparse Solvers are an important part
    of numerical computing as they provide a general robust way of solving large
    linear systems without and are capable of taking advantage of both high compute
    throughput and memory bandwidth of the GPUs.
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+
+about:
+  home: https://developer.nvidia.com/cudss
+  license: LicenseRef-NVIDIA-End-User-License-Agreement
+  license_file: LICENSE
+  license_url: https://docs.nvidia.com/cuda/cudss/license.html
+  summary: >-
+    NVIDIA cuDSS is an optimized, first-generation
+    GPU-accelerated Direct Sparse Solver library for solving linear systems
+    with very sparse matrices.
```
kvoronin, line 184:
```suggestion
    with sparse matrices.
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libcudss_commlayer_nccl.so.{{ soversion }}  # [linux]
+        - test -L $PREFIX/lib/libcudss_commlayer_nccl.so.{{ somajor }}  # [linux]
+    about:
+      summary: >-
+        The NCCL communications layer module for NVIDIA cuDSS
+
+  - name: libcudss-commlayer-mpi
```
kvoronin, line 140: I've learned recently about conda having an mpi lock ( = single mpi per environment). Will `libcudss-commlayer-mpi` be takling advantage of that? E.g., uf the user first installs an incompatible MPI, will we be able to report smth if the user attempt later to install `libcudss-commlayer-mpi`?

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+    NVIDIA cuDSS is an optimized, first-generation
+    GPU-accelerated Direct Sparse Solver library for solving linear systems
+    with very sparse matrices.
+  description: >-
+    NVIDIA cuDSS is an optimized, first-generation
+    GPU-accelerated Direct Sparse Solver library for solving linear systems
+    with very sparse matrices. Direct Sparse Solvers are an important part
+    of numerical computing for real-time applications like autonomous
+    driving and process simulation, where increasing complexity and high
+    throughput demands a robust direct solver.
```
carterbox, line 191:
```suggestion
  description: >-
    NVIDIA cuDSS is an optimized, first-generation
    GPU-accelerated Direct Sparse Solver library for solving linear systems
    with sparse matrices. Direct Sparse Solvers are an important part
    of numerical computing as they provide a general robust way of solving large
    linear systems without and are capable of taking advantage of both high compute
    throughput and memory bandwidth of the GPUs.
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
+        # FIXME: Get actual version ranges from kiril
+        - nccl >=2.0,<3  # [linux]
```
carterbox, line 129:
```suggestion
        - nccl  # collect channel pinnings
        - nccl >=2.0,<3
```

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ compiler('cxx') }}
+        - {{ compiler('cuda') }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
+        # FIXME: Get actual version ranges from kiril
```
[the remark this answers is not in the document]
carterbox: I need to know which minor version CUDSS is built against too. Otherwise, it is possible that CUDSS is compiled against a new NCCL before NCCL is released on conda-forge and then the NCCL binary will be missing a feature that CUDSS needs.

recipes/libcudss/meta.yaml
```diff
@@ -0,0 +1,198 @@
...
+        - if not exist %LIBRARY_BIN%\\cudss64_{{ somajor }}.dll exit 1  # [win]
+
+  - name: libcudss-dev
+    build:
+      run_exports:
+        - {{ pin_subpackage("libcudss" ~ somajor, max_pin="x.x.x") }}
+    files:
+      - lib/libcudss.so  # [linux]
+      - lib/cudss.lib  # [win]
+      - include/cudss*
```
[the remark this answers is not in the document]
kvoronin: I treat the src/ more like "auxiiliary" than as "examples" (because we do have pure examples which show how to use the library).
Do I get it right that you suggest to do one more package so that to get these src/ the user would do `conda install libcudss, libcudss-examples` or di you mean smth else?
We can expect a significant amount of users who would want to build their own communication layer so ideally I'd have it available "by default" if it is allowed by the common practices.

recipes/libcudss/meta.yaml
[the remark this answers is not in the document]
kvoronin: It was NCCL 2.21(.5). I think we don't use any features above the major version 2.0.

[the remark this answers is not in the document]
kvoronin: But i don't think we rely on any super-new features of NCCL (hopefully using an older version will not run into the fixed bugs since the last official release).

recipes/libcudss/meta.yaml
carterbox: The `libcudss-commlayer-mpi` package requires openmpi, so conda will switch to openmpi if another mpi is installed in the environment or it will raise an error because it is not able to find an mpi that is compatible with all the requested packages.

carterbox:
I believe the only thing remaining is the upstream patch for building without static libraries installed.

carterbox:
Waiting for next release of cudss in December 2024.

carterbox:

kvoronin (approved):
LGTM

adibbley (approved):

PR #27704, opened 2024-09-26
Add swmmanywhere

cheginit:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/swmmanywhere/meta.yaml
+++ b/recipes/swmmanywhere/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 18690d31bf59a41edc2caada4302c47f41e7146c8e779f2ff0864788bcda68dc
 
 build:
-  noarch: python
+  skip: true  # [osx]
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
 
--- a/recipes/swmmanywhere/meta.yaml
+++ b/recipes/swmmanywhere/meta.yaml
@@ -16,12 +16,12 @@ build:
 
 requirements:
   host:
-    - python >=3.10
+    - python
     - hatch-vcs
     - hatchling
     - pip
   run:
-    - python >=3.10
+    - python
     - cdsapi
     - cytoolz
     - geopandas >=1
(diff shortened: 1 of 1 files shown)


cheginit:
@dalonsoa and @barneydobson, please confirm that you are willing to be maintainers as well.

dalonsoa:
Yes, that's OK. 

To make sure I get this right, because I've no experience with `conda-forge`, this is just some sort of proxy for PyPI, right?

barneydobson:
Yes happy to

cheginit:
@conda-forge/help-python, this is ready for review, thanks!

recipes/swmmanywhere/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/swmmanywhere-{{ version }}.tar.gz
+  sha256: 18690d31bf59a41edc2caada4302c47f41e7146c8e779f2ff0864788bcda68dc
+
+build:
+  skip: true  # [osx or py<310]
```
xhochy, line 13: Why isn't this noarch?

cheginit: Unfortunately, the `pyswmm` on `conda-forge` is very old and is not maintained anymore. Its osx version cannot be installed due to version conflict for `libopenmp` which is dep of `swmm`. I opened an issue on `pyswmm`'s feedstock and asked if they can update it. I am not sure if they will since it hasn't been updated since 4 years ago.

cheginit: I just opened a new issue in their main repo.

ocefpaf: So it would result in a non-solvable installation, right?

It is OK for noarch package to fails with a missing dependency. We prefer noarch b/c it is faster to build and the package will be ready when the missing dependency is available on the missing platform.

cheginit: Ah, I see. Makes sense. I will remove it.

cheginit: @ocefpaf per your suggestion, I added noarch. Builds except for osx pass.

PR #17082, opened 2021-11-26
WIP: package torchaudio

h-vetinari:
Necessary for #14474, #17129

* [ ] figure out `RuntimeError: No such operator torchaudio::rnnt_loss`
* [ ] figure out segfaults on linux
* [ ] figure out test failures regarding libarosa
* [ ] package https://github.com/as-ideas/DeepPhonemizer
* [ ] Fix https://github.com/conda-forge/kaldi_io-feedstock/issues/1
* [ ] rebuild sox with patches mentioned [upstream](https://github.com/pytorch/audio/issues/1269#issuecomment-1188423233), see https://github.com/conda-forge/sox-feedstock/pull/27

PR #27715, opened 2024-09-28
add torchaudio

hadim:
Checklist
- [x] Build number is 0.


hadim:
Trying to revamp https://github.com/conda-forge/staged-recipes/pull/17082

hadim:

recipes/torchaudio/recipe.yaml
```diff
@@ -0,0 +1,167 @@
...
+  string: ${{ string_prefix }}py${{ python | version_to_buildstring }}h${{ hash }}_${{ build_number }}
+  variant:
+    use_keys:
+      # use cuda from the variant config, e.g. to build multiple CUDA variants
+      - cuda
+    # this will down-prioritize the cuda variant versus other variants of the package
+    down_prioritize_variant: ${{ 1 if cuda else 0 }}
+  dynamic_linking:
+    rpaths:
+      - lib/
```
hmaarrfk, line 31: This should no longer be needed

hadim:
@hmaarrfk LGTM for me, but please let me know if you think something is missing.

hmaarrfk:
Looks fine, but to be honest, I'm not sure on where we stand with rattler recipes. Are we accepting them yet? or just building out infrastructure for them?

h-vetinari (commented):
Thanks for picking this up again! Looks quite good, but I want to see how the tests are doing, at least a reasonable subset.

h-vetinari:
> Looks fine, but to be honest, I'm not sure on where we stand with rattler recipes. Are we accepting them yet?

Yes we are (as of https://github.com/conda-forge/conda-forge-ci-setup-feedstock/pull/337). There's [a few](https://github.com/search?q=org%3Aconda-forge%20path%3Arecipe%2Frecipe.yaml&type=code) already

recipes/torchaudio/recipe.yaml
```diff
@@ -0,0 +1,164 @@
...
+      source:
+        - test/
+        - examples/
+    script:
+      # remove example tests that are not importable out-of-tree
+      # - rm -rf test/torchaudio_unittest/example
+      # - CI=true pytest -v test/
+
+      # Note: Tests take too long to run and some are not importable out-of-tree
+      - echo "Hello World"
```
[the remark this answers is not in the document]
hadim: I enabled them, let's see what happens.

recipes/torchaudio/recipe.yaml
```diff
@@ -0,0 +1,164 @@
...
+  number: ${{ build_number}}
+  skip:
+    - win
+  string: ${{ string_prefix }}py${{ python | version_to_buildstring }}h${{ hash }}_${{ build_number }}
+  variant:
+    use_keys:
+      # use cuda from the variant config, e.g. to build multiple CUDA variants
+      - cuda
+    # this will down-prioritize the cuda variant versus other variants of the package
+    down_prioritize_variant: ${{ 1 if cuda else 0 }}
```
hadim, line 28: I don't think `cuda` is actually set, and the whole cuda logic is probably wrong here.

Revisit once https://github.com/conda-forge/xformers-feedstock/pull/33 has been merged and use it as a cuda template, maybe.

Tobias-Fischer:
This might help as a temporary workaround for the ffmpeg issues: https://github.com/pytorch/audio/pull/3837

hadim:
Thank you @Tobias-Fischer. What I don't get is why ffmpeg is not detected during the tests.

_I won't have the time to wrap this PR in the short term, so if anyone needs it, go ahead. Otherwise, I might come back to it at a later time._

recipes/torchaudio/recipe.yaml
```diff
@@ -0,0 +1,172 @@
...
+        - requests
+        - hypothesis
+        - inflect
+        - kaldi_io
+        - parameterized
+        - pysoundfile
+        - transformers
+        - unidecode
+        - inflect
+        - sox
```
Tobias-Fischer, line 154:
```suggestion
        - sox
        - sentencepiece
        - pytorch-lightning
```

Tobias-Fischer:
Any chance to give me write access to your fork @hadim?

recipes/torchaudio/build.sh
```diff
@@ -0,0 +1,44 @@
...
+  export USE_CUDA="false"
+  export BUILD_CUDA_CTC_DECODER="false"
+fi
+
+export BUILD_CPP_TEST="false"
+export BUILD_SOX="true"
+export BUILD_RIR="true"
+export BUILD_RNNT="true"
+export BUILD_ALIGN="true"
+export USE_FFMPEG="true"
```
Tobias-Fischer, line 37:
```suggestion
export FFMPEG_ROOT="${PREFIX}"
```

Tobias-Fischer: This will fix the ffmpeg issues - otherwise it is being vendoered.

hadim:
> Any chance to give me write access to your fork @hadim?

Should be good now.

Tobias-Fischer:
@hmaarrfk @h-vetinari - do you think it would be acceptable if we disable ffmpeg integration? I saw it’s  disabled in the torchvision feedstock. Same for sox which seems to cause issues.

hmaarrfk:
Yes. The torchvision deeper integration is stuck on ffmpeg 4 and was always broken IMO. 

I don't really know how well they tested it upstream.

h-vetinari:
Yeah, fine by me to disable. Sox had some other issues (there's some patching involved that upstream explicitly ruled out IIRC), some references can be found in my old PR and on the sox feedstock

hadim:
Well done @Tobias-Fischer. It looks like it's working.

The tests take about one hour per variant. Are you sure you want to take much time for building the package? I would maybe suggest downsizing the number of tests to be run. That would remove some pressure on the c-f CI but also reduce the testing loop of the feedstock itself.

But it's really up to you, and it's just a suggestion!

_(also, do you mind removing me from the maintainer list, as I don't think I'll have the bandwidth to maintain it in the future?)_

hmaarrfk:
I'm not sure if this kind of test skip strategy is preferred to you:
https://github.com/conda-forge/scikit-image-feedstock/blob/main/recipe/meta.yaml#L73

Tobias-Fischer:
@hadim: Thanks for the feedback! The stress on CIs should not be too bad once it's a feedstock, as separate python versions will be built in separate CI jobs. That said, I'll temporarily use just a single python version from my next commit onward to reduce the stress here.

@hmaarrfk: Yes, the current solution is ugly. I don't know yet how we can do something like the strategy you linked to with the new recipe.yaml. @wolfv - do you have a suggestion?

hmaarrfk:
Oh. I guess just maybe write one test per line in the test script?

that way we can read and understand each skip line by line.

recipes/torchaudio/run_tests.sh
```diff
@@ -25,4 +25,37 @@ export TORCHAUDIO_TEST_ALLOW_SKIP_IF_NO_RIR="true"
...
+# test_masking_iid
+# test_mvdr_0_ref_channel
+# test_rtf_mvdr
+# test_souden_mvdr
+
+# Segfault on CI (probably due to limited memory):
+# test_pitch_shift_shape_2
+# test_paper_configuration
+
+pytest -v test/torchaudio_unittest/ -k "not TestAutogradLfilterCPU and not test_deemphasis and not rnnt and not ray_tracing and not test_create_mel and not test_forced_align and not hubert_large and not hubert_xlarge and not wavlm_large and not test_masking_iid and not test_mvdr_0_ref_channel and not test_rtf_mvdr and not test_souden_mvdr and not test_pitch_shift_shape_2 and not test_paper_configuration"
```
hmaarrfk, line 61:
```suggestion
# Output 0 of UnbindBackward0 is a view and is being modified inplace. 
# This view is the output of a function that returns multiple views. 
# Such functions do not allow the output views to be modified inplace. 
# You should replace the inplace operation by an out-of-place one.
tests_to_skip="TestAutogradLfilterCPU"
tests_to_skip="test_deemphasis or ${tests_to_skip}"

# 'torchaudio' object has no attribute 'rnnt_loss'
tests_to_skip="rnnt or ${tests_to_skip}"
# 'torchaudio' object has no attribute 'ray_tracing'
tests_to_skip="ray_tracing or ${tests_to_skip}"
# ValueError: invalid version number '0.10.2.post1'
tests_to_skip="test_create_mel or ${tests_to_skip}"

# RuntimeError: torchaudio.functional._alignment.forced_align Requires alignment extension, but TorchAudio is not compiled with it.         Please build TorchAudio with alignment support.
tests_to_skip="test_forced_align or ${tests_to_skip}"

# Very slow on CI:
tests_to_skip="hubert_large or ${tests_to_skip}"
tests_to_skip="hubert_xlarge or ${tests_to_skip}"
tests_to_skip="wavlm_large or ${tests_to_skip}"
tests_to_skip="test_masking_iid or ${tests_to_skip}"
tests_to_skip="test_mvdr_0_ref_channel or ${tests_to_skip}"
tests_to_skip="test_rtf_mvdr or ${tests_to_skip}"
tests_to_skip="test_souden_mvdr or ${tests_to_skip}"

# Segfault on CI (probably due to limited memory):
tests_to_skip="test_pitch_shift_shape_2 or ${tests_to_skip}"
tests_to_skip="test_paper_configuration or ${tests_to_skip}"
pytest -v test/torchaudio_unittest/ -k "not (${tests_to_skip})"
```

h-vetinari:
For test skips, I definitely prefer the approach used in skimage (and elsewhere), where each component of the `not (x or y or a)` string we're building gets a comment or reference about why that test us skipped

Tobias-Fischer:
> For test skips, I definitely prefer the approach used in skimage (and elsewhere), where each component of the `not (x or y or a)` string we're building gets a comment or reference about why that test us skipped

Isn’t that what we do now already? We group them into different failure cases.

Tobias-Fischer:
@h-vetinari @hmaarrfk would you like to become maintainers as well!

hmaarrfk:
no ;)

h-vetinari:
Given the work I did on this in https://github.com/conda-forge/staged-recipes/pull/17082, sure.

wolfv:
Yeah there are some secret ways that could make something like this possible ... I am not yet sure if I like it or not! :)

Note: this doesn't seem to work with the version of `minijinja` we are currently using, but I can try to fix that asap:


```yaml
context:
  tests: |
    ${{ 
      (
        (['test1', 'test2', 'test3'] if ppc64le else []) +
        ['test1', 'test2']
      ) | join(" ")
    }}
```

This would render to a (string) list. I am unsure about comments in Jinja expressions.

Otherwise one could also argue to write a more explicit "script" either in Python or in Bash that would give the list of tests to run. Downside: it's not in the recipe itself, upside it is a proper programming language.

wolfv:
If the tests run only on Unix you could also use something like:


```
script:
  - echo "foo" > tests.txt
  - if: ppc64le
    then:
      - echo "bar baz bing" >> tests.txt

```

wolfv:
Argh, sorry, acidentally slipped and closed the PR! Apologies 😞

wolfv:
And if you really want (although I am not sure we will keep this "feature") you can do:

```yaml
context:
  fuuu: |
    {% set bla = "xxx" %}
    ${{ bla | upper }}
```

h-vetinari:
> although I am not sure we will keep this "feature"

This kind of freedom would be very useful in some of the more complicated recipes that you've mostly descoped so far (AFAIU), because you care primarily about the bulk of simpler recipes for now. Is `{% ... %}` still "pure YAML"? I cannot find it in the spec.

In any case, I'd like to not lose the ability to do commented test skips (at least talking about `pytest -k`) within the recipe. Whether that works as before with some variables that get extended or whether the skips get saved into a `skips.xyz` file (or whatever) doesn't matter to me, but I want it to be obvious from the recipe that there are test skips, what they are exactly, and the reasoning for each skip in a comment.

wolfv:
@h-vetinari yes, it is pure YAML because it is contained in a multi-line text block. :)

Also when I tested, the setted variable did not escape "outside" so maybe this can be allowed.

Tobias-Fischer:
Does anyone have an idea why it can’t find nvcc for cuda 12? It looks in PREFIX instead of BUILD_PREFIX despite setting the suggested environment variable CUDACXX

recipes/torchaudio/recipe.yaml
```diff
@@ -0,0 +1,180 @@
...
+    env:
+      cuda_compiler_version: ${{ cuda_version | default('None') }}
+
+requirements:
+  build:
+    - if: build_platform != target_platform
+      then:
+        - python
+        - cross-python_${{ target_platform }}
+        - pytorch
```
hadim, line 47: I added pytorch in the build section during my early tests but I know some recipes only has it in the host section such as here: https://github.com/conda-forge/xformers-feedstock/blob/main/recipe/meta.yaml#L50

Maybe try to remove it?

Tobias-Fischer: All good - fixed it by passing through CMAKE_ARGS :)

Tobias-Fischer:
I think this is finally ready for review :)

@hadim: Could you please remove the Draft label from the PR?

Tobias-Fischer:
Note that cuda 11.8 failed with "No space left on device (os error 28)" which will not be an issue once it's converted to a feedstock with separate builds.

Tobias-Fischer:
Thanks @hmaarrfk for the review!

PR #27801, opened 2024-10-06
Add QEMU-user with execve: Requires sysroot 2.28

MementoRC:
Checklist
- [x] Build number is 0.


MementoRC:
@conda-forge/help-c-cpp Hello, Ready for review. This recipe needs __glibc 2.28 in order to pass the tests, it is set in the `conda_build_config.yaml` but does not seem to take effect in this recipe

MementoRC:
@h-vetinari Did you mean to comment here? The other PR is highly experimental trying to build a qemu-user-aarch64 on OSX, which seem to almost work, and incidentally on windows (but that is kind of a fight the devil type of masochism - I profoundly dislike windows)

h-vetinari:
> @h-vetinari Did you mean to comment here

Well you managed to get my attention with the "Fails due to missing __glibc 2.28 in tests" in the title. ;)

MementoRC:
That was not my intent. This is the same recipe, limited to `linux` because it is the only OS that supports user-mode. It is ready to go. It does fail due to:
```
  - nothing provides __glibc >=2.28,<3.0.a0 needed by qemu-user-execve-aarch64-9.1.0-ha1f11d0_0
```
I separated linux from the other recipe, because it is not clear to me yet that it will work (or be acceptable) for the other OSses,

So, in essence, your comment on the other PR is much more related to this recipe than the other one.

MementoRC:

MementoRC:
@conda-forge/help-c-cpp Hello, Ready for review.
@xhochy This is the QEMU build we need for the `zig-feedstock`

MementoRC:
@xhochy I decided to transfer to rattler-build since the conda-forge team seems busy, but then I encountered an odd issue that I would like @wolfv to take a look at if he has some time. I know I did some 'don't do that', in particular, I don't know how to enable the `experimental` mode of rattler-build in the recipe folder (I tried conda_build_config, and variants)
Better ask for forgiveness than permission ... that's why it's reverted to draft - Hopefully, we can get it resolve soon
Thanks

xhochy:
> experimental mode

This won't be available on conda-forge.

> I decided to transfer to rattler-build since the conda-forge team seems busy

I don't think this will help with review speed. The recipe here is very complex and it will always take a lot of time to review, that's why I haven't commented here recently.

MementoRC:
> 
> > I decided to transfer to rattler-build since the conda-forge team seems busy
> 
> I don't think this will help with review speed. The recipe here is very complex and it will always take a lot of time to review, that's why I haven't commented here recently.

No disrespect, I only meant that after my passing commit, I felt I would continue working on the recipe (and prepare work on the feedstock) as time seemed to be available to do so. As I did, I found flaws anyway.

Are you then recommending that I revert to a conda-build since 'experimental' will not be available - That would actually alleviate this puzzling issue with the test - I will start reverting. The recipe needed to be improved anyway to share resources with qemu-system and after that `qemu-user-arm64 on osx-64`

MementoRC:
@xhochy I reverted to the conda-build. Ideally, I wanted to test that I can execute the execve() call from a compiled code in the test section, but there seems to be a complex memory related limitation of the docker container, I tried several local changes to docker, but could not resolve it (I left it in comments, in case you have some ideas)

MementoRC:
@xhochy after a long debugging using sysroot 2.17, I was not able to find a workaround for the failed memory allocation when trying to execute a cross-compiled binary. It seems to stem from missing `MAP_FIXED_NOREPLACE` (that I patched with the 2.28 value) - Any suggestions are welcome
For now I will revert to 'need 2.28'

recipes/qemu-user-execve/patches/0001-intercept-qexecve.patch
```diff
@@ -0,0 +1,25 @@
+--- qemu/linux-user/main.c
```
xhochy, line 1: What is the source of these patches?

recipes/qemu-user-execve/conda_build_config.yaml
```diff
@@ -0,0 +1,20 @@
...
+  - s390x-conda-linux-gnu-gcc
+target_emulated_platform:
+  - aarch64
+  - ppc64le
+  - s390x
+zip_keys:
+  - - target_emulated_platform
+    - target_emulated_gcc
+
+docker_image:
```
xhochy, line 19: Do you still need this? conda-forge switched to using an Alma9 image as the default.

MementoRC: I thought I did, this recipe fails because it uses sysroot 2.17 that does not have the definition of MAP_FIXED_NOREPLACE in `mmap.h`. The value in the linux64.yaml seem to still be the cos7 (remember, I often don't really know what I am doing o.O)

MementoRC: Ok, sorry, I forgot to remove the logging I used locally, so the error does not show in the log. next commit

recipes/qemu-user-execve/patches/0001-intercept-qexecve.patch
xhochy: Yes, if you have not written them by yourself, they should have the initial authors attached to them. Still, if you had written them, they should reference you as the author.

MementoRC: Got it.

So, the purpose of this recipe was mostly to resolve the ZIG cross-build. This patch was never accepted by the QEMU dev team. They argued that proot and binfmt_misc already served that purpose, but I find that it is simpler to use qemu+execve to resolve the ZIG build - Just want to make sure you know the full context
I only proposed this recipe after you suggested that it could be useful

recipes/qemu-user-execve/patches/0001-intercept-qexecve.patch
```diff
@@ -0,0 +1,42 @@
+This patch is extracted from Joel Holdsworth's patch submission.
```
xhochy, line 1: Can you please extract the patches in the `git format-patch` form? You should be able to download them in mbox format and `git am` them.

MementoRC: This is new to me, I would need to research this a bit - The patches were not merged (or even PR??) but just submitted for discussion, I found them in a mail discussion

MementoRC: Would using proot a preferable solution to the problem I was trying to resolve?

MementoRC: I am working on a qemu-system recipe (no patch) in order to try it as a qemu-user for OSX (I think Wolf V. needs it). It has overlapping artifact that I try to make cleanly separated and complementary, if you have the inkling a feedback would be appreciated (https://github.com/conda-forge/staged-recipes/pull/28603). It builds on linux/osx (it did on windows, but since getting a windows OS in VM is a bit above my understanding, I skip it for now)

xhochy (approved):
🎉

Issue #27817, opened 2024-10-08
Error missing stdlib(c) registering new recipe with recipe.yaml


minrk:
### Bug:

The `registering outputs` step is [failing](https://github.com/conda-forge/admin-requests/actions/runs/11230135146/job/31216904557#step:6:3771) for scifem which uses `recipe.yaml` with missing `stdlib('c')`:

```
registering outputs...

 ╭─ Finding outputs from recipe
 │ Found 1 variants
 │ Build variant: scifem-0.2.5-h81b34b9_0
 │ 
 │ ╭─────────────────┬──────────╮
 │ │ Variant         ┆ Version  │
 │ ╞═════════════════╪══════════╡
 │ │ build_platform  ┆ linux-64 │
 │ │ target_platform ┆ linux-64 │
 │ ╰─────────────────┴──────────╯
 │
 ╰─────────────────── (took 0 seconds)
Error:   × Failed to parse recipe

Error:   × Parsing: failed to render Jinja expression: undefined value: No stdlib
  │ found for language: c
  │ You should add `c_stdlib` to your variant config file. (in <string>:1)
    ╭─[30:7]
 29 │     - ${{ compiler('c') }}
 30 │     - ${{ stdlib('c') }}
    ·       ─────────┬────────
    ·                ╰─┤ undefined value: No stdlib found for language: c
    ·                  │ You should add `c_stdlib` to your variant config file.
 31 │     - ${{ compiler('cxx') }}
    ╰────

Traceback (most recent call last):
  File "/home/runner/work/admin-requests/admin-requests/.github/workflows/scripts/create_feedstocks.py", line 461, in <module>
    _, pkg_names, _ = parse_package_and_feedstock_names(feedstock_dir, use_container=False)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/Miniforge3/lib/python3.12/site-packages/conda_forge_feedstock_ops/parse_package_and_feedstock_names.py", line 53, in parse_package_and_feedstock_names
    return _parse_package_and_feedstock_names_local(feedstock_dir)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #27818, opened 2024-10-08
switch scifem to meta.yaml

minrk:
v1 recipe merged as #27758 but didn't finish creating the feestock (#27817).

I may have been premature to try a v1 recipe with https://github.com/conda-forge/conda-forge.github.io/issues/2308 not finished yet.

This should unblock creation of a merged recipe

cc @xhochy as the approver of the original #27758

I don't think this is urgent, if folks would rather fix the v1 issue than merge this.


minrk:
scifem repo finished creating

PR #27839, opened 2024-10-10
add diffsitter, dioxus, dua-cli, earthlyls, eva

danielnachun:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/diffsitter/meta.yaml
+++ b/recipes/diffsitter/meta.yaml
@@ -1,0 +1,38 @@
+{% set name = "diffsitter" %}
+{% set version = "0.8.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/afnanenayet/diffsitter
+  git_rev: v{{ version }}
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('rust') }}
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - git-lfs
+    - cargo-bundle-licenses
+
+test:
+  commands:
+    - diffsitter --help
+
+about:
+  home: https://github.com/afnanenayet/diffsitter
+  license: MIT
+  license_file:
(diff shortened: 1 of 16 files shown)


danielnachun:
@conda-forge/staged-recipes @conda-forge/help-rust This is now ready for review.

recipes/diffsitter/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set name = "diffsitter" %}
+{% set version = "0.8.4" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/afnanenayet/diffsitter
+  git_rev: v{{ version }}
```
ocefpaf, line 10: Try to use tags and checksum when possible. One can modify a tag/release in place on GitHub and, without the checksum, we may never know that it changed.

```suggestion
  url: https://github.com/afnanenayet/diffsitter/archive/refs/tags/v{{ version }}.tar.gz
  sha256: 3440e9e1c4e2e359092ec1cc35812d53efe9e56535148af140bbb131c59b500c
```

danielnachun: I forgot to mention in my comments, but I tried this already before submitting and it won't work here because it needs to check out some git submodules.  They should really be providing a source tarball with those submodules in it, but unfortunately they don't.

ocefpaf: Ah, git-submodules, we meet again my old enemy!

That makes for an even worse scenario as parts of the source code are downloaded/activate later. Everything should be in a tarball. Can you add a comment on this above the git_url and open an issue upstream requesting a complete source distribution tarball?

danielnachun: It turns out `diffsitter` does actually have a separate source archive with each release that does include the checked out submodules which I had completely missed before.  I've updated the recipes to use this archive instead.

xhochy (approved):

PR #27853, opened 2024-10-12
add x13as

h-vetinari:


h-vetinari:
Getting compilation errors:
```
$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran -O2  -c -o adpdrg.o adpdrg.f
adpdrg.f:409:72:

  409 |      &               'after end of LS sequence (LSS)',T)
      |                                                                        1
Error: More actual than formal arguments in procedure call at (1)
```

h-vetinari:
No idea why `-lgfortran -L$PREFIX/lib` cannot find libfortran on osx; `libgfortran5` (which [contains](https://conda-metadata-app.streamlit.app/?q=conda-forge%2Fosx-64%2Flibgfortran5-13.2.0-h2873a65_3.conda&with_broken=true) the library, at that path) is definitely present in the host env...
```
ld: library not found for -lgfortran
```

h-vetinari:
Who would have thought that for a crusty fortran recipe, the first passing architecture would be windows?! 😂

Well, for a given value of "passing" at least... this fortran program does not seem to have a `--help` option to test basic execution, and the options that do exist aren't exactly self-explanatory (thanks fortran 😝):
```
(%PREFIX%) %SRC_DIR%>x13as_ascii --help 
Fortran STOP

 X-13ARIMA-SEATS Seasonal Adjustment Program
 Version Number 1.1 Build 61
 Execution began  Oct 12, 2024  11.55.40 

 ERROR: Program option --help not defined.  Valid program options are 
        -I, -O, -M, -D, -C, -S, -N, -V, -P, -R, -Q, -W, -G, -T.
```
If anyone has ideas for a simple test invocation, please let me know.

recipes/x13as/build.sh
```diff
@@ -0,0 +1,36 @@
...
+    export LDFLAGS="$LDFLAGS -L$CONDA_BUILD_SYSROOT/../lib"
+    # needs to explicitly link glibc & libm
+    export LDFLAGS="$LDFLAGS -L$CONDA_BUILD_SYSROOT/lib64 -lc -lm"
+    # also needs compiler runtime
+    export LDFLAGS="$LDFLAGS -lgcc_s"
+else
+    export LDFLAGS="$LDFLAGS -framework CoreFoundation"
+fi
+# both platforms need to link libgfortran
+export LDFLAGS="$LDFLAGS -lgfortran"
```
h-vetinari, line 25: @isuruf, I was wondering if you'd have an idea here why `libgfortran` isn't found on osx, resp. `libgcc_s` on linux. After a bit of necessary massaging, I'm using `FFLAGS` and `LDFLAGS` from the fortran compiler activation, but for some reason `$PREFIX/lib` doesn't seem to be picked up correctly.

It's a bit weird that this is all so manual, but I guess that's because there's no real build orchestrator at play here, and it's just a hand-written upstream makefile. Still, it's surprising that a fortran compiler needs help to find libgfortran. 🙄 

Note: in contrast to autotools or at least CMake, `LDFLAGS` isn't being passed to the compiler, but to the linker itself; hence we need to strip `-Wl,` etc.

It _could_ maybe have something to do with having to use an older gfortran...? Doesn't sound very plausible, but I'm a bit out of ideas. Unfortunately, anything newer than gfortran 10 (inclusive) fails to compile the sources...

h-vetinari:
Thanks for the help! Guess it was my own fault that I switched the `LINKER=` to the actual linker and not the compiler driver 🙈

h-vetinari:
> If anyone has ideas for a simple test invocation, please let me know.

Still don't know how to test the CLI properly, but at least the binaries don't return a non-zero exit code for unknown flags, and it shows that they can be called 🤷 

So this PR is now ready for review @conda-forge/help-c-cpp

xhochy (approved):

PR #27901, opened 2024-10-16
Add python packages markdown-word-count, pelican-fontawesome and pelican-myst-reader

anjos:
Checklist
- [x] Build number is 0.


recipes/markdown-word-count/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - setuptools
+  run:
+    - python >=3.6
+
+test:
+  imports:
+    - mwc
+  commands:
+    - pip check
+    #- mwc --help
```
zklaus, line 32: Why is this commented out?

zklaus (approved):
I think this is ready for review.

The failing Windows build is (iiuc) due to emojis in the README, which get's read in the `setup.py` to be added as the long_description. Because no encoding is specified in the call to `open`, the default encoding is used, which for Linux comes from the locale settings and ends up being the correct utf-8 in our build environment, but for Windows defaults to cp1251, which cannot deal with the utf-8 encoded emojis.

Since these noarch packages will only be built on Linux, that is not a problem.

@anjos, to make this more resilient, you may want to suggest to upstream to add the encoding to the `open` call, or, better yet, to switch to a modern pyproject.toml based setup.

recipes/markdown-word-count/meta.yaml
anjos: This app has no help message.  It requires a markdown file to provide a zero status output, however none is shipped with the package.  I could simulate one, but I think this would be beyond the purpose of the package?

anjos:
> @anjos, to make this more resilient, you may want to suggest to upstream to add the encoding to the `open` call, or, better yet, to switch to a modern pyproject.toml based setup.

Done, and cross-linked to this issue.

anjos:
@conda-forge/help-python ready for review!  Thanks in advance.

recipes/markdown-word-count/meta.yaml
anjos: Thanks for this review.  I implemented a simple test as requested on the last push.

anjos: The test does not work because the program sets the exit code `!= 0` always, despite the fact it works...  I'll check it later today.

anjos: This is now fixed, and ready for review @zklaus

anjos:
@zklaus: ready for final review.  All suggestions were taken onboard. Thanks a lot!

zklaus (approved):
LGTM

anjos:
After a second thought, I think we should not merge these here.  Some of these packages are simply not being maintained.

PR #27946, opened 2024-10-20
pypi.io -> PyPI.org in example recipe

Carreau:
The io domain has not been used from quite some time; so this will avoid (at least) one redirect.

Note that https://pypi.org/packages/*... redirects to https://files.pythonhosted.org/$1, so maybe we should avoid a second redirect an do that replacement as well ?

I know it's a single redirect and it's not much, but should this be auto-applied to all existing recipes as well ?

### DIFF
--- a/recipes/example-v1/recipe.yaml
+++ b/recipes/example-v1/recipe.yaml
@@ -20,7 +20,7 @@ package:
   version: ${{ version }}
 
 source:
-  url: https://pypi.io/packages/source/${{ name[0] }}/${{ name }}/${{ name }}-${{ version }}.tar.gz
+  url: https://pypi.org/packages/source/${{ name[0] }}/${{ name }}/${{ name }}-${{ version }}.tar.gz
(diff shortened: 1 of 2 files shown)


Carreau:
> If these changes are intentional (and you aren't submitting a recipe), please add a `maintenance` label to the PR.

Note that this message should likely be updated as external contributor can't add labels like this.

jaimergp:
> should this be auto-applied to all existing recipes as well ?

This can be added as a hint in the linter. Also, not sure if `grayskull` is still defaulting to `.io` sources.

jaimergp (approved):

Carreau:
Should we then do a full redirect and replace by `files.pythonhosted.org/<stuff>` ? 

Also any pointer as to where the linter rule would be ?

jaimergp:
I think we are fine with just .org. Given that the equivalent PR in conda-smithy was merged, I'll get this one in too, as is.

PR #28051, opened 2024-10-29
add jupyter-(docprovider|collaboration-ui) 1.0.0

bollwyvl:
Checklist
- [x] Build number is 0.

References:
- unblocks https://github.com/conda-forge/jupyter-collaboration-feedstock/pull/18

### DIFF
--- a/recipes/jupyter-collaboration-ui/meta.yaml
+++ b/recipes/jupyter-collaboration-ui/meta.yaml
@@ -1,0 +1,52 @@
+{% set version = "1.0.0" %}
+{% if 1 %}{% set jupyterlab = "jupyterlab >=4.2.0,<5" %}{% endif %}
+
+package:
+  name: jupyter-collaboration-ui
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/j/jupyter-collaboration-ui/jupyter_collaboration_ui-{{ version }}.tar.gz
+  sha256: 853c9498bcd1bdec4d4d9c53bf831b7e59be3935bd8f41d178ba4026e93f5a76
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation --disable-pip-version-check
+  number: 0
+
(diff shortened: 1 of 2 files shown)


bollwyvl:
@conda-forge/help-python @conda-forge/jupyter-collaboration ready for review

davidbrochart (approved):
Thanks.

recipes/jupyter-collaboration-ui/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set version = "1.0.0" %}
+{% if 1 %}{% set jupyterlab = "jupyterlab >=4.2.0,<5" %}{% endif %}
```
ocefpaf, line 2: Do we really need this? It will became a complication as we move to the new recipe format.

bollwyvl: Do I need the version so people don't get broken stuff? Yes. Does it need the `if`? No, but I'm tired of fixing bot dupes, and this seems to confuse it enough that it doesn't touch things it shouldn't.

If the new format is preferred, I can just use that instead... last time I excitedly tried, i was told it wasn't ready to be used (even though the example was in here)....

ocefpaf: We are in slowly transition. I would not merge this if it was another maintainer but I'm sure you will be active and update this when the new format rolls out... Still I recommend removing as much jinja as possible from your existing and future recipes.

bollwyvl: > removing as much jinja as possible

Oh, that's my goal when living that conda-forge life. I'd use native YAML anchors, but multi-outputs _hate_ that. But having made enough "forgot to update a python version" errors, DRY wins. I think the new format will improve the situation (even though I think it could have used more existing techniques like JSON-e, patch, etc. and don't like that it doesn't declare a `$schema` for completion) but it's a big net step forward, and the significant speed improvements are certainly welcome!

PR #28181, opened 2024-11-12
Add skelly_viewer

roaldarbol:
Checklist
- [x] Build number is 0.


roaldarbol:
Seems to be an issue with `xcb`/`libxcb`. I've tried a few things, but can't really figure out what needs to be added. The error seems to be this:
```
qt.qpa.xcb: could not connect to display 
qt.qpa.plugin: From 6.5.0, xcb-cursor0 or libxcb-cursor0 is needed to load the Qt xcb platform plugin.
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: xcb, vnc, minimal, eglfs, vkkhrdisplay, minimalegl, offscreen, linuxfb.
```

As far as I can tell `xcb-cursor0` or `libxcb-cursor0` should be provided by `xcb-util-cursor` or `libxcb` on CentOS. I tried adding `libxcb-cursor0`, but IO don't think that's the correct name for installing with `yum`. 

@conda-forge/help-python Does anyone know how to proceed from here?

danielnachun:
Try setting environment variable `QT_QPA_PLATFORM=minimal` in your test - this should hopefully avoid trying to spawn X session in a headless environment.  You might still need the trick I suggested in your [other PR](https://github.com/conda-forge/staged-recipes/pull/28177#discussion_r1859974945) if it still won't return 0.

roaldarbol:
> Try setting environment variable `QT_QPA_PLATFORM=minimal` in your test - this should hopefully avoid trying to spawn X session in a headless environment

Thanks, I'll try that! Do you reckon I should still try with the minimal amount of dependencies in `yum_requirements` or is there no harm in having too many?

danielnachun:
Yes, let's try not to add unnecessary `yum_requirements`.  It looks like you need OpenGL support but I think you can just add `xorg-libx11` as a Linux-only dependency instead of using `libX11-devel` as a `yum_requirement`.

I'm not totally sure about this, but I feel like recently I have gotten OpenGL packages to work with only `mesa-libGL` and maybe `mesa-dri-drivers`.  However all the `mesa` packages you have are probably fine, as is `libglvnd`.  I'm hoping we can get `libglvnd` into conda-forge at some point and just use that to resolve host OpenGL implementations at run time, but that isn't available yet.

roaldarbol:
I think we've also got the issue here that `skelly_viewer --help` doesn't actually exist, so it runs `init` which attempts to open a window and the build basically stalls.

recipes/skelly-viewer/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  run:
+    - python
+    - pytest
+    - numpy ==1.26.2
+    - rich ==13.7.0
+    - pyside6 >=6.7.2
+    - opencv ==4.8.*
+    - pandas ==2.1.*
+    - matplotlib-base ==3.8.2
+    - xorg-libx11  # [linux]
```
ocefpaf, line 34: This one can probably be OS-noarch. See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

roaldarbol: Ah, that makes sense! I'll try that.

roaldarbol: Ithe conda-forge-linting is now failing; it seems I can't make build it this way whilst in staged-recpies since it requires changes to `conda-forge.yml`. Is that right @ocefpaf?

roaldarbol: Seems all the builds work, it's just the linting that's failing.

ocefpaf: Please see step 2 in the instructions I sent you above.

roaldarbol: I have, and my understanding of that point is that you need to configure `noarch_platforms ` in `conda-forge.yml` - but I’m not supposed to do change anything outside the recipe, thus my confusion. I’m aware it’s surely my understanding that’s lacking which is why I’m asking for help. :-)

ocefpaf: You add that file in the recipe directory,

roaldarbol: Aaaah! That’s really not clear from the documentation, neither from the link you sent, [Configuring conda-forge.yml](https://conda-forge.org/docs/maintainer/conda_forge_yml/) or anywhere else I’ve been looking. I might open a PR for that. Thanks for the help! :-)

roaldarbol: @ocefpaf I've now added the `conda-forge.yml` file to the `recipes` directory, but am still failing the linter. Any more suggestions?

recipes/conda-forge.yml
```diff
@@ -0,0 +1,2 @@
+noarch_platforms:
+  - linux_64
```
ocefpaf, line 2: This file should be in the recipe directory.

roaldarbol: Like I did in https://github.com/conda-forge/staged-recipes/pull/28181/commits/72b92e6591f71981c2e0e58b56f7aafb4289a486 which also failed? Otherwise I'm not sure I know where you mean?

ocefpaf: Can you please re-do it so we can check?

roaldarbol: Done. 👍

PR #28193, opened 2024-11-13
Add recipe for apache-solr

antarcticrainforest:
Checklist
- [x] ~~If static libraries are linked in, the license of the static library is packaged.~~
- [x] Build number is 0.


antarcticrainforest:
@conda-forge/help-java , ready for review.

antarcticrainforest:
@conda-forge/help-java, ping team.

antarcticrainforest:

antarcticrainforest:
@conda-forge/help-java how can we proceed?

recipes/apache-solr/LICENSE.txt
```diff
@@ -0,0 +1,226 @@
+                                 Apache License
```
bollwyvl, line 1: If this _really_ isn't in the distribution (unlikely for a flagship Apache project), I recommend getting it (and the NOTICE) from:

```yaml
- url: https://github.com/apache/solr/blob/releases/solr/{{ version }}/LICENSE.txt
  sha256: ... # only have to get it once, the bot will update it in the future
- url: https://github.com/apache/solr/blob/releases/solr/{{ version }}/NOTICE.txt
  sha256: ... # only have to get it once, the bot will update it in the future
```

recipes/apache-solr/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+requirements:
+  run:
+    - openjdk >=11  # Solr requires Java to run
+
+test:
+  commands:
+    - solr start -e cloud -noprompt  # [unix]
+    - solr stop -all  # [unix]
+    - solr.cmd start -e cloud -noprompt  # [win]
```
bollwyvl, line 23: don't need selectors here: `sold.cmd` will be on `%PATH%`, so can just be `solr`

recipes/apache-solr/meta.yaml
```diff
@@ -0,0 +1,37 @@
+{% set name = "apache-solr" %}
```
bollwyvl, line 1: probably don't  need `name` variable, as it's only ever used once

recipes/apache-solr/LICENSE.txt
antarcticrainforest: Ok. I've added instructions to the recipe to download those files from the solr git repository.

recipes/apache-solr/meta.yaml
antarcticrainforest: True - fair point.

recipes/apache-solr/meta.yaml
```diff
@@ -0,0 +1,37 @@
...
+source:
+  - url: https://downloads.apache.org/solr/solr/{{ version }}/solr-{{ version }}.tgz
+    sha256: 38548b86fa4e3c87883875952da124bf7d742cb8f7b25d37a1176833588e8552
+  - url: https://raw.githubusercontent.com/apache/solr/refs/heads/main/LICENSE.txt
+    sha256: 7feb4036d88cfa5fae13a0d55e7638a700ef77d2b4e6fcebaacac889081a7923
+  - url: https://raw.githubusercontent.com/apache/solr/refs/heads/main/NOTICE.txt
+    sha256: 15b37dbc8a5b68547c4a9d93fc1e8a72609e2d31d59e5d6f57241cee988e073a
+build:
+  number: 0
+  noarch: generic
```
xhochy, line 15: You need noarch_variants (see the docs) for Windows and Unix. The generated package isn't the same.

xhochy (changes requested):

danielnachun:
This recipe probably doesn't need a separate `build.sh` and `bld.bat` recipes at all.  I'd suggest instead building this from the source tarball in the upstream with `gradle`: https://github.com/apache/solr.  It seems this repository already provides separate [bash](https://github.com/apache/solr/blob/main/solr/bin/solr) and [batch](https://github.com/apache/solr/blob/main/solr/bin/solr.cmd) wrappers that could simply be symlinked to `${PREFIX}/bin`.

It turns out conda is smart enough to understand how to set up proper symbolic links on Windows from a package built on linux, and convenient `${PREFIX}/bin` is included in the path on Windows when an environment is active.  This isn't well documented at the moment but I'm planning to submit extensive documentation about this soon.  In the mean time if you'd like me to provide you with a `build.sh` that should work here I'm happy to try to do so.

antarcticrainforest:
> This recipe probably doesn't need a separate `build.sh` and `bld.bat` recipes at all. I'd suggest instead building this from the source tarball in the upstream with `gradle`: https://github.com/apache/solr. It seems this repository already provides separate [bash](https://github.com/apache/solr/blob/main/solr/bin/solr) and [batch](https://github.com/apache/solr/blob/main/solr/bin/solr.cmd) wrappers that could simply be symlinked to `${PREFIX}/bin`.
> 
> It turns out conda is smart enough to understand how to set up proper symbolic links on Windows from a package built on linux, and convenient `${PREFIX}/bin` is included in the path on Windows when an environment is active. This isn't well documented at the moment but I'm planning to submit extensive documentation about this soon. In the mean time if you'd like me to provide you with a `build.sh` that should work here I'm happy to try to do so.

Thanks for suggestion, I'll adjust the `build.sh` file and delete the `bld.bat`. Just out of curiosity, why do you think we should build solr from scratch again. Wouldn't it simply be enough to extract the pre-build tar.gz and copy the files that need to be copied + creating symlinks?

antarcticrainforest:
@danielnachun how can I tell this thing to only call the build recipe on linux or at least *not* on windows?

recipes/apache-solr/meta.yaml
antarcticrainforest: I've tried to run the build stage only for linux, as that should be enough. But it seems I don't understand the skip directive.

danielnachun:
You don't need to skip Windows here, for `noarch: generic` packages it's okay for it to fail.  You don't need to use any selectors here, just use one test:
```
    - solr start -e cloud -noprompt
    - solr stop -all
```

danielnachun:
Also, I will try to provide you with commands to actually build the JAR from source with `gradle` and download that the third party licenses shortly!

danielnachun:
Some important things to note:
- The full snapshot is written to `${PREFIX}/libexec/apache-solr` rather than `${PREFIX}/share/solr` - there is not really a clear standard yet for where to put JDK/.NET packages, but since they often produce folder hierarchies which deviate significantly from the standard Unix system, `libexec` is a natural place for the output to live, and then we can symlink or create wrapper scripts for executables.
- The `generateLicenseReport` is essential to properly downloading the third-party licenses for the package which are saved to `build/reports/dependency-license/` as has now been added to the `meta.yaml`.  The standard practice of simply redistributing a pre-built JAR file without its third party licenses is not really compliant with the conda-forge policies around third-party licenses which requires them to be redistributed along side the main package.

danielnachun:
I'll also note here that we've been having a discussion about improving documentation for submitting recipes for specific language ecosystems, and I'm planning to provide more comprehensive instructions for JDK, .NET and NodeJS packages in the next few weeks.  These packages actually integrate quite well into conda-forge, but their build process is radically different from Python or compiled languages and it took me months of effort to figure out things like third-party license downloads, devendoring native binaries etc.  I hoping in the near future we'll be able to direct users to specific instructions for these languages that will get them going in the right direction.

antarcticrainforest:
I am not sure what to do. The tests are failing because the docker image running them doesn't come with ps. This is rather unusual for a "real" Linux system not running in a docker container.

```console
docker run -it quay.io/condaforge/linux-anvil-x86_64:alma9 which ps
which: no ps in (/opt/conda/bin:/opt/conda/condabin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/home/conda/bin)
```

I think, I could add `procps-ng` as runtime dependency for linux but I think in 99% of the cases this should not be necessary. Is there a way to switch the default docker image where the tests are run at? For example:

```console
docker run -it quay.io/condaforge/mambaforge which ps 
Unable to find image 'quay.io/condaforge/mambaforge:latest' locally
latest: Pulling from condaforge/mambaforge
d9802f032d67: Pull complete 
f45ad57e79fe: Pull complete 
a38f0a2db3b3: Pull complete 
Digest: sha256:24da8189ddb53ae7e635f5db914eaa8f7f02790f455a82c4b5fcae7f02063be7
Status: Downloaded newer image for quay.io/condaforge/mambaforge:latest
/usr/bin/ps
```

danielnachun:
That last fix worked!  @xhochy I think this is ready to go.

xhochy:
> I'll also note here that we've been having a discussion about improving documentation for submitting recipes for specific language ecosystems, and I'm planning to provide more comprehensive instructions for JDK, .NET and NodeJS packages in the next few weeks

Happy to dump knowledge too if we can settle on the right place for these template recipes!

xhochy (approved):

PR #28225, opened 2024-11-15
Add openpmix, prrte

minrk:
currently bundled as dependencies of openmpi

related to #27988 which adds libfabric, also currently bundled as part of multiple mpi providers.

Once these and libfabric are packaged and built on all platforms, we can add them as dependencies of openmpi.

Each library has 3 outputs:

- `libpmix`, `libprrte` have libraries, most files (config in etc, help/error messages in share)
- `pmix-bin`, `prrte` have executables in `bin/`
- `libpmix-devel`, `libprrte-devel` have headers. I don't normally separate these, but both packages install a disproportionate number of headers in e.g. `include/pmix/src`

openmpi will depend on prrte-bin and libpmix, but _not_ pmix-bin and not (directly) libprrte.
Currently, the openmpi package installs as if _all_ of these packages are present (bin, lib, devel).

FWIW, ubuntu has `libpmix`, `libpmix-bin`, and `libpmix-dev`. Ubuntu does not appear to package prrte separately from openmpi. Fedora has [pmix](https://packages.fedoraproject.org/pkgs/pmix/pmix/) (libs), `pmix-tools` (bin), and `pmix-devel`, and [prrte](https://packages.fedoraproject.org/pkgs/prrte/prrte/) (bin), `prrte-libs` (lib), and `prrte-devel`.

xref https://github.com/conda-forge/openmpi-feedstock/issues/178

cc @leofang @dalcinl @jakirkham @j34ni who have been participating in related discussions, if anyone wants to chime in or join as a maintainer


j34ni:
@minrk If there is no known use case to have the binaries, then I would suggest to have a package with only the libraries, and to reflect that explicitly in the name (like libpmix)

j34ni:
@minrk I cloned your repo and I am trying to fix the recipe to add missing dependencies but I get a lot errors when building locally using `python3 build-locally.py`  (about syntax with the `$` and `%` that should not be there, the `recipe.yaml` that should be called `meta.yaml`, something about CI not being defined, etc.): can I ask what you are using to build since you managed to make it go that far whereas when for me it fails straight away?

minrk:
@j34ni that's great, thank you! I've been trying out [rattler-build](https://prefix.dev/blog/rattler_build_on_conda_forge) for new recipes, so to build linux64, run:

```
rattler-build build -r recipes/openpmix -m .ci_support/linux64.yaml
```

 But I've discovered working on this one that it doesn't yet have sufficient support for split-output recipes (at least until https://github.com/conda-forge/conda-smithy/issues/2057 is addressed so we can use [this](https://prefix-dev.github.io/rattler-build/latest/multiple_output_cache/), so I'll switch back to conda-build if we want to separate lib from bin packages, which I think makes sense.

minrk:
ok, I have everything working and a local build of openmpi using external libfabric, pmix, and prrte on osx-arm64 from these recipes.

I split out the headers into `-devel` packages because for some reason both these packages put a huge amount into `include/pmix|prte/src` (roughly half the install size of pmix is headers). The `prte/src` don't appear to be used by anything, but prrte needs the `pmix/src` to build.

So there are now these packages:

openpmix:
- libpmix (lib, etc, help)
- libpmix-devel (headers, pkg-config)
- pmix-bin (bin, man)

prrte:
- libprrte (lib, etc, help)
- libprrte-devel (headers)
- prrte-bin (bin, man)

with the following dependencies:

- libpprte has:
  - host dep on libpmix-devel
  - run dep on libpmix (via run_exports)
- openmpi has
  - host dep on libpmix-devel, libprrte-devel, prrte-bin
  - run dep on libpmix (via run_exports), prrte-bin (_not_ libprrte)

One thing I have discovered is that perhaps it doesn't make sense to have split libprrte/prrte-bin, since the only thing that's ever going to use it is openmpi, and it requires prrte-bin, not libprte, and doesn't actually link the library. If we switch back to a single `prrte` package, I'd probably exclude the `include/prte/src` headers entirely, since they don't seem to be used by anything, unlike the ones in pmix.

minrk:
apparently `compute_build_graph.py` cannot handle multi-output recipes, because it successfully finds the dependency of libprrte on libpmix-devel, then proceeds to build libprrte before libpmix-devel:

```
/home/conda/staged-recipes-copy/recipes/prrte {'platform': 'linux', 'arch': '64', 'label': 'linux-64'} <conda_build.config.Config object at 0x7f0ded966d80> False
rendering /home/conda/staged-recipes-copy/recipes/prrte for linux-64
WARNING: Setting build arch. This is only useful when pretending to be on another arch, such as for rendering necessary dependencies on a non-native arch. I trust that you know what you're doing.
WARNING: Setting build platform. This is only useful when pretending to be on another platform, such as for rendering necessary dependencies on a non-native platform. I trust that you know what you're doing.
WARNING:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:Dependency libprrte, version any is not installable from your channels: ['/home/conda/staged-recipes/build_artifacts', 'conda-forge'] with subdir linux-64.  Seeing if we can build it...
WARNING:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:Dependency libprrte, version any is not installable from your channels: ['/home/conda/staged-recipes/build_artifacts', 'conda-forge'] with subdir linux-64.  Seeing if we can build it...
/home/conda/staged-recipes-copy/recipes/openpmix {'platform': 'linux', 'arch': '64', 'label': 'linux-64'} <conda_build.config.Config object at 0x7f0ded966d80> False
rendering /home/conda/staged-recipes-copy/recipes/openpmix for linux-64
WARNING:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:Dependency libpmix, version any is not installable from your channels: ['/home/conda/staged-recipes/build_artifacts', 'conda-forge'] with subdir linux-64.  Seeing if we can build it...
WARNING:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:Dependency libpmix, version any is not installable from your channels: ['/home/conda/staged-recipes/build_artifacts', 'conda-forge'] with subdir linux-64.  Seeing if we can build it...
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: libprrte-3.0.7-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: [MatchSpec("gcc_linux-64=13"), MatchSpec("sysroot_linux-64=2.17"), MatchSpec("gxx_linux-64=13")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: [MatchSpec("libevent=2.1.12"), MatchSpec("libhwloc=2.11.2"), MatchSpec("libpmix-devel")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: libprrte-devel-3.0.7-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: [MatchSpec("libprrte")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: prrte-3.0.7-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: [MatchSpec("gcc_linux-64=13"), MatchSpec("sysroot_linux-64=2.17"), MatchSpec("gxx_linux-64=13")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: [MatchSpec("libevent=2.1.12"), MatchSpec("libhwloc=2.11.2"), MatchSpec("libpmix-devel"), MatchSpec("libprrte")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: [MatchSpec("libprrte")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: libpmix-5.0.4-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: [MatchSpec("gcc_linux-64=13"), MatchSpec("sysroot_linux-64=2.17"), MatchSpec("gxx_linux-64=13"), MatchSpec("make"), MatchSpec("pkg-config"), MatchSpec("python==3.12")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: [MatchSpec("libevent=2.1.12"), MatchSpec("libhwloc=2.11.2"), MatchSpec("zlib=1")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: ['pkg-config']
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: libpmix-devel-5.0.4-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: []
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: [MatchSpec("libpmix")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: ['pkg-config']
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:node: pmix-bin-5.0.4-on-linux-64
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   build: [MatchSpec("gcc_linux-64=13"), MatchSpec("sysroot_linux-64=2.17"), MatchSpec("gxx_linux-64=13")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   host: [MatchSpec("libpmix")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   run: [MatchSpec("libpmix")]
INFO:/home/conda/staged-recipes-copy/.ci_support/compute_build_graph.py:   test: []
Computed that there are 2 distributions to build from 2 recipes
Resolved dependencies, will be built in the following order:
    prrte-split-3.0.7-on-linux-64
    openpmix-5.0.4-on-linux-64
```

 This doesn't happen if the pmix top-level recipe name matches one of the outputs, which generally causes problems and should be avoided, but we can fix that on the feedstock.

minrk:

j34ni:
@minrk I do not understand: without automake and autoconf in the requirements my local build fails, anyway is {{ compiler('cxx') }} of any use?

minrk:
Good catch, I removed cxx, I just assumed it would be needed.

autotools should not generally be needed to build ~anything from source unless you need to ship patches for autoconf files. I did have to add [this touch](https://github.com/conda-forge/staged-recipes/pull/28225/files#diff-deeee96f1c97d237fc4055734be3a7548e3bdeca83ab519a5140de4988c5a972R4) to prevent autotools being invoked because there appears to be something wrong with the timestamps in the pmix tarball. But the builds here on CI and my local builds don't invoke autotools. I'm not sure what would cause your builds to trigger it. Can you share output?

minrk:
Yeah, that's exactly the issue fixed by the `touch` in these builds here. The dependency check is deciding (incorrectly) that autotools needs to regenerate configure or some other files, which it does based on the timestamps of the generated files and the files used to generate them. A `touch` on the generated files should convince it the files are up-to-date and prevent the invocation of autotools. Unfortunately, I don't know why it's coming to a different decision some places and not others. If you do a `ls -la` in the checkout directory, you might see which files have the wrong timestamps on them.

recipes/openpmix/meta.yaml
```diff
@@ -0,0 +1,131 @@
+{% set name = "openpmix" %}
+{% set version = "5.0.4" %}
+
+package:
+  # don't usually want top-level name to match any outputs,
```
danielnachun, line 5: Comments should be removed from the `meta.yaml` - if there are TODOs to track, open an issue when the feedstock is created to track these.  Other wise the recipe becomes difficult to parse programmatically and contributors have to open the `meta.yaml` to see what needs to be done instead of seeing a separate issue.

recipes/openpmix/meta.yaml
```diff
@@ -0,0 +1,131 @@
...
+    - zlib
+
+outputs:
+  - name: libpmix
+    files:
+      # libpmix has everything except for specific exclusions
+      include:
+        - '**/*'
+      exclude:
+        # these are in -bin
```
danielnachun, line 42: While these kinds of comments may be helpful in development, they should be removed for the final `meta.yaml`.

recipes/prrte/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+        - {{ stdlib('c') }}
+      host:
+        - libevent
+        - libhwloc
+        - libpmix-devel
+        - {{ pin_subpackage('libprrte', exact=True) }}
+      run:
+        - {{ pin_subpackage('libprrte', exact=True) }}
+    test:
+      script: test-bin.sh
```
danielnachun, line 101:
```suggestion
      script:
        - prte_info
        - prun --do-not-connect --allow-run-as-root -n 2 sh -c 'echo hi'
```
I think your test script is short enough it should be possible to just run it here.

recipes/prrte/meta.yaml
```diff
@@ -0,0 +1,119 @@
...
+
+
+about:
+  home: https://prrte.org
+  summary: PMIx Reference RunTime Environment
+  description: |
+    PRRTE is the PMIx Reference RunTime Environment
+  license: BSD-3-Clause
+  license_family: BSD
```
danielnachun, line 110: The `license_family` key is now redundant and can be removed because the `license` key is already linted to be a valid SPDX identifier.

recipes/openpmix/build.sh
```diff
@@ -0,0 +1,17 @@
+set -ex
+
+# touch generated files to prevent unnecessary invocation of automake
+touch Makefile.in configure
+
+# show available configure options
+# ./configure --help
```
danielnachun, line 7:
```suggestion
```
This commented out code appears only be needed for debugging purpose and should be removed.

recipes/openpmix/meta.yaml
```diff
@@ -0,0 +1,131 @@
...
+      commands:
+        - pmix_info
+
+about:
+  home: https://pmix.org
+  summary: OpenPMIx Implementation of the PMIx Standard
+  description: |
+    Reference Implementation of the Process Management Interface Exascale (PMIx) standard
+  license: BSD-3-Clause
+  license_family: BSD
```
danielnachun, line 122: The `license_family` key is now redundant and can be removed because the `license` key is already linted to be a valid SPDX identifier.

danielnachun (changes requested):
The overall structure here looks good to me - I just made some minor suggestions for improvements.

recipes/openpmix/meta.yaml
minrk: This is the only place we have to write a todo because the feedstock doesn't exist to open issues, I can remove it once the feedstock is created, since it is only here to workaround a bug in staged recipes, but there's nowhere else to put it for now.

Can I also ask where the idea that all comments should be discouraged in meta.yaml is coming from? I don't think it's true or a good practice and I haven't seen it suggested before. Comments help make clear, maintainable recipes, and should generally be encouraged.

minrk: Thanks! Hopefully those best practices will encourage code comments for clear, maintainable recipes :)

recipes/openpmix/meta.yaml
minrk: Since you mentioned documenting best practices, I'll provide a bit of background on why I think these (and other similar comments) should stay, and indeed be actively encouraged in best practices for recipes across conda-forge, in case it's helpful to your writing.

Most recipes are maintained in a tiny fraction of maintainers' time, and are often interacted with very infrequently, sometimes months or years between changes. Any deviation from the simplest recipes or bits that prompt "why is this here?" should have comments explaining what's going on. In this way, meta.yaml the same as any other code or config. Many times over the years, I've seen (and made!) what look like innocuous 'clean up' changes to recipes because something odd or unusual in meta.yaml was not explained, its rationale not remembered and not self-evident, and then removed or changed in a way which ultimately broke the recipe (examples include host dependencies that aren't linked, but required for build to succeed, parts of the recipe where changes must be kept in sync with another part of the recipe, etc.). Comments help explain these things and help maintainers keep recipes in good health, since it should not be expected that maintainers have all the history and context of the recipe in mind when it's time to make a contribution.

As a specific example, this particular comment helps maintainers and reviewers understand what should be packaged in which output vs excluded entirely, and what needs to be kept in sync. For example, if these exclusions are changed it is likely the `-bin` output inclusions are likely to need changes as well.

carterbox (changes requested):
The upstream project seems to be aware of sonames and provides separate ABI and API versions. However, the sonames of the binaries built here don't match what upstream claims they should be in their VERSION files. Maybe the project is configured incorrectly?

https://github.com/openpmix/openpmix/blob/20ad9b16fe49ede4a76e7489bdf0e2ea05c3de95/VERSION#L122
https://github.com/openpmix/prrte/blob/v3.0.7/VERSION#L92C1-L93C1

Also, I think the comments added to this recipe are appropriate. Comments are valid yaml, so I would expect the only time they conflict with parsing of the recipe is if the comments are in-line like selectors instead of above-line like you have here.

recipes/openpmix/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    test:
+      requires:
+        - pkg-config
+      commands:
+        - test -f $PREFIX/lib/libpmix${SHLIB_EXT}
+        - test ! -f $PREFIX/lib/libpmpix.a
+        - test ! -f $PREFIX/bin/pmix_info
+        - test ! -d $PREFIX/include/pmix/src
+        - test ! -d $PREFIX/share/doc/pmix
+        - pkg-config --cflags --libs pmix
```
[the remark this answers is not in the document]
minrk: thanks, this test was before I split off the devel packages. I would have expected it to fail.

minrk:
> However, the sonames of the binaries built here don't match what upstream claims they should be in their VERSION files. 

What version mismatch do you see? pmix is [producing](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1104064&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff&l=3761) libpmix.2.13.4.so, which is what you get from the (arcane) libtool version input of 15:4:13 -> (15-13).13.4, and same for prrte where 3:7:0 gives (3-0).0.7.

carterbox:
> What version mismatch do you see?

Maybe I don't understand libtool versioning. I thought that the sonames should be literally:

```
libprrte_so_version=3:7:0
libpmix_so_version=15:4:13
```

But I will defer to you.

Can we use the soname in the package name instead of pinning run_exports to the minor API version? e.g. `libprrt3` and `libpmix2`.

dalcinl:
> Maybe I don't understand libtool versioning. I thought that the sonames should be literally:

That's definitely NOT the way libtool versioning works.

> Can we use the soname in the package name instead of pinning run_exports to the minor API version? e.g. `libprrt3` and `libpmix2`.

That may be a good idea for lib* packages, I believe Debian uses a similar approach. However, pinning is still needed. When new APIs are added in a backward compatible way, the SONAME version  does not change. How do you prevent an old incompatible version from being installed if not by pinning? 

Then, what's the point of adding the soname to the package name? IMHO the the only reason to have libpmix2 is to be able to install  both libpmix1 and libpmix2 in the same conda environment.  Such thing may be a reasonable expectation for a Linux distro installing multiple libraries system-wide, but I think such feature is less useful within the context of throw away conda environments. I'm not opposed to it, but I wanted to point out it may not be that useful in the end.

carterbox:
> How do you prevent an old incompatible version from being installed if not by pinning?

By "instead of pinning [...] to the minor API version" I meant letting `max_pin` be the major version (which is the default behavior) instead of restricting `max_pin` to the minor version. That wasn't very clear of me, sorry.

In my experience, adding the soname to the package name lets you use looser pinning on the API version because the package name change will protect you from ABI breaks if a library breaks the ABI without breaking the API. A concrete example, one could break an ABI by changing the memory layout of a struct without breaking the API which (in my experience) is something that image/video codec libraries tend to do.

carterbox:
> I do think it would have to contain libpmix.so.2 and not libpmix.so, which would move to devel. 

Yes. That's exactly correct.

I think this PR is ready to merge. Please post an explicit decision about whether you want libpmix2 before merging. Since, if you decide to do so later, you probably need a repo patch.

minrk:
Yeah, I don't want to do the soversion package until there's a clearer agreement from conda-forge core on how to approach ABI tracking. I opened https://github.com/conda-forge/conda-forge.github.io/issues/2401 to start that discussion.

Depending on what approach we take, repodata patches may or may not be needed, but I don't anticipate it happening any time soon. If we go with `libpmix2`, I don't think any patches would be needed.

PR #28236, opened 2024-11-16
Add linux-aarch64 and osx-arm64 configs for local debugging

jaimergp:
We won't use these in our CI, but it's nice to have for local debugging if you are running Apple Silicon. This way you don't have to pay the emulation cost for `linux-64` on Docker.


jezdez (approved):

minrk:
Nice! FWIW, I've been working aroudn this myself for a long time with `-m ../some-other-feedstock/.ci_support/osx_arm64_.yaml`

PR #28246, opened 2024-11-17
Adding iterativefiltering package

EmanuelePapini:
added iterativefiltering to recipes

Checklist
- [ ] Build number is 0.


EmanuelePapini:
@conda-forge/help-python
Conda check fails in building for mac OS X. I have no idea of why that happens

EmanuelePapini:
@conda-forge/help-python my recipe for iterativefiltering is ready for review

recipes/iterativefiltering/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/iterativefiltering-{{ version }}.tar.gz
+  sha256: 02a89ea3a05aaddf3497a9928da515e4e5a42ba4b1e4868dbc80f03362733deb
+
+build:
+  noarch: python
```
synapticarbors, line 13: Please follow the convention for `noarch: python` described in:

synapticarbors (changes requested):

recipes/iterativefiltering/meta.yaml
EmanuelePapini: I have modified the .yaml file according to the documentation.

synapticarbors: Please have another look at the docs I linked. Under `host` and under the `requires` section of `test` it should be `- python {{ python_min }}`

EmanuelePapini: Ok, but I need to specify (for compatibility reasons with numba) that the version must be earlier than 3.13. I thought that the way I did it was correct. How can I do that in the .yaml file?
Also, I would like to note that in the documentation you pointed me at, the use of python_min is not mandatory (since it uses the noun "should" instead of "must") or am I mistaken?
Also, is there a python_max variable that can be used?

EmanuelePapini: @synapticarbors I applied the changes regarding python {{ python_min }}, but left the constraint on the python version (<3.13).

synapticarbors: `host` should be `- python {{ python_min }}`. It's an exact match so we don't worry about the upper bound. The same goes for `test.requires`. Under `run`, you need `- python >={{ python_min }},<3.13`. The way you currently have it specified will only allow users to use `python_min`.

EmanuelePapini: Thank you @synapticarbors. I have applied the changes and committed them.

EmanuelePapini:
@synapticarbors I believe I did all the corrections. Could you check if the package is now ready to be published?

synapticarbors:
Sorry for the delay on merging this and thanks for the reminder

PR #28267, opened 2024-11-18
Add x-transformers

iamthebot:
Checklist
- [x] Build number is 0.


lucidrains:
i agree to be maintainer for x-transformers

recipes/x-transformers/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+package:
+  name: x-transformers
+  version: ${{ version }}
+
+source:
+  url: https://pypi.org/packages/source/x/x-transformers/x_transformers-${{ version }}.tar.gz
+  sha256: eb342ffc5b86193354f27dcdbfd3e13746dd102bf5ddffa21f3286fdc07bde0e
+
+build:
```
synapticarbors, line 13: I think this should be `noarch: python`. We ignore the issue with pytorch missing on windows given that this package does not actually have any platform dependencies. You can get rid of the `skip` section, and then use the `python_min` in the way described in https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-python

synapticarbors (changes requested):

recipes/x-transformers/recipe.yaml
iamthebot: I just updated the recipe to be noarch again (and use the python_min CFEP). The issue, however, is the windows build pipeline still runs since it's a noarch package (and hence fails the staged-recipes build). Linux and OSX are fine.

Also I noticed that despite using `python_min` as directed, the linter still complains that I'm not using it. Is that because this is a rattler-build based recipe?

Let me know your thoughts :)

iamthebot: @synapticarbors

recipes/x-transformers/recipe.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - einops >=0.8.0
+    - packaging >=21.0.0
+
+tests:
+  - script:
+      - pytest -vvv tests
+    requirements:
+      run:
+        - pytest
+        - python >=${{ python_min }}
```
synapticarbors, line 39: For the tests this should be `- python ${{ python_min }}`, so that it tests the lowest allowable version of python. Can you also add an import test? See:
https://prefix-dev.github.io/rattler-build/latest/reference/recipe_file/#python-tests

recipes/x-transformers/recipe.yaml
synapticarbors: We just let it fail on Windows assuming the error is caused by the missing pytorch (which is the case here)

PR #28268, opened 2024-11-19
feat: add atlas-schema into conda-forge

kratsg:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/atlas-schema/meta.yaml
+++ b/recipes/atlas-schema/meta.yaml
@@ -12,7 +12,7 @@ source:
 
 build:
   noarch: python
-  script: {{ PYTHON }} -m pip install . -vv
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation 
   number: 0
 
 requirements:
--- a/recipes/atlas-schema/meta.yaml
+++ b/recipes/atlas-schema/meta.yaml
@@ -26,7 +26,7 @@ requirements:
     - coffea >=2024.10.0
     # below are for coffea[dask]
     - distributed >=2024.3.0
-    - "bokeh !=3.0.*,>=2.4.2"
+    - bokeh !=3.0.*,>=2.4.2
 
 test:
   imports:
--- a/recipes/atlas-schema/meta.yaml
+++ b/recipes/atlas-schema/meta.yaml
@@ -14,7 +14,7 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
-  skip: true  # [win]
+  #skip: true  # [win]
 
 requirements:
   host:
--- a/recipes/atlas-schema/meta.yaml
+++ b/recipes/atlas-schema/meta.yaml
@@ -14,7 +14,6 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv
   number: 0
-  #skip: true  # [win]
 
 requirements:
   host:
--- a/recipes/atlas-schema/meta.yaml
+++ b/recipes/atlas-schema/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "atlas-schema" %}
+{% set version = "0.1.0" %}
+{% set python_min = "3.9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/atlas_schema-{{ version }}.tar.gz
+  sha256: 50f8fd29e958f00ff5fe8b0f2235e281952e5391e893bb38e61c2c50751b7766
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  skip: true  # [win]
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - hatchling >=0.7
+    - hatch-vcs >=0.3.0
+  run:
+    - python >={{ python_min }}
+    - coffea >=2024.10.0
+    # below are for coffea[dask]
+    - distributed >=2024.3.0
+    - "bokeh !=3.0.*,>=2.4.2"
+
+test:
+  imports:
+    - atlas_schema
+  commands:
+    - pip check
+  requires:
+    - python {{ python_min }}
+    - pip
+
+about:
+  home: https://github.com/scipp-atlas/atlas-schema/
+  summary: Helper python package for ATLAS Common NTuple Analysis work.
+  license: Apache-2.0
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - kratsg


kratsg:

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/atlas_schema-{{ version }}.tar.gz
+  sha256: 50f8fd29e958f00ff5fe8b0f2235e281952e5391e893bb38e61c2c50751b7766
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv # --no-deps --no-build-isolation
+  number: 0
+  skip: true  # [win]
```
matthewfeickert, line 16: > * `noarch` packages can't have skips with selectors. If the selectors are necessary, please remove `noarch: python`

@kratsg this linter error is for the `# [win]` but the conda-forge team is okay with this happening for Windows. So this can be ignored.

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/atlas_schema-{{ version }}.tar.gz
+  sha256: 50f8fd29e958f00ff5fe8b0f2235e281952e5391e893bb38e61c2c50751b7766
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv # --no-deps --no-build-isolation
```
matthewfeickert:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv
```

These are already enforced by the build tool (e.g. `conda-build`, `rattler-build`).

matthewfeickert:
@kratsg you can keep it `noarch`.

recipes/atlas-schema/meta.yaml
kratsg: this came from grayskull, so it wasn't obvious if i can drop it, given the conda documentation recommended grayskull.

recipes/atlas-schema/meta.yaml
matthewfeickert: They will tell you it is fine. I was given this advice by the conda-forge/core team to avoid doing builds for Linux and macOS when they're equivalent.

kratsg:
> @kratsg you can keep it `noarch`.

done

matthewfeickert (approved):

matthewfeickert:
@kratsg, btw, for future testing the conda-forge team made it even easier now to do local builds before opening up a project by adding a `pixi` project config. So you now can just do

https://github.com/conda-forge/staged-recipes/blob/566507ceb7b3ae32c805043d89ad80343724fa90/README.md#L49-L51

for the next project you add. :rocket:

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - hatchling >=0.7
+    - hatch-vcs >=0.3.0
+  run:
+    - python >={{ python_min }}
+    - coffea >=2024.4.1
+    - distributed >= 2024.3.0
```
matthewfeickert, line 28:
```suggestion
    - distributed >=2024.3.0
```

matthewfeickert (changes requested):

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+    - python {{ python_min }}
+    - pip
+    - hatchling >=0.7
+    - hatch-vcs >=0.3.0
+  run:
+    - python >={{ python_min }}
+    - coffea >=2024.10.0
+    - distributed >=2024.3.0
+    # below are for coffea[dask]
+    - distributed >= 2024.3.0
```
matthewfeickert, line 30:
```suggestion
    - distributed >=2024.3.0
```

@kratsg, you typoed the spacing

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  host:
+    - python {{ python_min }}
+    - pip
+    - hatchling >=0.7
+    - hatch-vcs >=0.3.0
+  run:
+    - python >={{ python_min }}
+    - coffea >=2024.10.0
+    # below are for coffea[dask]
+    - distributed >= 2024.3.0
```
matthewfeickert, line 29:
```suggestion
    - distributed >=2024.3.0
```

@kratsg each time you force push you keep having the same typo in your spacing.

recipes/atlas-schema/meta.yaml
matthewfeickert: The issue is the first one.

> If the reason for the skip is a missing dependency, then remove this, let it fail, and the package won't be installable on Windows until that dependency is available.

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/atlas_schema-{{ version }}.tar.gz
+  sha256: 50f8fd29e958f00ff5fe8b0f2235e281952e5391e893bb38e61c2c50751b7766
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+  #skip: true  # [win]
```
matthewfeickert, line 17:
```suggestion
```

You can fully remove this line, as I remember now that a noarch build just builds on x86 Linux as the noarch bit makes other builds redundant. So there's no need to have this put back in the future.

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,49 @@
+{% set name = "atlas-schema" %}
+{% set version = "0.1.0" %}
```
ocefpaf, line 2: There is a 0.2.0 on PyPI. Do you need the 0.1? If not, please add the latest one.

matthewfeickert: It would be nice to have parity between PyPI and conda-forge, so having v0.1 could be useful for us.

recipes/atlas-schema/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/atlas_schema-{{ version }}.tar.gz
+  sha256: 50f8fd29e958f00ff5fe8b0f2235e281952e5391e893bb38e61c2c50751b7766
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
```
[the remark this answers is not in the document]
ocefpaf: We had minor conda-build breakages where it was. I really don't care about such small details that people feel so strongly. I have zero stake on this package. Go for it, if you ever get pip packages in your env, you know why.

kratsg:
@ocefpaf - all changes commited/made.

recipes/atlas-schema/meta.yaml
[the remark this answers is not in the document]
matthewfeickert: Sorry, I think I came across in the wrong manner. I thought that this problem only affected packages that had multiple build outputs, and that for all other situations this had gotten patched so that the build tool set all the flags that were required.

@kratsg, as I don't fully understand things, you should apply the `--no-deps --no-build-isolation` patch back.

recipes/atlas-schema/meta.yaml
matthewfeickert: This is a really helpful summary / checklist. I might open an Issue/PR to add this to the linter later.

recipes/atlas-schema/meta.yaml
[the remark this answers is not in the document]
ocefpaf: The reason why we have that set on grayskull is b/c we chose to be explicit and avoid regression on tools downstream. We could set those flags in conda-forge but the pip reversed logic and yet another hidden layer of options, would make this awkward. IMO, we should be explicit here to ensure the users know that is happening. Not long ago some people came to me surprised that we use no build isolation in conda-forge as if it was new... It isn't.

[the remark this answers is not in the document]
matthewfeickert: > IMO, we should be explicit here to ensure the users know that is happening. Not long ago some people came to me surprised that we use no build isolation in conda-forge as if it was new... It isn't.

Just for context, I always assumed that the fact that `grayskull` had this recommendation meant that `conda-forge` **wasn't** using `--no-build-isolation` as my assumption is that "If I have to specify something as the user then this means that it is my responsibility and that it isn't being dealt with elsewhere."

So I took away the opposite of the intended message.

PR #28276, opened 2024-11-19
fix multi-output dependency detection

minrk:
collapse_subpackage_nodes only handled outgoing dependencies (not incoming) before deleting nodes so some edges were lost.

by updating edges for both directions, relationships between recipes should be preserved.

for example with the graph:

```
[recipe-A, recipe-B, output-A, output-B]
output-A -> output-B
```

it updates to:

```
[recipe-A, recipe-B, output-A]
output-A -> recipe-B
```

then deletes the `output-A` node, implicitly deleting the only edge along with it, losing the relationship between the recipes. This PR updates _both_ sides of the subpackage edges to point to the recipe before deleting a node for a subpackage.

seen in #28225 where two multi-output packages are added with one dependency between them. If the top-level recipe name doesn't match an output, the packages are built in the wrong order because the subpackage node is deleted. This patch fixes the build order in that PR.


minrk:
As far as I know this remains relevant, but has received no review

PR #28395, opened 2024-11-28
Adding conda package for drug2cell tool

Mennayousef:
I am attempting to create a conda recipe package for drug2cell tool.

### DIFF
--- a/recipes/drug2cell/meta.yaml
+++ b/recipes/drug2cell/meta.yaml
@@ -18,11 +18,11 @@ build:
 
 requirements:
   host:
-    - python >={{ python_min }}  # Corrected spacing
+    - python >= {{ python_min }}  # Corrected spacing
     - pip
     - setuptools  # Required build backend
   run:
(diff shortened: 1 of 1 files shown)


Mennayousef:
maintainer

mfansler (changes requested):
Thanks for spending time on this! Unfortunately, I'm not for merging this as it stands.

Conda Forge packages should be properly versioned, either through PyPI or GitHub Releases rather from a commit hash, which is mutable. My recommendation is either that one gets the lab to commit to making releases of some form, or instead move this to Bioconda, which aligns with the intent of the package and also has a lower bar for recipes standards.

mfansler:
Just noticed that this was merged on Bioconda (https://github.com/bioconda/bioconda-recipes/pull/52452). As a `noarch` recipe, there will not be any advantage to shifting to Conda Forge, even with the version issue addressed. As such, I will close this.

PR #28413, opened 2024-11-30
Add imgui-bundle and other related packages

hmaarrfk:
Checklist
- [ ] Build number is 0.


hmaarrfk:
lol, out of memory error for windows...

hmaarrfk:
What a twist of fate. In 1.5.2 i could get linux + osx but not windows, now i can get windows only.

pthom:
Well done, you are not that far! 

---

#### linux_x64 failure: "_imgui_bundle.cpython-310-x86_64-linux-gnu.so: undefined symbol: glMatrixMode"

It seems like the *build succeeds*, but subsequently only the *test fails* when testing (because *OpenGL is not installed* on headless build servers).

This is an issue I already faced, and I applied several solutions for it:

**In pyproject.tom, there are fixes for cibuildwheel**

https://github.com/pthom/imgui_bundle/blob/0b6b9b6aef6e3e78bf30458057dd68ced2721e6a/pyproject.toml#L138-L142

```
[tool.cibuildwheel.linux]
before-all = "yum install -y libXcursor-devel libXi-devel libXinerama-devel libXrandr-devel mesa-libGL-devel"
[[tool.cibuildwheel.overrides]]
select = "*-musllinux*"
before-all = "apk add xorg-server-dev libxcursor-dev libxi-dev libxinerama-dev libxrandr-dev mesa-dev mesa-gl"
```

**in .github/workflows/pip.yml, there are fixes for GitHub'CI

https://github.com/pthom/imgui_bundle/blob/0b6b9b6aef6e3e78bf30458057dd68ced2721e6a/.github/workflows/pip.yml#L31-L37

```
    - name: install xorg-dev & start Xvfb (ubuntu only)
      if: ${{ matrix.platform == 'ubuntu-latest' }}
      run: |
        sudo apt-get update && sudo apt-get install -y xorg-dev libgl1-mesa-dev libglfw3-dev libglew-dev
        sudo apt-get install -y xvfb
        Xvfb :99 &
        echo "DISPLAY=:99.0" >> $GITHUB_ENV
```

pthom:
**Failure on osx_x64**

```
CMake Error at /Users/runner/bld/imgui-bundle_1733066354361/_build_env/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
    Could NOT find OpenGL (missing: OPENGL_gl_LIBRARY OPENGL_INCLUDE_DIR)
    /Users/runner/bld/imgui-bundle_1733066354361/_build_env/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
    /Users/runner/bld/imgui-bundle_1733066354361/_build_env/share/cmake-3.31/Modules/FindOpenGL.cmake:579 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
    external/immvision/immvision/src/immvision_gl_loader/CMakeLists.txt:30 (find_package)
```

where /immvision_gl_loader/CMakeLists.txt:30 contains

```
    find_package(OpenGL REQUIRED)
```

This one is strange, since this is the correct way to look for OpenGL with Mac.

You could 
- try to comment out these lines (but imgui_bundle_cmake/internal/add_imgui_bundle_bindings.cmake:87 will do the same find_package a little later)
- try to call find_package(OpenGL REQUIRED) very soon in the top CMakeLists

I have this in pyproject.toml, for cibuildwheel, but I don' know if it is related:
```
environment = { MACOSX_DEPLOYMENT_TARGET="14.0" }
```

As a last resort, you could do something close to:

```
if (APPLE)
    target_compile_options(immvision_gl_loader PUBLIC -framework OpenGL)
elseif (NOT EMSCRIPTEN)
    find_package(OpenGL REQUIRED)
    target_link_libraries(immvision_gl_loader PUBLIC OpenGL::GL)
endif()

```

hmaarrfk:
> linux_x64 failure: "_imgui_bundle.cpython-310-x86_64-linux-gnu.so: undefined symbol: glMatrixMode"

The thing it, it also occurs locally when I test. So it isn't related to the packages that are installed on the server (yet).
I think it has to do with the fact that you have in many location 

```
target_link_library(xxxx PUBLIC OpenGL::GL)
```
and it should be replaced with

```
target_link_library(xxxx, PRIVATE OpenGL::GL)
```


> environment = { MACOSX_DEPLOYMENT_TARGET="14.0" }

I can try to bump things, but 14.0 is quite new..... I wonder what would be triggering such a high version. we typically use 10.13 on x64 and whats strangest is that this was all working on 1.5.2 but broke on the 1.6.0 update.

hmaarrfk:
> As a last resort, you could do something close to:

```
    target_compile_options(immvision_gl_loader PUBLIC -framework OpenGL)
```

i'm not sure we are allowed to link to framework on conda-forge.

hmaarrfk:
@pthom also, thank you for chiming in here to help out!!! I'm excited to see all the changes between 1.5.2 and 1.6.0 that aimed to help maintainability!

pthom:
I'm interested in integrating some (or all) of the patches you had to do in CMake, possibly by surrounding them with

```cmake
if (CONDA_BUILD)
   ...
endif()
```

But this is not urgent :-)

hmaarrfk:
The other challenge with `imgui` is that it actively encourages vendoring code, (i.e. just copy the source into your project) which again isn't really fun when you are trying to package things for easy reuse.

recipes/imgui-bundle/meta.yaml
```diff
@@ -0,0 +1,109 @@
...
+    - rm -rf external/glfw/          # [unix]
+    - del /s /q external\glfw\       # [win]
+    - python -m pip install . -vvv
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+    - cmake
```
hmaarrfk, line 50:
```suggestion
    - cmake
    - glad2-cmake
```

hmaarrfk:
@pthom is there a commit or two that you can point me to where you changed the OpenGL linkages?

pthom:
> @pthom is there a commit or two that you can point me to where you changed the OpenGL linkages?

Hum, there might be a reason I just found in an issue which I am addressing right now: https://github.com/pthom/hello_imgui/issues/128

I just pushed [this commit](https://github.com/pthom/imgui_bundle/commit/58315e2dede62b022ae5a9ecb8f543b7b4c1ee36#diff-93593f0c350d115f3cd91e49afddf938164daee8d2240628d25ba42809d08646L1) which changes the way glad is build (and now build only once. My bad!):

See
* related [commit in Hello ImGui](https://github.com/pthom/hello_imgui/commit/b53e2b475d62174b086aae4eb1c52f05e34c068a)
* related [commit in ImmVision](https://github.com/pthom/immvision/commit/b73cc6d38e16b35d63e9f9c6ed8abeb938f59f90)

Unrelated: to address this particular user's issue, I probably will advise him to regenerated glad bindings with the GL_KHR_debug extension. Do you know if the glad package provided with Conda includes it (this is very much unimportant, though).

hmaarrfk:
> Do you know if the glad package provided with Conda includes it

No honestly this feedstock taught me about glad. 

You can make a pr there. PRs are free and helpful!!

pthom:
There is also this commit I made while switching to nanobind:

https://github.com/pthom/imgui_bundle/commit/824e7b20609bdfe8a4a9de0a0ce8728b7176878a#diff-89325aec0584aa614224b7af29dd9cb922f78830603c8abdeafc8ccd79331c77R50-R53

I had to link manually the bindings with OpenGL. It was not necessary with pybind11.

hmaarrfk:
Note to self, read up on how to use conda-forge's glad cmake's stuff
https://github.com/conda-forge/glad-feedstock/issues/7#issuecomment-1920716321

recipes/imgui-bundle/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - glad2-cmake
+    - ninja
+    # CMake will complain if it can't find make
+    - make             # [unix]
+  host:
+    - python
+    - pip
+    - glad2
+    - glad2-cmake
```
hmaarrfk, line 53:
```suggestion
```

recipes/imgui-bundle/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - {{ stdlib('c') }}
+    - cmake
+    - glad2-cmake
+    - ninja
+    # CMake will complain if it can't find make
+    - make             # [unix]
+  host:
+    - python
+    - pip
+    - glad2
```
hmaarrfk, line 52:
```suggestion
```

recipes/imgui-bundle/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+    - del /s /q external\glfw\       # [win]
+    - {{ PYTHON }} -m pip install . -vvv
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+    - cmake
+    - glad2-cmake
```
hmaarrfk, line 45:
```suggestion
```

pthom:
> Note to self, read up on how to use conda-forge's glad cmake's stuff [conda-forge/glad-feedstock#7 (comment)](https://github.com/conda-forge/glad-feedstock/issues/7#issuecomment-1920716321)

See here for the glad settings used in hello imgui and by extension in the bundle : https://github.com/pthom/hello_imgui/blob/master/external/OpenGL_Loaders/glad/include/glad/glad.h

hmaarrfk:
Something is just weird with nanobind and the file pybind_imgui_backends.cpp 

it seems if i comment it out from the nanobind module, the symbolds glMatrixMode and related are no longer visible through `nm -D`.

I'm not sure what is triggering this behavior.

it also seems that you don't trust your litgen to generate the bindings for that file, so something seems suspicious????

However much I try to have cmake add the flags `-lGL` it seems that if they aren't at the end of the c++ command, then GL simply isn't getting linked to..... I had to resort locally to adding it as a CFLAG.... then somehow it ends up at the end of the command.

recipes/imgui-bundle/singel_patch.patch
```diff
@@ -0,0 +1,854 @@
...
+     endif()
+ endfunction()
+ 
+-# See https://github.com/pthom/imgui_bundle/issues/261 and https://chatgpt.com/share/66ffb718-6408-8004-be5b-9e74064a8709
+-function(ibd_disable_system_libraries_for_python)
+-    if(IMGUI_BUNDLE_BUILD_PYTHON AND
+-       (NOT "$ENV{CMAKE_TOOLCHAIN_FILE}" MATCHES "vcpkg" AND NOT DEFINED CONAN_BUILD)  # Not using vcpkg or conan
+-       )
+-        message(STATUS "ibd_disable_system_libraries_for_python: Disabling system libraries for python bindings")
+-        set(CMAKE_FIND_FRAMEWORK NEVER CACHE STRING "" FORCE)
```
hmaarrfk, line 685: ps @pthom  I think things like "FORCE" are really hurtful for system integrators. I don't really think any of this is necessary and is typically handled by cibuildwheel (or at least I thought it was)

pthom:
> Something is just weird with nanobind and the file pybind_imgui_backends.cpp
> 
> it seems if i comment it out from the nanobind module, the symbolds glMatrixMode and related are no longer visible through `nm -D`.
> 
> I'm not sure what is triggering this behavior.
> 
> it also seems that you don't trust your litgen to generate the bindings for that file, so something seems suspicious????
> 
> However much I try to have cmake add the flags `-lGL` it seems that if they aren't at the end of the c++ command, then GL simply isn't getting linked to..... I had to resort locally to adding it as a CFLAG.... then somehow it ends up at the end of the command.

Hum, this is strange. The included files seems innocuous enough. Imgui_impl_sdl2 could be removed.
I’ll try to re-add it to the auto generated files. I do not think it will change much.

By the way this file is for a very rare « intermediate » use case: most people either use the full C++ runner (HelloImgui) or go for a full python backend (in which case they will not use this either). The intermediate use case is when you want to run via a python loop but still rely on ImGui to talk to glfw and OpenGL. This means it could be ignored with not much impact.

recipes/imgui-bundle/singel_patch.patch
pthom: Sorry for that. I’ll study if it can be removed.
I remember adding it because skbuild + pip was still catching some local libraries when building wheels that are to be distributed.

hmaarrfk: yeah i've struggled with the flags before. so long as they can be overwritten it is typically ok.

pthom:
The [failed build on win_64](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1098209&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de) seems to be due to lunasvg. lunasvg is not required if plutosvg is being used.
```
WARNING: Tests failed for lunasvg-3.0.1-he0c23c2_0.conda - moving package to D:\bld\broken
```

hmaarrfk:
> lunasvg is not required if plutosvg is being used.

as I understand it plutosvg is preferred? Didn't you delete the references to lunasvg?

pthom:
Yes plutosvg is preferred.
I did not remove references to lunasvg for backward compatibility, for long time users

hmaarrfk:
well this is a good sign, things are green. Great job working through the hacks for windows and nanobind.... i saw you issue. Is it OK that I conditionally apply the hacks only to windows?

pthom:
> well this is a good sign, things are green. Great job working through the hacks for windows and nanobind.... i saw you issue. Is it OK that I conditionally apply the hacks only to windows?

I did apply it on my side: see
https://github.com/pthom/imgui_bundle/blob/1c5c211b7583708718498df955679068d2ef71ad/imgui_bundle_cmake/internal/add_imgui_bundle_bindings.cmake#L13

pthom:
> > I did apply it on my side: see
> 
> i mean i mostly added a `if(WIN32)`.

Which is ok

pthom:
Great job!  I’ll try to test it also. Is it possible to download & test it from the CI ?

pthom:
I will test the package and let you know.
By the way, I did a checkout of your fork, and I am now testing whether some of the patches may be applied to imgui bundle.

PS: I just did a donation, your help was tremendous!

hmaarrfk:
> whether some of the patches may be applied to imgui bundle.

awesome, yeah the current status is messy and I won't merge until i get the patches a little more organized.

> PS: I just did a donation, your help was tremendous!

Thats awesome and very generous.


One thing we could do, is to create a shared library `imgui_bundle.so` and follow all the best shared library practices. Then we would link to it with the python wrappers.

One advantage for conda-forge is that we have the tooling to share the compilation results between multiply python versions.

Users are then able to use bot the C++ lib and the python lib. 

When i first saw imgui-bundle, i thought it was only Python. Now i learn that you care about c++ users too! ^_^.

pthom:
Here are some news after some testing on my side. I can build sucessesfully on linux, with the changes below:

I plan to release a corrective new version of imgui_bundle soon to assist you with this release.

I pushed some modifications on my side:


https://github.com/pthom/imgui_bundle/commit/d34be42e (HEAD -> main, origin/main) cmake python: improve site-package detection (see #287) [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/55edaf00 cmake: ibd_disable_system_libraries_for_python_for_pure_pip disabled for conda [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/68c288fb add_imgui_bundle_bindings: _nanobind_hack_disable_forceinline iif WIN32 [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/d0666bab imgui_bundle_conda.cmake: set Python_EXECUTABLE=python [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/45173052 add_imgui.cmake install(... lib) (instead of ./lib/) [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/5c1420f9 immapp/code_utils.h: review includes [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/35c7e7d5 _glfw_set_search_path.py: use conda shared lib if running within conda [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/eee60225 CMake: immvision can use nlohmann_json from find_package if found [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/8f7c0eff CMake: if conda IMGUI_BUNDLE_PYTHON_USE_SYSTEM_LIBS=ON (impacts freetype and glfw) [Pascal Thomet]
https://github.com/pthom/imgui_bundle/commit/34195064 Suppress broken old conda recipe [Pascal Thomet]


As a consequence, the patches below are not required anymore if you update to the latest commit on imgui_bundle. 
```
#     - 0001-Do-not-force-GLFW-to-be-recompiled.patch
#     - 0002-Do-not-force-static-freetype.patch
#     # https://github.com/pthom/imgui_bundle/pull/284
#     - 0003-Do-not-force-linking-to-non-existent-nlohmann_json-s.patch
#     - 0005-Find-glfw-correctly-for-conda-forge.patch
#     # Link privately to opengl to support conda-forge's linking then using sysstem libraries
#     # - 0006-private_link_to_opengl.patch
```


As far as the patch `singel_patch.patch` is concerned, I was able to build without it:

- you will see that I included parts of it in my commits
- I did not include the repository wide transformation of public dependencies into private, since I think the build
  IMHO, this has to be studied case by case: for example, it is acceptable that hello_imgui has a public dependency on imgui.

 
**In order to get the build to work, I had to add this to metal.yaml:**

* Add CONDA_BUILD 

This is used by https://github.com/pthom/imgui_bundle/blob/main/imgui_bundle_cmake/imgui_bundle_conda.cmake
```
script:
    - export CONDA_BUILD=ON   # [unix]
    - set CONDA_BUILD=ON      # [win]
```

* Add freetype to build/requirement/host: (it was not added, which is strange). The build seems to work without it, but this is not normal
```
build:
   requirement:
      host:
         - freetype
```

* Add some lib to run requirements, as recommended by GPT o1 (I am in no way familiar with this :-)
Among those, freetype is required, because the build failed without it (this was before adding to host requirement also).

However, GPT o1 told me this strange sentence: "freetype and glfw do not define run_exports, so they need to also be placed in run requirements"

```
build:
   requirement:
      run:
        - freetype
        - glfw
        - libgl
```


This is very much still work in progress.

hmaarrfk:
> However, GPT o1 told me this strange sentence: "freetype and glfw do not define run_exports, so they need to also be placed in run requirements"

https://github.com/conda-forge/freetype-feedstock/blob/57fd5866321ebf628b0c22dd4ae46bb205ccef33/recipe/meta.yaml#L18

be weary of GPT's recommendations

pthom:
Just some quick feedback on the packages I downloaded from the CI.

I could not test them on my mac (which is a M1: there are only x64 versions).

I tested it on linux, and it worked (but I had to manually install glfw, pydantic, libopencv and pillow):

```bash
conda create -n test_bundle python=3.12 -y 
conda activate test_bundle 

conda install conda_pkgs_linux/imgui-bundle-1.6.0-py312h01b9dd5_1.conda  # downloaded from CI
```

But I guess this package does not (yet, because not integrated in conda ?) resolve dependencies:
```bash
python -c "import imgui_bundle"  #  ==> ImportError: libopencv_core.so.410: cannot open shared object file
```

So I had to do
```bash
conda install -y libopencv glfw pydantic
```
And then I could import imgui_bundle.

Then, I tried the demo (i.e run `imgui_bundle_demo`)

```bash
imgui_bundle_demo # ==> ModuleNotFoundError: No module named 'PIL'
```

So I had to 
```bash
conda install -y pillow
imgui_bundle_demo
```

And then the demo worked perfectly, which is great!

I'll test on windows tomorrow.

As far as patches on imgui_bundle are concerned, I do not have idea of other patches to add at the moment (do you have suggestions?).

Concerning your idea of a shared library for imgui_bundle, there is an install step for imgui_bundle. I think it produces a static library at the moment.

hmaarrfk:
When you install a local package I noticed that conda doesn’t pull in the dependencies. 

I’ve also omitted some demo dependencies to keep the package thin. I guess that’s not the conda way. I could add a package. But I’m not too worried about this. 

We can build M1 packages on the feedstock. 

I want to trim down my patches. They touch a lot of things that I’m not too happy about.

pthom:
In tried to install the windows package today, and I reached a dead-end for Windows and MacOS ARM:

`conda install glfw` works under linux, but fails under windows and under macOS ARM (PackageNotFoundError).


Under windows, running `python -c "import imgui_bundle"` leads to a very unhelpful windows error message ("The specified module cannot be found")

IMHO this is also related to our remarks on pyGLFW in the [imgui_bundle discussion about cmake](https://github.com/pthom/imgui_bundle/issues/290#issuecomment-2520099056).

hmaarrfk:
can you provide me with the output of 

```
conda info
```
on windows and osx.

i fear you have "Anaconda" there and not conda-forge's channels.

hmaarrfk:
> `conda install glfw` works under linux, but fails under windows and under macOS ARM (PackageNotFoundError).


Basically our windows logs show that glfw can in fact be found ;)

on windows:
```
2024-12-03T22:13:26.7721136Z     glfw:                                 3.4-hcfcfb64_0                   conda-forge
```

you can search for `gflw:` to find that line.

pthom:
> i fear you have "Anaconda" there and not conda-forge's channels.

You are right, running `conda config --add channels conda-forge` was enough to install glfw.
Sorry, I am a noob with conda

---

Now I can install successfully on Windows x64, and run "import imgui_bundle".
I also tested that running  a simple hello world GUI application works, and that running an applications which will use opencv (through immvision) works.

However, I could not test the full demo under windows (imgui_bundle_demo.exe) because of a path issue:
```
C:\... > imgui_bundle_demo
Fatal error in launcher: Unable to create process using "D:\bld\imgui-bundle_XXXXX\_h_env\python.exe" .... This file can not be found
```
(I guess the path D:\bld\imgui-bundle_XXXXX\ comes from the build process)

I can run the full demo successfully under Linux (and all widgets seem to work).

hmaarrfk:
> However, I could not test the full demo under windows (imgui_bundle_demo.exe) because of a path issue:

hmm this is strange, i'll give it a look.

recipes/imgui-bundle/meta.yaml
```diff
@@ -0,0 +1,125 @@
...
+  version: {{ version }}
+
+source:
+  # url: https://github.com/pthom/imgui_bundle/archive/refs/tags/v{{ version }}.tar.gz
+  # sha256: ffae861ff9fd47c95cc56807f028d5fe78a700f92a2a43c43f31ea2358542a68
+  git_url: https://github.com/pthom/imgui_bundle.git
+  git_tag: v{{ version }}
+  entry_points:
+    - demo_imgui_bundle = imgui_bundle.demos_python.demo_imgui_bundle:main
+    - imgui_bundle_demo = imgui_bundle.demos_python.demo_imgui_bundle:main
```
hmaarrfk, line 14: @pthom i just added these 3 lines and this should resolve the issues with the windows demo. gotta wait 3 hours for the new builds though.

pthom:
### Full test of the windows conda package

I just tested the package on Windows x64 with python 3.12, and it works great.

#### Additional required installs to import

so that "import imgui_bundle" works, I had to manually install:
`conda install glfw  pydantic libopencv`

(this is probably temporary, when the package is not deployed)

#### Additional required installs to launch the demo
so that the demo works, I had to install `conda install pillow opencv`

Those packages should be installed together with it IMHO (at least pillow, and perhaps opencv. See discussion about OpenCV below)

Note: For the imGuizmo demo PyGLM and munch are also required, but this is a very small part of the demo, and the messages are clear for the user

#### Additional demos in the "immediate apps" tabs of "imgui_bundle_demos"

There are lots of apps that test various scenarios there. The packages that they require do not need to be installed together with imgui_bundle

* custom_background requires `conda install pyopengl`

* demo_matplotlib requires `conda install matplotlib`

* imgui_example_glfw_opengl3: requires the installation "pip install glfw" on top of "conda install glfw"
   (this is the mixed backend I explained: c++ will do the rendering. It works perfectly after installing PyGLFW)

* example_python_backend_glfw3: works after installing "pip install glfw" (i.e PyGLFW)

* example_python_backend_sdl2: works after install "conda install sdl2" + "pip install pysdl2"
(note: the pip package pysdl2 does not provide a dll, only ctypes bindings. As a consequence, this used the sdl2 dll from conda)

* example_python_backend_pyglet: work after installing "pip install pyglet" ("conda install pyglet" fails with conflicts)

**In summary, all the examples do work!!!**


---

### Points of Interest

#### glfw / conda vs pipy are not the same package

In order to run some demos that use the python glfw bindings, one has to run `conda install glfw` and `pip install glfw` (ie pyGLFW). 

Nothing to worry about, but a bit strange for users, I think.

#### A note about OpenCV being linked statically (pip version) vs dynamically

Nothing to be really worried about, but worth mentioning:

**In the pip version**
In the pip version of imgui_bundle, opencv is linked statically, since it is only used internally by one module (immvision).
It is not a requirement at runtime: all the demos and modules, except immvision, will work if "pip opencv-python" is not installed.

**In the conda version**
Since OpenCV is linked dynamically, it becomes a runtime requirement.
Also, one has to install `conda install opencv` at runtime (not only `libopencv`) because otherwise an error "OpenCV loader: missing configuration file" is raised.

hmaarrfk:
> In order to run some demos that use the python glfw bindings, one has to run conda install glfw and pip install glfw (ie pyGLFW).

this will be fixed when the package is deployed.

>  In the pip version of imgui_bundle, opencv is linked statically, since it is only used internally by one module (immvision).

static linking is a no-no in conda-forge. We can ship C dependencies in Conda in ADDITION to python dependencies. Conda is BUILT for enabling this kind of shared linking. 

> Since OpenCV is linked dynamically, it becomes a runtime requirement.
> Also, one has to install conda install opencv at runtime (not only libopencv) because otherwise an error "OpenCV loader: missing configuration file" is raised.

libopencv will be automatically installed in deployed packages.

# Optional dependencies such as Pillow

I'm very weary of making Pillow and "Python Opencv" (vs libopencv) a runtime dependency. This kind of "demo stuff" is fun for literally the first day. Then you never use it again. Users may be interested in using pillow, or they may not be. Same with opencv.

So I'm OK with documenting this dependency:

```
To get started either use pip

pip install imgui_bundle

or conda-forge

# Opencv and Pillow are optional dependencies for the demo applications
conda install imgui-bundle opencv pillow
```

Conda(-forge) is meant to be batteries included, but it gets frustrated once you are done with the "training wheels" to slim down the packages. One can use "split packages" but this is generally more mental load than simply "install opencv, pillow to run the demo python applications".

hmaarrfk:
This package is nearly ready to go (thanks for testing so thoroughly).

I still want to slim down my "singel_patch.path" file to reduce it, or at least to get it to be a little more organized.

I generally suggest that people that are interested in "rushing a release" just upload to their own channels, so I"m going to take my own tip for now.

pthom:
### Full test of the linux x64 conda package

I just tested the package on Linux x64 with python 3.12, and it works great.

**I will mention  below  only the  differences with the Windows test**

#### Additional required installs
As for  windows, I had to manually install:
`conda install glfw  pydantic opencv`


**Additional required installs to launch the demo:  pillow and libopengl**

As for  windows, I had to install `conda install pillow`

**However `conda install libopengl`** is also required, otherwise many apps will fail to start with an error message "Failed to initialize OpenGL loader!",  which  actually happens in:

```cpp
int imgl3wInit(void)
{
    int res = open_libgl();
    if (res)
        return res;
    atexit(close_libgl);
    return imgl3wInit2(get_proc);
}
```

#### A failure with freetype:

The demo "demo_docking.py" available in the "Immediate Apps"tab of the demo  tries  to load a font for  which freetype support inside imgui is required.

I get  this error:

```
  File "/home/pascal/miniforge3/envs/test_bundle/lib/python3.12/site-packages/imgui_bundle/demos_python/demos_immapp/demo_docking.py", line 95, in load_fonts
    app_state.color_font = hello_imgui.load_font("fonts/Playbox/Playbox-FREE.otf", 24., font_loading_params_color)
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: IM_ASSERT( false && "FontLoadingParmas.loadColor requires freetype (IMGUI_ENABLE_FREETYPE)" )   ---   hello_imgui_font.cpp:197

```

I suspect this issue is also present on Windows.

pthom:
> libopencv will be automatically installed in deployed packages.

`conda install  libopencv`  might not  not enough I'm afraid. It leads to  a difficult to diagnose error message:  
some  python apps which *do not require the  opencv-python* package will      fail to start with a message "Bad OpenCV configuration"

pthom:
...  PS :  I could not test on Mac,  due to a missing  osx_arm64  packae

hmaarrfk:
> `conda install libopencv` might not not enough I'm afraid. It leads to a difficult to diagnose error message:
> some python apps which _do not require the opencv-python_ package will fail to start with a message "Bad OpenCV configuration".



I think for you to get the best experience is to download the packages, use 

```
anaconda login
anaconda upload package
conda config --add channels YOUR_CHANNEL_NAME
```

and then install the package "using standard conda tools".

I just don't see `import cv2` in any non-demo location.
https://github.com/search?q=repo%3Apthom%2Fimgui_bundle+cv2+language%3APython&type=code&l=Python

There are some references in the `pyi` file, but again, that is for development. People can install `opencv` which includes the python stuff.

The package name is different for conda-forge than the one on pypi.

pthom:
> `conda install libopencv` might not not enough I'm afraid. It leads to a difficult to diagnose error message:
> some python apps which _do not require the opencv-python_ package will fail to start with a message "Bad OpenCV configuration".

I cannot reproduce this issue anymore, even when deleting, then recreating an environment. I guess it was a one time issue.

pthom:
Hi Mark,

I worked on trying the recipe on my Mac (arm64), and this lead me to the following interesting findings, which I share with you below.

But first, thanks a lot for your hard work!
I know this is a huge effort, and I'm sorry that imgui_bundle is so complex to build, especially because - given the numerous dependencies it has - supporting multiple platforms becomes a bit of a challenge.

### Adaptation on recipes in order to produce conda packages for macos-arm64:
I was fighting to get the build to work on my Mac, and I found that the changes below are required for it to work on Mac arm processors:

**1. In the recipe imgui_bundle**
```diff
diff --git a/recipes/imgui-bundle/conda_build_config.yaml b/recipes/imgui-bundle/conda_build_config.yaml
index 60aa719400..a0efe03bc2 100644
--- a/recipes/imgui-bundle/conda_build_config.yaml
+++ b/recipes/imgui-bundle/conda_build_config.yaml
@@ -1,5 +1,6 @@
 # 10.15 is needed for std::filesystem::exists
-c_stdlib_version:           # [osx and x86_64]
-  - '10.15'                 # [osx and x86_64]
-MACOSX_SDK_VERSION:         # [osx and x86_64]
-  - '10.15'                 # [osx and x86_64]
+# 11.0 is needed for arm64
+c_stdlib_version:           # [osx]
+  - '11.0'                  # [osx]
+MACOSX_SDK_VERSION:         # [osx]
+  - '11.0'                  # [osx]
```

The reason is that support for ARM was introduced in the SDK 11.0 (and I could not get it to work with distinct numbers for x86_64 and arm64).

**2. in the recipe plutovg**
```diff
diff --git a/recipes/plutovg/conda-forge.yml b/recipes/plutovg/conda-forge.yml
index aca8bc3972..c4c26a1840 100644
--- a/recipes/plutovg/conda-forge.yml
+++ b/recipes/plutovg/conda-forge.yml
@@ -1,5 +1,3 @@
-build_platform:
-  osx_arm64: osx_64
 provider:
   linux_aarch64: azure
   linux_ppc64le: azure
```

This one is perhaps optional for the final deployment. However, it makes it impossible to compile on a arm Mac.


### Remaining patches to apply:
Below I studied the patches that are still needed to build imgui_bundle's recipe.

**patches that are not needed anymore**
*if using latest commit on imgui_bundle, those patches are not required anymore*
0001-Do-not-force-GLFW-to-be-recompiled.patch
0002-Do-not-force-static-freetype.patch
0003-Do-not-force-linking-to-non-existent-nlohmann_json-s.patch
0006-Enable-shared-libraries-for-conda-forge.patch

**0005-Find-glfw-correctly-for-conda-forge.patch**
Still needed. Do you have guidance on what I should do in the repo to make this patch unnecessary?

**0007-Remove-optional-dependencies-to-slim-package.patch**
Still needed (and I think it is a good idea to keep it like this for conda).

**0008-Adjust-compilation-of-immapp.patch (patch for plutovg)**
*This is the patch that aims at replacing the internal version of plutosvg with the conda-forge version.*

At the moment, some adaptations are needed:
    - only plutovg is provided as a recipe. Actually, we need plutovg **and** plutosvg.
    - plutosvg shall be compiled with support for freetype (PLUTOSVG_HAS_FREETYPE)
    - imgui shall be compiled with support for plutosvg (IMGUI_ENABLE_FREETYPE_PLUTOSVG)
I did a commit that simplifies the logic in hello_imgui_build_lib.cmake:
==> See https://github.com/pthom/hello_imgui/commit/1aab7451bc2f177d4d18a71743d03f4167bf4147#diff-cfb85203a084c5a2879783bfd64865bf9456d01c48d92515432a4b1e11dad515R502-R516

(although my advice would be perhaps to keep using a static internal version of plutosvg for the time being:
plutosvg is brand new, so that I suspect its API may vary so that it is not yet ready for a stable recipe).

**singel-patch**
I can build on my mac without it when using the latest version of imgui_bundle.


### status on my mac arm64
I could compile it, and all tests and demos pass! The demo which uses freetype also works!
This is if I compile with the latest imgui_bundle commit (which still uses a statically linked version of plutosvg).

---

PS: ping me when you want me to create a tag on imgui_bundle. I can create one (or several) version tag(s) for your tests (1.6.1 or example).

hmaarrfk:
> (although my advice would be perhaps to keep using a static internal version of plutosvg for the time being:
plutosvg is brand new, so that I suspect its API may vary so that it is not yet ready for a stable recipe).

Is your advice also to use the plutvg as an internally compiled static library?

One of the challenges is that we should add every license file, which starts to become hard to manage.

hmaarrfk:
> singel-patch
> I can build on my mac without it when using the latest version of imgui_bundle.

I don't really doubt you can build on your mac, the challenge is that you "leak symbols" by linking "PUBLIC". this causes future downstream projects to think they can access all the symbols from libraries such as OpenCV and GL "through" yours. 

When you are interesting in building a system, this makes it fragile.

So for maintenance, things should in general be linked "privately"

pthom:
> > (although my advice would be perhaps to keep using a static internal version of plutosvg for the time being:
> > plutosvg is brand new, so that I suspect its API may vary so that it is not yet ready for a stable recipe).
> 
> Is your advice also to use the plutvg as an internally compiled static library?
> 
> One of the challenges is that we should add every license file, which starts to become hard to manage.

Yes my advice was to link the pair Plutovg + plutosvg as an internal static library. But I understand your concern about license files

hmaarrfk:
> Yes my advice was to link the pair Plutovg + plutosvg as an internal static library. But I understand your concern about license files

I'll try to keep the unvendoring in clean patches (if i can) and we can keep this as an option if maintainability becomes a problem.

hmaarrfk:
> PS: I just created a tag v1.6.1 with the latest changes. Ping me later if you want me to create another.

One thing that might help more than a tag is an archive with all the git submodules.

Using git + conda-forge is kinda frowned uppon

PR #28461, opened 2024-12-05
CI: add missing config argument to _get_or_render_metadata

### LINKED ISSUE #28445: Build fails with `python_min is undefined` in PRs with multiple packages
### Bug:

At the moment, you can skip specifying `set python_min...` at the top of the yaml file if your package is strictly not Python 3.10+, but the CI fails if -
1. you do not specify the value for `python_min`, and
2. your PR has multiple interdependent packages.
 
After some digging, I am confident that the failure does not depend on the presence or absence of `noarch`, instead, it happens with every PR having multiple packages. For instance, see the failures and discussions in https://github.com/conda-forge/staged-recipes/pull/28405 and https://github.com/conda-forge/staged-recipes/pull/28431.

cc: @conda-forge/staged-recipes

d70-t:
Within `.ci_support/compute_build_graph.py`,`_get_or_render_metadata` is called from two sites, but when called from `_buildable`, the `config` argument was not passed along. Apparently, it's not possible to build multiple interdependent `noarch: python` packages in one PR without this `config` argument, but the build seems to work when applying the change as proposed in this PR.

I have to admit, that I'm new to conda-forge and didn't yet understand what exactly is passed along in this `config` argument, but I guessed that the `python_min` variable would be part of it.
I also have to admit, that I don't know if a PR to this repo would actually be the right way to fix this issue, I'd be glad for some guidelines if there'd be a better place.

Fixes #28445

### DIFF
--- a/.ci_support/compute_build_graph.py
+++ b/.ci_support/compute_build_graph.py
@@ -453,7 +453,7 @@ def _buildable(name, version, recipes_dir, worker, config, finalize):
                     packagename_re.match(dirname)))
     metadata_tuples = [m for path in likely_dirs
                         for (m, _, _) in _get_or_render_metadata(os.path.join(recipes_dir,
-                                                                 path), worker, finalize=finalize)]
+                                                                 path), worker, config=config, finalize=finalize)]
 
     # this is our target match
     ms = MatchSpec(" ".join([name, _fix_any(version, config)]))


d70-t:
The linter rightfully complains about me changing something else than a receipe, but I guess that was intended. As I'm new to conda-forge, I'd hope that I interpreted the PR instructions correctly when asking:

PR #28490, opened 2024-12-09
Add recipe for python-vesin

Luthaf:
Checklist
- [ ] ~If static libraries are linked in, the license of the static library is packaged.~
- [ ] ~Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md).~
- [x] Build number is 0.

### DIFF
--- a/recipes/vesin/meta.yaml
+++ b/recipes/vesin/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "vesin" %}
+{% set version = "0.3.0" %}
+
+package:
+  name: python-{{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+  sha256: bb325eb9f41e6632600379b1babd26264859d68b2bd1a8bff22ef23efbd8a689
+
+build:
+  skip: true  # [py<39]
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}
+    - cmake
+    - make  # [linux]
+  host:
+    - python
+    - setuptools
+    - wheel
+    - pip
+  run:
+    - python
+    - numpy
+
+test:
+  imports:
+    - vesin
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  summary: Computing neighbor lists for atomistic system
+  license: BSD-3-Clause
+  license_file: LICENSE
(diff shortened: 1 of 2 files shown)


Luthaf:
@conda-forge/help-python-c, ready for review! Just to be clear, there is no way to indicate that a project does not depend on Python version, but does need to be built once per os/architecture combination? `noarch: python` seems to assume pure Python code, while this project also contains C++, but uses ctypes to call the functions instead of going through `libpython`.

recipes/vesin/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}
+    - cmake
+    - make  # [linux]
+  host:
+    - python
```
xhochy, line 24: This is a bit nasty, but may solve your problem by putting everything into build. I may lead to problems with cross-compiling though.
```suggestion
    - python >=3.9
```

recipes/vesin/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - setuptools
+    - wheel
+    - pip
+  run:
+    - python
```
xhochy, line 29:
```suggestion
    - python >=3.9
```

Luthaf: Should this be `python >=3.9` or `python >={{ python_min }}`?

recipes/vesin/meta.yaml
Luthaf: I'm not sure I understand? What I'm trying to achieve is reduce the number of variants built (i.e. only build for `os x arch` instead of `os x arch x python-version`). You are saying that not having a `python` dependency in `host` would do this?

I think I'll need to cross compile to build packages for linux-arm64 and macos-arm64, so maybe this is something I could experiment with in a branch of the feedstock?

xhochy: Yes, maybe do that and let's start with building out for every Python version. My suggestion should lead to what you want but for cross-compiled versions, it will be more complicated.

We should soon have support for `abi3`-based packages. You could wait for that and use that approach: https://github.com/conda/ceps/pull/86 Eventhough in your case, it isn't `abi3` but a similar approach.

Luthaf: Thanks for the link. Yes `abi3` seems like a good idea here, it's not quite the same mechanism but should yield the same results anyway!

recipes/vesin/meta.yaml
Luthaf: Sounds like the linter does not like either of these and fails with `Non noarch packages should have python requirement without any version constraints.`

xhochy (approved):

PR #28587, opened 2024-12-15
add aurelia-cli, framework7-cli, preact-cli, pwa-builder

danielnachun:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/aurelia-cli/build.sh
+++ b/recipes/aurelia-cli/build.sh
@@ -1,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
+# Create package archive and install globally
+npm pack --ignore-scripts
+npm install -ddd \
+    --global \
+    --build-from-source \
+    ${SRC_DIR}/${PKG_NAME}-${PKG_VERSION}.tgz
+
+# Create license report for dependencies
+pnpm install
+pnpm-licenses generate-disclaimer --prod --output-file=third-party-licenses.txt
--- a/recipes/aurelia-cli/build.sh
+++ b/recipes/aurelia-cli/build.sh
@@ -4,11 +4,13 @@ set -o xtrace -o nounset -o pipefail -o errexit
 
 # Create package archive and install globally
 if [[ "${target_platform}" == "osx-arm64" ]]; then
-  export npm_config_arch="arm64"
+    export npm_config_arch="arm64"
 fi
 
-rm $PREFIX/bin/node
-ln -s $BUILD_PREFIX/bin/node $PREFIX/bin/node
+if [[ "${build_platform}" != "${target_platform}" ]]; then
+    rm $PREFIX/bin/node
+    ln -s $BUILD_PREFIX/bin/node $PREFIX/bin/node
+fi
 
 npm pack --ignore-scripts
 npm install -ddd \
--- a/recipes/aurelia-cli/build.sh
+++ b/recipes/aurelia-cli/build.sh
@@ -3,6 +3,13 @@
 set -o xtrace -o nounset -o pipefail -o errexit
 
 # Create package archive and install globally
+if [[ "${target_platform}" == "osx-arm64" ]]; then
+  export npm_config_arch="arm64"
+fi
+
+rm $PREFIX/bin/node
+ln -s $BUILD_PREFIX/bin/node $PREFIX/bin/node
+
 npm pack --ignore-scripts
 npm install -ddd \
     --global \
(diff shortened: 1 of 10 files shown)


danielnachun:
@conda-forge/staged-recipes @conda-forge/help-nodejs This is now ready for review.

Unlike the previous PRs, these NodeJS recipes do have some native code, and Windows has been disabled for now due to the complexities of building native code through `npm` on that platform.

recipes/aurelia-cli/build.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
```
xhochy, line 4: nodejs package with compiled code should have this boilerplate. Maybe we need to put this into a `{{ compiler('nodejs') }}` at one point. 

```suggestion
if [[ "${target_platform}" == "osx-arm64" ]]; then
  export npm_config_arch="arm64"
fi
# Don't use pre-built gyp packages
export npm_config_build_from_source=true

rm $PREFIX/bin/node
ln -s $BUILD_PREFIX/bin/node $PREFIX/bin/node
```

recipes/aurelia-cli/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - make
+    - nodejs
+    - pnpm
+    - pnpm-licenses
```
xhochy, line 23: If there is compiled code, nodejs should also be in build
```suggestion
    - pnpm-licenses
  host:
    - nodejs
```

recipes/aurelia-cli/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  number: 0
+  skip: True  # [win]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - make
+    - nodejs
```
xhochy, line 21:
```suggestion
    - nodejs  # [build_platform != target_platform]
```

xhochy (changes requested):
Only commented on the first package but the review applies to all.

recipes/aurelia-cli/build.sh
danielnachun: Perhaps a `${NPM_ARGS}` variable would make sense for this case?  I can propose what should go in there in the `nodejs` feedstock for future submissions, though for now let's just set the boilerplate.

Also, according to here: https://github.com/swc-project/node-swc/issues/12, adding the flag `--build-from-source` to `npm install` is equivalent to setting `export npm_config_build_from_source=true`.  I have no strong preference for using a flag versus an environment variable, as long as we are consistent, but it's probably redundant to use both in these recipes.

I'll also note here that `--build-from-source` doesn't actually always guarantee that native **dependencies** will be built from source.  I carefully chose these packages because their native dependencies will build from source if the necessary build tools and other dependencies are available.  I currently have about 25 NodeJS packages where I haven't been to force rebuilding some native dependencies from source, though I'm steadily making progress.  I'm happy to share more info about that somewhere if it would be useful.

xhochy: > Perhaps a ${NPM_ARGS} variable would make sense for this case?

Yes, that sounds better. The `compiler('nodejs')` implementation would have the benefit that it could also do the symlinking for us.

recipes/preact-cli/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: preact-cli
+  version: {{ version }}
+
+source:
+  url: https://registry.npmjs.org/preact-cli/-/preact-cli-{{ version }}.tgz
+  sha256: 029392799c88e65b2004c6569c7046fafa6547c83e4df265ad8a4675d95ba173
+
+build:
+  number: 0
+  skip: True  # [win]
```
xhochy, line 13: On the feedstock, you should be able to have a `skip: true  # [nodejs != '20']` and can then remove all of the pins below.

xhochy (approved):

xhochy:
@conda-forge-admin please relint

recipes/preact-cli/meta.yaml
danielnachun: This worked here as well!  I've also applied this to the other PR

xhochy (approved):

PR #28588, opened 2024-12-15
add awk-language-server, gatsby-cli, spago

danielnachun:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/gatsby-cli/meta.yaml
+++ b/recipes/gatsby-cli/meta.yaml
@@ -14,9 +14,9 @@ build:
 
 requirements:
   build:
-    - {{ compiler('c')}}
-    - {{ compiler('cxx')}}
-    - {{ stdlib('c')}}
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
     - findutils
     - jq
     - make
@@ -27,9 +27,9 @@ requirements:
     - nodejs
     - xsel  # [linux]
 
-tests:
-  - script:
-      - gatsby --help
+test:
+  commands:
+    - gatsby --help
 
 about:
   license: MIT
@@ -37,10 +37,9 @@ about:
     - LICENSE
     - third-party-licenses.txt
   summary: The best React-based framework with performance, scalability and security built in.
-  homepage: https://github.com/gatsbyjs/gatsby-cli
-  repository: https://github.com/gatsbyjs/gatsby-cli
+  home: https://github.com/gatsbyjs/gatsby-cli
+  dev_url: https://github.com/gatsbyjs/gatsby-cli
 
 extra:
   recipe-maintainers:
     - danielnachun
-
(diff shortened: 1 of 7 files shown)


danielnachun:
@conda-forge/staged-recipes @conda-forge/help-nodejs This is now ready for review.

Unlike the previous PRs, these NodeJS recipes do have some native code, and Windows has been disabled for now due to the complexities of building native code through `npm` on that platform.

xhochy (changes requested):
See https://github.com/conda-forge/staged-recipes/pull/28587#pullrequestreview-2509977842

danielnachun:
@xhochy `awk-language-server` has a hard dependency on NodeJS 20 but it seems it's not respecting that now that we've put NodeJS in both build and host requirements and build fails because the NodeJS 22 is being installed into the build dependencies.  Should I be specifying the version constraints differently?

danielnachun:
Actually the problem is that the change suggested in https://github.com/conda-forge/staged-recipes/pull/28587#pullrequestreview-2509977842 replaces the host NodeJS with the build NodeJS unconditionally.  I think we should only do this when `target_platform == "osx-arm64"` or perhaps more generically when `build_platform != target_platform`.  Let me know what you think.

xhochy:
Yes, we should only do this if we're cross-compiling. This worked in the other recipes as I had `nodejs` in build and host (both) unconditionally.

recipes/gatsby-cli/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  number: 0
+  skip: True  # [win]
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - findutils
+    - gcc  # [linux]
```
xhochy, line 21: You should never need to depend on `gcc` in recipes. If really required, add an explanative comment.
```suggestion
```

xhochy (changes requested):

recipes/gatsby-cli/meta.yaml
danielnachun: The issue here is that the the `msgpackr-extract`, a dependency of `gatsby-cli`, hardcodes `gcc` as the compiler executable name rather than using CC on Linux: https://github.com/kriszyp/msgpackr-extract/blob/db8f5b97e7aaab0bcad95f10582915d417153689/binding.gyp#L3.  

Would it be better in this scenario to use symlink from `CC` to `gcc` in `$BUILD_PREFIX/bin`?  I'm not sure what our established best practice is for this problem.

xhochy: > Would it be better in this scenario to use symlink from CC to gcc in $BUILD_PREFIX/bin? I'm not sure what our established best practice is for this problem.

Yes, that would at least be my preferred solution.

danielnachun: Sorry I fell behind on this, see the last commit for the requested change!

danielnachun:
@xhochy can you take a look now? Thanks!

xhochy (approved):

PR #28593, opened 2024-12-16
Add Python package 'crystmatch'

Fangcheng-Wang:
I am the author and the only maintainer of `crystmatch`, a Python package introduced in my paper [FC Wang, QJ Ye, YC Zhu, and XZ Li, Physical Review Letters 132, 086101 (2024)](https://arxiv.org/abs/2305.05278).

Checklist
- [x] Build number is 0.

I guarantee that I will maintain `crystmatch` continuously.


recipes/crystmatch/meta.yaml
```diff
@@ -0,0 +1,55 @@
+{% set name = "crystmatch" %}
+{% set version = "1.1.2" %}
+{% set python_min = "3.9" %}
```
synapticarbors, line 3: You don't need to set this since it's already the global pinned version that conda-forge uses.

recipes/crystmatch/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - setuptools >=42
+    - wheel
+    - numpy
```
synapticarbors, line 25: Just as a general note, all of the requirements below `wheel` (except for `pip`), are only needed because you put them in the `requires` section under `build-system` in your upstream `pyproject.toml`. Looking at the `setup.py` they do not appear to be strictly required to "build" the package. They are only really runtime dependencies. If I'm correct about this, it might be beneficial in a future release to fix that upstream.

recipes/crystmatch/meta.yaml
Fangcheng-Wang: Thank you for your kind advice! I have removed these lines now.

recipes/crystmatch/meta.yaml
Fangcheng-Wang: You are correct. I removed these lines in `meta.yaml` as well as those in `pyproject.toml`. Thanks!

PR #28625, opened 2024-12-17
Add `marray-python` recipe

mdhaber:
I (mdhaber) am willing to be listed in the maintainer section.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/marray-python/meta.yaml
+++ b/recipes/marray-python/meta.yaml
@@ -1,5 +1,6 @@
 {% set name = "marray-python" %}
 {% set version = "0.0.5" %}
+{% set python_min = "3.10" %}
 
 package:
   name: {{ name|lower }}
--- a/recipes/marray-python/meta.yaml
+++ b/recipes/marray-python/meta.yaml
@@ -1,4 +1,4 @@
-{% set name = "marray" %}
+{% set name = "marray-python" %}
 {% set version = "0.0.5" %}
 
 package:
--- a/recipes/marray-python/meta.yaml
+++ b/recipes/marray-python/meta.yaml
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/marray-{{ version }}.tar.gz
+  url: https://pypi.org/packages/source/m/marray/marray-{{ version }}.tar.gz
   sha256: 09fa2325fe6fca2ce005aaa250152a95ae5e61b341d8d443600a78b95bfe06b0
 
 build:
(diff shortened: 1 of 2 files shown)


mdhaber:
@bluescarni it [looks like](https://github.com/conda-forge/marray-feedstock/blob/8f313282cbb92d4b075f7eebfedd2009ea32873d/recipe/meta.yaml#L48) you are the recipe maintainer for [`marray`](https://github.com/conda-forge/marray-feedstock/tree/main). We (some SciPy/NumPy/XArray maintainers) are developing a package that adds support for masks (missing data) to any [Python Array API Standard](https://data-apis.org/array-api/latest/) compliant array backend, and we have been calling it `marray`, too. It is currently released on PyPI as such, and we were hoping to release it on conda-forge, too, but it looks like we have a name conflict. Is your project still active? If not, could we use the name in the future?

lucascolley:
cc @bjoern-andres also

lucascolley:
Would you like to go ahead with naming the feedstock and package `python-marray` @mdhaber? I'll make core aware of the situation on Zulip in case they have any advice

recipes/marray-python/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "marray-python" %}
+{% set version = "0.0.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/marray-{{ version }}.tar.gz
```
mdhaber, line 9:
```suggestion
  url: https://pypi.org/packages/source/m/marray/marray-{{ version }}.tar.gz
```

recipes/marray-python/meta.yaml
```diff
@@ -0,0 +1,43 @@
+{% set name = "marray-python" %}
+{% set version = "0.0.5" %}
```
mdhaber, line 2:
```suggestion
{% set version = "0.0.5" %}
{% set python_min = "3.10" %}
```

recipes/marray-python/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+{% set python_min = "3.10" %}
```
mdhaber, line 3: @lucascolley If I'm reading CI correctly, looks like the previous CI run was trying to install Python 3.9 but our `pyproject.toml` insists on `>=3.10`. I'm not sure if it really matters, but we probably should leave `pyproject.toml` alone since we're well past Python 3.9 according to SPEC 0. However, we probably shouldn't require a higher minimum Python version than SciPy does if we want to use this there. Does changing this here look OK?

lucascolley: Seems fine to only support SPEC-0 windows to me

lucascolley:

lucascolley:

PR #28636, opened 2024-12-18
Add UMNN

matthewfeickert:
* Add UMNN v1.71 by Antoine Wehenkel (@AWehenkel).
   - c.f. https://github.com/AWehenkel/UMNN
* c.f. https://github.com/bayesiains/nflows/issues/82

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/umnn/meta.yaml
+++ b/recipes/umnn/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "umnn" %}
+{% set version = "1.71" %}
+{% set python_min = "3.7" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/umnn-{{ version }}.tar.gz
+  sha256: bdd41d941a5d904e2217a960a9584922afad8068304976dc6fb0245e4f834996
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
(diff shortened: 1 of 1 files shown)


matthewfeickert:

recipes/umnn/meta.yaml
```diff
@@ -0,0 +1,46 @@
+{% set name = "umnn" %}
+{% set version = "1.71" %}
+{% set python_min = "3.7" %}
```
ocefpaf, line 3: Be aware that conda-forge min supported Python is 3.9.

matthewfeickert: Thanks for the fast merge, @ocefpaf!

Indeed, that's why I set the `python_min` here to try to be as close to PyPI as could be supported. My understanding (please correct me if I'm wrong here) is that the current global pins (so supported versions) are [Python 3.9 to 3.12](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L826-L841) and that these are the default supported versions, but, if you wanted to go outside those supported defauts that's still okay for `noarch: python` packages as long as you don't expect any support there. Is that correct?

ocefpaf: It is correct. The warning is mostly "if this breaks, it is unlikely we'll supported it." (It should work though.)

matthewfeickert: Great and thanks very much for the info! I really appreciate the work that you and the other core team members do to keep everything running so smoothly.

PR #28644, opened 2024-12-19
Add a recipe for openlb 1.7

martin-g:
Checklist
- [x] Build number is 0.


martin-g:
@conda-forge/help-c-cpp  this is ready for review

carterbox (changes requested):
Thanks for following our naming convention by appending "-static" to the package name. However, we have a strong preference for dynamic linking on this channel unless there is significant performance uplift from static linking. Could you please patch this project to generate shared libraries instead?

Also, it doesn't seem like you are shipping any header files? How are downstream packages to link against this library without header files or find this library without a package configuration file?

martin-g:
@carterbox Thank you for the review! I've applied your suggestions! Please take a new look when you have a moment!

martin-g:
The linter issue is because I switched to recipe v2 (rattler-build).

xhochy:
Let's see whether this fixes the linter.

recipes/openlb/recipe.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - zlib
+    - tinyxml
+  run:
+  run_exports:
+    - ${{ pin_subpackage(name, upper_bound='x') }}
+
+tests:
+  - script:
+      - if: unix
+        then: test -f ${PREFIX}/lib/libolbcore.so
```
xhochy, line 41: Shouldn't this be?
```suggestion
        then: test -f ${PREFIX}/lib/libolbcore${SHLIB_EXT}
```

martin-g: Thanks!
Fixed it!

xhochy (approved):

Issue #28111, opened 2024-11-05
Package request: `libnvshmem`


jakirkham:
### Package name

`libnvshmem`

### Package version

Newest

### Package website

Available via `redist` linked below

### Package availability

https://developer.download.nvidia.com/compute/nvshmem/redist/

### Additional comments

cc @conda-forge/cuda (who will pick this up)

### Package is not available

- [x] The package is not available on conda-forge.

### No previous issues or open PRs

- [x] No previous issue exists and no PR has been opened.

carterbox:
This library seems to be open source and thus compileable with the channels toolchain. Is there a reason to use the redist?

carterbox:
Notes about publishing this recipe:

# Source Hash

This package is source-available, but the source download from NVIDIA is hidden behind a developer login, and the download is unique to the user. Each time you download, the hash of the file changes, so it is not suitable for building automatically in a feedstock.

# Dependencies

Some of the optional dependencies are not yet available on the channel including:

- [x] libfabric >= 1.15.0.0  https://github.com/conda-forge/staged-recipes/pull/27988
- [ ] PMIx >= 3.1.5 https://github.com/conda-forge/staged-recipes/pull/28225
- [ ] shmem?
- [ ] gdrcopy Maybe a kernel-driver is required which we can't ship
- [ ] Others... mellanox, nvlink, etc

carterbox:
TODO: Contact library team to discuss what features they think are appropriate for this conda package.

jakirkham:
cc @billysuh7 (who is picking this up)

sisodia1701:
Build shared with nvshmem and waiting for feedback.

carterbox:
[Done.](https://github.com/conda-forge/libnvshmem-feedstock)

PR #28647, opened 2024-12-19
nvshmem 3.1.7

billysuh7:
Checklist
- [X] Build number is 0.

Xref: https://github.com/conda-forge/staged-recipes/issues/28111

Please also refer to https://docs.nvidia.com/nvshmem/release-notes-install-guide/install-guide/abstract.html for NVSHMEM requirements and how much this package were able/unable to meet.


recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+        - {{ pin_compatible("cuda-version", min_pin="x", max_pin="x") }}
+        - {{ pin_subpackage("libnvshmem-dev", exact=True) }}
+        - nccl
+        - ucx
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+    test:
+      commands:
+        - test -f $PREFIX/lib/libnvshmem_device.a
+        - test -f $PREFIX/lib/libnvshmem.a
```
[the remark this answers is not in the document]
seth-howell: Can we also add the bitcode library now? That might be easier than doing another pass.The first 3.2 builds are in URM now:

https://urm.nvidia.com/ui/repos/tree/General/sw-nvshmem-generic-local/NVSHMEM/gpu_comms_cuda12.0_compatibility/3.2.2/libnvshmem_cuda12-linux-x86_64-3.2.2.tar.gz

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,204 @@
...
+{% set extension = "tar.xz" %}
+
+{% set soname = version.split(".")[0] %}
+{% set cuda_major = environ.get("cuda_compiler_version", "12.6").split(".")[0] %}
+
+package:
+  name: libnvshmem-split
+  version: {{ version }}
+
+source:
```
seth-howell, line 13: I have checked to make sure the files are in the required places.

Let me run an additional test to make sure the binaries can find the libraries correctly.

recipes/libnvshmem/meta.yaml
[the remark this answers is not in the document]
billysuh7: Size is about 60MB which isn't too too bad, but I understand conda generally discourage static packages so I guess I'll take 'em out. About the bitcode - this PR is for nvshmem 3.1.7, so once this has gone through vetting and is released for conda, I can work on the next version and include bitcode at that time.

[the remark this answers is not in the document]
seth-howell: Sorry, just libnvshmem.a. libnvshmem_device.a likely has to stay for jit compiling. This is something that's been communicated to me from cuBLASMp.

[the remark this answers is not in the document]
billysuh7: I see. Anyway, I just realized that out of 130MB, the static conda package only takes up 12MB so it is inconsequential either way :) For now I'll take out libnvshmem.a

recipes/libnvshmem/meta.yaml
seth-howell: I have confirmed the following:

Various NVSHMEM Bootstraps can be used, as long as the LD_LIBRARY_PATH is properly specified to include the desired packages.
The Bootstraps can be found automatically by the library
The performance tests in src can be compiled against NVSHMEM. The kitmaker libraries work properly.

From my end, the configuration is good.

recipes/libnvshmem/meta.yaml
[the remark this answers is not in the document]
adibbley: Typically we have been fine adding any static library to `-static` outputs, was there another reason we would take out `libnvshmem.a` for now?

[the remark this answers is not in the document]
seth-howell: libnvshmem.a is a deprecated library and not used. libnvshmem_device.a is what will be used by our users. I don't think the extra space is worth it. I would prefer to not change the recipe right now, and 
We have a bitcode library (libnvshmem_device.bc) coming in 2.1. I believe it's something Billy has planned for an additional layer to come later.

recipes/nvshmem/meta.yaml
```diff
@@ -0,0 +1,39 @@
+{% set version = "3.1.7" %}
```
carterbox, line 1: Why a separate recipe for this metapackage? Are there plans to add additional packages to the nvshmem umbrella?

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+  sha256: 4e03e40d16770a5bdbcefeb7f18579bdfc07aaa7b09a4825a7ab9b5122f69567  # [linux64 and (cuda_compiler_version or "").startswith("11")]
+  sha256: 128c62dd19926f58d1ff69290917dd50f3fb1690567f52d3965cd8fbfb9b2412  # [linux64 and (cuda_compiler_version or "").startswith("12")]
+  sha256: 03e4404e888609bc98f496453896acc34c9ca7bbc39cc32b41d677bcaab71c49  # [aarch64 and (cuda_compiler_version or "").startswith("12")]
+
+build:
+  number: 0
+  skip: true  # [not (linux64 or aarch64)]
+  skip: true  # [cuda_compiler_version in (None, "None") or (aarch64 and (cuda_compiler_version or "").startswith("11"))]
+
+outputs:
```
carterbox, line 24:
```suggestion
build:
  number: 0
  skip: true  # [not (linux64 or aarch64)]
  skip: true  # [cuda_compiler_version in (None, "None") or (aarch64 and (cuda_compiler_version or "").startswith("11"))]

requirements:
  build:
    - cf-nvidia-tools 1  # [linux]

outputs:
```

recipes/libnvshmem/build-libnvshmem.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p $PREFIX/lib/cmake/nvshmem/
```
carterbox, line 5:
```suggestion
set -e

check-glibc bin/* lib/*

mkdir -p $PREFIX/lib/cmake/nvshmem/
```
https://github.com/conda-forge/cf-nvidia-tools-feedstock/tree/main/recipe

recipes/libnvshmem/build-libnvshmem.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
carterbox, line 1: Rename this file build.sh

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+        - "*libfabric1*" # omitted as dependency - used for both Slingshot and EFA NICs. They have a custom libfabric installation and plugin
+        - "*libmlx5*"    # omitted as dependency - needed for ibgda and ibdevx transport, installed as part of the Mellanox OFED
+        - "*libmpi*"     # omitted as dependency - part of openmpi
+        - "*libpmix*"    # omitted as dependency - part of openmpi
+        - "*liboshmem*"  # omitted as dependency - part of openmpi
+        - "*rdma-core*"  # omitted as dependency - MOFED replaces rdma-core
+      run_exports:
+        - {{ pin_subpackage("libnvshmem" ~ soname, max_pin="x") }}
+      script: |
+        cp -rv lib $PREFIX/
```
carterbox, line 40:
```suggestion
```

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+      missing_dso_whitelist:
+        - "*libcuda.so*" # driver
+        - "*libnvshmem_host.so.{{ soname }}" # avoids: ERROR (libnvshmem-dev,bin/perftest/device/pt-to-pt/shmem_put_bw): lib/libnvshmem_host.so.3 not found in any packages. We DO test for its existence though.
+        - "*libmpi*"
+        - "*libpmix*"
+        - "*liboshmem*"
+    run_exports:
+      - {{ pin_subpackage("libnvshmem" ~ soname, max_pin="x") }}
+    script: build-libnvshmem.sh
+    script_interpreter: /bin/bash
```
carterbox, line 104:
```suggestion
```

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+        - test -L $PREFIX/lib/nvshmem_transport_ibdevx.so
+        - test -L $PREFIX/lib/nvshmem_transport_ibgda.so
+        - test -L $PREFIX/lib/nvshmem_transport_ibrc.so
+        - test -L $PREFIX/lib/nvshmem_transport_libfabric.so
+        - test -L $PREFIX/lib/nvshmem_transport_ucx.so
+
+  - name: libnvshmem-static
+    build:
+      script: |
+        cp -rv lib/libnvshmem_device.a $PREFIX/lib/
```
carterbox, line 167:
```suggestion
```

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+  - name: libnvshmem-dev
+    build:
+      missing_dso_whitelist:
+        - "*libcuda.so*" # driver
+        - "*libnvshmem_host.so.{{ soname }}" # avoids: ERROR (libnvshmem-dev,bin/perftest/device/pt-to-pt/shmem_put_bw): lib/libnvshmem_host.so.3 not found in any packages. We DO test for its existence though.
+        - "*libmpi*"
+        - "*libpmix*"
+        - "*liboshmem*"
+    run_exports:
+      - {{ pin_subpackage("libnvshmem" ~ soname, max_pin="x") }}
```
carterbox, line 102:
```suggestion
      run_exports:
        - {{ pin_subpackage("libnvshmem" ~ soname, max_pin=None) }}
```
I think this needs to be indented one level?

Also, the max_pin should be None if the project actually uses the soname to indicate ABI compatability.

recipes/nvshmem/meta.yaml
carterbox: Technically, this package is redundant because you will get the same deps from installing libnvshmem-static.

carterbox (changes requested):

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+        - {{ compiler('cuda') }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_compiler_version }}
+        - cuda-cudart-dev   # [(cuda_compiler_version or "").startswith("12")]
+        - cudatoolkit       # [(cuda_compiler_version or "").startswith("11")]
+        - ucx
+      run:
+        - {{ pin_compatible("cuda-version", min_pin="x", max_pin="x") }}
```
carterbox, line 58: Enhanced CUDA compatibility is not implemented correctly for this output. You need to ignore exports from the CUDA compiler.

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+    script: build-libnvshmem.sh
+    script_interpreter: /bin/bash
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_compiler_version }}
```
carterbox, line 112: Please check that cuda_compiler_version is defined when you haven't used the compiler package in the build requirements.

recipes/libnvshmem/meta.yaml
billysuh7: Done. Thanks!

recipes/libnvshmem/build-libnvshmem.sh
billysuh7: Done. Thanks!

recipes/libnvshmem/meta.yaml
billysuh7: I removed `run_exports`: section only, but I still need the `script:` section for this particular package

recipes/libnvshmem/meta.yaml
billysuh7: Please see my earlier comment on this

recipes/libnvshmem/meta.yaml
billysuh7: I am keeping this section for the same reason, but modified the pattern to add all static libs in order for cmake test to pass. Otherwise I'd get 

```
CMake Error at /home/conda/staged-recipes/build_artifacts/libnvshmem-split_1739198615629/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/cmake/nvshmem/NVSHMEMTargets.cmake:190 (message):
    The imported target "nvshmem::nvshmem" references the file
  
       "/home/conda/staged-recipes/build_artifacts/libnvshmem-split_1739198615629/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/lib/libnvshmem.a"
  
    but this file does not exist. 
```

recipes/libnvshmem/meta.yaml
billysuh7: Thank you for the headsup! Hopefully fixed now. The `cuda-cudart` dependency is no longer constrained to any version, leaving that decision up to `cuda-version`.

recipes/libnvshmem/meta.yaml
billysuh7: Done

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,203 @@
...
+        - test -L $PREFIX/lib/nvshmem_bootstrap_pmi.so
+        - test -L $PREFIX/lib/nvshmem_bootstrap_pmi2.so
+        - test -L $PREFIX/lib/nvshmem_bootstrap_pmix.so
+        - test -L $PREFIX/lib/nvshmem_bootstrap_shmem.so
+        - test -L $PREFIX/lib/nvshmem_bootstrap_uid.so
+        - test -L $PREFIX/lib/nvshmem_transport_ibdevx.so
+        - test -L $PREFIX/lib/nvshmem_transport_ibgda.so
+        - test -L $PREFIX/lib/nvshmem_transport_ibrc.so
+        - test -L $PREFIX/lib/nvshmem_transport_libfabric.so
+        - test -L $PREFIX/lib/nvshmem_transport_ucx.so
```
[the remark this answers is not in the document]
billysuh7: Done

recipes/libnvshmem/meta.yaml
billysuh7: Fixed indentation and set `max_pin` to `None`. But I wonder how we can enforce ABI compatibility strictly ? Are leaving the responsibility to the end user to understand what they are doing based on the `soname` ?

recipes/libnvshmem/build-libnvshmem.sh
carterbox: Typically, I use a global `build.sh` to copy everything to $PREFIX, then use `files:` in each output to select which files go to each output. Here you are using `build: script:` in each output to move the files for each output separately. This is allowed, but unexpected. It's also irregular because two outputs are using scripts defined in the meta.yaml and one is using a separate bash script.

https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#specifying-files-to-include-in-output

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,225 @@
...
+        - "*libmpi*"     # omitted as dependency - part of openmpi
+        - "*libpmix*"    # omitted as dependency - part of openmpi
+        - "*liboshmem*"  # omitted as dependency - part of openmpi
+        - "*rdma-core*"  # omitted as dependency - MOFED replaces rdma-core
+      script: |
+        cp -rv lib $PREFIX/
+    files:
+      - lib/libnvshmem_host.so.*
+      - lib/nvshmem_bootstrap*.so.*
+      - lib/nvshmem_transport*.so.*
```
carterbox, line 45:
```suggestion
      script: |
        mkdir -p $PREFIX/lib  # maybe two lines 
        cp -v lib/*.so.* $PREFIX/lib/
```
This is redundant. You should be able to do this in one or two line(s).

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,225 @@
...
+        - test -L $PREFIX/lib/nvshmem_transport_ibrc.so
+        - test -L $PREFIX/lib/nvshmem_transport_libfabric.so
+        - test -L $PREFIX/lib/nvshmem_transport_ucx.so
+
+  - name: libnvshmem-static
+    build:
+      script: |
+        cp -rv lib/libnvshmem*.a $PREFIX/lib/
+    files:
+      - lib/libnvshmem*.a
```
carterbox, line 177:
```suggestion
      script: |
        cp -rv lib/libnvshmem*.a $PREFIX/lib/
```
See other comment about possibly needing to make $PREFIX/lib before you try to copy into it.

recipes/libnvshmem/build-libnvshmem.sh
```diff
@@ -0,0 +1,16 @@
...
+$BUILD_PREFIX/bin/check-glibc bin/* lib/* bin/examples/* bin/perftest/device/coll/* bin/perftest/device/pt-to-pt/* bin/perftest/host/coll/* bin/perftest/host/init/* bin/perftest/host/pt-to-pt/*
+
+mkdir -p $PREFIX/lib/cmake/nvshmem/
+
+cp -rv bin $PREFIX/
+cp -rv include/ $PREFIX/
+cp -rv lib/cmake/ $PREFIX/lib/
+cp -rv lib/libnvshmem_host.so $PREFIX/lib
+cp -rv lib/nvshmem_bootstrap*.so $PREFIX/lib
+cp -rv lib/nvshmem_transport*.so $PREFIX/lib
```
carterbox, line 14:
```suggestion
cp -rv lib/*.so $PREFIX/lib
```
There should only be shared libraries that we want to install in the source tar, right?

recipes/libnvshmem/build-libnvshmem.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -e
+
+$BUILD_PREFIX/bin/check-glibc bin/* lib/* bin/examples/* bin/perftest/device/coll/* bin/perftest/device/pt-to-pt/* bin/perftest/host/coll/* bin/perftest/host/init/* bin/perftest/host/pt-to-pt/*
```
carterbox, line 5: "$BUILD_PREFIX/bin/check-glibc" If you need to full path for this tool to work it's because you're using this script in an unintended way. I suggest moving this call to check-glibc into `build.sh` and moving the dependency on `cf-nvidia-tools` to the top-level requirements.

The install steps can stay in this script.

carterbox (changes requested):

recipes/libnvshmem/meta.yaml
carterbox: If we are building libnvshmem version 1.2.3 then this template would be de-templated to something like:

```
- libnvshmem1 >=1.2.3
```

The `soname` is already a promise of ABI compatibility because of the way that unix searches for shared libraries. Any library named `foo.so.4` is assumed to be backward compatible with any earlier `foo.so.4`. Therefore, any library looking for libnvshmem5 can run with the latest version of whatever libnvshmem5 is available. This is why there is no upper bound.

The `libnvshmem3` package is not intended to be installed directly by humans. All downstream users should be installing `libnvshmem-dev`. If you think end users of this package want to be able to install the shared libraries directly without any of the headers or config, we can add a non-versioned `libnvshmem` output.

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,225 @@
...
+  sha256: 128c62dd19926f58d1ff69290917dd50f3fb1690567f52d3965cd8fbfb9b2412  # [linux64 and (cuda_compiler_version or "").startswith("12")]
+  sha256: 03e4404e888609bc98f496453896acc34c9ca7bbc39cc32b41d677bcaab71c49  # [aarch64 and (cuda_compiler_version or "").startswith("12")]
+
+build:
+  number: 0
+  skip: true  # [not (linux64 or aarch64)]
+  skip: true  # [cuda_compiler_version in (None, "None") or (aarch64 and (cuda_compiler_version or "").startswith("11"))]
+
+outputs:
+  - name: libnvshmem{{ soname }}
```
carterbox, line 25:
```suggestion
  - name: libnvshmem
    requirements:
      run:
        - {{ pin_subpackage("libnvshmem" ~ soname, max_pin=None) }}

  - name: libnvshmem{{ soname }}
```
If you want a non-ABI-versioned metapackage. part 1 of 2.

recipes/libnvshmem/meta.yaml
```diff
@@ -0,0 +1,225 @@
...
+        - {{ compiler('cuda') }}  # [(cuda_compiler_version or "").startswith("12")]
+        - cuda-cudart-dev         # [(cuda_compiler_version or "").startswith("12")]
+      missing_dso_whitelist:
+        - "*libcuda.so*" # driver
+        - "*libnvshmem_host.so.{{ soname }}" # avoids: ERROR (libnvshmem-dev,bin/perftest/device/pt-to-pt/shmem_put_bw): lib/libnvshmem_host.so.3 not found in any packages. We DO test for its existence though.
+        - "*libmpi*"
+        - "*libpmix*"
+        - "*liboshmem*"
+      run_exports:
+        - {{ pin_subpackage("libnvshmem" ~ soname, max_pin=None) }}
```
carterbox, line 107:
```suggestion
      run_exports:
        - {{ pin_subpackage("libnvshmem", max_pin=None) }}
        - {{ pin_subpackage("libnvshmem" ~ soname, max_pin=None) }}
```
If you want a non-ABI-versioned metapackage. part 2 of 2.

recipes/libnvshmem/meta.yaml
billysuh7: Thanks for the explanation. This helps!

recipes/libnvshmem/meta.yaml
billysuh7: I've decided to have global `build.sh` as you suggested so this has been taken out

recipes/libnvshmem/build-libnvshmem.sh
billysuh7: I will go with global `build.sh` after experimenting with both ways, global vs per-output install to `$PREFIX`. Thanks!

recipes/libnvshmem/meta.yaml
billysuh7: This will be replaced with global `build.sh`

recipes/libnvshmem/build-libnvshmem.sh
billysuh7: Yes - and I will have global `build.sh `with greatly simplified file patterns to copy

recipes/libnvshmem/meta.yaml
billysuh7: I won't bother with this after your explanation of how OS searches for ABI libs.

recipes/libnvshmem/meta.yaml
billysuh7: I won't bother with this after your explanation of how OS searches for ABI libs.

PR #28721, opened 2024-12-29
Adding chafa package

peterjc:
Extracted from stage-recipes pull request #26240 by @pavelzw which combined this with a recipe for fastfetch (see discussion there). If this looks good we can bump from v1.40.0 to the current v1.40.5 before/after merging.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/chafa/build.sh
+++ b/recipes/chafa/build.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# By installing from a release tar-ball (see meta.yaml), we avoid the
+# need for ./autogen.sh and additional dependencies
+
+./configure \
+    --disable-debug \
+    --disable-dependency-tracking \
+    --prefix="${PREFIX}" \
+    --disable-silent-rules
+make install
--- a/recipes/chafa/build.sh
+++ b/recipes/chafa/build.sh
@@ -5,6 +5,8 @@ set -euo pipefail
 # By installing from a release tar-ball (see meta.yaml), we avoid the
 # need for ./autogen.sh and additional dependencies
 
+export PKG_CONFIG_PATH="${BUILD_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
+
 ./configure \
     --disable-debug \
     --disable-dependency-tracking \
--- a/recipes/chafa/build.sh
+++ b/recipes/chafa/build.sh
@@ -5,8 +5,6 @@ set -euo pipefail
 # By installing from a release tar-ball (see meta.yaml), we avoid the
 # need for ./autogen.sh and additional dependencies
 
-export PKG_CONFIG_PATH="${BUILD_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
-
 ./configure \
     --disable-debug \
     --disable-dependency-tracking \
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -11,7 +11,7 @@ source:
 
 build:
   number: 0
-  skip: true # [win]
+  skip: true  # [win]
 
 requirements:
   build:
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -11,6 +11,7 @@ source:
 
 build:
   number: 0
+  skip: true # [win]
 
 requirements:
   build:
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -19,6 +19,7 @@ requirements:
     - {{ stdlib('c') }}
     - pkg-config
     - make
+    - freetype
   host:
     - freetype
     - glib
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -34,19 +34,6 @@ requirements:
     - gdk-pixbuf  # [osx]
     - libintl-devel  # [osx]
   run:
-    - freetype
-    - cairo
-    - glib
-    - libjpeg-turbo
-    - librsvg
-    - libtiff
-    - libwebp
-    - libavif
-    - libjxl
-    - libpng  # [linux]
-    - zlib  # [linux]
-    - gdk-pixbuf  # [osx]
-    - gettext  # [osx]
 
 test:
   commands:
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -63,3 +63,5 @@ about:
 extra:
   recipe-maintainers:
     - pavelzw
+    - eunos-1128
+    - peterjc
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -1,0 +1,50 @@
+{% set name = "chafa" %}
+{% set version = "1.14.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/hpjansson/{{ name }}/releases/download/{{ version }}/chafa-{{ version }}.tar.xz
+  sha256: 670e55c28b5ecd4c8187bd97f0898762712a480ec8ea439dae4a4836b178e084
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    # - {{ stdlib('c') }}
+    - pkg-config
+    - make
+  host:
+    - freetype
+    - glib
+    - libjpeg-turbo
+    - librsvg
+    - libtiff
+    - libwebp
+  run:
+    - freetype
+    - glib
+    - libjpeg-turbo
+    - librsvg
+    - libtiff
+    - libwebp
+
+test:
+  commands:
+    - chafa --version
+
+about:
+  home: https://hpjansson.org/chafa
+  summary: Versatile and fast Unicode/ASCII/ANSI graphics renderer
+  license: LGPL-3.0-or-later
+  license_family: LGPL
+  license_file: COPYING
+  dev_url: https://github.com/hpjansson/chafa
+
+extra:
+  recipe-maintainers:
+    - pavelzw
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -32,7 +32,7 @@ requirements:
     - libpng  # [linux]
     - zlib  # [linux]
     - gdk-pixbuf  # [osx]
-    - gettext  # [osx]
+    - libintl-devel  # [osx]
   run:
     - freetype
     - cairo
--- a/recipes/chafa/meta.yaml
+++ b/recipes/chafa/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "chafa" %}
-{% set version = "1.14.0" %}
+{% set version = "1.14.5" %}
 
 package:
   name: {{ name|lower }}
@@ -7,7 +7,7 @@ package:
 
 source:
   url: https://github.com/hpjansson/{{ name }}/releases/download/{{ version }}/chafa-{{ version }}.tar.xz
-  sha256: 670e55c28b5ecd4c8187bd97f0898762712a480ec8ea439dae4a4836b178e084
+  sha256: 7b5b384d5fb76a641d00af0626ed2115fb255ea371d9bef11f8500286a7b09e5
 
 build:
   number: 0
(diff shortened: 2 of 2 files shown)


recipes/chafa/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  url: https://github.com/hpjansson/{{ name }}/releases/download/{{ version }}/chafa-{{ version }}.tar.xz
+  sha256: 670e55c28b5ecd4c8187bd97f0898762712a480ec8ea439dae4a4836b178e084
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    # - {{ stdlib('c') }}
```
pavelzw, line 18:
```suggestion
    - {{ stdlib('c') }}
```

peterjc:
Linux 64 failed:

```
configure: WARNING: unrecognized options: --disable-debug
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a race-free mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking whether make supports the include directive... yes (GNU style)
checking for x86_64-conda-linux-gnu-gcc... $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc accepts -g... yes
checking for $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc option to enable C11 features... none needed
checking whether $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc understands -c and -o together... yes
checking dependency style of $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc... none
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for wchar.h... yes
checking for minix/config.h... no
checking whether it is safe to define __EXTENSIONS__... yes
checking whether _XOPEN_SOURCE should be defined... no
checking whether build environment is sane... yes
checking whether to enable maintainer-specific portions of Makefiles... no
checking for an ANSI C-conforming const... yes
checking for x86_64-conda-linux-gnu-gcc... (cached) $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc
checking whether the compiler supports GNU C... (cached) yes
checking whether $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc accepts -g... (cached) yes
checking for $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc option to enable C11 features... (cached) none needed
checking whether $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc understands -c and -o together... (cached) yes
checking dependency style of $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cc... (cached) none
checking how to run the C preprocessor... $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-cpp
checking for x86_64-conda-linux-gnu-ar... $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-ar
checking the archiver ($BUILD_PREFIX/bin/x86_64-conda-linux-gnu-ar) interface... ar
checking build system type... x86_64-conda-linux-gnu
checking host system type... x86_64-conda-linux-gnu
checking how to print strings... printf
checking pkg-config is at least version 0.9.0... yes
checking for GLIB... yes
checking for FREETYPE... no
configure: error: You need freetype2-devel (or libfreetype6-dev on Debian) to build command-line tools, or pass --without-tools to build without.
```

The windows failure is simpler - there is no batch file so probably we need to skip Windows initially?

```
WARNING: No files or script found for output chafa
number of files: 0
```

hpjansson:
Hi, Chafa's `configure.ac` does this:

`PKG_CHECK_MODULES(FREETYPE, [freetype2 >= 2.0.0],,` (...)

It looks for `freetype2.pc` in the pkg-config search path. It won't build without. You need to have the development release of FreeType installed - it provides the `.pc` file. You can verify it on the command line by running `pkg-config --path freetype2`.

FreeType is pretty low in the stack, so maybe conda-forge already has it. I did a search and found [commits](1e33dd163d76e7f18a337985ec195c55b462732e), but GH claims the file can't be viewed. Not sure why.

If FreeType can't be provided, I may be able to turn it into an optional dependency, but it's kind of a last resort, since your build would then be missing functionality that's available in the Chafa CLI elsewhere.

Let me know if there's more I can do to help. Thanks for packaging Chafa!

peterjc:
A quirk of how conda-forge works means this repository https://github.com/conda-forge/staged-recipes/ only contains new potential packages - once accepted they each get their own repository.

I believe we'd using conda-forge's ``freetype`` package https://github.com/conda-forge/freetype-feedstock (and possibly the ``freetype-py`` package https://github.com/conda-forge/freetype-py-feedstock too?).

That's already listed as a host and run requirement though... but we can look for the ``freetype2.pc`` file next, thanks!

hpjansson:
Ok, sounds good. Chafa only needs the C library, so you should be fine with just `freetype`.

peterjc:
I wonder if this is breaking due to https://github.com/conda-forge/freetype-feedstock/issues/57 ? There's an open pull request which looks to fix that but as part of much more ambitious work...

hpjansson:
Chafa doesn't use `freetype-config`, though -- it uses the first `pkg-config` it can find in your path. I don't know why the wrong one is being found (should be the same as `which pkg-config` would find). Maybe you can work around it by doing one of the following before running `configure`:

* Ensuring `PATH` is set so the `bin` dir containing the correct instance of `pkg-config` is searched first.
* Setting the `PKG_CONFIG` env var to the path of Conda's `pkg-config`.

or

* Setting the `PKG_CONFIG_PATH` env var to the path where Conda's `freetype2.pc` file can be found.

`./configure --help` lists these and other env vars and options that can affect the build.

hpjansson:
Hmm, actually it may be finding the right `pkg-config` after all. Looking at the pipeline, it says:

`checking for x86_64-conda-linux-gnu-pkg-config... $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-pkg-config`

Maybe `config.log` will have more details, if you can extract it after the failed configure run.

eunos-1128:
Is there any progress?

peterjc:
The last attempt was a month ago so the CI logs have gone. I'll rebase this now so you can take a look at the failure @eunos-1128 and hopefully you can spot something I missed.

peterjc:
I think the key part of the failing log is:

```
...
checking pkg-config is at least version 0.9.0... yes
checking for GLIB... yes
checking for FREETYPE... no
configure: error: You need freetype2-devel (or libfreetype6-dev on Debian) to build command-line tools, or pass --without-tools to build without.
```

i.e. The same as quoted earlier. We have declared a dependency on ``freetype2`` which I think means either the build script needs more help to find the header files, or that package isn't exporting all the headers chafa needs.

Quoting @hpjansson from earlier, one of his suggestions was doing this before ``./configure`` is run:

> Setting the `PKG_CONFIG_PATH` env var to the path where Conda's `freetype2.pc` file can be found.

I don't know where `freetype2.pc` should be with conda-forge...

eunos-1128:
@peterjc 

How about adding the [freetype](https://anaconda.org/anaconda/freetype) package from conda-forge to the build section?

hpjansson:
Looks like it's finding GLib, which is also done by `pkg-config` invocation. So both `pkg-config` and the `glib-2.0.pc` file are being found. `freetype2.pc` should be in the same directory as `glib-2.0.pc` (typically `/usr/lib64/pkgconfig/` on vanilla Linux).

It's interesting that this seems to only affect the linux_64 build. Are there other packages that use `freetype2.pc`?

eunos-1128:
@peterjc 

I've submitted a PR below in which `PKG_CONFIG_PATH` is fixed and zlib and some other packaged are added.

It seems zlib is required on build. Even if `PKG_CONFIG_PATH` is fixed, it fails without zlib.

https://github.com/peterjc/staged-recipes/pull/2

eunos-1128:
@hpjansson 

> Looks like it's finding GLib, which is also done by pkg-config invocation. So both pkg-config and the glib-2.0.pc file are being found. freetype2.pc should be in the same directory as glib-2.0.pc (typically /usr/lib64/pkgconfig/ on vanilla Linux).
>
> It's interesting that this seems to only affect the linux_64 build. Are there other packages that use freetype2.pc?

macOS has the zlib library built into the system, so zlib is usually available “by default”, but minimal Linux installations and build environments (e.g. containers) may not include development headers (zlib-devel or libz-dev) The following is an example of a case where the development header (zlib-devel or libz-dev) is not included

I believe that is the cause of the error.

hpjansson:
> @hpjansson
> 
> macOS has the zlib library built into the system, so zlib is usually available “by default”, but minimal Linux installations and build environments (e.g. containers) may not include development headers (zlib-devel or libz-dev).
> 
> I believe that is the cause of the error.

I see. Chafa doesn't have a direct dependency on zlib, but some of our dependencies do. I think it's their responsibility to indicate the correct link flags in .pc files. I took a look at the `freetype2.pc` files in conda-forge, and it looks like the main difference between the osx and linux ones is that while the osx one has this:

```
Requires.private:
Libs.private: -lpng16 -lz
```

the linux one has this:

```
Requires.private: zlib, libpng
Libs.private: 
```

eunos-1128:
@hpjansson 

I found that [Homebrew's chafa formula](https://github.com/Homebrew/homebrew-core/blob/f5e8c5cd0b52d6e1f49a3f215acd894bf048d0c0/Formula/c/chafa.rb) have some packages missing in this PR, so added them as dependencies.

What do you think about them?

```ruby
  depends_on "pkgconf" => :build
  depends_on "cairo"
  depends_on "freetype"
  depends_on "glib"
  depends_on "jpeg-turbo"
  depends_on "jpeg-xl"
  depends_on "libavif"
  depends_on "librsvg"
  depends_on "libtiff"
  depends_on "webp"

  on_macos do
    depends_on "gdk-pixbuf"
    depends_on "gettext"
  end
```

hpjansson:
Cairo, GDK-Pixbuf and gettext are superfluous. We don't use any of them directly - the former two are dependencies of librsvg, and gettext gets pulled in by GLib. Apart from that, the list looks good :+1:

eunos-1128:
@peterjc 

Now ready for review?

recipes/chafa/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  home: https://hpjansson.org/chafa
+  summary: Versatile and fast Unicode/ASCII/ANSI graphics renderer
+  license: LGPL-3.0-or-later
+  license_family: LGPL
+  license_file: COPYING
+  dev_url: https://github.com/hpjansson/chafa
+
+extra:
+  recipe-maintainers:
+    - pavelzw
```
pavelzw, line 65: @eunos-1128 @peterjc would you like to be maintainers as well for this package?

peterjc: I'm willing, but a bit out of my comfort zone here (Python packaging is safer ground).

eunos-1128: I'm fine either way:)

recipes/chafa/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# By installing from a release tar-ball (see meta.yaml), we avoid the
+# need for ./autogen.sh and additional dependencies
+
+export PKG_CONFIG_PATH="${BUILD_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH:-}"
```
xhochy, line 8: Why is this needed? You should not be pulling anything from $BUILD_PREFIX here.

recipes/chafa/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - libjpeg-turbo
+    - librsvg
+    - libtiff
+    - libwebp
+    - libavif
+    - libjxl
+    - libpng  # [linux]
+    - zlib  # [linux]
+    - gdk-pixbuf  # [osx]
+    - gettext  # [osx]
```
xhochy, line 35:
```suggestion
    - libintl-devel  # [osx]
```

recipes/chafa/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
```
xhochy, line 49: These should probably all be pulled in by their respective `run_exports`
```suggestion
```

xhochy (changes requested):

recipes/chafa/build.sh
peterjc: @eunos-1128 you added this line - were you seeing issues with pkgconfig?

eunos-1128: @peterjc 

I added the line according to https://github.com/conda-forge/staged-recipes/pull/28721#issuecomment-2608503329

> Setting the PKG_CONFIG_PATH env var to the path where Conda's freetype2.pc file can be found.

As I tried conda-build again without the line, however, I found it not required.

Please delete the line.

peterjc:
@xhochy I think this addressed your requests, thank you.

xhochy (approved):

PR #28748, opened 2025-01-03
Add pybammsolvers

prady0t:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/pybammsolvers/build.sh
+++ b/recipes/pybammsolvers/build.sh
@@ -1,0 +1,28 @@
+set -euxo pipefail
+
+### Compile and install SUITESPARSE ###
+# SuiteSparse is required to compile SUNDIALS's
+# KLU solver.
+
+SUITESPARSE_DIR=suitsparse
+for dir in SuiteSparse_config AMD COLAMD BTF KLU
+do
+    make -C $SUITESPARSE_DIR/$dir library
+    make -C $SUITESPARSE_DIR/$dir install INSTALL=/usr
+done
+
+mkdir -p build_sundials
+cd build_sundials
+KLU_INCLUDE_DIR=/usr/local/include
+KLU_LIBRARY_DIR=/usr/local/lib
+SUNDIALS_DIR=sundials                  
+cmake -DENABLE_LAPACK=ON\
+      -DSUNDIALS_INDEX_SIZE=32\
+      -DEXAMPLES_ENABLE:BOOL=OFF\
+      -DENABLE_KLU=ON\
+      -DENABLE_OPENMP=ON\
+      -DKLU_INCLUDE_DIR=$KLU_INCLUDE_DIR\
+      -DKLU_LIBRARY_DIR=$KLU_LIBRARY_DIR\
+      -DCMAKE_INSTALL_PREFIX=/usr\
+      ../$SUNDIALS_DIR
+make install
--- a/recipes/pybammsolvers/meta.yaml
+++ b/recipes/pybammsolvers/meta.yaml
@@ -1,0 +1,54 @@
+{% set name = "pybammsolvers" %}
+{% set version = "0.0.4" %}
+{% set sundials_version = "6.5.0" %}
+{% set suitsparse_version = "6.0.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/pybammsolvers-{{ version }}.tar.gz
+    sha256: 6b76fb61e68af34b3a4803edc8020f508b10195f67857d8010e40f78cbbe2768
+  - url: https://github.com/LLNL/sundials/archive/v{{ sundials_version }}.tar.gz
+    sha256: 7557cdb8fe2a9b56fa0c259aa4e2c2dc488a1c79d482bb7943e27b7d76f6931b
+    folder: sundials
+  - url: https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/refs/tags/v{{ suitsparse_version }}.tar.gz
+    sha256: 7111b505c1207f6f4bd0be9740d0b2897e1146b845d73787df07901b4f5c1fb7
+    folder: suitsparse
+
+  
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python >=3.9,<3.13
+    - setuptools >=64
+    - wheel
+    - casadi >=3.6.7  # [not win]
+    - cmake  # [not win]
+    - pip
+  run:
+    - python >=3.9,<3.13
+    - casadi
+    - numpy <2.0
+
+test:
+  imports:
+    - pybammsolvers
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  summary: Python interface for the IDAKLU solver
+  license: BSD-3-Clause
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - prady0t


prady0t:
@agriyakhetarpal @Saransh-cpp @arjxn-py @kratman 
Locally I have problems with `gfortran`

kratman:
Ok I will take a look on Sunday

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,28 @@
+set -euxo pipefail
+
+### Compile and install SUITESPARSE ###
+# SuiteSparse is required to compile SUNDIALS's
+# KLU solver.
```
agriyakhetarpal, line 5:
```suggestion
### Compile and install SUITESPARSE ###
# SuiteSparse is required to compile SUNDIALS's
# KLU solver.
```

Nitpicking: if you do wish to keep comments for future reference in this recipe, maybe consolidating them in one place is one way to go. That said, I don't think the comments themselves will add much, could you add links to the docs here?

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,28 @@
...
+    make -C $SUITESPARSE_DIR/$dir library
+    make -C $SUITESPARSE_DIR/$dir install INSTALL=/usr
+done
+
+mkdir -p build_sundials
+cd build_sundials
+KLU_INCLUDE_DIR=/usr/local/include
+KLU_LIBRARY_DIR=/usr/local/lib
+SUNDIALS_DIR=sundials                  
+cmake -DENABLE_LAPACK=ON\
```
agriyakhetarpal, line 19:
```suggestion
mkdir -p build_sundials
cd build_sundials
KLU_INCLUDE_DIR=/usr/local/include
KLU_LIBRARY_DIR=/usr/local/lib
SUNDIALS_DIR=sundials
cmake -DENABLE_LAPACK=ON\
```

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,28 @@
...
+      -DSUNDIALS_INDEX_SIZE=32\
+      -DEXAMPLES_ENABLE:BOOL=OFF\
+      -DENABLE_KLU=ON\
+      -DENABLE_OPENMP=ON\
+      -DKLU_INCLUDE_DIR=$KLU_INCLUDE_DIR\
+      -DKLU_LIBRARY_DIR=$KLU_LIBRARY_DIR\
+      -DCMAKE_INSTALL_PREFIX=/usr\
+      ../$SUNDIALS_DIR
+make install
```
agriyakhetarpal, line 28:
```suggestion
make install

```

(nit: no newline at the end of the file)

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - setuptools >=64
+    - wheel
+    - casadi >=3.6.7  # [not win]
+    - cmake  # [not win]
+    - pip
+  run:
+    - python >=3.9,<3.13
+    - casadi
+    - numpy <2.0
+
```
agriyakhetarpal, line 38: We can use conda-forge's `{{ python_min }}` syntax here, at least for the minimum version of Python:

See https://github.com/conda-forge/cfep/blob/main/cfep-25.md

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    sha256: 7111b505c1207f6f4bd0be9740d0b2897e1146b845d73787df07901b4f5c1fb7
+    folder: suitsparse
+
+  
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
```
agriyakhetarpal, line 26: You'll also need` stdlib('c')` (libc), `compiler('c++')`, and  `compiler('fortran')` here – does that help with your issue? Or, are you currently testing your changes using `build_locally.py` and Docker?

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set name = "pybammsolvers" %}
+{% set version = "0.0.4" %}
+{% set sundials_version = "6.5.0" %}
+{% set suitsparse_version = "6.0.3" %}
```
agriyakhetarpal, line 4:
```suggestion
{% set suitesparse_version = "6.0.3" %}
```

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,28 @@
+set -euxo pipefail
+
+### Compile and install SUITESPARSE ###
+# SuiteSparse is required to compile SUNDIALS's
+# KLU solver.
+
+SUITESPARSE_DIR=suitsparse
```
agriyakhetarpal, line 7:
```suggestion
SUITESPARSE_DIR=suitesparse
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+source:
+  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/pybammsolvers-{{ version }}.tar.gz
+    sha256: 6b76fb61e68af34b3a4803edc8020f508b10195f67857d8010e40f78cbbe2768
+  - url: https://github.com/LLNL/sundials/archive/v{{ sundials_version }}.tar.gz
+    sha256: 7557cdb8fe2a9b56fa0c259aa4e2c2dc488a1c79d482bb7943e27b7d76f6931b
+    folder: sundials
+  - url: https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/refs/tags/v{{ suitsparse_version }}.tar.gz
+    sha256: 7111b505c1207f6f4bd0be9740d0b2897e1146b845d73787df07901b4f5c1fb7
+    folder: suitsparse
```
agriyakhetarpal, line 18:
```suggestion
  - url: https://github.com/DrTimothyAldenDavis/SuiteSparse/archive/refs/tags/v{{ suitesparse_version }}.tar.gz
    sha256: 7111b505c1207f6f4bd0be9740d0b2897e1146b845d73787df07901b4f5c1fb7
    folder: suitesparse
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - python >=3.9,<3.13
+    - setuptools >=64
+    - wheel
```
agriyakhetarpal, line 30:
```suggestion
    - setuptools >=71
```

Shouldn't be needed anymore, see https://github.com/pybamm-team/PyBaMM/pull/4652 and https://github.com/pybamm-team/pybammsolvers/pull/14

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - casadi >=3.6.7  # [not win]
+    - cmake  # [not win]
```
agriyakhetarpal, line 32:
```suggestion
    - casadi >=3.6.7
    - cmake
```

Our packaging is going to differ here because we won't use Vcpkg and MSVC – we'll use Clang that `conda-forge` provides, so we'll need CasADi and CMake too.

recipes/pybammsolvers/build.sh
agriyakhetarpal, line 1: Not to be addressed right away, but we'll also need a `build.bat` equivalent of this script for Windows.

agriyakhetarpal (changes requested):
Thanks @prady0t, looks like a good start! https://github.com/conda-forge/staged-recipes/pull/28748#issuecomment-2569878449 would need to be addressed as far as the recipe's static analysis goes, and then I've left other comments below about how the packaging of the recipe should go:

recipes/pybammsolvers/meta.yaml
agriyakhetarpal: Though, it would be better to compile CasADi (just the C++ interface, that is – which means, with `WITH_PYTHON3` set to `OFF` ) on our own as well so that we guarantee ABI compatibility with CasADi's conda-forge package.

We could then patch these lines: https://github.com/pybamm-team/PyBaMM/blob/a5616c960ac4769db7ff98a04628a04789e8bf7e/CMakeLists.txt#L119-L139 to turn off our lookup for the CasADi Python interface. The reason for this is that we wouldn't want to link against PyPI dependencies with conda-forge ones; both follow different methodologies for ABI compatibility. Let's revisit this once we have some progress and when we have SuiteSparse and SUNDIALS being compiled correctly...

recipes/pybammsolvers/build.sh
prady0t: Yes I'm aware. Once the `sh` script works, I'll try replication the steps for `bat` file.

recipes/pybammsolvers/meta.yaml
prady0t: Yes, I can compile Sundials now. I don't think it's necessary to have ` stdlib('c')` here, a c compiler should automatically get standard c libraries. Also `{{ stdlib("c") }}` causes : 
```
└─ c_osx-arm64 =* * does not exist (perhaps a typo or a missing channel).
```
But I may be doing something wrong. I'm commenting out this step until further discussion, as I can still compile without it.

I'm currently only using `conda-build`. Should I choose docker instead?

prady0t: I'm sorry I didn't see the lint bot comments. We do need  `{{ stdlib("c") }}`

recipes/pybammsolvers/meta.yaml
prady0t: For some reason this fails for `conda-build` locally, so can't test this part locally.

recipes/pybammsolvers/build.sh
prady0t: Completely forgot about this. Let me resolve this in the next commits.

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,28 @@
...
+
+### Compile and install SUITESPARSE ###
+# SuiteSparse is required to compile SUNDIALS's
+# KLU solver.
+
+SUITESPARSE_DIR=suitesparse
+for dir in SuiteSparse_config AMD COLAMD BTF KLU
+do
+    make -C $SUITESPARSE_DIR/$dir library
+    make -C $SUITESPARSE_DIR/$dir install INSTALL=$PREFIX
```
prady0t, line 11: I've used $PREFIX here and for the `Sundials` installation as well because there were permission problems with `usr`. Let me know if this is not preferable.

prady0t:
We have some progress, I can now compile both `Suitesparse` and `Sundials`. The failure now is `conda-build` cannot import `pybammsolvers` which suggests the build was not completed.

recipes/pybammsolvers/build.sh
prady0t: I can still see the CI fails due to the same problem.

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - pip
+
+about:
+  summary: Python interface for the IDAKLU solver
+  license: BSD-3-Clause
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - prady0t
```
kratman, line 56: This should be the IDAKLU maintainers group from pybamm team

kratman (commented):
Just starting to take a look at this repo. I will share anything I find

recipes/pybammsolvers/meta.yaml
agriyakhetarpal: It's common for `conda-forge` feedstocks for packages to be maintained by folks other than the upstream package maintainers, so I will suggest keeping both – unless you did mean that, of course.

recipes/pybammsolvers/build.sh
agriyakhetarpal: Yes, $PREFIX is the ideal location to store them, `usr/` is not user-writeable on most CI runners.

recipes/pybammsolvers/meta.yaml
agriyakhetarpal: Yes. I think it wouldn't hurt to try using Docker, since getting it to work on Linux before macOS would be nice, and it being isolated doesn't modify your local development environment.

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('fortran') }}
+    - {{ stdlib("c") }}
+    - {{ compiler('cxx') }}
+  host:
+    - python {{ python_min }}
+    - setuptools >=71
+    - casadi >=3.6.7
+    - cmake
```
agriyakhetarpal, line 34:
```suggestion
    - cmake
    - make
```
 
CI says the 'make' command is not found, maybe this works?

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('fortran') }}
+    - {{ stdlib("c") }}
+    - {{ compiler('cxx') }}
+  host:
```
agriyakhetarpal, line 30:
```suggestion
  host:
        host:
    - if: osx
       then:
         - llvm-openmp
    - if: linux
      then:
        - libgomp
```

agriyakhetarpal: We'll probably need this later

recipes/pybammsolvers/meta.yaml
agriyakhetarpal: Once we switch to a higher version for SuiteSparse, we could likely configure CMake/Ninja as the generators...

recipes/pybammsolvers/meta.yaml
prady0t: Should we do 

```yaml
  host:
    - llvm-openmp  # [osx]
    - libgomp  # [linux]
   
  ```
   
 instead? Current changes give syntax errors.

Comment (prady0t) [reply recipes/pybammsolvers/meta.yaml:30]: Doing this just so that CI passes. I'm happy to make this syntax (your suggestion) work.

Comment (agriyakhetarpal) [reply recipes/pybammsolvers/meta.yaml:30]: Ah, nope, yours is better – we should use it indeed.

File: recipes/pybammsolvers/meta.yaml [hunk 15]
```diff
@@ -0,0 +1,59 @@
...
+    - llvm-openmp  # [osx]
+    - libgomp  # [linux]
```
agriyakhetarpal, line 32:
```suggestion
    - libgomp  # [linux]
    - libblas
```

agriyakhetarpal (commented):
It proceeds to compile now but fails to find a BLAS distribution:

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+{% set version = "0.0.4" %}
+{% set sundials_version = "6.5.0" %}
+{% set suitesparse_version = "6.0.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/pybammsolvers-{{ version }}.tar.gz
```
agriyakhetarpal, line 11:
```suggestion
  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
```

Already defined above

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('fortran') }}
+    - {{ stdlib("c") }}
+    - {{ compiler('cxx') }}
+  host:
+    - llvm-openmp  # [osx]
+    - libgomp  # [linux]
+    - libblas
+    - libcblas
+    - liblapack
```
agriyakhetarpal, line 35:
```suggestion
    - liblapack
    - pkg-config
```

agriyakhetarpal:
How about we don't attempt to compile SuiteSparse and directly add SUNDIALS as a host dependency – it's [linked with KLU](https://github.com/conda-forge/sundials-feedstock/blob/bd8b70c3d07d845eae73222010e3cef7a4120d80/recipe/build-sundials.sh#L41) like we need it to be, and [version 6.5.0 is available](https://anaconda.org/conda-forge/sundials/files?version=6.5.0) as well. That would be much easier :) [Chaste](https://github.com/Chaste/Chaste/blob/02dd6fef6e5bb11d86d0e3148a5a52c1eb65e0be/pychaste/src/py/conda/recipe/meta.yaml) and [Octave](https://github.com/conda-forge/octave-feedstock/blob/1981bd35675500a181c61a15f4ef73450950c08b/recipe/meta.yaml#L47) use SUNDIALS in this manner.

prady0t:
> How about we don't attempt to compile SuiteSparse and directly add SUNDIALS as a host dependency – it's [linked with KLU](https://github.com/conda-forge/sundials-feedstock/blob/bd8b70c3d07d845eae73222010e3cef7a4120d80/recipe/build-sundials.sh#L41) like we need it to be, and [version 6.5.0 is available](https://anaconda.org/conda-forge/sundials/files?version=6.5.0) as well. That would be much easier :) [Chaste](https://github.com/Chaste/Chaste/blob/02dd6fef6e5bb11d86d0e3148a5a52c1eb65e0be/pychaste/src/py/conda/recipe/meta.yaml) and [Octave](https://github.com/conda-forge/octave-feedstock/blob/1981bd35675500a181c61a15f4ef73450950c08b/recipe/meta.yaml#L47) use SUNDIALS in this manner. It would also circumvent our BLAS requirement that we're not able to fix here.

That would remove most of the components inside `build.sh` right?

agriyakhetarpal:
> That would remove most of the components inside `build.sh` right?

Yes

prady0t:
Removing all contents of `build.sh` and just relying on pre-compiled sundials and pip installation, gives 
```
  CMake Error: The source directory "/opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1736447926947/work" does not appear to contain CMakeLists.txt.
  ```

**Comment by agriyakhetarpal** [comment]
Yeah, you'll need to add the file through a patch... many of the extra packaging changes are included with https://github.com/pybamm-team/pybammsolvers/pull/7, but it requires a round of review and a subsequent merge/release.

**Comment by agriyakhetarpal** [comment]
https://www.anaconda.com/blog/patching-source-code-to-conda-build-recipes or https://pyodide.org/en/stable/development/new-packages.html#generating-patches might help

**Comment by prady0t** [comment]
Can we get our pybammsolver's `stdist` from GitHub [release](https://github.com/pybamm-team/pybammsolvers/releases) instead? It has the `CmakeList.txt` file as well? I tried doing it locally but it failed because of : 

```
    The source directory

      /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1736533690295/work/pybind11

    does not contain a CMakeLists.txt file.
   
```

**Comment by agriyakhetarpal** [comment]
That can be an option we can try, but in this case, this error comes from `pybind11` and not PyBaMM. if you were to add `pybind11` as a host dependency here and configure it (see the modifications to `setup.py` and `CMakeLists.txt` in my old closed PR at pybamm-team/PyBaMM#3560) through a patch, I think it would all work.

**Comment by prady0t** [comment]
I tried using the original idea (adding CMakeList.txt). It still cannot find `pybind11`. Fails with : 

```
  CMake Error at CMakeLists.txt:82 (pybind11_add_module):
    Unknown CMake command "pybind11_add_module".
```

**Comment by agriyakhetarpal** [comment]
That sounds fixable, did you add `pybind11` as a build-time dependency in `pyproject.toml` as well?

File: recipes/pybammsolvers/meta.yaml [hunk 18]
```diff
@@ -0,0 +1,57 @@
...
+    - pkg-config
+    - python {{ python_min }}
+    - setuptools >=71
+    - casadi >=3.6.7
+    - sundials
+    - pip
```
agriyakhetarpal, line 36:
```suggestion
  host:
    - llvm-openmp  # [osx]
    - libgomp  # [linux]
    - libblas
    - libcblas
    - liblapack
    - pkg-config
    - python {{ python_min }}
    - setuptools >=71
    - cmake
    - casadi >=3.6.7
    - sundials
    - pip
```

prady0t:
I was able to find casadi but some packages `nvector/nvector_openmp.h` cannot be found.

```
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/Solution.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/Solution.hpp:4:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolverOpenMP_solvers.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolverOpenMP_solvers.hpp:4:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolverOpenMP.hpp:4:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolver.hpp:4:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolverGroup.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolverGroup.hpp:4:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolver.hpp:4:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolver.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IDAKLUSolver.hpp:4:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/SolutionData.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/SolutionData.hpp:5:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IdakluJax.cpp:1:
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/IdakluJax.hpp:4:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector            */
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
  In file included from /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.cpp:1:
  /opt/miniconda3/envs/pybammsolvers-build/conda-bld/pybammsolvers_1737224830610/work/src/pybammsolvers/idaklu_source/common.hpp:10:10: fatal error: 'nvector/nvector_openmp.h' file not found
     10 | #include <nvector/nvector_openmp.h>  /* access to openmp N_Vector  
```

agriyakhetarpal:
Thanks! It might be good to see what the CI jobs need in comparison to your local build, as it is difficult for us to look at your build the full logs.

As for finding `nvector_openmp.h` file, that should be present in the solvers source code. As the logs are truncated, I'm not sure why it fails – in the full logs, is there a longer stack trace? If it fails to find OpenMP, I'm not sure why that would be the case. This is because SUNDIALS does build with it, at least on Linux: https://github.com/conda-forge/sundials-feedstock/blob/bd8b70c3d07d845eae73222010e3cef7a4120d80/recipe/build-sundials.sh#L7-L11

prady0t:
Mostly, the CI logs represent my local build failures. For osx, I think OpenMP was not found. Even inside `$PREFIX/include/nvector`, I cannot find OpenMP related files and these are its only contents: 

```
nvector_manyvector.h
nvector_serial.h

```
Same for Windows builds. For linux now we cannot find `klu`. Might it be possible KLU linked Sundials is not available across all platforms?

agriyakhetarpal:
From https://github.com/conda-forge/sundials-feedstock/blob/bd8b70c3d07d845eae73222010e3cef7a4120d80/recipe/build-sundials.sh#L7-L11, SUNDIALS builds without OpenMP for Darwin, but builds with it for Linux. And, checking `build.bat`, it uses OpenMP for Windows. So we would indeed need to switch to a custom build of SUNDIALS for all platforms, where we properly link against both KLU and OpenMP (llvm-openmp on macOS and libgomp on Linux).

prady0t:
Sundials is build successfully. I've used pre-build Suitesparse from conda-forge and linked KLU against it. I don't know about the error in CI 

```
    raise OverLinkingError(overlinking_errors)
conda_build.exceptions.OverLinkingError: <exception str() failed>
```
Locally it fails as it's not able to run tests, even though all steps pass without failure.

agriyakhetarpal:
The overlinking means that some dependencies are included as a part of the runtime, which aren't needed – for example, SUNDIALS and SuiteSparse are required only at build time, and not at runtime.

Also, is the IDAKLU solver being compiled? I ask because the CMakeLists.txt file was removed, so I'm not sure if it is. Could you add a test for the IDAKLU solver import?

prady0t:
I missed that. Thanks! But tests for importing pybamsolvers are still failing?

agriyakhetarpal:
I'm not sure if I see an import failure in the logs, though. Could you copy that, if you have some logs from a local build?

Also, @Saransh-cpp, @kratman, and I discussed this PR briefly in-person today. Would it be fine with you if we were to take over this PR in case you feel it will take some time over the next few weeks or so? Eric is planning for a v25.2 release at the end of February, and we will be two releases behind by then. I think contributing to the same PR should be fine for preserving authorship. I also have access to your fork already, so I can help push a few changes next week. If it's a yes from you, please grant Eric access in a few weeks as well.

I think the plan of action at the moment will be as follows:
1. get the CMakeLists.txt file back
2. patch CMakeLists.txt to look for the CasADi libs + headers under the prefix dirs, wherever they are (with CasADi set as a host dependency so that the libs and headers are available)
3. remove whichever overlinked dependencies are left
4. test that the `import pybammsolvers` statement works across all platforms.

prady0t:
Yes, it's totally fine! I've already sent the access to all of you.

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - make
+    - casadi >=3.6.7
+    - pybind11
+    - pip
+    - suitesparse
+  run:
+    - python >={{ python_min }}
+    - casadi
+    - numpy <2.0
+
```
agriyakhetarpal, line 50:
```suggestion
  run:
    - python >={{ python_min }}
    - casadi
    - numpy <2.0
    - suitesparse

```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - casadi >=3.6.7
+    - pybind11
+    - pip
+    - suitesparse
+  run:
+    - python >={{ python_min }}
+    - casadi
+    - numpy <2.0
+    - suitesparse
+
```
agriyakhetarpal, line 51:
```suggestion
  run:
    - python >={{ python_min }}
    - casadi
    - numpy <2.0
    - suitesparse
    - libklu
    - libopenblas

```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    # - liblapack
+    - pkg-config
+    - python {{ python_min }}
+    - setuptools >=71
+    - cmake
+    - make
+    - casadi >=3.6.7
+    - pybind11
+    - pip
+    - suitesparse
```
agriyakhetarpal, line 45:
```suggestion
  host:
    - llvm-openmp  # [osx]
    - libgomp  # [linux]
    - pkg-config
    - python {{ python_min }}
    - setuptools >=71
    - cmake
    - make
    - casadi >=3.6.7
    - pybind11
    - pip
    - suitesparse
    - libklu
    - libopenblas
```

agriyakhetarpal:
There are errors like:

```
  ERROR (pybammsolvers,lib/libsundials_sunlinsollapackband.so.4.5.0): Needed DSO lib/libopenblas.so.0 found in ['conda-forge/linux-64::libopenblas==0.3.28=pthreads_h94d23a6_1']
  ERROR (pybammsolvers,lib/libsundials_sunlinsollapackband.so.4.5.0): .. but ['conda-forge/linux-64::libopenblas==0.3.28=pthreads_h94d23a6_1'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

in the logs, which we'll need to fix. I noticed SuiteSparse offers `libklu` as a separate package as its recipe contains multiple outputs. Please bear with me while I try to make more changes.

agriyakhetarpal:
Okay, I fixed the overlinking issue with SuiteSparse and OpenBLAS – we now need to compile `pybammsolvers`.

agriyakhetarpal:
I think this should be unblocked enough for you to proceed with the following steps, @prady0t. The next thing on the list would be to point to the KLU headers correctly – they will be in `$PREFIX/include`, so we can patch these lines: https://github.com/pybamm-team/pybammsolvers/blob/791677d64aff3b62025a62e85e97bb73736a0391/setup.py#L14-L16 so that we look for them in the right location.

prady0t:
Cannot find `sundials_legacy_wrapper.hpp` but it was removed. We might have to wait for a new release (the latest release still has it).

Saransh-cpp (commented):
```
undefined symbol: IDASensEEtolerances 
```
Hmm, I think the workflows cannot find `IDAS`. Maybe we need to mirror https://github.com/pybamm-team/pybammsolvers/blob/main/FindSUNDIALS.cmake?

agriyakhetarpal:
Yes, the `.cmake` files for SUNDIALS and SuiteSparse will need to be included. Thanks for raising this, @Saransh-cpp.

I've been mostly helping @prady0t out on this from the sidelines, but I can take a closer look here in mid-April if things don't progress by then.

agriyakhetarpal (commented):
The missing symbols are likely coming from the fact that we're packaging `pybammsolvers` for version 0.0.4, which contains the `sundials_legacy_wrapper` header, but it's being removed through the patch here – it shouldn't be removed.

We can drop the patch and bump to pybammsolvers 0.1.0 as it dropped the file anyway.

Saransh-cpp (commented):
I tried patching `FindSUNDIALS.cmake` but it doesn't work. We still get the same error. For some reason, CMake never sets (or is unable to set) `SUNDIALS_INCLUDE_DIR`, and I suspect that this is the reason for not being able to find IDAS - 
```
  SUNDIALS found in:
```

But, on the other hand, we do not get any error on the IDAS variables used above `IDASensEEtolerances`. I have not looked into this, but this would be good to investigate too.

I give up for today 😞

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,74 @@
+{% set name = "pybammsolvers" %}
+{% set version = "0.1.0" %}
+{% set sundials_version = "6.5.0" %}
+{% set suitesparse_version = "6.0.3" %}
```
Saransh-cpp:
```suggestion
{% set version = "0.3.3" %}
{% set sundials_version = "7.3.0" %}
{% set suitesparse_version = "7.8.3" %}
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+{% set sundials_version = "7.3.0" %}
+{% set suitesparse_version = "7.8.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+    sha256: cd7214d6a7d226feabedf82497f99caf3238e6ebce2aec16236bf7a5cea72d2c
```
Saransh-cpp:
```suggestion
    sha256: 2e1b8bacfe62b3433b48095df77528fabbabc37c321e17ab4e5a1a0a5bfbeb2c
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+source:
+  - url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz
+    sha256: 2e1b8bacfe62b3433b48095df77528fabbabc37c321e17ab4e5a1a0a5bfbeb2c
+  - url: https://github.com/LLNL/sundials/archive/v{{ sundials_version }}.tar.gz
+    sha256: 7557cdb8fe2a9b56fa0c259aa4e2c2dc488a1c79d482bb7943e27b7d76f6931b
+    folder: sundials
+  - url: https://github.com/pybind/pybind11/archive/refs/tags/v2.13.6.tar.gz
+    sha256: e08cb87f4773da97fa7b5f035de8763abc656d87d5773e62f6da0587d1f0ec20
+    folder: pybind11
```
Saransh-cpp:
```suggestion
    sha256: 97d0ae56391812013928143d5226a2ea789ff0bd657a8a6eeebf9e26d60186c0
    folder: sundials
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,74 @@
...
+  host:
+    - llvm-openmp  # [osx]
+    - libgomp  # [linux]
+    - pkg-config
+    - python
+    - setuptools >=71
+    - cmake
+    - make
+    - casadi >=3.6.7
+    - pybind11
```
Saransh-cpp, line 44:
```suggestion
    - pybind11[global]
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+  host:
+    - llvm-openmp  # [osx]
+    - libgomp  # [linux]
+    - pkg-config
+    - python
+    - setuptools >=71
+    - cmake
+    - make
+    - casadi >=3.6.7
+    - pybind11[global]
```
Saransh-cpp:
```suggestion
    - pybind11
```

Saransh-cpp:
Spent some time on this today. Everything on osx and linux is working now, and the only blocker imo is https://github.com/pybamm-team/pybammsolvers/pull/77. I can write a `.bat` script once there is a new release and everything works as expected.

Saransh-cpp:
Given how messy this PR has become, I have opened a new cleaner one - #31795. @prady0t could you please close this? Thanks!

PR #28777, opened 2025-01-07
add synspace

hadim:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/synspace/recipe.yaml
+++ b/recipes/synspace/recipe.yaml
@@ -13,7 +13,7 @@ source:
 build:
   number: 0
   noarch: python
-  script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  script: python -m pip install . -vv
 
 requirements:
   host:
--- a/recipes/synspace/recipe.yaml
+++ b/recipes/synspace/recipe.yaml
@@ -1,0 +1,52 @@
+context:
+  name: synspace
+  version: 0.3.0
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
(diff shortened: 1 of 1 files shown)


hadim:
Needed for https://github.com/conda-forge/exmol-feedstock/pull/22

hadim:

The linter error seems to be a false positive, since the correct `python_min` version is used during the tests.

recipes/synspace/recipe.yaml
```diff
@@ -0,0 +1,51 @@
...
+  version: ${{ version }}
+
+source:
+  url: https://pypi.org/packages/source/${{ name[0] }}/${{ name }}/synspace-${{ version }}.tar.gz
+  sha256: d0e004041e8e9ea401b699241d8f18574d8a4893e179260316ff8e5d1ff67515
+
+build:
+  number: 0
+  noarch: python
+  script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
synapticarbors, line 16:
```suggestion
  script: python -m pip install . -vv
```

I think with these new recipes we don't need `${{ PYTHON }}`, which is also reflected in the example recipe for v1

synapticarbors (commented):
Please note that the autotick machinery isn't working yet for v1 recipes so until that's completed the resulting feedstock will not recieve automatic version update PRs

recipes/synspace/recipe.yaml
hadim: I think it's needed for cross compilation purposes, but it won't be needed since its `noarch` so we can remove it.

PR #28794, opened 2025-01-08
NEW: Add cf-nvidia-tools

carterbox:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/cf-nvidia-tools/bin/check-glibc
+++ b/recipes/cf-nvidia-tools/bin/check-glibc
@@ -91,8 +91,8 @@ function glibc-check () {(
     "compatible" )
       local first second
       local -a sorted
-      first="$(glibc version-normalize ${1})" ; shift
-      second="$(glibc version-normalize ${1})" ; shift
+      first="$(glibc version-normalize "${1}")" ; shift
+      second="$(glibc version-normalize "${1}")" ; shift
 
       mapfile -t sorted < <(
         printf "%s\n" "${first}" "${second}" |
@@ -115,8 +115,8 @@ function glibc-check () {(
     "status" )
       local first second
       local -a sorted
-      first="$(glibc version-normalize ${1})" ; shift
-      second="$(glibc version-normalize ${1})" ; shift
+      first="$(glibc version-normalize "${1}")" ; shift
+      second="$(glibc version-normalize "${1}")" ; shift
       mapfile -t sorted < <(
         printf "%s\n" "${first}" "${second}" |
         sort -Vr
@@ -187,9 +187,9 @@ fi
 
 for file in "$@"; do
     if [[ -f "$file" && ! -L "$file" ]]; then  # Ensure it's a file and not a link
-      BINARY_GLIBC_VERSION="$(glibc-detect req $file)"
+      BINARY_GLIBC_VERSION="$(glibc-detect req "$file")"
       echo "binary glibc ${BINARY_GLIBC_VERSION} <= recipe glibc ${RECIPE_GLIBC_VERSION} <= system glibc ${SYSTEM_GLIBC_VERSION} $file"
-      BINARY_IS_COMPATIBLE="$(glibc-check compatible $BINARY_GLIBC_VERSION $RECIPE_GLIBC_VERSION)"
+      BINARY_IS_COMPATIBLE="$(glibc-check compatible "$BINARY_GLIBC_VERSION" "$RECIPE_GLIBC_VERSION")"
       if [[ $BINARY_IS_COMPATIBLE == "false" ]] ; then
         echo "The binary is not compatible with the recipe's glibc pinning."
         exit 1
--- a/recipes/cf-nvidia-tools/bin/check-glibc
+++ b/recipes/cf-nvidia-tools/bin/check-glibc
@@ -1,0 +1,199 @@
+#!/usr/bin/env bash
+
+
+function glibc () {(
+
+  # a dispatching function for
+  local mode="${1}" ; shift
+  case "${mode}" in
+    ## 1. detecting the glibc version
+    "detect" )
+      glibc-detect "${@}"
+    ;;
+    ## 2. checking a pair of glibc versions for compatibility
+    "check" )
+      glibc-check "${@}"
+    ;;
+    "version-normalize" )
+      local version="${1}" ; shift
+      IFS="." read -ra components <<< "${version}"
+      while [[ "${#components[@]}" -lt "3" ]] ; do
+        components+=("0")
+      done
+      version="$(printf "%s." "${components[@]}")"
+      version="${version%"."*}"
+      if [[ -n "${version}" ]] ; then
+        echo "${version}"
+      fi
+    ;;
+  esac
+)}
+
+function glibc-detect () {(
+  # a dispatching function for detecting the glibc version
+  local mode="${1}" ; shift
+  ## modes:
+  case "${mode}" in
+    ## 1. "env" - unpack the value of a named variable
+    ### e.g. 'glibc-detect "env" "c_stdlib_version"'
+    #### returns the value of "c_stdlib_version"
+    #### it is equivalent to echo "${c_stdlib_version}"
+    "env" )
+      local named_variable="${1}"
+      echo "${!named_variable}"
+    ;;
+    ## 2. "req" - detect the glibc version required by a library
+    ### e.g. 'glibc-detect "req" "/path/to/library"'
+    #### returns the minimum required version of glibc for the library
+    "req" )
+      glibc-detect "library" "${@}" |
+      head --lines=1
+    ;;
+    ## 3. "system" - detect the version of glibc on the system
+    ### e.g. 'glibc-detect "system"'
+    #### returns the version of glibc on the system as reported by 'ldd --version'
+    "system" )
+      ldd --version |
+      awk 'NR==1 {print $NF}'
+    ;;
+    ## 4. "library" - detect all versions of glibc in a library
+    ### e.g. 'glibc-detect "library" "/path/to/library"'
+    #### returns all versions of glibc in the symbol table of the library
+    "library" )
+      local version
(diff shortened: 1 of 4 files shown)


carterbox:
@conda-forge/cuda, please review

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+  - path: bin
+    target_directory: bin
+
+build:
+  skip:
+    - not linux
+  noarch: generic
+  script:
+    - ls -lah
+    - cp -r bin $PREFIX/bin
```
jakirkham, line 19:
```suggestion
    - cp -r bin "$PREFIX/bin"
```

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+requirements:
+  run:
+    - __unix
+    - patchelf >=0.12
+
+tests:
+  - script:
+    - export c_stdlib_version="2.17.0"
+    - check-glibc ${CONDA_PREFIX}/lib/*.so.*
```
jakirkham, line 30:
```suggestion
    - check-glibc "${CONDA_PREFIX}/lib/*.so.*"
```

Also would it make sense to have a default search argument or would we rather be explicit like in this test

Lastly should we add a library test dependency to ensure there is one library we can check?

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+about:
+  summary: NVIDIA's feedstock validation and linting tools for conda-forge
+  description: >
+    This package contains CLI tools for validating and linting NVIDIA's conda recipes on
+    conda-forge. For a description of the tools see the README in the package feedstock. The
+    tools are hosted directly in the feedstock; there is no external source code repository
+    for these tools at this time.
```
jakirkham, line 38: Do we want this to be one long string or do want to have it wrapped a certain way?

FWIW find this website to be a useful resource when playing with YAML string formatting: https://yaml-multiline.info

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - cp -r bin $PREFIX/bin
+  number: 0
+
+requirements:
+  run:
+    - __unix
+    - patchelf >=0.12
+
+tests:
+  - script:
```
jakirkham, line 28: Would it make sense to have a `shellcheck` step here?

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - export c_stdlib_version="2.17.0"
```
jakirkham, line 29: Should we do some kind of help check?

```suggestion
    - check-glibc
    - export c_stdlib_version="2.17.0"
```

jakirkham (commented):
Thanks Daniel! 🙏

Really appreciate you putting the time into creating this testing resource. This will be very helpful when testing packages and catching any missed changes.

Had a few questions below

recipes/cf-nvidia-tools/recipe.yaml
carterbox: I don't want to provide a default argument because in production, we don't want to test all of the binaries in PREFIX, just the NVIDIA redists. There is no glob that I can think of that satisfies this condition.

recipes/cf-nvidia-tools/recipe.yaml
carterbox: I want one long string. anaconda.org handles the wrapping of long lines.

recipes/cf-nvidia-tools/recipe.yaml
carterbox: Never heard of shellcheck before now. I'll add it.

recipes/cf-nvidia-tools/recipe.yaml
carterbox: Done.

recipes/cf-nvidia-tools/recipe.yaml
carterbox: No double quotes because we want the terminal to expand the glob expression not pass a literal glob expression.

carterbox:
@conda-forge/staged-recipes, please review

recipes/cf-nvidia-tools/bin/check-glibc
```diff
@@ -0,0 +1,199 @@
...
+      first="$(glibc version-normalize "${1}")" ; shift
+      second="$(glibc version-normalize "${1}")" ; shift
+
+      mapfile -t sorted < <(
+        printf "%s\n" "${first}" "${second}" |
+        sort -Vr
+      )
+      if [[ "${first}" == "${second}" ]] ; then
+        echo "true"
+        return 0
```
billysuh7, line 103: Is the return value of 1 or 0 ever used ? I found that when this file was used in cudnn feedstock, and when it returned 1 (incompatible), it would exit build.sh right away due to "set -e" in build.sh, without printing the message "The binary is not compatible with the recipe's glibc pinning". This does not happen here in this feedstock as it does not have "set -e" though.

recipes/cf-nvidia-tools/bin/check-glibc
```diff
@@ -0,0 +1,199 @@
...
+
+if [ "$c_stdlib_version" = "0.0.0" ]; then
+    echo "c_stdlib_version is not set!"
+    exit 1
+fi
+
+for file in "$@"; do
+    if [[ -f "$file" && ! -L "$file" ]]; then  # Ensure it's a file and not a link
+      BINARY_GLIBC_VERSION="$(glibc-detect req "$file")"
+      echo "binary glibc ${BINARY_GLIBC_VERSION} <= recipe glibc ${RECIPE_GLIBC_VERSION} <= system glibc ${SYSTEM_GLIBC_VERSION} $file"
```
billysuh7, line 191: Maybe I am missing something - is the 2nd half of the check i.e. "recipe glibc ${RECIPE_GLIBC_VERSION} <= system glibc ${SYSTEM_GLIBC_VERSION}" being run ? The next line, glibc-check compatible "$BINARY_GLIBC_VERSION" "$RECIPE_GLIBC_VERSION", compares binary glibc and recipe glibc only I think.

carterbox: I think you are correct. We don't check if the system glibc version is >= the recipe version. This was important for debugging when developing this tool, but it doesn't actually affect the correctness of the recipe.

It is perfectly valid to *build* a package that requires glibc 5 on a system that only has glibc 2. It's just that you won't be able to *run* or *install* the binaries you have built because the package will require __glibc >= 5.

In summary, printing the system glibc version is printed for informational purposes and may help with debugging if tests which try to load the libraries fail during the test phase.

I could add a check and echo a warning message if system glibc is < recipe glibc, but definitely not raise an error.

recipes/cf-nvidia-tools/bin/check-glibc
carterbox: This is a good point! You're saying that return causes the entire script to abort instead of just moving up one level. I will refactor to use the strings only.

carterbox: I've stripped unused parts of the API and added a test to ensure that the script exits 1 when appropriate.

recipes/cf-nvidia-tools/bin/check-glibc
```diff
@@ -0,0 +1,153 @@
...
+  local mode="${1}" ; shift
+  case "${mode}" in
+    ## 1. detecting the glibc version
+    "detect" )
+      glibc-detect "${@}"
+    ;;
+    ## 2. checking a pair of glibc versions for compatibility
+    "check" )
+      glibc-check "${@}"
+    ;;
```
billysuh7, line 16: I don't believe these two modes are used. Are they for future extension ?

recipes/cf-nvidia-tools/bin/check-glibc
billysuh7: Thanks for clarifying! Yeah I think it's better to give a warning if system glibc is < recipe glibc, or at least simply print SYSTEM_GLIBC_VERSION value on a separate line as dry information - right now it may give impression to the user that system glibc is being checked against recipe glibc version

recipes/cf-nvidia-tools/bin/check-glibc
carterbox: Removed.

recipes/cf-nvidia-tools/bin/check-glibc
carterbox: I have changed the logging message to only print the system glibc once per call to check-glibc

billysuh7 (approved):
LGTM

recipes/cf-nvidia-tools/recipe.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  - path: bin
+    target_directory: bin
+
+build:
+  skip:
+    - not linux
```
carterbox, line 15:
```suggestion
```

h-vetinari:
For some history / cross-references on this, @carterbox and I seemed to work on the same idea pretty much concurrently, though my attempt (https://github.com/conda-forge/libcufile-feedstock/commit/65e6c5d4135179fd296a54765e661e995078647b) was far less elaborate than what ended up being the seed of this PR in https://github.com/conda-forge/cudnn-feedstock/pull/97 (which was apparently also inspired by some pre-existing script).

Cool that this is getting automated more broadly! (that glibc check might have even deserved a more generic name, because it's not just nvidia that has repackaged binaries; in any case, this will be useful!)

PR #28828, opened 2025-01-13
add tiledbsoma

hadim:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/tiledbsoma/recipe.yaml
+++ b/recipes/tiledbsoma/recipe.yaml
@@ -111,7 +111,7 @@ outputs:
 
   # python bindings
   - package:
-      name: tiledbsoma
+      name: tiledbsoma-py
 
     build:
       script: |
--- a/recipes/tiledbsoma/recipe.yaml
+++ b/recipes/tiledbsoma/recipe.yaml
@@ -198,3 +198,4 @@ extra:
   recipe-maintainers:
     - hadim
     - jdblischak
+    - johnkerl
--- a/recipes/tiledbsoma/recipe.yaml
+++ b/recipes/tiledbsoma/recipe.yaml
@@ -197,3 +197,4 @@ extra:
   feedstock-name: tiledbsoma
   recipe-maintainers:
     - hadim
+    - jdblischak
--- a/recipes/tiledbsoma/recipe.yaml
+++ b/recipes/tiledbsoma/recipe.yaml
@@ -1,0 +1,199 @@
+context:
+  name: tiledbsoma
+  version: 1.15.3
+
+recipe:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/single-cell-data/TileDB-SOMA/archive/refs/tags/${{ version }}.zip
+  sha256: 1c9b60b14c900e252c30586fcadaeddd455d58df624112ecea62eef24212dc61
+  patches:
+    # NOTE:
+    # - fix failing includes
+    # - fix failing linking configuration for tiledbsoma-cli and spdlog
+    - fix-lib-build.patch
+
+build:
+  number: 0
+  skip: win
+
+outputs:
+  # C++ library
+  - package:
+      name: libtiledbsoma
+
+    build:
+      script: |
+        # Configure
+        cmake -G Ninja \
+          ${CMAKE_ARGS} \
+          -DCMAKE_BUILD_TYPE="Release" \
+          -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+          -DDOWNLOAD_TILEDB_PREBUILT=OFF \
+          -DOVERRIDE_INSTALL_PREFIX=OFF \
+          -DFORCE_BUILD_TILEDB=OFF \
+          -DTILEDBSOMA_BUILD_PYTHON=ON \
+          -DTILEDBSOMA_BUILD_CLI=ON \
+          -DTILEDBSOMA_ENABLE_TESTING=OFF \
+          -DTILEDBSOMA_ENABLE_WERROR=OFF \
+          -DTileDB_DIR="${PREFIX}" \
+          -B build/ \
+          -S $SRC_DIR/libtiledbsoma
+
+        # Compile
+        cmake --build build/ -j ${CPU_COUNT}
+
+        # Install
+        cmake --build build/ -j ${CPU_COUNT} --target install-libtiledbsoma
+
+    requirements:
+      build:
+        - ${{ compiler('cxx') }}
+        - ${{ stdlib("c") }}
+        - cmake >=3.21
+        - ninja
+        - pkg-config
+      host:
+        - tiledb
+        - spdlog >=1.15
+        - fmt >=11.0
+      run_exports:
+        - ${{ pin_subpackage('libtiledbsoma', exact=True) }}
+
+    tests:
+      - script:
+          - tdbsoma
+      - package_contents:
+          files:
+            - lib/cmake/tiledbsoma/TileDBSomaConfig.cmake
+            - lib/cmake/tiledbsoma/TileDBSomaTargets-release.cmake
+            - lib/cmake/tiledbsoma/TileDBSomaTargets.cmake
+          bin:
+            - tdbsoma
+          lib:
+            - tiledbsoma
+          include:
+            - tiledbsoma/nanoarrow/nanoarrow.h
+            - tiledbsoma/nanoarrow/nanoarrow.hpp
+            - tiledbsoma/nlohmann/json.hpp
+            - tiledbsoma/reindexer/reindexer.h
+            - tiledbsoma/soma/array_buffers.h
+            - tiledbsoma/soma/column_buffer.h
+            - tiledbsoma/soma/enums.h
+            - tiledbsoma/soma/logger_public.h
+            - tiledbsoma/soma/managed_query.h
+            - tiledbsoma/soma/soma_array.h
+            - tiledbsoma/soma/soma_collection.h
+            - tiledbsoma/soma/soma_context.h
+            - tiledbsoma/soma/soma_dataframe.h
+            - tiledbsoma/soma/soma_dense_ndarray.h
+            - tiledbsoma/soma/soma_experiment.h
+            - tiledbsoma/soma/soma_geometry_dataframe.h
+            - tiledbsoma/soma/soma_group.h
+            - tiledbsoma/soma/soma_measurement.h
+            - tiledbsoma/soma/soma_multiscale_image.h
+            - tiledbsoma/soma/soma_object.h
+            - tiledbsoma/soma/soma_point_cloud_dataframe.h
+            - tiledbsoma/soma/soma_scene.h
+            - tiledbsoma/soma/soma_sparse_ndarray.h
+            - tiledbsoma/soma/span/span.hpp
+            - tiledbsoma/tiledbsoma
+            - tiledbsoma/tiledbsoma_export.h
+            - tiledbsoma/utils/arrow_adapter.h
+            - tiledbsoma/utils/common.h
+            - tiledbsoma/utils/fastercsx.h
+            - tiledbsoma/utils/parallel_functions.h
+            - tiledbsoma/utils/stats.h
+            - tiledbsoma/utils/util.h
+            - tiledbsoma/utils/version.h
+
+  # python bindings
+  - package:
+      name: tiledbsoma
+
+    build:
+      script: |
+        # Set the version
+        echo "${PKG_VERSION}" > ${SRC_DIR}/apis/python/RELEASE-VERSION
+
+        # Define prefix for tiledb and tiledbsoma
+        export TILEDB_PATH="$PREFIX"
+        export TILEDBSOMA_PATH="$PREFIX"
+
+        # Use Ninja
+        export CMAKE_GENERATOR="Ninja"
+
+        # Build the python binding
+        ${{ PYTHON }} -m pip install apis/python/ -vv --no-deps --no-build-isolation
+
+    requirements:
+      build:
+        - if: build_platform != target_platform
+          then:
+            - python
+            - cross-python_${{ target_platform }}
+            - pybind11 >=2.10.0
+
+        - ${{ compiler('cxx') }}
+        - ${{ stdlib("c") }}
+        - cmake >=3.21
+        - ninja
+        - pkg-config
+        - git
+      host:
+        - python
+        - setuptools >=70.1
+        - pip
+        - pybind11 >=2.10.0
+        # libraries
+        - libtiledbsoma
+        - tiledb
+      run:
+        - python
+        - anndata >=0.10.1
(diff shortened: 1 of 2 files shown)


hadim:

---

The OSX build works fine, but the linux one is failing with

```bash
 │ │ [35/38] Building CXX object src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/util.cc.o
 │ │ [36/38] Linking CXX shared library src/libtiledbsoma.so
 │ │ [37/38] Building CXX object src/CMakeFiles/tiledbsoma-cli.dir/cli/cli.cc.o
 │ │ [38/38] Linking CXX executable src/tdbsoma
 │ │ FAILED: src/tdbsoma 
 │ │ : && $BUILD_PREFIX/bin/x86_64-conda-linux-gnu-c++ -fvisibility-inlines-hidden -fmessage-length=0 -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/libtiledbsoma-1.15.3 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix  -O3 -DNDEBUG -Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -L$PREFIX/lib    -Wl,--dependency-file=src/CMakeFiles/tiledbsoma-cli.dir/link.d src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/reindexer/reindexer.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/managed_query.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_array.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_group.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_object.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_collection.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_experiment.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_measurement.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_scene.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_geometry_dataframe.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_point_cloud_dataframe.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_multiscale_image.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_context.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_dataframe.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_dense_ndarray.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/soma_sparse_ndarray.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/array_buffers.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/soma/column_buffer.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/arrow_adapter.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/logger.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/stats.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/util.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/version.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/external/src/thread_pool/thread_pool.cc.o src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/external/src/thread_pool/status.cc.o src/CMakeFiles/TILEDBSOMA_NANOARROW_OBJECT.dir/external/src/nanoarrow/nanoarrow.c.o src/CMakeFiles/tiledbsoma-cli.dir/cli/cli.cc.o -o src/tdbsoma  -Wl,-rpath,$SRC_DIR/build/libtiledbsoma/src:::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::  src/libtiledbsoma.so  $PREFIX/lib/libtiledb.so.2.27  -lpthread  $PREFIX/lib/libspdlog.so.1.15.0  -lpthread  $PREFIX/lib/libfmt.so.11.0.2 && :
 │ │ $BUILD_PREFIX/bin/../lib/gcc/x86_64-conda-linux-gnu/13.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: src/tdbsoma: hidden symbol `_ZN6spdlog6logger11set_patternENSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEENS_17pattern_time_typeE' in src/CMakeFiles/TILEDB_SOMA_OBJECTS.dir/utils/logger.cc.o is referenced by DSO
 │ │ $BUILD_PREFIX/bin/../lib/gcc/x86_64-conda-linux-gnu/13.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: final link failed: bad value
```

I tried a few things, but I can't make it work (either locally or in the CI). The issue happens when linking the CLI while the linking works for the library itself.

hadim:
@conda-forge/help-python-c @conda-forge/help-c-cpp ready for review!

johnkerl:
Hello @hadim!

I would note that we have https://github.com/TileDB-Inc/tiledbsoma-feedstock on the `tiledb` channel -- if this is helpful to you.

hadim:
> Hello @hadim!
> 
> I would note that we have https://github.com/TileDB-Inc/tiledbsoma-feedstock on the `tiledb` channel -- if this is helpful to you.

Thanks! I noticed that feedstock indeed, but it seems best to have it on c-f for easier maintenance and allow other deps to depend on it (https://github.com/conda-forge/pytdc-feedstock for example).

johnkerl:
> Thanks! I noticed that feedstock indeed, but it seems best to have it on c-f for easier maintenance and allow other deps to depend on it (https://github.com/conda-forge/pytdc-feedstock/pull/32 for example).

💯 agreed! Glad to see you had already found our channel for reference. :)

Thank you!

hadim:
@johnkerl should I add you as maintainer?

johnkerl:
@hadim let's discuss with @ihnorton and @jdblischak -- and @jdblischak would be a great maintainer (better than I)

I would note that we use the name:

* `tiledbsoma-py` -- Python
* `r-tiledbsoma` -- R
* `libtiledbsoma` -- C++

I note that this current PR uses the name `tiledbsoma` for Python

hadim:
Noted I will switch to `tiledbsoma-py` then and the C++ lib is already called `libtiledbsoma`.

---

Anyone who wants to be a maintainer should explicitly say it in that PR, and I will add him.

jdblischak:
@hadim please add me as a recipe maintainer. Thanks!

johnkerl:
@hadim please add me as a recipe maintainer. Thanks!

recipes/tiledbsoma/recipe.yaml
```diff
@@ -0,0 +1,200 @@
...
+  version: ${{ version }}
+
+source:
+  url: https://github.com/single-cell-data/TileDB-SOMA/archive/refs/tags/${{ version }}.zip
+  sha256: 1c9b60b14c900e252c30586fcadaeddd455d58df624112ecea62eef24212dc61
+  patches:
+    # NOTE:
+    # - fix failing includes
+    # - fix failing linking configuration for tiledbsoma-cli and spdlog
+    - fix-lib-build.patch
```
jdblischak, line 16: Could you please elaborate on how your patch works? We've had trouble in the past with linking errors related to fmt/spdlog for not only our libtiledbsoma recipe but also for some of our other conda recipes for packages that link against libtiledb. Essentially I am wondering if we need to upstream your patch to SOMA (and maybe to some of our other software) in order to alleviate future fmt/spdlog issues.

recipes/tiledbsoma/recipe.yaml
```diff
@@ -0,0 +1,200 @@
...
+
+    requirements:
+      build:
+        - ${{ compiler('cxx') }}
+        - ${{ stdlib("c") }}
+        - cmake >=3.21
+        - ninja
+        - pkg-config
+      host:
+        - tiledb
```
jdblischak, line 59: Note that we strictly pin `tiledb` in our recipe since we often migrate minor releases of tiledb in tandem with the client packages. I think you should be able to instead use the conda-forge migrations, but if a migration PR fails, it's likely because it needs to be combined with the latest tiledbsoma update PR.

hadim: `tiledb` is pinned at c-f level at https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/1ae12810d4b2f2104a3862a1d11113fb8b4981a7/recipe/conda_build_config.yaml#L960 and so this will be automatically managed for this recipe.

Is that what you mean?

jdblischak: Yes, exactly. But in our recipe we directly pin `tiledb` in the recipe requirements, which gives us more control. Just wanted to give you the heads up on the difference

recipes/tiledbsoma/recipe.yaml
hadim: To be honest, it was a lot of trial and error to make it work. I am not sure 100% sure why it is needed here and not upstream. I guess it's due to a few differences in the build tool chain in between your feedstocks and the c-f ones?

The important difference I can see here is that the build uses `fmt` and `spdlog` from the c-f channel, which is the preferred way to do it in c-f.

Sorry, I can't provide you more hints here whether you should upstream this patch or not. Maybe try to un-vendor `fmt` and `spdlog` and only rely on the ones from c-f?

recipes/tiledbsoma/recipe.yaml
jdblischak: > As a tip, when building conda packages outside c-f, I still try to use the c-f pinning YAML file, so I guarantee (sort of!) a compat with c-f.

It's more complex than that. We still use the c-f pinning files because we use conda-smithy to manage our feedstocks. In the case of the tiledb pin, we just make it more strict than the global pin by manually pinning it in the recipe. In other words, we maintain a strict 1:1 relationship between minor releases of tiledb and its downstream clients. The global c-f pinning and its migrations can't handle that level of detail. It will open a tiledb migration PR against the previous minor release of SOMA, which in many cases will fail due to new features that were added. However, tiledb is more stable now, so hopefully this won't be a problem as often going forward.

hadim: Ok, thanks for the details.

If that causes an issue on the c-f package, it's all good, since we can always override the default pinning in the feedstock as needed.

martin-g (approved):

recipes/tiledbsoma/recipe.yaml
jdblischak: >  it was a lot of trial and error to make it work. I am not sure 100% sure why it is needed here and not upstream. 

@hadim thanks for the explanation and advice.

> The important difference I can see here is that the build uses fmt and spdlog from the c-f channel, which is the preferred way to do it in c-f.

Interesting. While our CMake files support bootstrapping fmt/spdlog in a non-conda-build, our intention is for our feedstock builds to install all dependencies from conda-forge. That is why we list fmt/spdlog in the [host requirements](https://github.com/TileDB-Inc/tiledbsoma-feedstock/blob/6b56770bf1171f3d74ab4ec1124fb254c35788f9/recipe/meta.yaml#L50-L51).

Looking at our most recent feedstock build ([commit](https://github.com/TileDB-Inc/tiledbsoma-feedstock/commit/6b56770bf1171f3d74ab4ec1124fb254c35788f9), [Azure build](https://dev.azure.com/TileDB-Inc/CI/_build/results?buildId=41701&view=results)), it looks like we are successfully linking against fmt from conda-forge but not spdlog:

```
   INFO (libtiledbsoma,lib/libtiledbsoma.so): Needed DSO lib/libfmt.so.11 found in conda-forge/linux-64::fmt==11.0.2=h434a139_0
WARNING (libtiledbsoma): run-exports library package conda-forge/linux-64::spdlog==1.15.0=h10c9db5_0 in requirements/run but it is not used (i.e. it is overdepending or perhaps statically linked? If that is what you want then add it to `build/ignore_run_exports`)
```

Whereas in your [build](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1150399&view=results) for this PR with the new recipe format, it looks like you are successfully linking against fmt _and_ spdlog:

```
│ │ New relative path: $ORIGIN/
│ │ Patched dynamic section of "/tmp/libtiledbsomaE15RFd/lib/libtiledbsoma.so"
│ │ 
│ │ [lib/libtiledbsoma.so] links against:
│ │  ├─ lib/libgcc_s.so.1 (libgcc)
│ │  ├─ ld-linux-x86-64.so.2 (system)
│ │  ├─ libpthread.so.0 (system)
│ │  ├─ lib/libspdlog.so.1.15.0 (spdlog)
│ │  ├─ lib/libfmt.so.11.0.2 (fmt)
│ │  ├─ lib/libstdc++.so.6.0.33 (libstdcxx)
│ │  ├─ libc.so.6 (system)
│ │  └─ lib/libtiledb.so.2.27 (tiledb)
│ │ 
│ │ [bin/tdbsoma] links against:
│ │  ├─ libc.so.6 (system)
│ │  ├─ lib/libfmt.so.11.0.2 (fmt)
│ │  ├─ lib/libtiledbsoma.so (package)
│ │  ├─ lib/libspdlog.so.1.15.0 (spdlog)
│ │  ├─ lib/libtiledb.so.2.27 (tiledb)
│ │  ├─ ld-linux-x86-64.so.2 (system)
│ │  ├─ libpthread.so.0 (system)
│ │  ├─ lib/libgcc_s.so.1 (libgcc)
│ │  └─ lib/libstdc++.so.6.0.33 (libstdcxx)
│ │ Post-processing done!
```

This suggests that we do need to fix something upstream to support both conda and non-conda builds.

hadim:
@conda-forge/help-python-c @conda-forge/help-c-cpp ready for review!

xhochy (approved):

hadim:
Thank you @xhochy !

PR #28830, opened 2025-01-14
[MNT] Support `TARGET_ARCH` for local osx build

hadim:
It allows building `osx-arm64` package locally. Without it the target-platform is set to `osx-64` which set the wrong compilation flags.


hadim:
@conda-forge/core: please review

PR #28837, opened 2025-01-14
Add Windmapper

Chrismarsh:
Add windmapper


Chrismarsh:
@conda-forge/help-python this is ready for review

recipes/windmapper/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - gdal
+    - mpi4py
+    - windninja
+
+test:
+  imports:
+    - windmapper
+  requires:
+    - pip
+
```
ocefpaf, line 46:
```suggestion
  requires:
    - pip
    - python {{ python_min }}
  commands:
    - pip check
```

recipes/windmapper/LICENSE
```diff
@@ -0,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
```
ocefpaf, line 1: This file is in the tarball, no need to add it here again.

recipes/windmapper/meta.yaml
Chrismarsh: There is currently no `--help`

ocefpaf: Is there any commad that we can use to test the entry-point?

Chrismarsh: It doesn't at the moment. Is that required?

ocefpaf: Not required, but nice to have.

recipes/windmapper/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - cloudpickle
+    - gdal
+    - mpi4py
+    - windninja
+
+test:
+  imports:
+    - windmapper
+  requires:
+    - pip
```
ocefpaf, line 45: Please add the pip-check command test.

recipes/windmapper/meta.yaml
Chrismarsh: I'll make sure to add it in the future

PR #28846, opened 2025-01-16
Adding Fandango, a python library for configuring and managing Tango Control Systems

sergirubio:
The Tango Control System is a collaboration between several scientific institutes for an open source control system, most of them using conda-forge. Fandango is used by many of these institutes to create, configure and manage their control system from command line.

@beenje 

Checklist
- [ ] Build number is 0.


recipes/fandango/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+test:
+  imports:
+    - fandango
+  commands:
+    - pip check
+  requires:
+    - python {{ python_min }}
+    - pip
+    - future
+    - setuptools
```
beenje, line 37: future and setuptools aren't needed here.
(future is a run requirement so it will be installed automatically).

recipes/fandango/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools    
+    - future    
+  run:
+    - python >={{ python_min }}
+    - future
```
beenje, line 26: This matches what you have in your setup.py but it seems some dependencies are missing.
At least `pytango` should be defined here.

```
import: 'fandango'
Unable to import fandango.*tango modules
```

Just a question: shouldn't that raise an error instead of just printing some message?
The import test passes but that sounds incorrect. I guess you can't do much without tango installed.

beenje (changes requested):
You can't have [scripts](https://gitlab.com/tango-controls/fandango/-/blob/master/setup.py?ref_type=heads#L218) in your `setup.py` for a noarch package.

Those should be converted to entry points.

recipes/fandango/meta.yaml
sergirubio: Hi Benjamin, this is intentional

We do use fandango in some systems where the tango control system is not available, in extraction apis from archiving or web clients.

We just print the warning but allow to use the rest of functionality that does not depend on pytango

recipes/fandango/meta.yaml
sergirubio: ok, I will update that

recipes/fandango/meta.yaml
sergirubio: I will modify the scripts to be entry points and then proceed with the merge

I'll make a new release for fandango and update yaml

sergirubio:
Hi @beenje , I have created a new 16.0.0 version replacing scripts by entry points, I tested it on an environment with python 3.12.

recipes/fandango/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/fandango-{{ version }}.tar.gz
+  sha256: a4530b04d67f8b42e6f4924ded1ba146a63c4d1fd2fcae2584d36c64bf94a625
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
```
beenje, line 16: You need to explicitly define the entry points here:

```
  entry_points:
    - fandango = fandango.scripts.fandango_cli:main
    - tango2csv = fandango.scripts.tango2csv:main
    - tango2json = fandango.scripts.tango2json:main
    - tango_host = fandango.scripts.tango_host:main
    - tango_monitor = fandango.scripts.tango_monitor:main
    - tango_servers = fandango.scripts.tango_servers:main
```

recipes/fandango/meta.yaml
beenje: Still need to remove future and setuptools

recipes/fandango/meta.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - future    
+  run:
+    - python >={{ python_min }}
+    - future
+
+test:
+  imports:
+    - fandango
+  commands:
+    - pip check
```
beenje, line 37: Would be nice to test some entry points here. I'd add:

```suggestion
    - pip check
    - fandango --help
```

I guess other entry points require extra dependencies, so you can skip them.

beenje (changes requested):
The entry points have to be defined explicitly. See my other comments.

After that, it should be good!

ovallcorba:
> Hi! This is the staged-recipes linter and I found some lint.
> 
> File-specific lints and/or hints:
> 
> * `recipes/fandango/meta.yaml`:
>   
>   * lints:
>     
>     * The following maintainers have not yet confirmed that they are willing to be listed here: ovallcorba. Please ask them to comment on this PR if they are.

Hi! its ok! thanks!

beenje:
The recipe looks good to me.

But looking at the logs, you have some syntax errors in 2 files:

```
  *** Error compiling '$PREFIX/lib/python3.10/site-packages/fandango/scripts/tango2json.py'...
    File "$PREFIX/lib/python3.10/site-packages/fandango/scripts/tango2json.py", line 55
      devs.extend([d for d in server_devs if not exported or fn.check_device(d)])]
                                                                                 ^
  SyntaxError: unmatched ']'
  
  *** Error compiling '$PREFIX/lib/python3.10/site-packages/fandango/scripts/tango_monitor.py'...
    File "$PREFIX/lib/python3.10/site-packages/fandango/scripts/tango_monitor.py", line 149
      def main()
                ^
  SyntaxError: expected ':'
```

Those 2 entry points are broken.

You might want to make a new release to fix that?
I'd recommend running a linter on your code to avoid this in the future.

sergirubio:
Thanks @benjee, I passed flake8 and fixed the errors in a new 16.0.3 version.

PR #28864, opened 2025-01-18
Adding junifer-data

synchon:
Checklist
- [x] Build number is 0.


synchon:
Confirming my listing to be the maintainer.

synchon:
@conda-forge/staged-recipes ready for review! Please ping @conda-forge/help-python

ocefpaf:
Both macOS and Windows are failing with:

```
pip check
annexremote 1.2.1 is not supported on this platform
```

Is this package Linux only? If so, can you do a OS-noarch here? See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies for more info.

synchon:
> Both macOS and Windows are failing with:
> 
> ```
> pip check
> annexremote 1.2.1 is not supported on this platform
> ```
> 
> Is this package Linux only? If so, can you do a OS-noarch here? See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies for more info.

Hi, thanks for the review! Here's the recipe link for a package I maintain with a subset of the requirements: https://github.com/conda-forge/junifer-feedstock/blob/main/recipe/meta.yaml . Any reason why that doesn't fail while building?

ocefpaf:
Ideally `annexremote` should do this as well, thy have a noarch package that is installable on platforms that it should not. To prevent further delays here, you can make this package noarch and linux only with:

recipes/junifer-data/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - setuptools-scm >=8
+    - pip
+  run:
+    - python >={{ python_min }}
+    - click >=8.1.3,<8.2
+    - datalad >=1.0.0,<1.2.0
+    - lazy_loader ==0.4
+    - __linux  # [linux]
+    - __osx  # [osx]
+    - __win  # [win]
```
ocefpaf: You do not want osx and windows, right? Also, check step 2 in the docs link. You are missing a file specifying the platforms.

synchon: The package should run on all platforms and it runs as intended when installed directly via `pip` on windows and osx.
Ah yes I see the missing `noarch_platforms`.

synchon: Is there a way I can ignore `annexremote` requirement on osx and windows?

ocefpaf: Is `annexremote` is an optional dependency?

synchon: It's not a direct dependency of the package so can be considered optional.

ocefpaf: If so, I recommend removing it from the package. Or you can consider a multiple output recipe with a junifer-data-(name of the extra upstream).

synchon: Thanks for the suggestions! The issue is that I don't have `annexremote` as a dependency in the package recipe, it's something that is being required when installing some other package (most likely, `datalad`). The strange thing is that in other recipe where `datalad` is required, `annexremote` issue doesn't pop up.

recipes/junifer-data/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - python >={{ python_min }}
+    - click >=8.1.3,<8.2
+    - datalad >=1.0.0,<1.2.0
+    - lazy_loader ==0.4
+
+test:
+  imports:
+    - junifer_data
+  commands:
+    - junifer-data --help
```
ocefpaf, line 35:
```suggestion
    - junifer-data --help
    - pip check
```

recipes/junifer-data/meta.yaml
ocefpaf: 1. datalad is not noarch, but it should
2. the real issue is that there are versions of annexremote should be linux only and it seems that some versions aren't

I see you removed it now. Let's see if the CIs pick the a solvable env this time.

recipes/junifer-data/meta.yaml
synchon: The `pip check` would raise the `annexremote` issue again, wouldn't it?

ocefpaf: We need to fix that upstream. Leave it here and I'll work our way out of this mess. If you ignore it we just pass the problem along to the next package.

synchon: Alright that's fair. Thanks for your help!

synchon: Added `pip check` now.

ocefpaf: After checking both datalad and annexremote I don't think it is worth my time trying to fix them The former is Python 3.9 only, which is reaching EoL soon, and the annexremote seems to have fixed it it for latest buids, meaning datalad can add a pin to newer versions and escape this mess.

TL;DR comment out the pip-check, do not remove it, and add a link to these comments so people facing issues in the future can track what happened and try to fix it later.

synchon: > TL;DR comment out the pip-check, do not remove it, and add a link to these comments so people facing issues in the future can track what happened and try to fix it later.

Done.

PR #28914, opened 2025-01-24
Add papylio

ivoseverins:
Checklist
- [x] Build number is 0.


ivoseverins:
Not sure whether the staged-recipes linter error is related to this pull request. Other pull requests seem to have the error as well.

@conda-forge/help-python, ready for review

ivoseverins:
@conda-forge/help-python, please let me know in case you require are changes for merging.

recipes/papylio/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/papylio-{{ version }}.tar.gz
+  sha256: d7187e587427fee9b45d635f0e59ab412338ff3dbefc49b87f87057842b8c3b0
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
```
ocefpaf, line 14: This recipe can be noarch.

ivoseverins: Thank you for the quick reply.
I tried to make the recipe noarch again, however, it does not seem to work as this line is linux specific:
` - libgl-devel  # [linux]`
This package is needed for opencv on linux and is only available for linux, but not on the windows and mac.
See: https://conda-forge.org/docs/maintainer/maintainer_faq/#how-do-i-fix-the-libglso1-import-error

ocefpaf: You can remove that pre-processor selector and let it build on Linux only. Windows will fail but the noarch package will be installable there too.

recipes/papylio/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/papylio-{{ version }}.tar.gz
+  sha256: d7187e587427fee9b45d635f0e59ab412338ff3dbefc49b87f87057842b8c3b0
+
+build:
+  noarch: python
+  skip: true # [not linux]
```
ocefpaf, line 14:
```suggestion
```

recipes/papylio/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+requirements:
+  host:
+    - python {{ python_min }}
+    - setuptools >=61.0.0
+    - setuptools-scm >=8
+    - wheel
+    - pip
+    - libgl-devel
+  run:
+    - python >=3.9,<3.10
```
ocefpaf, line 27:
```suggestion
    - python {{ python_min }},<3.10
```

recipes/papylio/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - numba >=0.56,<0.57
+    - pomegranate ==0.14.8
+    - matchpoint >=1.0,<1.1
+    - objectlist >=1.0,<1.1
+
+test:
+  imports:
+    - papylio
+  requires:
+    - pip
```
ocefpaf, line 60:
```suggestion
test:
  imports:
    - papylio
  commands:
    - pip check
  requires:
    - python {{ python_min }}
    - pip
```

ivoseverins: Due to differences in naming between some of the pypi and conda-forge packages (e.g. matplotlib and matplotlib-base) pip check gave an error earlier.

ocefpaf: The current error is:

```
    import cv2
ImportError: libGL.so.1: cannot open shared object file: No such file or directory
```

That is probably b/c you need a yum requirement file like this one https://github.com/conda-forge/pygridgen-feedstock/blob/main/recipe/yum_requirements.txt

Let's use pip-check so we can see what are the errors first and if we can fix them.

ivoseverins:
On windows it now gives the error:
```
papylio 0.7 requires pyside2, which is not installed.
papylio 0.7 has requirement h5netcdf<1.1,>=1.0, but you have h5netcdf 0.0.0.
```
And on osx
```
papylio 0.7 has requirement h5netcdf<1.1,>=1.0, but you have h5netcdf 0.0.0.
papylio 0.7 has requirement pomegranate==0.15; sys_platform != "win32", but you have pomegranate 0.14.8.
```

The error for pomegranate I can understand, because version 0.15 is only available on pypi, not on conda-forge, so the specification in the meta.yml file is different from the package on pypi.

Not sure why the errors for the other packages occur, because in the list of installed packages they seem to be installed:
```
    pyside2:                 5.13.2-py39hc084194_8        conda-forge
    h5netcdf:                1.0.2-pyhd8ed1ab_0           conda-forge

```

ocefpaf:
Looks like pip-check is failing:

```
+ pip check
papylio 0.7 has requirement h5netcdf<1.1,>=1.0, but you have h5netcdf 0.0.0.
papylio 0.7 has requirement pomegranate==0.15; sys_platform != "win32", but you have pomegranate 0.14.8.
```

One is a bug in an older version of `h5netcdf`, unlikely to be fixed. the other one looks like you could fix here by using `pomegranate==0.15` for both Linux and Windows.

PS: You will need to skip pip-check anyways, please add a comment in the recipe with this info so we can address it in the future when the dependencies change.

PR #28924, opened 2025-01-24
add colour-checker-detection

carlodri:
https://github.com/colour-science/colour-checker-detection

Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/colour-checker-detection/meta.yaml
+++ b/recipes/colour-checker-detection/meta.yaml
@@ -1,0 +1,49 @@
+{% set name = "colour-checker-detection" %}
+{% set version = "0.2.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/colour_checker_detection-{{ version }}.tar.gz
+  sha256: 68dd0d48aa0f4355ecd0640eee52d3b53c90b24a7d966f422f5513de1bc39e83
+
+build:
+  entry_points:
+    - colour-checker-detection-inference = colour_checker_detection.scripts.inference:segmentation
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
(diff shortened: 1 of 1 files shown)


carlodri:
@conda-forge/staged-recipes this is ready for review/merge

recipes/colour-checker-detection/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/colour_checker_detection-{{ version }}.tar.gz
+  sha256: 68dd0d48aa0f4355ecd0640eee52d3b53c90b24a7d966f422f5513de1bc39e83
+
+build:
+  entry_points:
+    - colour-checker-detection-inference = colour_checker_detection.scripts.inference:segmentation
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
```
ocefpaf, line 18:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation 
```

carlodri: Thanks for the correction @ocefpaf. I was not sure about this since I didn't see those arguments in the example recipe, so I removed them from the grayskull-generated recipe...

synapticarbors: I'm actually confused about this, since my impression was that these flags don't need to be set since they are being handled by conda-build via env variables:

https://github.com/conda/conda-build/blob/cf35214183b6e549675874ed37c7e01c6b7322dc/conda_build/build.py#L3014-L3044

Additionally since they aren't in the example, it seems like we are saying it's not best practice to use those flags.

carlodri: That's what I thought too..

ocefpaf: Those are necessary when using multiple outputs and to be safe in case of regressions. They should go away as we move to rattler build, until then, better add them.

PR #28931, opened 2025-01-25
Add vllm

maresb:
Very rough draft. I will almost certainly require help.

Opened on the advice of @h-vetinari in https://github.com/conda-forge/xformers-feedstock/issues/42

Direct and transitive dependencies:
- [x] compressed-tensors #28778
- [x] partial-json-parser #28793 
- [x] outlines-core #28863
- [x] outlines https://github.com/conda-forge/outlines-feedstock/pull/33
- [x] airportsdata #28861
- [x] depyf #28865
- [x] libnacl #28930 
- [x] tensorizer #28932
- [ ] xgrammar #28866

Checklist
- [x] Build number is 0.


maresb:
Interesting, we're getting different results between CUDA 11.8 and 12.0.

Both fail in the following command:

```python
['cmake', '$SRC_DIR', '-G', 'Ninja', '-DCMAKE_BUILD_TYPE=RelWithDebInfo', 
'-DVLLM_TARGET_DEVICE=cuda', '-DVLLM_PYTHON_EXECUTABLE=$PREFIX/bin/python', 
'-DVLLM_PYTHON_PATH=$PREFIX/lib/python3.9/site-packages/pip/_vendor/pyproject_hooks/_in_process:$PREFIX/lib/python39.zip:$PREFIX/lib/python3.9:$PREFIX/lib/python3.9/lib-dynload:$PREFIX/lib/python3.9/site-packages:$PREFIX/lib/python3.9/site-packages/setuptools/_vendor', 
'-DFETCHCONTENT_BASE_DIR=$SRC_DIR/.deps', '-DNVCC_THREADS=1',
'-DCMAKE_JOB_POOL_COMPILE:STRING=compile', '-DCMAKE_JOB_POOLS:STRING=compile=2']
```

12.0 fails earlier at CUDA detection:

```
 │ │       -- Caffe2: Found protobuf with new-style protobuf targets.
 │ │       -- Caffe2: Protobuf version 28.2.0
 │ │       -- Could NOT find CUDA (missing: CUDA_INCLUDE_DIRS) (found version "12.0")
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:31 (message):
 │ │         Caffe2: CUDA cannot be found.  Depending on whether you are building Caffe2
 │ │         or a Caffe2 dependent library, the next warning / error will give you more
 │ │         info.
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
 │ │       
 │ │       
 │ │       CMake Error at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:90 (message):
 │ │         Your installed Caffe2 version uses CUDA but I cannot find the CUDA
 │ │         libraries.  Please set the proper CUDA prefixes and / or install CUDA.
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
```

11.8 gets further:

```
 │ │       -- Caffe2: Found protobuf with new-style protobuf targets.
 │ │       -- Caffe2: Protobuf version 28.2.0
 │ │       -- Found CUDA: /usr/local/cuda (found version "11.8")
 │ │       -- The CUDA compiler identification is NVIDIA 11.8.89 with host compiler GNU 11.4.0
 │ │       -- Detecting CUDA compiler ABI info
 │ │       -- Detecting CUDA compiler ABI info - done
 │ │       -- Check for working CUDA compiler: $PREFIX/bin/nvcc - skipped
 │ │       -- Detecting CUDA compile features
 │ │       -- Detecting CUDA compile features - done
 │ │       -- Found CUDAToolkit: /usr/local/cuda/include (found version "11.8.89")
 │ │       -- Caffe2: CUDA detected: 11.8
 │ │       -- Caffe2: CUDA nvcc is: /usr/local/cuda/bin/nvcc
 │ │       -- Caffe2: CUDA toolkit directory: /usr/local/cuda
 │ │       -- Caffe2: Header version is: 11.8
 │ │       -- Found Python: $PREFIX/bin/python (found version "3.9.21") found components: Interpreter
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:140 (message):
 │ │         Failed to compute shorthash for libnvrtc.so
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
 │ │       
 │ │       
 │ │       CMake Warning (dev) at $PREFIX/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:441 (message):
 │ │         The package name passed to `find_package_handle_standard_args` (nvtx3) does
 │ │         not match the name of the calling package (Caffe2).  This can lead to
 │ │         problems in calling code that expects `find_package` result variables
 │ │         (e.g., `_FOUND`) to follow a certain pattern.
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:174 (find_package_handle_standard_args)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
 │ │       This warning is for project developers.  Use -Wno-dev to suppress it.
 │ │       
 │ │       -- Could NOT find nvtx3 (missing: nvtx3_dir)
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:180 (message):
 │ │         Cannot find NVTX3, find old NVTX instead
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
 │ │       
 │ │       
 │ │       -- USE_CUDNN is set to 0. Compiling without cuDNN support
 │ │       -- USE_CUSPARSELT is set to 0. Compiling without cuSPARSELt support
 │ │       -- USE_CUDSS is set to 0. Compiling without cuDSS support
 │ │       -- USE_CUFILE is set to 0. Compiling without cuFile support
 │ │       -- Automatic GPU detection failed. Building for common architectures.
 │ │       -- Autodetected CUDA architecture(s): 3.5;5.0;8.0;8.6;8.9;9.0;8.9+PTX;9.0+PTX
 │ │       -- Added CUDA NVCC flags for: -gencode;arch=compute_35,code=sm_35;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_89,code=sm_89;-gencode;arch=compute_90,code=sm_90;-gencode;arch=compute_89,code=compute_89;-gencode;arch=compute_90,code=compute_90
 │ │       -- Found Torch: $PREFIX/lib/libtorch.so
 │ │       -- CUDA target architectures: 3.5;5.0;8.0;8.6;8.9;9.0
 │ │       -- CUDA supported target architectures: 8.0;8.6;8.9;9.0
 │ │       -- FetchContent base directory: $SRC_DIR/.deps
 │ │       CMake Error at $PREFIX/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:943 (message):
 │ │         error: could not find git for clone of cutlass-populate
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
 │ │         CMakeLists.txt:29 (ExternalProject_Add)
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:84 (find_package)
 │ │       
 │ │       
 │ │       -- USE_CUDNN is set to 0. Compiling without cuDNN support
 │ │       -- USE_CUSPARSELT is set to 0. Compiling without cuSPARSELt support
 │ │       -- USE_CUDSS is set to 0. Compiling without cuDSS support
 │ │       -- USE_CUFILE is set to 0. Compiling without cuFile support
 │ │       -- Automatic GPU detection failed. Building for common architectures.
 │ │       -- Autodetected CUDA architecture(s): 3.5;5.0;8.0;8.6;8.9;9.0;8.9+PTX;9.0+PTX
 │ │       -- Added CUDA NVCC flags for: -gencode;arch=compute_35,code=sm_35;-gencode;arch=compute_50,code=sm_50;-gencode;arch=compute_80,code=sm_80;-gencode;arch=compute_86,code=sm_86;-gencode;arch=compute_89,code=sm_89;-gencode;arch=compute_90,code=sm_90;-gencode;arch=compute_89,code=compute_89;-gencode;arch=compute_90,code=compute_90
 │ │       -- Found Torch: $PREFIX/lib/libtorch.so
 │ │       -- CUDA target architectures: 3.5;5.0;8.0;8.6;8.9;9.0
 │ │       -- CUDA supported target architectures: 8.0;8.6;8.9;9.0
 │ │       -- FetchContent base directory: $SRC_DIR/.deps
 │ │       CMake Error at $PREFIX/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:943 (message):
 │ │         error: could not find git for clone of cutlass-populate
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/share/cmake-3.31/Modules/ExternalProject.cmake:3041 (_ep_add_download_command)
 │ │         CMakeLists.txt:29 (ExternalProject_Add)
 │ │       
 │ │       
 │ │       -- Configuring incomplete, errors occurred!
```

h-vetinari:
Thanks @maresb! I had forgotten that there's already #24710, perhaps @mediocretech would be interested in collaborating?

W.r.t. CUDA, we need to move on from 12.0 here, which isn't used anywhere else in conda-forge anymore - it's just that staged-recipes seems to have been forgotten in the context of https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/6630.

maresb:
Oh, I didn't notice that effort, thanks @h-vetinari! Although that's old it looks like @rongou is eager to help! :rocket: 

Do you think that CUDA 12.0 is actually causing a problem here? I was thinking (i.e. wildly guessing) that we need to patch `CMakeLists.txt`, but I've never used cmake. :disappointed:

h-vetinari:
Mainly I want to avoid redundant work. As soon as https://github.com/conda-forge/staged-recipes/pull/28938 is in and we have merged main here, I'll be happy to take a look what's going on.

h-vetinari:
In any case, you'll have to address
```
 │ │       CMake Error at $PREFIX/share/cmake-3.31/Modules/ExternalProject/shared_internal_commands.cmake:943 (message):
 │ │         error: could not find git for clone of cutlass-populate
```

maresb:
Woah, it's compiling on CUDA 11.8 until it runs out of memory and crashes. Maybe I can add some swap:

```
...
 │ │ Building wheels for collected packages: vllm
 │ │   Building wheel for vllm (pyproject.toml): started
 │ │   Building wheel for vllm (pyproject.toml): still running...
...
 │ │   Building wheel for vllm (pyproject.toml): still running...
##[warning]Free memory is lower than 5%; Currently used: 95.80%
##[warning]Free memory is lower than 5%; Currently used: 95.80%
##[warning]Free memory is lower than 5%; Currently used: 95.80%
##[warning]Free memory is lower than 5%; Currently used: 95.80%
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
```

maresb:
Ah, hmm, I just added swap to conda-forge.yml. Not sure how that's supposed to work here on staged-recipes. :thinking:

h-vetinari:
> On CUDA 12.x I'm hitting the error:

We're (now) aware of the CUDA-angle of https://github.com/conda-forge/pytorch-cpu-feedstock/issues/333



> On 11.8, after adding `git` as a host dependency, compilation starts but it runs out of memory. I tried to add swap by editing `conda-forge.yml`, but it didn't apply to the CUDA builds.

https://github.com/conda-forge/staged-recipes/pull/28979

maresb:
I would have hoped to get more out of setting `VERBOSE=1`.

It's supposed to add the flag `-DCMAKE_VERBOSE_MAKEFILE=ON`. Maybe it's not relevant at this stage?

Here's the corresponding Python code to go from the envvar to get the flag:

https://github.com/vllm-project/vllm/blob/a1fc18c030e4d0466f2b23cb7dd4d11ce4b85603/vllm/envs.py#L138-L140

https://github.com/vllm-project/vllm/blob/a1fc18c030e4d0466f2b23cb7dd4d11ce4b85603/setup.py#L132-L134

shermansiu:
Hmm, it still appears broken after https://github.com/conda-forge/pytorch-cpu-feedstock/pull/339.

```
Could NOT find CUDA (missing: CUDA_INCLUDE_DIRS) (found version "12.6")
```

Is `CUDA_INCLUDE_DIRS` properly set?

shermansiu:
The CUDA 11.8 build probably fails because it's out of disk space, but that's just speculation:

`##[warning]Free disk space on / is lower than 5%; Currently used: 95.08%` (x5)

`##[warning]Free memory is lower than 5%; Currently used: 96.11%` (x5)

maresb:
Hey @shermansiu, great to have you around!!!

I'm a bit lost since I'm not very familiar with CUDA.

I was just now having some trouble getting the CI to rerun the CUDA builds, but rebasing seems to have fixed it.

Also, post-rebase things seem to be proceeding slightly further for 12.6:

```
 │ │       -- Caffe2: Found protobuf with new-style protobuf targets.
 │ │       -- Caffe2: Protobuf version 28.3.0
 │ │       -- Unable to find cublas_v2.h in either "$PREFIX/targets/x86_64-linux/include" or "$PREFIX/math_libs/include"
 │ │       -- Found CUDAToolkit: $PREFIX/targets/x86_64-linux/include (found version "12.6.85")
 │ │       -- Check for working CUDA compiler: $PREFIX/bin/nvcc - skipped
 │ │       -- Detecting CUDA compile features
 │ │       -- Detecting CUDA compile features - done
 │ │       -- Unable to find cublas_v2.h in either "$PREFIX/targets/x86_64-linux/include" or "$PREFIX/math_libs/include"
 │ │       -- Caffe2: CUDA detected: 12.6.85
 │ │       -- Caffe2: CUDA nvcc is: $PREFIX/bin/nvcc
 │ │       -- Caffe2: CUDA toolkit directory:
 │ │       -- Caffe2: Header version is: 12.6
 │ │       CMake Error at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:107 (get_target_property):
 │ │         get_target_property() called with non-existent target "CUDA::nvrtc".
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       
 │ │       
 │ │       -- Found Python: $PREFIX/bin/python (found version "3.9.21") found components: Interpreter
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:116 (message):
 │ │         Failed to compute shorthash for libnvrtc.so
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       
 │ │       
 │ │       CMake Warning (dev) at $PREFIX/share/cmake-3.31/Modules/FindPackageHandleStandardArgs.cmake:441 (message):
 │ │         The package name passed to `find_package_handle_standard_args` (nvtx3) does
 │ │         not match the name of the calling package (Caffe2).  This can lead to
 │ │         problems in calling code that expects `find_package` result variables
 │ │         (e.g., `_FOUND`) to follow a certain pattern.
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:154 (find_package_handle_standard_args)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       This warning is for project developers.  Use -Wno-dev to suppress it.
 │ │       
 │ │       -- Could NOT find nvtx3 (missing: nvtx3_dir)
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:160 (message):
 │ │         Cannot find NVTX3, find old NVTX instead
 │ │ Failed to build vllm
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       
 │ │       
 │ │       -- USE_CUDNN is set to 0. Compiling without cuDNN support
 │ │       -- USE_CUSPARSELT is set to 0. Compiling without cuSPARSELt support
 │ │       -- USE_CUDSS is set to 0. Compiling without cuDSS support
 │ │       -- USE_CUFILE is set to 0. Compiling without cuFile support
 │ │       -- Added CUDA NVCC flags for:
 │ │       CMake Warning at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:22 (message):
 │ │         static library kineto_LIBRARY-NOTFOUND not found.
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:120 (append_torchlib_if_found)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       
 │ │       
 │ │       -- Found Torch: $PREFIX/lib/libtorch.so
 │ │       CMake Error at CMakeLists.txt:122 (message):
 │ │         Can't find CUDA or HIP installation.
```

I'm not too sure what this means or how to fix it.

shermansiu:
Hmm, I'd like to build the recipe locally to diagnose this further, but at a glance, the following line looks a bit concerning:

```
 -- Unable to find cublas_v2.h in either "$PREFIX/targets/x86_64-linux/include" or "$PREFIX/math_libs/include"
```

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,108 @@
...
+  - packaging
+  - wheel
+  - jinja2
+  - pytorch * cuda*
+  - cmake
+  - git
+  - ${{ stdlib('c') }}
+  - ${{ compiler('c') }}
+  - ${{ compiler('cxx') }}
+  - ${{ compiler('cuda') }}
```
h-vetinari, line 38: All this (+ninja) should move to the build environment.

h-vetinari (commented):
Well, you need more than just `{{ compiler("cuda") }}` to get what all the CUDA components you need.

Look like you need at minimum
```yaml
    - cuda-version =={{ cuda_compiler_version }}
    - cuda-cudart-dev
    - cuda-nvrtc-dev
    - libcublas-dev
```
in the host environment. Also note that we're still figuring out an issue with nvtx, see https://github.com/conda-forge/pytorch-cpu-feedstock/issues/357

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,122 @@
...
+    entry_points:
+    - vllm  =  vllm.scripts:main
+  skip: cuda_compiler_version == "None"
+
+requirements:
+  build:
+  - cmake
+  - ninja
+  - git
+  - ${{ stdlib('c') }}
```
shermansiu, line 28:
```suggestion
  - ${{ stdlib('c') }}
    - nvtx-c
```

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,122 @@
...
+  version: ${{ version }}
+
+source:
+- url: https://pypi.org/packages/source/v/vllm/vllm-${{ version }}.tar.gz
+  sha256: 91521ca3e7629e2a40a4fceea3290cbde34c3ea6cb291ee9f84ad2ce752e65be
+
+build:
+  number: 0
+  script: |
+    python use_existing_torch.py
```
shermansiu, line 16:
```suggestion
    python use_existing_torch.py
    mkdir -p $SRC_DIR/third_party/NVTX/c
    ln -s $CONDA_PREFIX/include $SRC_DIR/third_party/NVTX/c/include
```

shermansiu (changes requested):
This seems to resolve the nvtx issue, but then it complains about not being able to find `kineto`. 

Using `USE_KINETO=0` doesn't seem to work because the existing PyTorch .cmake files in the environment already have kineto enabled.

`lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake`
```
if(ON)
  append_torchlib_if_found(kineto)
endif()
```

See:
- https://github.com/vllm-project/vllm/issues/12989
- https://github.com/pytorch/pytorch/issues/51026
- https://github.com/pytorch/builder/pull/650/files

recipes/vllm/recipe.yaml
shermansiu: Add `nvtx-c` as a dependency

recipes/vllm/recipe.yaml
shermansiu: Properly link up nvtx so the code can find it.

maresb:
Thanks @shermansiu for the suggestions!!!

We can create a patch if you think it would help to modify the cmake files.

h-vetinari:
> but then it complains about not being able to find `kineto`.

This is very likely related to https://github.com/conda-forge/pytorch-cpu-feedstock/issues/335 - could you make a comment there?

maresb:
> This is very likely related to [conda-forge/pytorch-cpu-feedstock#335](https://github.com/conda-forge/pytorch-cpu-feedstock/issues/335) - could you make a comment there?

@shermansiu, could you please handle this? I don't understand enough to write anything coherent.

maresb:
The current state seems to have regressed back to `-- Could NOT find CUDA (missing: CUDA_INCLUDE_DIRS) (found version "12.6")`. I will try temporarily switching back to the previous version.

maresb:
Ugh, I accidentally rebased away @shermansiu's suggestions. :man_facepalming:

maresb:
That's looking a bit better now.

By the way, please feel free to open parallel PRs to this one! I don't want to slow anyone down.

maresb:
I seem to be making some progress.

I started compiling for CPU. Despite including `libnuma` I am getting:

```
 │ │       $SRC_DIR/vllm/csrc/cpu/utils.cpp:2:12: fatal error: numa.h: No such file or directory
 │ │           2 |   #include <numa.h>
 │ │             |            ^~~~~~~~
```

Also I realized that they pull in a [heavily-patched](https://github.com/vllm-project/flash-attention/tree/d4e09037abf588af1ec47d0e966b237ee376876c) version of [flash-attn](https://github.com/conda-forge/flash-attn-feedstock).

Note that flash-attn seems really [difficult to compile](https://github.com/conda-forge/flash-attn-feedstock/blob/main/recipe/conda_build_config.yaml#L3C1-L4):

```yaml
github_actions_labels:
- cirun-openstack-cpu-xlarge
```

I'm not sure what we can do about that. :disappointed: 

On the CUDA side I managed to get further by patching all the CMake stuff to use the `CUDAToolkit` package instead of `CUDA`. (I saw that this was done in the conda-forge pytorch package, no idea if doing so is correct or not.) I'm now getting the following error, perhaps it's https://github.com/conda-forge/pytorch-cpu-feedstock/issues/357?

```
 │ │       CMake Error at cmake/utils.cmake:440 (target_link_libraries):
 │ │         Target "_C" links to:
 │ │       
 │ │           CUDAToolkit::cudart
 │ │       
 │ │         but the target was not found.  Possible reasons include:
 │ │       
 │ │           * There is a typo in the target name.
 │ │           * A find_package call is missing for an IMPORTED target.
 │ │           * An ALIAS target is missing.
 │ │       
 │ │       Call Stack (most recent call first):
 │ │         CMakeLists.txt:454 (define_gpu_extension_target)
 │ │       
 │ │       
 │ │       CMake Error at cmake/utils.cmake:440 (target_link_libraries):
 │ │         Target "_moe_C" links to:
 │ │       
 │ │           CUDAToolkit::cudart
 │ │       
 │ │         but the target was not found.  Possible reasons include:
 │ │       
 │ │           * There is a typo in the target name.
 │ │           * A find_package call is missing for an IMPORTED target.
 │ │           * An ALIAS target is missing.
 │ │       
 │ │       Call Stack (most recent call first):
 │ │         CMakeLists.txt:510 (define_gpu_extension_target)
 │ │       
 │ │       
 │ │       CMake Error at $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/public/cuda.cmake:162 (set_property):
 │ │         The link interface of target "torch::nvtoolsext" contains:
 │ │       
 │ │           CUDA::nvToolsExt
 │ │       
 │ │         but the target was not found.  Possible reasons include:
 │ │       
 │ │           * There is a typo in the target name.
 │ │           * A find_package call is missing for an IMPORTED target.
 │ │           * An ALIAS target is missing.
 │ │       
 │ │       Call Stack (most recent call first):
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Caffe2/Caffe2Config.cmake:86 (include)
 │ │         $PREFIX/lib/python3.9/site-packages/torch/share/cmake/Torch/TorchConfig.cmake:68 (find_package)
 │ │         CMakeLists.txt:81 (find_package)
 │ │       
 │ │       
 │ │       CMake Error at $SRC_DIR/flash-attention/cmake/utils.cmake:399 (target_link_libraries):
 │ │         Target "_vllm_fa2_C" links to:
 │ │       
 │ │           CUDAToolkit::cudart
 │ │       
 │ │         but the target was not found.  Possible reasons include:
 │ │       
 │ │           * There is a typo in the target name.
 │ │           * A find_package call is missing for an IMPORTED target.
 │ │           * An ALIAS target is missing.
 │ │       
 │ │       Call Stack (most recent call first):
 │ │         $SRC_DIR/flash-attention/CMakeLists.txt:205 (define_gpu_extension_target)
```

shermansiu:
> > but then it complains about not being able to find `kineto`.
> 
> This is very likely related to [conda-forge/pytorch-cpu-feedstock#335](https://github.com/conda-forge/pytorch-cpu-feedstock/issues/335) - could you make a comment there?

I left a comment!

h-vetinari:
> I left a comment!

Thanks! What I meant though was describing the situation here with vllm and why it wants to find kineto. That would be an argument to ship it in pytorch. I don't think we'll disable kineto support.

shermansiu:
Update: I think I've fixed most of the issues now but now that most of the package resolution has been solved, the build takes really long to build. I found a way to get it to work with CUDA instead of CUDAToolkit.

Could you please grant me access to modify https://github.com/maresb/staged-recipes/tree/vllm?

maresb:
> Could you please grant me access 

I just sent you an invite as a collaborator. Let me know if this doesn't work.

> Update: I think I've fixed most of the issues now

Incredible!!!

> the build takes really long to build

Do you think most of the build time is due to flash-attn? If so, I wonder if it would make sense to make a fork of [conda-forge/flash-attn](https://github.com/conda-forge/flash-attn-feedstock) to build the vllm fork of flash-attn?

> I found a way to get it to work with CUDA instead of CUDAToolkit.

Is there somewhere I can learn about the difference? I'm so confused with all this.

shermansiu:
> I just sent you an invite as a collaborator. Let me know if this doesn't work.

Invite accepted!

> Do you think most of the build time is due to flash-attn? If so, I wonder if it would make sense to make a fork of [conda-forge/flash-attn](https://github.com/conda-forge/flash-attn-feedstock) to build the vllm fork of flash-attn?

I'm not too sure (I don't want to kill the current run to find out), but it's certainly possible! 

I found this comment in the CMakeLists.txt file:
```
# IMPORTANT: This has to be the last thing we do, because vllm-flash-attn uses the same macros/functions as vLLM.
# Because functions all belong to the global scope, vllm-flash-attn's functions overwrite vLLMs.
# They should be identical but if they aren't, this is a massive footgun.
#
# The vllm-flash-attn install rules are nested under vllm to make sure the library gets installed in the correct place.
# To only install vllm-flash-attn, use --component _vllm_fa2_C (for FA2) or --component _vllm_fa3_C (for FA3).
# If no component is specified, vllm-flash-attn is still installed.
```

As a result, I don't think it's a good idea to split it off into a separate feedstock, at least based on the currently available information.

> Is there somewhere I can learn about the difference? I'm so confused with all this.

I primarily did a bunch of tests in the docker image. I found the following:

```
-- CUDA found: TRUE
-- CUDA version: 12.6
-- CUDA include dirs: /home/conda/staged-recipes/build_artifacts/bld/rattler-build_vllm_1739473579/host_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/targets/x86_64-linux/include
-- CUDA libraries: /home/conda/staged-recipes/build_artifacts/bld/rattler-build_vllm_1739473579/host_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/targets/x86_64-linux/lib/libcudart_static.a;Threads::Threads;dl;/home/conda/staged-recipes/build_artifacts/bld/rattler-build_vllm_1739473579/build_env/x86_64-conda-linux-gnu/sysroot/usr/lib/librt.so
-- CUDA root: 
nvcc warning : incompatible redefinition for option 'compiler-bindir', the last value of this option was used
-- CUDAToolkit found: TRUE
-- CUDAToolkit version: 12.6.85
-- CUDAToolkit include dirs: /home/conda/staged-recipes/build_artifacts/bld/rattler-build_vllm_1739473579/host_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold/targets/x86_64-linux/include
-- CUDAToolkit libraries: 
-- CUDAToolkit root: 
```

I suspect the issue was that PyTorch loaded CUDAToolkit but not CUDA, so when I loaded CUDA, it worked.

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,153 @@
...
+    export VERBOSE=1
+    export VLLM_TARGET_DEVICE=${{ vllm_target_device }}
+    # export VLLM_CMAKE_ARGS="-DVLLM_TARGET_DEVICE=${{ vllm_target_device }} -DCUDA_TOOLKIT_ROOT_DIR=$PREFIX"
+    ${{ PYTHON }} -m pip install . --no-build-isolation
+
+  python:
+    entry_points:
+    - vllm  =  vllm.scripts:main
+
+  skip: true  # [win]
```
maresb, line 38:
```suggestion
  skip: win
```

shermansiu: Thanks so much! I guess the README needs to be updated, haha

h-vetinari:
> Do you think most of the build time is due to flash-attn? If so, I wonder if it would make sense to make a fork of [conda-forge/flash-attn](https://github.com/conda-forge/flash-attn-feedstock) to build the vllm fork of flash-attn?

It would be much better to patch vllm to use the existing flash-attn. It's conceivable to build a special patched variant just for vllm, but the first approach should be to get away from the vendored stuff. I haven't looked at how intrusive their patches are, if they're benign, we might be able to carry them on the flash-attn feedstock directly. Best of course would be to work with upstream flash-attn to get these patches merged so the problem solves itself.

maresb:
Ya, there are some very large patches. I don't understand what they are for, but it's hard to imagine they are for nothing.

shermansiu:
Sounds good! I'm not sure what the existing practices for Conda-forge wheels are, but locally, I was able to finish building the wheel for Linux (CPU) and get tests to pass for it. For Linux (Cuda 12.6), I got the wheel to build and a few tests to pass, but I didn't go through everything yet.

I have a really busy schedule so I don't think I'll have the time over the rest of this weekend to work on Cuda 11.8. (e.g. Replacing cuda-cudart with cudatoolkit or something). I also couldn't get either of the two osx Docker images to run locally because micromamba was missing.

Looking forward to seeing this being added to Conda-forge!

shermansiu:
I'm also linking this issue because it's relevant for building Linux (CPU): https://github.com/vllm-project/vllm/issues/9683

shermansiu:
Is there any way to increase the amount of memory when building it out?

maresb:
Thanks @shermansiu for your amazing work on this! It's great to see a successful Linux CPU build.

> Is there any way to increase the amount of memory when building it out? The config should say that the instance is self-hosted, right? Through GitHub Actions?

Comparing with flash-attn https://github.com/conda-forge/staged-recipes/pull/26239, it looks like we'd need to open an admin request like [this](https://github.com/conda-forge/admin-requests/pull/1040/files).

h-vetinari:
That only works on the feedstock, not in staged recipe, but it should be possible if necessary.

maresb:
According to the docs, vLLM is [Linux-only](https://docs.vllm.ai/en/latest/getting_started/quickstart.html#prerequisites), so unless you have info to the contrary we should probably `skip: not linux`

shermansiu:
Generally, some documents recommend that we cross-compile from regular OS-X onto OS-X (Arm64).

maresb:
Here's my investigation to try and understand what's going on with the fork of flash-attention.

Here's the [current main](https://github.com/vllm-project/flash-attention/commits/95898bad1d6b2c1668e39bcaa7ce70c38270e194/) branch with my annotations.

vllm fork of flash-attention

The nominal patch seems to be [90eacc1](https://github.com/vllm-project/flash-attention/commit/90eacc1af2a7c3de62ea249e929ed5faccf38954). It also adds a CMake config, a `vllm_flash_attn` Python module, modifies the API (e.g. removing arguments from `mha_fwd` in `csrc/flash_attn/flash_api.cpp`), and adds a bunch of new code (most of which seems to relate to sparse tensors).

Between that and the currently-vendored commit [d4e0903](d4e09037abf588af1ec47d0e966b237ee376876c), there seem to be [bugfixes and error-reporting improvements](https://github.com/vllm-project/flash-attention/compare/90eacc1af2a7c3de62ea249e929ed5faccf38954..d4e09037abf588af1ec47d0e966b237ee376876c). The total patch is [here](https://github.com/vllm-project/flash-attention/compare/bc482cbf918c8cecd368ee54c306684645548e23..d4e09037abf588af1ec47d0e966b237ee376876c).

At least for the moment it seems to me like we won't be able to make use of the existing conda-forge flash-attention.

shermansiu:
Hmm, it seems like we might need to make a separate feedstock after all. It's definitely not ideal though.

The best situation would be for all of the changes to be merged into the upstream branch, so I guess I'll leave some feedback to the authors.

maresb:
I tried to run the CUDA build on a cloud instance (GCP g2-standard-4, 4xCPU, 16GB RAM, L4), and it crashed with the instance becoming unresponsive. I'm going to try and allocate a larger instance to see if I can get a successful private build.

maresb:
> Hmm, it seems like we might need to make a separate feedstock after all. It's definitely not ideal though.

I was wondering if this would even work?

From [here](https://github.com/vllm-project/vllm/blob/da833b0aeee979bd754e92f7ada3a6709cf51954/CMakeLists.txt#L574-L582):

> #### Build vLLM flash attention from source
>
> *IMPORTANT:* This has to be the last thing we do, because vllm-flash-attn uses the same macros/functions as vLLM.
> Because functions all belong to the global scope, vllm-flash-attn's functions overwrite vLLMs.
> They should be identical but if they aren't, this is a massive footgun.
>
> The vllm-flash-attn install rules are nested under vllm to make sure the library gets installed in the correct place.
> To only install vllm-flash-attn, use --component _vllm_fa2_C (for FA2) or --component _vllm_fa3_C (for FA3).
> If no component is specified, vllm-flash-attn is still installed.

So maybe flash-attention needs to be build afterwards? Like would we need to first build vLLM without flash-attention, and then make another build pass where we add it in? (Sorry for my likely nonsense, I'm inexperienced with compilation.)

maresb:
Any ideas for how to make things slightly more verbose? It's quite a drag to only see `Building wheel for vllm (pyproject.toml): still running...`.

shermansiu:
> I was wondering if this would even work?

> > Build vLLM flash attention from source
IMPORTANT: This has to be the last thing we do...

Yeah, I quoted this part yesterday too. I think it might be possible, but then the configuration would be really hacky and not ideal. https://github.com/conda-forge/staged-recipes/pull/28931#issuecomment-2660965154

Personally, I would much rather have several outputs in `recipe.yaml`, like we already do in both `flash-attn-feedstock` and `pytorch-cpu-feedstock`.

shermansiu:
> Any ideas for how to make things slightly more verbose? It's quite a drag to only see `Building wheel for vllm (pyproject.toml): still running...`.

No, unfortunately that's what I also see when I build it locally with a higher verbosity. The CUDA build _should_ take only ~50 minutes when there are adequate resources though.

maresb:
I'm building on a GCP instance with 16xCPU and 64GB. I did not record the first 30 minutes. I'm seeing RAM spikes up to 52GB. (Disk usage is along the right axis and is fluctuating between 65 and 70 GB (units should be kb, not %).

image

maresb:
@shermansiu, have you already managed to produce a CUDA build locally with rattler-build? Is that where you get the 50 minutes estimate, or was your estimate based on builds outside of the conda-forge infrastructure?

maresb:
Success!!!

Comparing the PyPI build (left) with my conda-forge build (right), my FA3 library is much bigger. I wonder why.

image

I will try to upload...

maresb:
https://anaconda.org/maresb/vllm

shermansiu:
> ave you already managed to produce a CUDA build locally with rattler-build? Is that where you get the 50 minutes estimate, or was your estimate based on builds outside o

That's from my local build!

shermansiu:
For OSX, it's able to check that the `BUILD_PLATFORM=osx-64` and `HOST_PLATFORM=osx-arm64`, but once it builds the Conda env, it keeps setting `target_platform` to `osx-64`, not `osx-arm64`. However, the `target_platform` field in `.ci_support/osx_arm64.yaml` is set correctly. This seems like a bug.

shermansiu:
Adding `vllm` here comes after the feedstock is created: https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/migrations/osx_arm64.txt

h-vetinari:
> Hmm, it seems like we might need to make a separate feedstock after all.

It shouldn't be a separate feedstock; we could build this on a dedicated branch on the flash-attn feedstock.

shermansiu:
> It shouldn't be a separate feedstock; we could build this on a dedicated branch on the flash-attn feedstock.

Sounds good! Not sure how that would work though, especially since it's supposed to overwrite things in vLLM. When we're not building vllm, I'm not even sure we need vllm-flash-attn: IIRC, I usually use regular flash-attn when installing through pip.

h-vetinari:
> When we're not building vllm, I'm not even sure we need vllm-flash-attn: IIRC, I usually use regular flash-attn when installing through pip.

That's actually good news, because obviously the patched `flash-attn` would not be possible to install side-by-side with the original one. But if it's only necessary at build time for `vllm`, then we can put the patched flash-attn into a label, and then _only_ pull that in at build time of vllm using
```yaml
channel_sources:
  - conda-forge/label/<whatever>,conda-forge
```
Note that non-default labels don't work correctly with rattler-build yet, see https://github.com/conda-forge/triton-feedstock/pull/42

shermansiu:
`vllm` needs to build `flash-attn` from source at the very end, then use those compiled files to overwrite some of its functions before the final wheel is ready.

I think that would make sense if vllm depended on a custom pre-built `flash-attn` binary, but because it needs to build it from scratch at the very end of compilation, I think it doesn't make sense to separate it out in this _particular_ scenario. (But in most other cases, I would agree that it should be part of the feedstock for the original package.)

h-vetinari:
Are you saying that the patched flash-attn _cannot_ be built in advance, and _must_ be built after vllm? If so, that's a pretty major obstacle, to the point that I'd say that upstream vllm has to improve this procedure (at least to allow prebuilding flash-attn, or better, not using a patched version in the first place) before we can package it.

shermansiu:
> Are you saying that the patched flash-attn _cannot_ be built in advance, and _must_ be built after vllm?

Yeah, that seems like what's going on right now.

shermansiu:
Also, even if we did merge all of the differences from vllm-flash-attn into flash-attn, I'm not sure it would still be easier to keep them in separate feedstocks: we don't actually run two separate commands to compile both wheels.

The single command to install vllm with pip itself triggers the build for vllm-flash-attn through CMakeFiles.txt, for CUDA builds.
https://github.com/vllm-project/vllm/blob/69e1d23e1eae4d33ac44fdab2257e29a542cd9a9/CMakeLists.txt#L592-L607

To be more specific, vllm does not build vllm-flash-attn after it has been fully built. Building vllm-flash-attn is the penultimate step before it copies over the compiled files and links everything together. It's important for the global scope during compilation to be shared so that the common functions defined in vllm-flash-attn can overwrite the ones defined in vllm.

h-vetinari:
> To be more specific, vllm does not build vllm-flash-attn after it has been fully built. Building vllm-flash-attn is the penultimate step before it copies over the compiled files and links everything together. It's important for the global scope during compilation to be shared so that the common functions defined in vllm-flash-attn can overwrite the ones defined in vllm.

This sounds horrible and needs to be engineered better, e.g. compiler vllm-flash-attn first and use the right functions in vllm right away, rather than relying on some replacement process.

shermansiu:
Sounds good!

shermansiu:
I was able to cross-compile the wheel for macOS (arm64). All that's left is to ensure that `opencv-python-headless` ships with an arm64 build.

shermansiu:
I don't think the current scripts in staged-recipes were tested to support cross-compilation.

The aarch64 one fails because it seems to depend on the setup from the newer Conda-smithy script (i.e. defining RECIPE_ROOT) and the Mac Arm64 one fails because cmake doesn't support Python 3.13? Something doesn't feel right...

shermansiu:
I was able to fix the MacOS compilation script, but the recipe_root definition I added is hacky because most other parts of the conda-forge ecosystem expect a single `recipe` directory for the feedstock.

https://github.com/conda-forge/conda-forge-ci-setup-feedstock/blob/d8b11fd85622f4d332cd099d91e8fa12562c9be6/recipe/cross_compile_support.sh#L36C20-L36C31

shermansiu:
## Summary:
- **MacOS (Arm64):** Awaiting `opencv-python-headless`. In the CI, it is able to build the wheel and gets stuck at this step
- **Linux 64 (Cuda 11.8):** It seems to work, but it runs out of disk space and crashes the image. When running it locally, it also keeps crashing my computer. I'll need to get access to better compute to ensure that this runs.

## TODO:
- Fix cross-compilation in `staged-recipes`. I'll need to add the fixes I made to the macOS run script and find a more permanent fix for the `RECIPE_ROOT` environment variable during cross-compilation (Linux AArch64). I'll also need to include instructions to set the `TARGET_ARCH` environment variable when cross-compiling from `osx-64` to `osx-arm64`.
- Get `opencv-python-headless` to work for Mac (Arm64).

maresb:
> I'll need to get access to better compute to ensure that this runs.

I can run this on a GCP instance. (Unfortunately I can't give you direct access.) But just let me know which builds you want. Just linux-64 cuda 11.8 for now?

shermansiu:
Yep! That's the only one left!

maresb:
Sorry, didn't manage it today. Will try tomorrow.

shermansiu:
Hmm, interesting! I've always been using `build-locally.py`.

maresb:
The pip check failed due to https://github.com/conda-forge/xgrammar-feedstock/pull/5. Adding it in by hand and rerunning...

maresb:
I finally got around to posting on prefix.dev's discord about the conditional evaluation issue I mentioned above that's blocking me from producing those builds:

image

maresb:
You can also see in these build runs that the CUDA versions have `pytorch-cpu` selected. :disappointed:

shermansiu:
My guess is that something changed in rattler-build [v0.36.0](https://github.com/prefix-dev/rattler-build/releases/tag/v0.36.0) that wasn't present in 0.35.9. It's also broken for Cuda 12 now.

shermansiu:
Maybe this is related? https://github.com/prefix-dev/rattler-build/pull/1385

Jinja variables are no longer coerced to be strings.

shermansiu:
Yep, I confirmed it. It has to do with how Jinja variables are no longer coerced to be strings, particularly Boolean values. https://github.com/prefix-dev/rattler-build/pull/1385

shermansiu:
Also, I guess `pybind11` will be removed as a dependency for `xgrammar` (as seen in your discussion in https://github.com/mlc-ai/xgrammar/pull/190)? As it was able to be built successfully and pass the pip checks somehow in the CI, I think we can remove it. Wdyt?

maresb:
> Also, I guess `pybind11` will be removed as a dependency for `xgrammar` (as seen in your discussion in [mlc-ai/xgrammar#190](https://github.com/mlc-ai/xgrammar/pull/190))? As it was able to be built successfully and pass the pip checks somehow in the CI, I think we can remove it. Wdyt?

Sorry, I missed your comment. I added it because my pip check was failing. (The real mystery is why any of the pip checks passed!)

There has been a v0.1.14 release on PyPI 4 hours ago, fixing the dependency issue. There still needs to be a conda-forge build, but I think that will be ready in as soon as a few hours.

shermansiu:
FYI: After I added more RAM, I was able to build CUDA 11.8 locally without `pybind11`. At its peak, my machine used 80-100 GB of RAM...

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,192 @@
...
+    - pybind11
+  - typing_extensions >=4.10
+  - filelock >=3.16.1
+  - partial-json-parser
+  - pyzmq
+  - msgspec
+  - gguf ==0.10.0
+  - importlib-metadata
+  - mistral-common >=1.5.0
+  - opencv-python-headless >=4.0.0  # mistral-common[opencv]
```
hmaarrfk, line 133:
```suggestion
  - opencv >=4.0.0  # mistral-common[opencv]
```

if you don't want qt to installed at build time you can:

1. Depend on `noqt` https://github.com/conda-forge/noqt-feedstock
2. Depend on `opencv =*=headless*`

hmaarrfk: as a runtime dependency, just depend on opencv, 

people have historically come to opencv from conda-forge as a "batteries included" package.

We've provided options for system integrators to install headless ,but by default it should still install the "qt version" for linux + windows users.

shermansiu: Sounds good! Thanks for your feedback!

shermansiu:
You're welcome! 😄

shermansiu:
I'm waiting for #29771 to get merged so that we can get `llguidance` on `conda-forge`. In the meantime, `xgrammar`'s 0.1.17 feedstock has also been pushed.

shermansiu:
#29817 will also provide `hf-xet` on Conda Forge, which is a new dependency of `vllm`.

shermansiu:
Now that both `llguidance` and `hf-xet` are up on Conda-forge, we can finally resume work on this!

shermansiu:
`hf-xet` doesn't support AArch64 or macOS arm64 builds because the vendored openssl doesn't support cross-compilation. Since it's one of the dependencies now, I'll probably drop support for those two here as well.

i.e. We'll probably just support CPU-only Linux and CUDA Linux (x86_64).

h-vetinari:
> Now that both `llguidance` and `hf-xet` are up on Conda-forge, we can finally resume work on this!

Great news, thanks for following up on this!

> because the vendored openssl doesn't support cross-compilation

Not directed at you personally, but we should 100% unvendor openssl in any package in conda-forge. OpenSSL has a very stable API/ABI, so vendoring such a critical library (which thus gets outdated) is IMO a verging-on-unacceptable security issue.

CC @conda-forge/core

shermansiu:
I guess we can work on all of the other platforms _after_ we get the feedstock up.

shermansiu:
At this point, the CUDA 12.6 build works locally, but there seems to be a problem with the `opentelemetry-exporter-prometheus` feedstock, which is unrelated (as it's not a direct dependency)

```
 │ │ Installing test environment
 │ │ ✔ Successfully updated the test environment
 │ │ INFO 07-09 05:54:10 [__init__.py:243] No platform detected, vLLM is running
 │ │  on UnspecifiedPlatform
 │ │ ✔ python imports test passed!
 │ │ opentelemetry-exporter-prometheus 0.48b0 has requirement opentelemetry-sdk~
 │ │ =1.27.0, but you have opentelemetry-sdk 1.34.1.
```

shermansiu:
It just doesn't work here because it runs out of RAM: a megabuild is needed for CUDA 12.6.

shermansiu:
@conda-forge/help-python-c This recipe is ready for review!

shermansiu:
@conda-forge/help-python-c Please review this when you have the time!

recipes/vllm/conda_build_config.yaml
```diff
@@ -0,0 +1,11 @@
+# https://github.com/conda-forge/.cirun
+github_actions_labels:  # [cuda_compiler_version != "None" and linux]
+- cirun-openstack-gpu-2xlarge  # [(cuda_compiler_version or '')[:2] == '12' and linux]
+- cirun-openstack-gpu-4xlarge  # [(cuda_compiler_version or '')[:2] == '11' and linux]
```
h-vetinari: we've dropped cuda 11, so this shouldn't be necessary.

recipes/vllm/conda_build_config.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+c_stdlib_version:          # [linux]
+  - "2.30"                 # [linux]
```
h-vetinari: Where does this requirement come from? We only provide 2.28 & 2.34

recipes/vllm/patches/remove_ninja_pip_requirement.txt
```diff
@@ -0,0 +1,9 @@
+diff --git a/requirements/common.txt b/requirements/common.txt
```
h-vetinari: For all the patches, it would be nice to have authorship info and a commit message. The best way to do that is to have a local git checkout of the upstream sources, create a branch (I usually call it `conda`), hard-reset it to the version tag you want to build on, apply the patches as regular git commits, and then do
```
git format-patch <tag> --no-signature
```

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+  - if: linux
+    then:
+    - patches/vllm-cpu-utils.patch
+  - if: is_cross_compiling
+    then:
+    - patches/vllm-cmake-args.patch
+  target_directory: vllm
+- url: https://github.com/vllm-project/flash-attention/archive/d637d8927a35922ce6f6c0dff6dd3f765ed71f3c.tar.gz
+  sha256: 3099add00c9938735b84319d176c5b239c0165e3f9be6540a7a3505cd897c7cd
+  target_directory: flash-attention
```
h-vetinari, line 28: Can you explain (in a comment at least) why we need to vendor flash-attention, rather than relying on the one from conda-forge. Presumably vllm needs a specific commit, then you should note where that commit comes from (e.g. a submodule upstream, or some file that contains the git rev).

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+
+build:
+  number: 0
+  script: |
+    sed -i.bak 's/set(TORCH_SUPPORTED_VERSION_CUDA "2.4.0")/set(TORCH_SUPPORTED_VERSION_CUDA "2.5.1")/g' flash-attention/CMakeLists.txt
```
h-vetinari: How does this match with the `pytorch_version: 2.6.0` further up?

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
+context:
+  version: 0.8.3
+  pytorch_version: 2.6.0
+  use_cuda: ${{ cuda_compiler_version != "None" }}
+  is_cuda_12: ${{ (cuda_compiler_version or '')[:2] == '12' }}
```
h-vetinari: cuda 11 is no more, so this can be simplified

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+    - setuptools >=74.1.1
+  - einops
+  - compressed-tensors ==0.9.2
+  - depyf ==0.18.0
+  - cloudpickle
+  - watchfiles
+  - python-json-logger
+  - scipy
+  - pytorch ==${{ pytorch_version }}
+  - if: use_cuda or ppc64le
```
h-vetinari: not sure what role `or ppc64le` is supposed to do here; we don't have pytorch builds for that.

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+    else:
+    - torchaudio
+    - torchvision
+  - if: use_cuda
+    then:
+    - if: match(python, ">3.9")
+      then:
+      - numba ==0.61
+      else:
+      - numba ==0.60.0
```
h-vetinari: stuff like this needs a comment IMO

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+  - huggingface_hub >=0.30.0
+  - hf-xet >=0.1.4
+  - tokenizers >=0.19.1
+  - protobuf
+  - fastapi >=0.115.0
+  - aiohttp
+  - openai >=1.52.0
+  - uvicorn-standard
+  - pydantic >=2.9
+  - prometheus_client >=0.18.0
```
h-vetinari: This number of dependencies becomes completely unmanageable if not alphabetized, e.g. when you're looking for the right line to update some version bounds for a new release.

(I didn't select all relevant lines; this is meant for the entire `run:` section)

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,206 @@
...
+  - pillow
+  - prometheus-fastapi-instrumentator >=7.0.0
+  - tiktoken >=0.6.0
+  - lm-format-enforcer >=0.10.9,<0.11
```
maresb:
```suggestion
  - lm-format-enforcer >=0.10.11,<0.11
```

recipes/vllm/recipe.yaml
maresb: Thanks @h-vetinari for the very thorough review!

Dependencies here are currently in Grayskull order:

```bash
grayskull pypi vllm==0.8.3
```

which preserves the upstream order. The upstream situation is basically these two requirements files: [common.txt](https://github.com/vllm-project/vllm/blob/v0.8.3/requirements/common.txt) and [cuda.txt](https://github.com/vllm-project/vllm/blob/v0.8.3/requirements/cuda.txt). Alternatively, you can look under `info.requires_dist` in [this PyPI endpoint](https://pypi.org/pypi/vllm/0.8.3/json).

I personally find upstream/grayskull ordering much easier to check than alphabetical. But if it's not alphabetical then we should at least explain all the above about where the ordering comes from in comments.

h-vetinari: Recipes are primarily for human consumption. It's fine to use tools like grayskull for support, but the end result needs to be readable, which this is not. I think grayskull at the very least needs an option to sort dependencies (which ideally should be the default, and I'd go as far as considering it a bug to not do it).

You can do
```python
foo = """<copy content of run:>"""
print("\n".join(sorted(foo.splitlines())))
# copy result back to recipe
```
by hand too of course. This is a job to be done once when updating the recipe, as opposed to 100s of time in the heads of contributors when reading the recipe.

recipes/vllm/conda_build_config.yaml
shermansiu: Done!

recipes/vllm/conda_build_config.yaml
shermansiu: My memory's a bit fuzzy, but I think it had to be higher than 2.28: I'll try bumping it to 2.34 instead then!

recipes/vllm/recipe.yaml
shermansiu: Done!

recipes/vllm/recipe.yaml
shermansiu: Removed!

recipes/vllm/recipe.yaml
shermansiu: Done!

recipes/vllm/recipe.yaml
shermansiu: Fixed!

recipes/vllm/recipe.yaml
shermansiu: Done!

recipes/vllm/recipe.yaml
shermansiu: Fixed! It was overlooked when I bumped the version for vLLM - sorry about that.

recipes/vllm/patches/remove_ninja_pip_requirement.txt
shermansiu: I'll get to this a bit later

shermansiu:
Thanks so much, h-vetenari, for your incredibly detailed review!

recipes/vllm/recipe.yaml
shermansiu: I'm not particularly opinionated on this either way, but I decided to sort them anyways. For the `run` section, I kept the `python` requirement at the top, though.

recipes/vllm/conda_build_config.yaml
h-vetinari: Can you try 2.28 first?

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,202 @@
...
+
+  python:
+    entry_points:
+    - vllm  =  vllm.scripts:main
+
+  skip:
+    - win
+    - osx and x86_64
+    # conda-forge torchaudio dropped support for CUDA 11.8
+    - cuda_compiler_version != "None" and match(cuda_compiler_version, "<12.0")
```
h-vetinari: obsolete skips

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,202 @@
...
+    - ray-cgraph >=2.43.0,!=2.44
+    - torchaudio ==${{ pytorch_version }}
+    - torchvision ==0.21.0
+    # v0.61 doesn't support Python 3.9. Required for N-gram speculative decoding
+    # Requirements taken from https://github.com/vllm-project/vllm/blob/296c6572dd1f76b31b93be19e550790afcfb8843/requirements/cuda.txt#L4-L5
+    - if: match(python, ">3.9")
+      then:
+      - numba ==0.61
+      else:
+      - numba ==0.60.0
```
h-vetinari: Isn't this whole distinction obsolete since we have a global `match(python, "<3.10")` skip? 🤔

recipes/vllm/patches/remove_ninja_pip_requirement.txt
shermansiu: Done!

recipes/vllm/recipe.yaml
shermansiu: Removed!

recipes/vllm/recipe.yaml
shermansiu: Yeah, it is! I updated the comment to explain the discrepancy between the upstream and Conda-Forge requirements and removed the conditional Numba 0.60 dependency.

recipes/vllm/conda_build_config.yaml
h-vetinari: I'd like to see if 2.28 works (or how it fails).

shermansiu: I think it was for `gettid`, but since we add the patch defining it manually anyways, we can change it back to 2.28

https://stackoverflow.com/questions/30680550/c-gettid-was-not-declared-in-this-scope

shermansiu: Kicked off a CUDA build: we'll see how that goes

shermansiu: I removed the manual `c_stdlib_version` specification and the CUDA build still worked!

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,197 @@
...
+  script: |
+    sed -i.bak 's/set(TORCH_SUPPORTED_VERSION_CUDA "2.4.0")/set(TORCH_SUPPORTED_VERSION_CUDA "${{ pytorch_version }}")/g' flash-attention/CMakeLists.txt
+    export VLLM_FLASH_ATTN_SRC_DIR=$SRC_DIR/flash-attention
+    cd vllm
+    python use_existing_torch.py
+    mkdir -p $SRC_DIR/vllm/third_party/NVTX/c
+    ln -s $PREFIX/include $SRC_DIR/vllm/third_party/NVTX/c/include
+    # export VERBOSE=1
+    export VLLM_TARGET_DEVICE=${{ vllm_target_device }}
+    ${{ PYTHON }} -m pip install . --no-build-isolation
```
h-vetinari: I would add some verbosity here; currently the logs only show an endless stream of
```
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
 │ │   Building wheel for vllm (pyproject.toml): still running...
 ```

File: recipes/vllm/conda_build_config.yaml [hunk 19]
```diff
@@ -0,0 +1,7 @@
+# https://github.com/conda-forge/.cirun
+github_actions_labels:  # [cuda_compiler_version != "None" and linux]
+- cirun-openstack-gpu-2xlarge  # [cuda_compiler_version != "None" and linux]
```
h-vetinari, line 3: Note that you'll have to open a PR to admin-requests following this [example](https://github.com/conda-forge/admin-requests/blob/main/examples/example-open-gpu-server.yml), and you'll need to follow the [instructions](https://github.com/Quansight/open-gpu-server#access) (more details in PR template) for getting the rights to interact with the server.

recipes/vllm/conda_build_config.yaml
```diff
@@ -0,0 +1,7 @@
...
+
+megabuild:
+- "true"  # [cuda_compiler_version != "None" and linux]
+- "false"  # [cuda_compiler_version == "None" or not linux]
```
h-vetinari: As it stands, this will likely not take effect yet (upon rendering the feedstock) as intended. But I don't see why this feedstock should do megabuilds - those are useful primarily if there's a common shared artefact (e.g. a C++ library) that's the same for all python versions, and where it would be wasteful to build it each time in different python jobs. This is the situation for `libtorch` vs. `pytorch`, but I can't see a comparable situation here.

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,197 @@
...
+
+  python:
+    entry_points:
+    - vllm  =  vllm.scripts:main
```
h-vetinari: If you add an entry point, you should add a test for it.

recipes/vllm/recipe.yaml
```diff
@@ -0,0 +1,197 @@
...
+    - cuda-nvrtc-dev
+    - libcublas-dev
+tests:
+- python:
+    imports:
+    - vllm
+    - if: linux and use_cuda
+      then:
+      - vllm.vllm_flash_attn
+    pip_check: true
```
h-vetinari, line 181: generally, I'd be happier if there was any testing beyond the bare import & pip check. Like maybe you can find a toy example (perhaps even directly in the upstream repo) that would be small enough to run as a test of basic operations.

recipes/vllm/patches/0003-Manually-define-gettid.patch
```diff
@@ -0,0 +1,23 @@
+From 2c319f3bad078d60759cc17e704fc3a9524a6091 Mon Sep 17 00:00:00 2001
+From: Sherman Siu <s8siu@uwaterloo.ca>
+Date: Mon, 21 Jul 2025 05:05:36 -0400
+Subject: [PATCH 3/4] Manually define gettid
```
h-vetinari, line 4: Now that we have a proper commit message, you can use it to add additional information, like the stackoverflow link you posted. Something like "gettid not defined [by default?] until glibc 2.30, see [link]".

You can add this to the actual commit you have in your local checkout by doing an interactive rebase, choose `edit` (instead of `pick`) for that commit, then run `git commit --amend` and add a body of the commit message (separated from the title by an empty line). Then regenerate the patches as before with `git format-patch ...`

recipes/vllm/conda_build_config.yaml
shermansiu: The main purpose is to get more RAM for the CUDA build - I'm not sure what the best approach is for this!

shermansiu:
Thanks for looking at this, @h-vetinari! I should be able to get to the rest of the things later in the week, hopefully!

recipes/vllm/conda_build_config.yaml
h-vetinari: That's not what megabuild means, though I can see why you might think so 😁 

Please remove this.

The [size](https://github.com/Quansight/open-gpu-server?tab=readme-ov-file#available-runners) of the machine (incl. the amount of RAM) is determined by `github_actions_labels`

shermansiu: Sounds good, thanks! :)

recipes/vllm/conda_build_config.yaml
shermansiu: Thanks for letting me know! :grin:

recipes/vllm/conda_build_config.yaml
shermansiu: So megabuild just means that the same server is used for each Python version build? I couldn't find any online documentation about it.

h-vetinari: > So megabuild just means that the same ~server~ CI job is used for each Python version build?

Fixed (the linux jobs are all run on the same server, but that's unrelated).

Building everything in one job is the only way to share libtorch across different pytorch builds (per python version).

> I couldn't find any online documentation about it.

That's not surprising, because it's nothing more than a local variable on the pytorch feedstock. 😉

recipes/vllm/recipe.yaml
shermansiu: Updated! Though there's still plenty of `Building wheel for vllm (pyproject.toml): still running...` messages

recipes/vllm/conda_build_config.yaml
shermansiu: Ahhh, thanks! :sweat_smile: I was wondering about that!

recipes/vllm/recipe.yaml
shermansiu: Done!

recipes/vllm/patches/0003-Manually-define-gettid.patch
shermansiu: Thanks for the useful tips about the interactive rebase, amending the commit, and using `git format-patch`! :grinning:

Done!

recipes/vllm/recipe.yaml
shermansiu: Done! I picked one of the tests that doesn't depend on external packages.

recipes/vllm/conda_build_config.yaml
shermansiu: Anyways, `megabuild` has been removed from the config!

shermansiu:
If there are no other requested changes, I'm good to have this merged! The pull request is no longer just a draft, but I don't have the permissions to change it.

shermansiu:
@maresb Please add "Closes #24710" and "fixes #29105" to the initial PR description so that we can close the other vllm PR and the package request issue automatically once this gets merged

shermansiu:
Anyways, the CUDA 12.6 build works locally and the tests pass:
```bash
 │ Installing test environment
 │ ✔ Successfully updated the test environment
 │ Testing commands:
 │ ============================= test session starts ==============================
 │ platform linux -- Python 3.10.18, pytest-8.4.1, pluggy-1.6.0
 │ rootdir: $PREFIX/etc/conda/test-files/vllm/2
 │ plugins: anyio-4.9.0
 │ collected 22 items
 │ vllm/tests/core/test_scheduler.py ......................                 [100%]
 │ ============================== 22 passed in 6.97s ==============================
 │
 ╰─────────────────── (took 72 seconds)
 ✔ all tests passed!
```

recipes/vllm/conda_build_config.yaml
h-vetinari: Have you done this already?

shermansiu: No, I thought it needed to be done after the feedstock was created. I'll do it now!

h-vetinari (approved):
This is still marked as a draft (intentional?), and you'll have to do the procedures to get the rights to the opengpu server, but the PR itself LGTM!

recipes/vllm/conda_build_config.yaml
h-vetinari: Open the relevant PRs please. :)

The one behind the second link should be done in advance, and we can merge the admin-requests one shortly after merging this one

shermansiu:
"Only those with [write access](https://docs.github.com/get-started/learning-about-github/access-permissions-on-github) to this repository can mark a draft pull request as ready for review." - I'm unable to change this!

recipes/vllm/conda_build_config.yaml
h-vetinari: Note that each maintainer needs to do this (if you want to be able to run CI jobs), so @maresb, please do [that one](https://github.com/Quansight/open-gpu-server#access) as well.

shermansiu: https://github.com/Quansight/open-gpu-server/pull/63
Once that's merged, I'll send in a pull request to `conda-forge/admin-requests`!

shermansiu:
> Please add "Closes https://github.com/conda-forge/staged-recipes/pull/24710" and "fixes https://github.com/conda-forge/staged-recipes/issues/29105" to the initial PR description so that we can close the other vllm PR and the package request issue automatically once this gets merged

Thanks @h-vetinari for adding that in! :smile:

recipes/vllm/conda_build_config.yaml
shermansiu: @maresb Here's the link that displays your GitHub ID, for the sake of convenience: https://api.github.com/users/maresb

shermansiu: Also, here's the admin-requests PR, even though it shouldn't be merged yet: https://github.com/conda-forge/admin-requests/pull/1596

maresb: Ok, I opened https://github.com/Quansight/open-gpu-server/pull/64, thanks so much for the clear instructions.

And sorry about the draft thing, I'm glad you were able to mark it as ready-for-review without me.

shermansiu: Both of the `open-gpu-server` PRs have been closed!

PR #28944, opened 2025-01-26
Add renode-cli

MementoRC:
Checklist
- [x] Build number is 0.


recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,79 @@
...
+          - patches/update-for-cmake-build.sh.patch
+          - patches/resolve-posix-unix.patch
+  - git: https://github.com/renode/renode-infrastructure.git
+    # rev: 9736cf57e9e9518fe0e53a58cecdb09aee404a4f  # no CMakeLists.txt
+    rev: 24c2b35b1b7fb46e47961c08c486fd57c2640b8b
+    target_directory: renode-infrastructure
+    patches:
+      - if: unix
+        then:
+          - patches/update-IronPython-Unittest_NET.csproj.patch
```
danielnachun, line 26: Please use https://github.com/renode/renode/releases/download/v1.15.3/renode_1.15.3_source.tar.xz as the source instead of the regular git tag.  That archive contains the full source code with all git modules checked out already, so you don't need to check out `renode-infrastructure` or any of the AntMicro submodules.  Your patches may need path updates.

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,121 @@
...
+  rm -rf ELFSharp && mv ELFsharp ELFSharp
+  update_antmicro_submodule e379e8ae696676afffed2f33dd8083855af00f2f FdtSharp
+  update_antmicro_submodule 33e5ab24eaaab488e9a94f1a40d5d6ae2f7f02f1 InpliTftpServer
+  update_antmicro_submodule c6514e99f2c35afec083b9f4a7eec3408c0081d1 Migrant
+  update_antmicro_submodule 2fe74fd257f6d6f86076c100952178f347098b3d Packet.Net
+  update_antmicro_submodule 33d24f1307d267c34b7f1439bc159a8825c5aa3d bc-csharp
+  update_antmicro_submodule bde21d04fbfc540989b7a0ac13a54eae8b756994 cctask
+  update_antmicro_submodule e8c2051dec56c5ccd3a4927b07a5740f34eed8c8 options-parser
+  update_antmicro_submodule 71af57ef4fec29e416f48160b8918057a58548a9 termsharp
+popd
```
danielnachun, line 47: I believe you'll be able to remove all of these when you've switch to the new source archive I suggested.

recipes/renode-cli/recipe.yaml
MementoRC: Great. Thank you for the pointer. Though, I end-up needing a different version of the `renode-infrastructure` as the one in the tarball does not support CMake and I could not figure how to get CCTask to compile the Cores
Can this be a `noarch` even though there is a gcc compilation stage?
I kept checking the other recipe in case you had gotten started, but as I eventually won that Csharp fight, I went ahead with the submission

recipes/renode-cli/build.sh
MementoRC: Would you accept to be a co-maintainer. I can tell that there are a lot of updates in `master`, the next release might be a tough one as well

danielnachun:
Since `build.sh` here is mostly a wrapper `dotnet publish`, I'm wondering if rather than trying to patch that script, if we can just figure out the right `dotnet publish` commands and call those directly.  This usually leads to a much simpler, easier to maintain recipe.

> @danielnachun Please take a look at: https://github.com/conda-forge/staged-recipes/pull/28944 - The recipe has some 'gcc' compilations that seem to integrate into DLLs, does that mean that it cannot be a noarch? Is dotnet that good of a magician?

I believe the native code originates from https://github.com/antmicro/tlib/tree/4a28cc4048ebcf4196a110218adff13253f1ba69.  Normally I would say that we should build this as a separate package and make it a dependency.  However, in this case that repository is untagged and is itself a maintenance fork of another untagged repository.  Ideally we would built this a split package with `tlib` as a native package and `recode-cli` as a `noarch` package that depends on it.  I don't have much experience making split packages but this one seems fairly simple.

Let me know if you have questions so far, and thanks for being flexible about how to do this.  We're still figuring out how to handle .NET best in conda-forge so there's a lot of best practices that still need to be worked out, especially for more complex packages like this one.

MementoRC:
Sounds good, but then I need to go back to the drawing board. I made a few split packages and will give it a try. I will ping the upstream maintainer to try to get a tag, they seem very thorough and conda-friendly - It's curfew over here, so I'll follow-up tomorrow

danielnachun:
Great!  I'll keep an eye on this issue in case you run into more challenges but this is great progress so far!

recipes/renode-cli/build.sh
danielnachun: You can add me as co-maintainer, this will be a useful recipe for figuring mixed C#/C packages in the future.

recipes/renode-cli/recipe.yaml
danielnachun: I see now there is additional C code in `renode-infrastructure` as well.  I think for now we might just be able to copy into the recipe directory and then place in the right folder so we can build the cores with CMake.  We should do our CMake invocation for building this code as well.  If there is a `.csproj` file that tries to call CCTask or otherwise recompile C code we've already built, we can just patch the `.csproj` file.

danielnachun: Also, to your question about `noarch` - we might be able make another split package for something like `renode-infrastructure-cores` that only ships the compiled C code for the cores.  However, we need to understand better how that cores and `tlib` are loaded by `renode` - if the integration between the C# and C code is too tight, then separating them into difference packages won't make sense.  

I've only dealt with a handful of .NET + native packages, but so far they just copied a `.so`/`.dylib` into a a folder and the tool just finds the library.  We'll see if such a strategy makes sense here, or if it just has to be native.

One difference I've gathered so far is that it seem that both .NET and JDK languages make far less use of their C APIs to interface with native code.  This is in contrast to Python, Perl, Ruby and Lua, which more tightly integrate native extensions in such a way that makes separating them not worth the effort.

MementoRC:
@danielnachun A little bit of butchering, but what are your thoughts so far?

MementoRC:
@danielnachun Now would be a good time to peak-a-boo! I tried to add a dynamic library loader to the code to load the `*.so` from the various `renode-cores` in the `renode-cli`. I don't know if it works.

danielnachun:
This is looking better but I am still kind of concerned at the amount of patching and modifications we're making here.  Some of this may be unavoidable, but right now this still seems difficult to maintain without a really deep understanding of this tool.

A couple thoughts so far:
- If we have to write a custom dynamic loader for the native core plugins just to get a `noarch` package, we should probably just accept this being a native package now.  I wasn't aware of those plugins when I initially suggested trying `noarch`, but I would rather we just try to modify the code and build system as little as possible.
- Have you tried the `dotnet publish` command yet?  I haven't had a chance to try it yet myself, but I'm thinking we'd need a lot less modification if we can just patch enough for that to work.

As I've mentioned before, this is a work in progress, but from what I've dealt with so far, things that are easy and reasonable to patch are:
- .NET version requirements
- Dependency versions if a version is outdated and causing a build to fail
- Forced usage of Mono on Linux and macOS

Making changes beyond those basic things runs a much higher risk of breaking automated updates.  This weekend I will take what you have and try to simplify as much as possible.  Thanks for your patience so far - I want to figure this out in a way that really makes it easy to maintain in the future!

MementoRC:
@conda-forge/staged-recipes, Hello ready for review

ocefpaf:
@danielnachun was the first reviewer here. I confess I don't really understand the recipes here. Is this OK to go @danielnachun ?

MementoRC:
@ocefpaf Thanks for the review. Sure, I did not realize @danielnachun was a conda reviewer, I had contacted him because he seemed to be a dotnet expert. Great thx

ocefpaf:
> @ocefpaf Thanks for the review. Sure, I did not realize @danielnachun was a conda reviewer, I had contacted him because he seemed to be a dotnet expert. Great thx

He is not an "official" conda-forge/staged-recipes reviewer, yet... However, he is doing an amazing job helping us out in situations where we lack expertise. I'm only waiting for his OK to merge this one.

danielnachun:
I've submitted a request staged-recipes so I can be an official reviewer, but that is still in progress!  In the meantime I'm trying to help where I can.  I'm working on a new review now.

MementoRC:
I had to drop windows, it builds but there seems to be something that prevents the build_env to be cleaned up, maybe a file gets erroneously locked

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
```diff
@@ -0,0 +1,95 @@
...
+OUTPUT_DIRECTORY="${SRC_DIR}/output"
+
+CONFIGURATION="Release"
+BUILD_PLATFORM="Any CPU"
+HEADLESS=true
+NET=true
+TFM="net${framework_version}"
+HOST_ARCH="i386"
+# Common cmake flags
+CMAKE_COMMON=""
```
danielnachun, line 17: Please reduce the usage of variables in these helper scripts only to those values which need to differ between platforms.  It makes the code logic very difficult to follow.  Use line breaks with `\ ` if the line is too long.  If you have entire arguments that are only needed on one platform, then just use a simple string rather than an array.  If we can simplify the logic here enough, we can just move this code into the main `build.sh` script.

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
```diff
@@ -0,0 +1,95 @@
...
+)
+
+if $HEADLESS
+then
+    BUILD_TARGET=Headless
+    PARAMS+=(p:GUI_DISABLED=true)
+else
+    BUILD_TARGET=Mono
+fi
+
```
danielnachun, line 35: Let's **not**  attempt to support alternate build options that are not needed for us to build this in conda-forge.  We should only support building a headless build that does not use Mono (which is now effectively dead anyway), and only support native packages here and not `noarch`.  The simpler we can make this recipe, the easier it will be for others to understand and help to maintain.

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
```diff
@@ -0,0 +1,95 @@
...
+</Project>
+EOF
+
+export DOTNET_CLI_TELEMETRY_OPTOUT=1
+CS_COMPILER="dotnet build"
+TARGET="${SRC_DIR}/Renode_NET.sln"
+BUILD_TYPE="dotnet"
+
+OUT_BIN_DIR="${SRC_DIR}/output/bin/${CONFIGURATION}"
+BUILD_TYPE_FILE="${OUT_BIN_DIR}/build_type"
```
danielnachun, line 51: More things here which probably don't need to be variables.

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
```diff
@@ -0,0 +1,95 @@
...
+  LLVM_LIB="libllvm-disas-aarch64"
+fi
+if [[ "${target_platform}" == "linux-"* ]] || [[ "${target_platform}" == "noarch" ]]; then
+  LIB_EXT="so"
+elif [[ "${target_platform}" == "osx-"* ]]; then
+  LIB_EXT="dylib"
+else
+  LIB_EXT="dll"
+fi
+cp lib/resources/llvm/$LLVM_LIB.$LIB_EXT $OUT_BIN_DIR/libllvm-disas.$LIB_EXT
```
danielnachun, line 95: We should only support Linux and macOS in this script, so we can just use the conda-defined `$SHLIB_EXT` here instead of all this more complex logic.

recipes/renode-cli/helpers/renode_build_with_dotnet.ps1
```diff
@@ -0,0 +1,71 @@
...
+$SRC_DIR = Resolve-Path "$Env:SRC_DIR"
+
+$OUTPUT_DIRECTORY = "$SRC_DIR\output"
+$CONFIGURATION = "Release"
+$BUILD_PLATFORM = "Any CPU"
+$HEADLESS = $true
+$NET = $true
+$TFM = "net$framework_version-windows"
+$HOST_ARCH = "i386"
+$CMAKE_COMMON = ""
```
danielnachun, line 14: See my previous comments about reducing usage of variables - if we simplify this logic enough, we won't even need a wrapper anymore.

recipes/renode-cli/helpers/renode_build_with_dotnet.ps1
```diff
@@ -0,0 +1,71 @@
...
+
+$PARAMS = @()
+if ($Env:CC) {
+    $PARAMS += "/p:CompilerPath=$Env:CC"
+    $PARAMS += "/p:LinkerPath=$Env:CC"
+}
+if ($Env:AR) {
+    $PARAMS += "/p:ArPath=$Env:AR"
+}
+$PARAMS += $args
```
danielnachun, line 24: See my previous comments about only supporting what is needed to build in conda-forge.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,83 @@
+context:
+  name: renode-cli
+  version: "1.15.3"
+  python_min: "3.6"
```
danielnachun, line 4: This variable is not used anywhere, and furthermore the oldest support version of Python is 3.9 - 3.6 has been EOL for quite a while now

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,83 @@
...
+          - patches/resolve-posix-unix.patch
+          - patches/update-unix-informational-version.patch
+        else:
+          - patches/fix-missing-env-common_copy_licenses.sh.patch
+          - patches/update-win-informational-version.patch
+
+build:
+  number: 0
+  skip:
+    - win
```
danielnachun, line 28: I assume this is disabled for testing?

recipes/renode-cores/helpers/renode_build_with_cmake.sh
```diff
@@ -0,0 +1,68 @@
...
+      cmake --build .
+
+      CORE_BIN_DIR=$CORES_BIN_PATH/lib
+      mkdir -p $CORE_BIN_DIR
+      if [[ "${target_platform}" == "osx-"* ]]; then
+          # macos `cp` does not have the -u flag
+          cp -v tlib/*.so $CORE_BIN_DIR/
+      else
+          cp -u -v tlib/*.so $CORE_BIN_DIR/
+      fi
```
danielnachun, line 64: Instead of using different logic for macOS and Linux, just add `coreutils` as build dependency.

recipes/renode-cores/helpers/renode_build_with_cmake.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
+CONFIGURATION="Release"
+HOST_ARCH="i386"
+CMAKE_COMMON=""
```
danielnachun, line 7: See my previous comments about not relying so much on variables and simplifying logic - I suspect as with the other helper scripts we probably don't actually need separate scripts if we simplify this enough.

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
```diff
@@ -0,0 +1,68 @@
...
+$cpuCount = (Get-CimInstance Win32_Processor).NumberOfLogicalProcessors
+
+$env:PATH = "${env:BUILD_PREFIX}/Library/mingw-w64/bin;${env:BUILD_PREFIX}/Library/bin;${env:PREFIX}/Library/bin;${env:PREFIX}/bin;${env:PATH}"
+$env:SRC_DIR = $env:SRC_DIR
+
+$CMAKE = (Get-Command cmake).Source
+
+$CONFIGURATION = "Release"
+$HOST_ARCH = "i386"
+$CMAKE_COMMON = ""
```
danielnachun, line 12: See my previous comments about not relying so much on variables and simplifying logic - I suspect as with the other helper scripts we probably don't actually need separate scripts if we simplify this enough.

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
```diff
@@ -0,0 +1,68 @@
...
+    param (
+        [string]$filePath,
+        [string]$flagToRemove,
+        [string]$flagToAdd
+    )
+    (gc $filePath) -replace $flagToRemove,$flagToAdd | sc $filePath
+}
+
+Update-CMakeLists "$CORES_PATH\tlib\CMakeLists.txt" "-fPIC" "-Wno-unused-function -Wno-maybe-uninitialized"
+Update-CMakeLists "$CORES_PATH\tlib\tcg\CMakeLists.txt" "-fPIC" "-Wno-unused-function -Wno-maybe-uninitialized"
```
danielnachun, line 37: It would be much simpler to just add `-Wno-unused-function -Wno-maybe-uninitialized` to the `CXXFLAGS` environment variable and then just use the replace function directly to remove `-fPIC` instead of writing a custom function like this.

Also please provide a comment explaining why `-fPIC` should be removed - there are very few cases where we wouldn't position independent code.

recipes/renode-cores/variants.yaml
```diff
@@ -0,0 +1,12 @@
...
+      then: sysroot
+    - if: osx
+      then: macosx_deployment_target
+
+m2w64_c_compiler:
+  - m2w64-toolchain
+m2w64_cxx_compiler:
+  - m2w64-toolchain
+m2w64_c_stdlib:
+  - m2w64-toolchain
```
danielnachun, line 12:
```suggestion
```
You shouldn't need this file in `conda-forge` - these defaults are already supposed to set in both staged-recipes and the feedstock.  You might need it to build locally, but it shouldn't be included in the recipe submission

recipes/renode-cli/variants.yaml
```diff
@@ -0,0 +1,7 @@
+c_stdlib:
+    - if: linux
+      then: sysroot
+    - if: osx
+      then: macosx_deployment_target
+    - if: not unix
+      then: vs
```
danielnachun, line 7:
```suggestion
```
You shouldn't need this file in `conda-forge` - these defaults are already supposed to set in both staged-recipes and the feedstock.  You might need it to build locally, but it shouldn't be included in the recipe submission.

danielnachun (changes requested):
This is going in the right direction - the patches now are much simpler and should be easier to maintain, though we should see if any can be upstreamed.  In particular we should suggest to the upstream that they at least make it easier to build without Mono, as that is now an unmaintained run time.

My main overall thought now is that there is still quite a bit of overengineering in some places.  We shouldn't be trying to build a piece of software for our recipe, just the bare minimum needed to for things to work in conda-forge.  If users want to customize the build differently, they should fork the feedstock rather than us trying to accommodate more complex logic.  Conda is not a strictly declarative package manager like Nix, but the closer we stick to declarative-style recipes, the easier it is to maintain in the long term.  Accessibility and readability matter a lot more here than the internal core of a complex piece of software.

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,94 @@
...
+mkdir -p ${PREFIX}/bin
+mkdir -p ${PREFIX}/libexec/${PKG_NAME}
+export PATH="${DOTNET_ROOT}/dotnet:${PATH}"
+
+install_prefix="${PREFIX}/opt/${PKG_NAME}"
+
+dotnet_version=$(dotnet --version)
+framework_version=${dotnet_version%.*}
+
+# Patch the project files to use the correct .NET version
```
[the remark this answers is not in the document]
MementoRC: No, I did not. Yes, the build was getting flooded with warnings and I could not catch the errors in the log. Let me remove both and remember why I used them

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,94 @@
...
+chmod +x tools/{building,packaging}/*.sh
+${RECIPE_DIR}/helpers/renode_build_with_dotnet.sh ${framework_version}
+
+# Install procedure
+mkdir -p $PREFIX/libexec/${PKG_NAME}
+cp -r output/bin/Release/net${framework_version}/* $PREFIX/libexec/${PKG_NAME}/
+
+mkdir -p $PREFIX/opt/${PKG_NAME}/scripts
+mkdir -p $PREFIX/opt/${PKG_NAME}/platforms
+mkdir -p $PREFIX/opt/${PKG_NAME}/tests
```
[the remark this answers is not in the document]
MementoRC: The original conda recipe from upstream placed them in `opt` and I shied away from changing it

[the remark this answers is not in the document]
danielnachun: I see, the structure of conda packages is such that probably shouldn't follow their directory structure.  Let's see how that change works.

recipes/renode-cli/build.sh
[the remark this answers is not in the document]
danielnachun: Alternatively if that `--property WarningLevel=0` isn't able to suppress the warnings, we can also redirect the warning output perhaps with something like `2>/dev/null`.  But let's see if my first suggestion works before trying alternatives.

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
MementoRC: I seem to remember that this -fPIC is redundant and creates an extra notification. Also, these CMakeLists.txt may not use the CXXFLAGS, but I should double-check

recipes/renode-cli/build.sh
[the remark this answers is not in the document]
MementoRC: Right, I preferred at the time not changed their design. I will not commit this particular one, but do it with a replace all on my fork

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,94 @@
...
+mkdir -p $PREFIX/opt/${PKG_NAME}/licenses
+
+cp .renode-root $PREFIX/opt/${PKG_NAME}/
+cp -r scripts/* $PREFIX/opt/${PKG_NAME}/scripts/
+cp -r platforms/* $PREFIX/opt/${PKG_NAME}/platforms/
+cp -r tests/* $PREFIX/opt/${PKG_NAME}/tests/
+cp -r tools/metrics_analyzer $PREFIX/opt/${PKG_NAME}/tools
+cp -r tools/execution_tracer $PREFIX/opt/${PKG_NAME}/tools
+cp -r tools/gdb_compare $PREFIX/opt/${PKG_NAME}/tools
+cp -r tools/sel4_extensions $PREFIX/opt/${PKG_NAME}/tools
```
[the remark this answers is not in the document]
MementoRC: Would a multi-output recipe be acceptable in this case?

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
MementoRC: Right, a lot of it is again from their original design, I carved out their build script (if I remember correctly). now that you semi-approve the overall structure, I probably can simplify a bunch as you suggested

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
MementoRC: Agreed

recipes/renode-cli/recipe.yaml
MementoRC: What do you mean? The win build fails as something from the build_env cannot be deleted, so I skip it for the staged-recipe, I was planning to solve and add it to the feedstock along with the other arch

danielnachun: I didn't see your other comment about disabling Windows build earlier.  For now, maybe we can leave it enabled so we can make sure things are still working after the cleanup.  And maybe someone will be able to figure why the build_env is clean up is failing as well.

recipes/renode-cli/build.sh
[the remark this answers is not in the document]
danielnachun: Yes, if that isn't too much trouble.  For now I think just `renode-python-tools` and `renode-models-analyzer` is sufficient - it's not worth making separate packages for the individual Python tools.

recipes/renode-cli/helpers/renode_build_with_dotnet.sh
danielnachun: Indeed, now that we know what it's doing we can simplify a lot more.  I'll check in once you've had a chance to do more cleanup - while this recipe will never be as simple as most other .NET recipes, I'm fairly sure we can get it to the point that it's pretty straight forward to follow even if it's a bit longer.

MementoRC:
Wow, I solved it, `dotnet` was apparently still running or locking up some DLLs ... almost there

MementoRC:
@danielnachun This may be a good point to give it another review. I tried to move to a single-file build, but for windows, I feared having to battle the .bat, so i redirect to a .ps1 - All arch are building and passing the tests. I made provision to separate the huge tests from the install files, it could easily be switched back if needed

@wolfv The tests framework is quite voluminous. In the original recipe, they are installed in the package folder (they used /opt), but I wonder if it is necessary to include them in the conda package since we are running the testsuite whenever we build

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+  version: ${{ version }}
+
+source:
+  - url: https://github.com/renode/renode/releases/download/v${{ version }}/renode_${{ version }}_source.tar.xz
+    sha256: 40c50afe8db86a9c63f4a86b93116a6babd43bf241cfaac47b2c0d2cbc2b15a8
+    patches:
+      - patches/update-IronPython-Renode_NET.csproj.patch
+      - patches/update-cores.template_NET.csproj.patch
+      - patches/update-IronPython-Unittest_NET.csproj.patch
+      # - patches/fix-from-master-Renode_NET.sln.patch # rattler seems to choke on leading tabs
```
danielnachun, line 16:
```suggestion
```
I think we can drop this patch as I don't believe the `.sln` file is used now in the build process.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - patches/fix-from-master-TermSharp_NET.csproj.patch
+      - patches/fix-from-master-Xwt.Gtk3_NET.csproj.patch
```
danielnachun, line 18: Is this patch still needed since we are building without the GUI?

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+source:
+  - url: https://github.com/renode/renode/releases/download/v${{ version }}/renode_${{ version }}_source.tar.xz
+    sha256: 40c50afe8db86a9c63f4a86b93116a6babd43bf241cfaac47b2c0d2cbc2b15a8
+    patches:
+      - patches/update-IronPython-Renode_NET.csproj.patch
+      - patches/update-cores.template_NET.csproj.patch
+      - patches/update-IronPython-Unittest_NET.csproj.patch
+      # - patches/fix-from-master-Renode_NET.sln.patch # rattler seems to choke on leading tabs
+      - patches/fix-from-master-TermSharp_NET.csproj.patch
```
danielnachun, line 17: I think it would be better to convert this patch to a simple `sed` replacement since it's a temporary fix for a dependency version that will probably be updated in a later release.  We can add a comment above the `sed` statement explaining that it can be removed once there is a new release that fixes the version.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - patches/fix-from-master-Xwt.Gtk3_NET.csproj.patch
+      - if: unix
+        then:
+          - patches/resolve-posix-unix.patch
+          - patches/update-unix-informational-version.patch
```
danielnachun, line 22: Is the script this is patching still used in the build process?  Usually `.csproj` files don't call shell scripts.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+        else:
+          - patches/fix-missing-env-common_copy_licenses.sh.patch
+          - patches/update-win-informational-version.patch
```
danielnachun, line 25: See my previous commend about the other `information-version` patch - if this Powershell script isn't being called by the build process, we don't need to patch it.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  - url: https://github.com/renode/renode/releases/download/v${{ version }}/renode_${{ version }}_source.tar.xz
+    sha256: 40c50afe8db86a9c63f4a86b93116a6babd43bf241cfaac47b2c0d2cbc2b15a8
+    patches:
+      - patches/update-IronPython-Renode_NET.csproj.patch
```
danielnachun, line 13: I think it would be better to convert this patch to a simple `sed` replacement since it's a temporary fix for a dependency version that will probably be updated in a later release. We can add a comment above the `sed` statement explaining that it can be removed once there is a new release that fixes the version.

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,86 @@
...
+      - patches/update-cores.template_NET.csproj.patch
+      - patches/update-IronPython-Unittest_NET.csproj.patch
```
danielnachun, line 15: Is this fix taken from an upstream change?  It looks like it's fixing a typing issue that isn't specific to building in conda-forge so we should report it if it's a generic problem with their code.

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env bash
+
+set -o xtrace -o nounset -o pipefail -o errexit
+
+# Consolidated script logic
+main() {
```
danielnachun, line 6: I'm not sure this script needs separate functions, as we're not iterating over these functions or passing special parameters.  We can just run everything sequentially.

recipes/renode-cli/helpers/renode_build_with_dotnet.ps1
```diff
@@ -0,0 +1,114 @@
...
+# Check for empty environment variables
+if ([string]::IsNullOrEmpty($SRC_DIR)) { throw "SRC_DIR is empty" }
+if ([string]::IsNullOrEmpty($PREFIX)) { throw "PREFIX is empty" }
+if ([string]::IsNullOrEmpty($PKG_NAME)) { throw "PKG_NAME is empty" }
+
+# Set environment variables
+$env:PATH = "${env:BUILD_PREFIX}/Library/mingw-w64/bin;${env:BUILD_PREFIX}/Library/bin;${env:PREFIX}/Library/bin;${env:PREFIX}/bin;${env:PATH}"
+
+# Function to install tests (defined before use in PowerShell)
+function install_tests {
```
danielnachun, line 20: Similar to my previous comment, I don't see the need to define functions in this script given that everything is just evaluated imperatively here anyway.

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,95 @@
...
+        tools/packaging/common_copy_licenses.sh ${SRC_DIR}/license-files linux
+    fi
+
+    # Create renode script (using heredoc)
+    cat > "${PREFIX}/bin/renode" <<EOF
+#!/bin/sh
+exec "\${DOTNET_ROOT}"/dotnet exec "\${CONDA_PREFIX}"/libexec/renode-cli/Renode.dll "\$@"
+EOF
+    chmod +x "${PREFIX}/bin/renode"
+
```
danielnachun, line 64: We still need to call `dotnet-licenses` here to get all the third party license used in the actual .NET DLLs.  The licenses copied by that helper script don't actually cover all of those.   I think you had this in a previous version of the script?

danielnachun (changes requested):
Things are looking really good now!  I made a few more suggestions to consolidate/simplify patching.  The only long term patching we'll have to carry for now is the changes needed to use `dotnet` instead of `mono` on Linux/macOS and the changes to disable building `renode-cores` inside of `renode`.  

I'll leave it up to you if you want to open an upstream issue about those changes - they should really consider dropping Mono altogether, and maybe they would be open to adding a flag for users to skip building `renode-cores` automatically so that users can build it separately.  But I don't think these changes need to be upstreamed for us to add the package to conda-forge.

danielnachun:
> @danielnachun This may be a good point to give it another review. I tried to move to a single-file build, but for windows, I feared having to battle the .bat, so i redirect to a .ps1 - All arch are building and passing the tests. I made provision to separate the huge tests from the install files, it could easily be switched back if needed
> 
> @wolfv The tests framework is quite voluminous. In the original recipe, they are installed in the package folder (they used /opt), but I wonder if it is necessary to include them in the conda package since we are running the testsuite whenever we build

I think deleting large test files is definitely the right call here, especially since as you said the test suite is always run while building anyway.  If a user really needs things from the test suite for developing some other piece of software, they can always download that from the repo directly.

recipes/renode-cli/recipe.yaml
MementoRC: I moved them into a patch so that when a new version is released, it will make the patching fail and I won't forget to remove it, whereas, I would certainly forget the `sed`.

danielnachun: That's fair, let's leave them as patches then!

MementoRC: Well, that was my motivation, but then the same can be done inside the script with a test on the PKG_VERSION, so I implemented that. I am working on your suggestions, I broke linux build somewhere and working on that currently, I should have an update soon for unix to transfer to my windows VM

recipes/renode-cli/build.sh
MementoRC: this is sufficient:
`    dotnet-project-licenses --input "$SRC_DIR/src/Renode/Renode_NET.csproj" -d "$SRC_DIR/license-files" -f "txt"
`?

recipes/renode-cli/helpers/renode_build_with_dotnet.ps1
MementoRC: Done. I don't quite understand, though, I usually use functions to clearly define a scope, like a list of parameters that are needed and also an easy way to add/remove part of the script. In the editor I can also easily collapse them and not be bothered by the boundaries when they are in one big blob

recipes/renode-cli/build.sh
MementoRC: Done

recipes/renode-cli/recipe.yaml
MementoRC: The type issue causes a fail at my end, I incorporated t into the script

recipes/renode-cli/recipe.yaml
MementoRC: Done

recipes/renode-cli/recipe.yaml
MementoRC: We do, there is an <Exec ...> into one of the .csproj that calls this, it assumes that it is a git repo and uses `git ...`

recipes/renode-cli/recipe.yaml
MementoRC: Done

recipes/renode-cli/recipe.yaml
MementoRC: It is needed, it caused a fail at my end, I moved it to the script

recipes/renode-cli/recipe.yaml
danielnachun: In that case as this needs to be a long-term fix, I think `sed` replacement would be better here (we can just use a dummy revision) so that we don't need to update so many patches.

MementoRC: Oh, got it now!

recipes/renode-cores/recipe.yaml
```diff
@@ -0,0 +1,89 @@
+context:
+  name: renode-cores
+  version: "1.15.3"
+  python_min: "3.6"
```
danielnachun, line 4:
```suggestion
```
This should removed as we don't use the variable anywhere in the recipe and the oldest supported Python version is 3.9

recipes/renode-cores/build.sh
```diff
@@ -0,0 +1,63 @@
...
+cp cmake-tlib/tcg/CMakeLists.txt "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/tcg"
+cp cmake-tlib/LICENSE "${RECIPE_DIR}/tlib-LICENSE"
+cp "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/softfloat-3/COPYING.txt" "${RECIPE_DIR}/softfloat-3-COPYING.txt"
+
+if [[ "${target_platform}" == "osx-arm64" ]]; then
+  # Apply OSX-specific patches
+  sed -i -E 's/add_definitions\(-Wno-error=clobbered\)/add_compile_options(-Wno-unknown-warning-option -Wno-clobbered)/' \
+    "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/CMakeLists.txt" \
+    "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/CMakeLists.txt" \
+    "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/tcg/CMakeLists.txt"
```
danielnachun, line 17: Can we just use:
```
export CXXFLAGS="${CXXFLAGS} -Wno-unknown-warning-option"`
```
Also setting these CXXFLAG does not need to be OS specific.

recipes/renode-cores/recipe.yaml
```diff
@@ -0,0 +1,89 @@
...
+  - url: https://github.com/renode/renode/releases/download/v${{ version }}/renode_${{ version }}_source.tar.xz
+    sha256: 40c50afe8db86a9c63f4a86b93116a6babd43bf241cfaac47b2c0d2cbc2b15a8
+  - if: unix
+    then:
+      - git: https://github.com/renode/renode-infrastructure.git
+        rev: "3fc2d5fe643068e595e875d9408cb4329522b229"
+        target_directory: cmake-renode-infrastructure
+      - git: https://github.com/antmicro/tlib.git
+        rev: "69fff75a0eba7471283b0b8db2e55e8388e284f6"
+        target_directory: cmake-tlib
```
danielnachun, line 20: Why is this only cloned on Unix?  Also, instead of using `git`, let's try to downloading the archive for this revision so we can have a check sum:

```
url: https://github.com/renode/renode-infrastructure/archive/3fc2d5fe643068e595e875d9408cb4329522b229.tar.gz
url: https://github.com/antmicro/tlib/archive/69fff75a0eba7471283b0b8db2e55e8388e284f6.tar.gz
```

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
```diff
@@ -0,0 +1,86 @@
...
+Copy-Item "$SRC_DIR/src/Infrastructure/src/Emulator/Cores/tlib/softfloat-3/COPYING.txt" "$Env:RECIPE_DIR/softfloat-3-COPYING.txt" -Force
+
+# Check weak implementations (using combined path)
+pushd $SRC_DIR/tools/building
+    & bash.exe -c ". './check_weak_implementations.sh'"
+popd
+
+$env:PATH = "${env:BUILD_PREFIX}/Library/mingw-w64/bin;${env:BUILD_PREFIX}/Library/bin;${env:PREFIX}/Library/bin;${env:PREFIX}/bin;${env:PATH}"
+$env:CXXFLAGS = "$env:CXXFLAGS -Wno-unused-function -Wno-maybe-uninitialized -Wno-error"
+$env:CFLAGS = "$env:CFLAGS -Wno-unused-function -Wno-maybe-uninitialized -Wno-error"
```
danielnachun, line 37: It would be better if we could explicitly disable warnings rather than setting `-Wno-error`.  Otherwise this can let problematic issues slip through.

danielnachun (changes requested):
`renode-cli` looks great now.  I have a few questions about `renode-cores` and then I think we'll be ready to go.

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
MementoRC: That is why I was removing the -fPIC, this is the issue, on windows the libraries are shared, so there is a warning saying so, which is automatically turned into a failure. I had the same concern

recipes/renode-cores/recipe.yaml
MementoRC: Possibly a fumble, the cloning is inside the script on windows

recipes/renode-cores/helpers/renode_build_with_cmake.ps1
MementoRC: That is a left-over, I actually still remove the -fPIC manually. Also, I actually ran this recipe on an osx-arm64 runner, but we can revisit when adding the archs after release to feedstock

recipes/renode-cores/recipe.yaml
MementoRC: Done

recipes/renode-cli/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+        - renode
+  - script:
+      - renode --help
+      - LOCAL_TEST_PREFIX=./test-bundle test-bundle/bin/renode-test test-bundle/share/renode-cli/tests/unit-tests
+    requirements:
+      run:
+        - robotframework 6.1
+        - psutil 5.9.3
+        - python
+        - pyyaml 6.0
```
danielnachun, line 49:
```suggestion
        - robotframework 6.1
        - psutil 5.9.3
        - python
        - pyyaml 6.0
```
How strict should these bounds be?  If possible it would be better to set ranges like `psutil >=5.9,<5.10`, and only if the latest version doesn't work.

recipes/renode-cli/build.sh
```diff
@@ -0,0 +1,95 @@
...
+cp -r ${SRC_DIR}/platforms/* "${PREFIX}/share/${PKG_NAME}/platforms/"
+
+# Remove Mono dynamic library
+find ${PREFIX}/libexec/${PKG_NAME} -name "libMonoPosixHelper.so" -exec rm -f {} +
+
+# if [[ "${target_platform}" == "osx-*" ]]; then
+#     tools/packaging/common_copy_licenses.sh ${SRC_DIR}/license-files macos
+# else
+#     tools/packaging/common_copy_licenses.sh ${SRC_DIR}/license-files linux
+# fi
```
danielnachun, line 65:
```suggestion
```
Let's remove commented out code we aren't using any more for the sake of readability.

recipes/renode-cores/build.sh
```diff
@@ -0,0 +1,67 @@
...
+# if [[ "${target_platform}" == "osx-arm64" ]]; then
+#   # Apply OSX-specific patches
+#   sed -i -E 's/add_definitions\(-Wno-error=clobbered\)/add_compile_options(-Wno-unknown-warning-option -Wno-clobbered)/' \
+#     "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/CMakeLists.txt" \
+#     "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/CMakeLists.txt" \
+#     "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/tcg/CMakeLists.txt"
+#   sed -i -E 's|    \$\{CMAKE_SOURCE_DIR\}|    \$\{CMAKE_SOURCE_DIR\} \$\{CMAKE_SOURCE_DIR\}/tlib/tcg|' \
+#     "${SRC_DIR}/src/Infrastructure/src/Emulator/Cores/tlib/tcg/CMakeLists.txt"
+# fi
+
```
danielnachun, line 25:
```suggestion
```
See previous comment about commented out code.

danielnachun (changes requested):
I made some very minor comments for a few things I didn't catch in my last review.  Once those are done I think we'll be all set.

recipes/renode-cli/recipe.yaml
MementoRC: Actually, the original recipe does not have requirements, but for robotframework. I may have had issues and tuned to the minimal revision., we can leave them open ended

recipes/renode-python-tools/build.sh
```diff
@@ -0,0 +1,14 @@
...
+set -o xtrace -o nounset -o pipefail -o errexit
+
+pushd ${SRC_DIR}/tools/dts2repl
+  cp LICENSE ${SRC_DIR}/dts2repl-LICENSE
+  ${PYTHON} -m pip install . --no-deps --no-build-isolation -vv
+popd
+
+# Install non-pip Python tools
+mkdir -p ${PREFIX}/share/renode-python-tools/
+cp -r ${SRC_DIR}/tools/{csv2resd,execution_tracer,gdb_compare,guest_cache,metrics_analyzer} ${PREFIX}/share/renode-python-tools/
```
danielnachun, line 12: There are couple issues with the way `renode-python-tools` is set up right now:
- Executable scripts like these should be installed in `bin`, not `share`.  It becomes difficult for users to directly execute these otherwise.  A symlink back into share can be made if `renode-cli` expects the scripts to be there, but I don't think that's even needed here.
- Several of these scripts have additional python dependencies, one of which need to be added to conda-forge
    - `csv2resd` - construct
    - `execution_tracer` - pyelftools
    - `gdbcompare` - psutil (not in conda-forge, but I'm planning to add this soon), pexpect, telnetlib3
    - `guest_cache` - tqdm
    - `metrics_analyzer` - matplotlib-base, pandas, numpy, pillow
- Only the `dts2repl` script is being tested, and this obscures the missing dependencies for the other scripts.
- `guest_cache` and `metrics_analyzer` have files that need to be installed in `site-packages` and their executable scripts have to be patched so they can find those files.

Given how complex this PR is already and that `renode-python-tools` isn't a dependency of the other two packages, I am going to strongly suggest here that we move `renode-python-tools` to a separate PR.  Furthermore we should either drop `gdbcompare` initially and add it later when `psutil` has been added to conda-forge, or we can wait until `psutil` has been added to submit that PR.

MementoRC: Ok, we have a bit of a disconnect. I understood that only `dts2repl` could be correctly installed/tested. The other 'resources' are provided for an interested user to figure out how to use them, that's why I made the separation into share

psutil is in conda-forge: https://anaconda.org/conda-forge/psutil/ (?)

But, I readily agree with separating it out

danielnachun: Whoops `psutil` is available, I was thinking of `psutils` which is not the same package!  Since it's possible for us to ship a `renode-python-tools` package that "just works", we should put in the effort to do that right in a new PR.

danielnachun (approved):
@ocefpaf we can merge this now!  @MementoRC Thank you for your patience in putting up with my pickiness here.  This is an incredibly complex recipe so it's essential that we follow the strictest standards to have any hope of being able to maintain it long term, and I think it's as simple as we can make it for now.

recipes/renode-python-tools/build.sh
danielnachun: When you resubmit the PR I'll show what to do.  It's pretty easy - the executable `.py` files go into `${PREFIX}/bin` and the supporting files go into `site-packages`.  There's no need for a `setup.py`, just a few `install` or `cp` commands and a little bit of `sed` magic.

MementoRC:
> @ocefpaf we can merge this now! @MementoRC Thank you for your patience in putting up with my pickiness here. This is an incredibly complex recipe so it's essential that we follow the strictest standards to have any hope of being able to maintain it long term, and I think it's as simple as we can make it for now.

Thank you for your time and help, I learned a lot.

PR #28958, opened 2025-01-27
Adding cppad_mixed to conda-forge

bradbell:
Checklist
- [ ] Build number is 0.


bradbell:
@conda-forge/c-cpp, ready for review!

This is the computational core for the python-c package dismod-at which I plan to add to conda-forge once cppad-mixed is available; see
https://dismod-at.readthedocs.io/latest/

bradbell:
@conda-forge/help-c-cpp this is ready for review

bradbell:
I tried converging the recipe from conda buid to rattler build using
```
conda-recipe-manager convert meta.yaml > recipe.yaml
```

The following commit was used to test this
https://github.com/conda-forge/staged-recipes/pull/28958/commits/83528bf5a42625acf8656ed91334cc40d6409a10

The windows test failed and I am having trouble recreating that failure on my windows machine. So I reverted back to the conda build version of this pull request.

recipes/cppad-mixed/build.sh
```diff
@@ -0,0 +1,40 @@
+#! /usr/bin/env bash
+set -e -u
+echo "build.sh: Start"
+#
+# PKG_CONFIG_PATH
+PKG_CONFIG_PATH="$BUILD_PREFIX/lib/pkgconfig"
```
xhochy, line 6: You should not be doing this. All requirements should be in the PREFIX itself.

recipes/cppad-mixed/build.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+# build
+mkdir build && cd build
+#
+# cmake
+cmake -S $SRC_DIR -B . \
+   -G 'Unix Makefiles' \
+   -D CMAKE_BUILD_TYPE=Release \
+   -D cmake_install_prefix="$PREFIX" \
+   -D cmake_search_prefix="$BUILD_PREFIX" \
```
xhochy, line 27: Setting these should not be necessary.

recipes/cppad-mixed/build.sh
```diff
@@ -0,0 +1,40 @@
...
+   -D extra_cxx_flags="$extra_cxx_flags" \
+   -D cmake_libdir=lib \
+   -D ldlt_cholmod=yes \
+   -D optimize_cppad_function=yes \
+   -D for_hes_sparsity=yes 
+#
+# check
+make -j4 check
```
xhochy, line 35:
```suggestion
make -j$CPU_COUNT check
```

xhochy (changes requested):

recipes/cppad-mixed/build.sh
bradbell: > 

I like to use -j4 because it seems that gnu parallel make gets confused when I run it with lots of processes (on the order of 16).

bradbell:
Do you know which version of Visual Studio conda uses ?

I have finally tracked down one source of my troubles testing on my windows machine. 
It seems that I am using the wrong version of visual studio. 
I can tell that the conda install version of cppad thinks that  `unsigned int` and `size_t` are different.

But when I run on my windows machine with Visual Studio 2022, cmake detects that these types are the same. This difference leads to an attempt to define the same function twice.

bradbell:
@xhochy 
I made all the changes to build.sh you suggested above except for removing
```
 -D cmake_install_prefix="$PREFIX" \
 ``
I left this in because the upstream source CMakeLists.txt requires that it be set (and my local tests fails if it do not include it). In the text below which appears in the file
https://github.com/bradbell/cppad_mixed/blob/master/CMakeLists.txt

```
required_definition( cmake_install_prefix PATH
   "prefix used by cmake for installing cppad_mixed"
)
```
i

xhochy:
We should be setting this as part of CMAKE_ARGS but it could be that casing is an issue here. Let's keep it then.

xhochy (approved):

PR #28987, opened 2025-01-30
Add multipers

DavidLapous:
Checklist
- [x] Build number is 0.


DavidLapous:
Dear @conda-forge/help-python-c 

I think this is almost ready, and it seems to build properly on linux + macos. 

I have some questions:
 - [disabling windows release] `multipers` is not yet compatible with windows. I did a
   ```yaml
   build:
     skip: true  # [win]
   ```
   I hope this is the correct way to do this (I didn't find this part in the doc)
 - [python version] for `host`, `build` and `run`, I added a 
   ```
    - python >=3.10, <3.13
   ```
   since `multipers` requires `>=3.10` and a dependency (`gudhi`) requires `<3.13`.
   The linter complains with 
   ```
   Non noarch packages should have python requirement without any version constraints.
   ```
   but I don't really understand this --- what I added are precisely version constraints.
 - [maintainers] it had an error in the first version, and it is fixed now (although the bot didn't update this).
   
Also, I have the hand on the source repo, so if changing something there can help this process, feel free to ask for a change there. I plan to make a new release before the end of this process anyway.

Thanks for your help and review !

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
+  number: {{ build }}
+  skip: true  # [win]
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ stdlib('c') }}
+    - {{ compiler('cxx') }}
+    - python >={{ python_min }},<{{ python_max }}
```
synapticarbors, line 27:
```suggestion
    - python
```

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - numpy >=2
+    - cython >=3  
+  host:
+    - python >={{ python_min }},<{{ python_max }}
```
synapticarbors, line 31:
```suggestion
    - python
```

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - numpy >=2
+    - cython >=3
+    - setuptools
+    - pip
+    - tbb
+    - tbb-devel
+    - libboost-python-devel
+    - libboost-devel
+  run:
+    - python >={{ python_min }},<{{ python_max }}
```
synapticarbors, line 41:
```suggestion
    - python
```

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - gudhi >=3.8
+    - tqdm
+    - scipy
+    - joblib
+    - matplotlib-base
+    - scikit-learn
+    - pot
+
+
+test:
```
synapticarbors, line 53: You should probably just have:

```
test:
   imports:
      - multipers
   requires:
      - pip
   commands:
      - pip check
```

```

File: recipes/multipers/LICENSE [hunk 5]
```diff
@@ -0,0 +1,21 @@
+MIT License
```
synapticarbors, line 1: You can remove this file since it's in the src distribution on `pypi` (assuming you switch to that as the source)

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+source:
+  git_url: https://github.com/DavidLapous/multipers.git
+  git_rev: tags/{{ version }}
+
+build:
+  script: |
+    {{ PYTHON }} setup.py build_ext -j${CPU_COUNT}
+    {{ PYTHON }} -m pip install . --no-deps --no-build-isolation -vv
+  number: {{ build }}
+  skip: true  # [win]
```
synapticarbors, line 20:
```suggestion
  skip: true  # [win or py<310]
```

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://github.com/DavidLapous/multipers.git
+  git_rev: tags/{{ version }}
+
+build:
+  script: |
+    {{ PYTHON }} setup.py build_ext -j${CPU_COUNT}
```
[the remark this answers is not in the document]
DavidLapous: The only advantage of this line is multithreaded build. `pip install` only does single-core compilations

recipes/multipers/meta.yaml
DavidLapous: For reasons that I don't understand yet, the `import multipers` doesn't work in the root directory (since python goes first to local files for the import), so I have to `cd` first.

recipes/multipers/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+{% set version = "v2.2.1" %}
+{% set build = "0" %}
+{% set python_min = "3.10" %}
+{% set python_max = "3.13" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
```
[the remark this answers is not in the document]
DavidLapous: One of the reasons that I'm building it here is that I have linked libraries, which are included in the pip package (after an `auditwheel`) which is not required here since they can just be added as dependencies.

My current build system for pip is also a conda environment (`cxx-compiler` included) which is (to my knowledge) not as clean as what conda-forge. In particular, as `c_stdlib_versions` can be better fixed with conda-forge, this should allow older systems to install this library.  

If these are not  "strong reasons" I can go back to the pypi package.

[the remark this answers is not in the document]
DavidLapous: I think I didn't understand your point:
If the sources are set to pypi, then will conda use directly the compiled version or the sources (`sdist`) ? If it's the sources, then my points are not relevant, and I'll switch to pypi sources.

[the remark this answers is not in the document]
synapticarbors: If you look in https://github.com/conda-forge/staged-recipes/blob/main/recipes/example/meta.yaml, you'll see that the canonical url we recommend is:

`url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name }}-{{ version }}.tar.gz`

So it's being built from the sdist.

recipes/multipers/meta.yaml
synapticarbors: That's strange. My understanding of our tests is that we don't pull anything from the sdist unless explicitly asked for and we install the package into a conda env based on the local build artifact. Basically you should just get the  conda package, the `run` dependencies, and anything that appears under `requires`.

DavidLapous: It seems to work indeed; thanks !

recipes/multipers/meta.yaml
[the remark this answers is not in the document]
DavidLapous: As mentioned below ([here](https://github.com/conda-forge/staged-recipes/pull/28987#issuecomment-2634065279)), I had to use a `url` on a github release since my github action produced corrupted sources on pip.

DavidLapous:
I think I know the culprit for the windows version bug; it seems that `multipers` has troubles compiling with old versions of `boost` (here `libboost-devel=1.84`) on windows, which I think are installed here since `gudhi` is there. 

One possibility is simply to remove `gudhi` from the dependency, or disable the windows build until the next version of `gudhi` is released. Since windows support on `multipers` is new, I think I will disable the windows build until it works.

DavidLapous:
Another question that I forgot to ask : 

During the compilation there is a `tbb` link. Should I do something to pin the `tbb` `run` version to the `build` version ?

PR #29064, opened 2025-02-07
Adding xgrammar

zbowling:
Checklist
- [x] Build number is 0.


zbowling:
@conda-forge/help-python-c @conda-forge/help-python This should be ready for review now.

maresb:
Nice!!!

Is it problematic to be vendoring these? https://github.com/mlc-ai/xgrammar/tree/v0.1.8/3rdparty

zbowling:
> Is it problematic to be vendoring these?

I'm not here. One is a header only include so there wouldn't be anything to link. Then dlpack I get from conda and use the -Isystem line to fix the included since not included in the tarball because of the submodule so it works. The last is googletest but since we are not running tests we don't need it.

maresb:
Oh very nice!!! Thanks a lot for the explanation!

What about [picojson](https://github.com/conda-forge/picojson-feedstock)? Should that be handled similarly to [here](https://github.com/conda-forge/jwt-cpp-feedstock/blob/7367f28847444a80d1b852ebf63b864436939826/recipe/meta.yaml#L12)?

I don't do C, and I was looking at this before, so it's really educational for me.

zbowling:
Ah, good catch. Because it wasn't a submodule, I didn't think much of it. Header-only includes libs are funny things since they don't link anything, statically or dynamically. The gain by unvendoring in that case is only that if the source is fixed in some way all the dependent packages can get rebuilt. 

That other case in that other feedstock was funny in that it was trying to *prevent* picking up the unvendored version so I don't know what they were doing there. Maybe it's a case that they conditionally not use it or maybe it's used a feature also disabled.

zbowling:
Ok, I dug into this after trying to use the conda-forge version. It turns out that picojson stopped being maintained in 2015, and now a series of hard forks exist across a dozen or so projects. A whole bunch of ML frameworks seem to be copying different updated versions from each other, making their own changes, and they are all slightly incompatible with each other now. So that explains why jwt-cpp-feedstock explicitly doesn't want to use the conda-forge version. Since it's a header only include you can just make it work by disabling warnings without disabling them everywhere. 

This looks like one of the latest forks folks are copying from, but the upstream project is dead-dead and has compile issues with modern C++ compilers because of const correctness. 
https://github.com/apache/tvm/tree/main/3rdparty/picojson

So in this case, we can look of it as a fork and not a vendored version for the sake of being vendored.

recipes/xgrammar/001-fix-macos-no-triton.patch
```diff
@@ -0,0 +1,43 @@
...
++    pass
+ 
+ """The dtype of the bitmask: int32."""
+ bitmask_dtype = torch.int32
+@@ -122,6 +128,8 @@ def apply_token_bitmask_inplace(
+         )
+ 
+     if logits.device.type == "cuda":
++        if apply_token_bitmask_inplace_triton is None:
++            raise ImportError("Triton is not installed.`.")
```
wolfv, line 40: Nitpick but there is a stray ``.`

wolfv (approved):

PR #29073, opened 2025-02-09
Add dp-accounting

rpreen:
Checklist
- [x] Build number is 0.


rpreen:

recipes/dp-accounting/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/dp-accounting-{{ version }}.tar.gz
+  sha256: bf7ffc4b8048a0bc44c3601a263a23dc6c4c37bed348f1dfe378eb808ceb3419
+  patches:
+    - set_version.patch
+    - set_requirements.patch
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
```
ocefpaf, line 17:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

recipes/dp-accounting/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/dp-accounting-{{ version }}.tar.gz
+  sha256: bf7ffc4b8048a0bc44c3601a263a23dc6c4c37bed348f1dfe378eb808ceb3419
+  patches:
+    - set_version.patch
+    - set_requirements.patch
```
ocefpaf, line 13: If you can, please open issues upstream for the missing files and add a comment here so we can track the evolution of these patches.

ocefpaf:
Not sure that is going on with osx here but we don't need it to merge this PR. The package is noarch.

rpreen:
OK, thanks you!

I have added an issue about the missing files: https://github.com/google/differential-privacy/issues/286

I noticed that the PyPI tar.gz files are named slightly differently in version 0.4.3 and 0.4.4 - where in 0.4.3 (this meta.yaml) the file name is "dp-accounting-0.4.3.tar.gz" but for 0.4.4 it is named "dp_accounting-0.4.4.tar.gz" - does something need to be changed here to deal with both cases because it's using {{ name }} in the url?

ocefpaf:
> I have added an issue about the missing files: [google/differential-privacy#286](https://github.com/google/differential-privacy/issues/286)

👍🏽 
 
> I noticed that the PyPI tar.gz files are named slightly differently in version 0.4.3 and 0.4.4 - where in 0.4.3 (this meta.yaml) the file name is "dp-accounting-0.4.3.tar.gz" but for 0.4.4 it is named "dp_accounting-0.4.4.tar.gz" - does something need to be changed here to deal with both cases because it's using `dp-accounting-{{ version }}.tar.gz` in the url?

The import will work with either `_` or `-`, but you may need to fix the URL. The package name is another problem... PyPI doesn't care and you can `pip install` it in any form, like the import, but conda does care b/c conda is python agnostic. You can:

1. choose one name and keep it
2. create multiple outputs and claim both names

For 2 see https://github.com/conda-forge/ioos_qc-feedstock/blob/main/recipe/meta.yaml for an example.

rpreen:
Ok, so the URL will just need to be changed for 0.4.4? I believe it's just an odd naming for 0.4.4 because all the prior versions use the "-" naming as here.

ocefpaf:
> Ok, so the URL will just need to be changed for 0.4.4? I believe it's just an odd naming for 0.4.4 because all the prior versions use the "-" naming as here.

I think PyPI started enforcing `_` for URLs. I wish they did the same for naming and installing... But at least we have only 1 URL to download the source now. It is a win.

TL;DR it is not the project's fault.

PR #29076, opened 2025-02-10
Add tensorflow-privacy

rpreen:
Checklist
- [x]  Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [x] Build number is 0.


manics:
I can reproduce the build failure on OS X when running `python setup.py build`. The source includes a file called `BUILD`. The build process tries to create a directory called `build` which works on Linux, but fails on OS X because the filesystem is case-insensitive (though case is maintained when displaying names). I think if you rename `BUILD` to `BUILD.txt` this should successfully build.

The Windows is failing with
```
  - nothing provides tensorflow-base >=2.14.0,<2.15.0 needed by tensorflow-probability-0.22.1-pyh4af843d_1
  - unsupported request
```
Builds for tensorflow2 are broken for Windows:
https://github.com/conda-forge/tensorflow-feedstock/issues/124
so unfortunately I think you'll need to skip Windows

manics:
I've opened an upstream issue for the `BUILD`/`build` problem https://github.com/tensorflow/privacy/issues/612

rpreen:

recipes/tensorflow-privacy/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+  url: https://github.com/tensorflow/privacy/archive/v{{ version }}.tar.gz
+  sha256: 86cbf17eb2f1a23defcdf1ac7ba18a504a88bc1b2423ca5c25c7509d703fefe2
+  patches:
+    - rename_build.patch
+    - set_path.patch
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: True  # [win or py>=312]
```
ocefpaf, line 18: Looks like this one can be noarch, right?

recipes/tensorflow-privacy/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+requirements:
+  build:
+    - python
+    - setuptools
+    - pip
+    - tensorflow >=2.4,<3.0
+    - tensorflow-probability >=0.22.0,<0.23.0
```
ocefpaf, line 26:
```suggestion
```

recipes/tensorflow-privacy/meta.yaml
rpreen: Yep, will change it back

PR #29097, opened 2025-02-11
feat: add YARR software package

kratsg:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/yarr/meta.yaml
+++ b/recipes/yarr/meta.yaml
@@ -1,0 +1,51 @@
+{% set name = "yarr" %}
+{% set version = "1.5.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.cern.ch/yarr/{{ name }}/-/archive/v{{ version }}/{{ name }}-v{{ version }}.tar.gz
+  sha256: 56dd0a42f5260470239f76b89aad016913e8267fbf3e10227abf1797618de18d
+
+build:
+  skip: true  # [win]
+  script:
+    - CXXFLAGS="${CXXFLAGS}" cmake ${CMAKE_ARGS} -DCMAKE_INSTALL_PREFIX=$PREFIX -S . -B build
+    - cmake --build build --clean-first --parallel="${CPU_COUNT}"
+    - cmake --install build
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - cmake
+    - make
+  host: []
+  run: []
+
+test:
+  requires:
+    - {{ compiler('cxx') }}
+  commands:
+    - scanConsole -h
+
+about:
+  home: https://gitlab.cern.ch/YARR/YARR
+  summary: 'Yet Another Rapid Readout'
+  description: |
+    YARR is a readout system based around the concept of moving intelligence
+    from the FPGA firmware into the host computer software. The role of the FPGA is
+    the one of a reconfigurable I/O interface and not anymore the one of a hardware
+    accelerator. YARR supports multiple types of FPGA platforms and readout chips.
+  license: GPL-2.0-or-later
+  license_family: GPL
+  license_file: LICENSE.txt
+  doc_url: https://yarr.web.cern.ch/yarr/
+  dev_url: https://gitlab.cern.ch/YARR/YARR
+
+extra:
+  recipe-maintainers:
+    - kratsg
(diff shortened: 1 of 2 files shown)


kratsg:
@conda-forge/help-c-cpp, ready for review! A few notes:

- linux looks great, compiles and links as completely expected, and the test runs fine
- osx does not look great, there are a few patches specifically for osx, however it's not clear to me why the linking is wildly different (see screenshots below) but it at least compiles and I want to get the feedstock in which can then be iterated upon

Screenshot 2025-02-12 at 09 58 43

Screenshot 2025-02-12 at 10 00 44

recipes/yarr/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - {{ stdlib('c') }}
+    - cmake
+    - make
+    - rsync
+  host: []
+  run: []
+
+test:
+  requires:
+    - {{ compiler('cxx') }}
```
chrisburr, line 41: Why do you require this?

kratsg: I don't, i'll drop it - thanks for the quick catch.

kratsg: fixed in [dacee40](https://github.com/conda-forge/staged-recipes/pull/29097/commits/dacee40e2da0b592953dd6fa561575a08075c2c6)

recipes/yarr/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  run: []
+
+test:
+  commands:
+    - test -f $PREFIX/bin/scanConsole
+    - test -f $PREFIX/share/yarr/configs/controller/emuCfg_rd53a.json
+    - scanConsole -h
+    - scanConsole -k
+    # needed for YARR to pick up the configs when it loads the emulator
+    - ln -s "$PREFIX/share/yarr/configs" ./configs  # [linux]
```
chrisburr, line 44: Is this expected? Or should the sources be patched to be more friendly to packaging?

kratsg: this is generally what users would need to do on some level when they pull this. the `emulator` requires looking for a `configs` path specifically with the information needed 

```
            "feCfg" : "configs/emulator/emu_fe0_rd53a.json"
```

and it's not trivial to patch the source here as this touches a few more places (and not really worthwhile, I don't think).

chrisburr: You could do something lik:

```
find $PREFIX/share/yarr/configs -name '*.json' -print -exec sed -i 's@"configs/@"'$PREFIX'/configs/@g' {} \;
```

Conda will then take care of replacing the path at install time to be correct.

kratsg: added it, that's a great idea and I did not know conda could do that :)

kratsg: @chrisburr seems like while the regex works, `conda` is stripping away the full prefix on install? I'm not fully sure why though...

kratsg: I can confirm that the regex works to replace as

```
{
    "ctrlCfg" : {
        "type": "emu_Rd53a",
        "cfg" : {
            "type" : "rd53a",
            "tx" : {
                "id" : 1337,
                "size" : 256
            },
            "rx" : {
                "id" : 1338,
                "size" : 256
            },
            "feCfg" : "$PREFIX/configs/emulator/emu_fe0_rd53a.json"
        }
    }
}
```

but looking at conda's docs: https://docs.conda.io/projects/conda-build/en/stable/resources/package-spec.html#package-metadata

it seems I need to be using the actual value of the prefix, not the `"$PREFIX"`, but the value of it. It also seems to indicate as such here: https://github.com/conda/conda-prefix-replacement

kratsg: will try this version

```
find $PREFIX/share/yarr/configs -name '*.json' -print -exec sed -i "s@\"configs/@\"$PREFIX/configs/@g" {} +
```

for the replacement and see if that works.

kratsg: I think the problem is a missing `share/yarr/` for the prefix substitution.. so let me try that first.

recipes/yarr/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - cmake
+    - make
+    - rsync
+    - sed
+  host: []
+  run: []
```
matthewfeickert, line 38:
```suggestion
```

Not wrong as in, but you can drop these keys if they're empty. I find that cleaner visually, but again, not wrong as is.

kratsg:

PR #29136, opened 2025-02-14
Adding wibl-python package

selimnairb:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/wibl-python/meta.yaml
+++ b/recipes/wibl-python/meta.yaml
@@ -1,0 +1,60 @@
+{% set name = "wibl-python" %}
+{% set version = "1.0.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/CCOMJHC/WIBL/archive/refs/tags/{{ name }}-{{ version }}.tar.gz
+  sha256: d4e6d4b50f0f674657c817531e34c8d29056ea1d6ec121bf54f4f8e32ea42012
+
+build:
+  entry_points:
+    - wibl = wibl.command.__main__:main
+  noarch: python
+  script: |
+    {{ PYTHON }} -m pip install ./wibl-manager
+    {{ PYTHON }} -m pip install ./wibl-python -vv
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+    - setuptools
+  run:
+    - python
+    - gdal
+    - matplotlib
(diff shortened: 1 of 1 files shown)


brian-r-calder:
Happy to be included as a maintainer.

recipes/wibl-python/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+  url: https://github.com/CCOMJHC/WIBL/archive/refs/tags/{{ name }}-{{ version }}.tar.gz
+  sha256: d4e6d4b50f0f674657c817531e34c8d29056ea1d6ec121bf54f4f8e32ea42012
+
+build:
+  entry_points:
+    - wibl = wibl.command.__main__:main
+  noarch: python
+  script: |
+    {{ PYTHON }} -m pip install ./wibl-python/wibl-manager
+    {{ PYTHON }} -m pip install csbschema==1.1.2
```
synapticarbors, line 19: You will have to create another recipe for `csbschema` as we do not permit installing for `pypi` directly.

selimnairb: Hi @synapticarbors. Thanks for catching this. Yeah, I came to that conclusion. Grayskull to the rescue...

selimnairb: @synapticarbors Added csbschema conda-forge dependency, waiting on CI. Thanks for reviewing these quickly!

synapticarbors:
Windows is failing because `gunicorn` is not compatible with windows. This contradicts the upstream metadata of the packages being platform independent:

https://github.com/CCOMJHC/WIBL/blob/main/wibl-python/wibl-manager/pyproject.toml#L18

It looks like it's only really used in the docker container, so it doesn't seem like a true dependency. Options would be to fix the upstream metadata (if that's appropriate), or patch it out of the `requirements.txt`, or remove it here (although that would probably cause the `pip check` to fail).

synapticarbors:
Or we ignore it since this is `noarch:python` and will build fine in the feedstock because that's linux-only, but it will cause issues for windows users who try to install it.

selimnairb:
> Or we ignore it since this is `noarch:python` and will build fine in the feedstock because that's linux-only, but it will cause issues for windows users who try to install it.

Hi @synapticarbors I restructured the upstream dependencies to be noarch-compliant for the default install of `wibl-manager`, which is all that this recipe (`wibl-python`) needs.

PR #29150, opened 2025-02-15
Add `lytekit`

nh13:
- [x] Depends on #29148
- [x] Depends on #29149

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/lytekit/meta.yaml
+++ b/recipes/lytekit/meta.yaml
@@ -1,5 +1,5 @@
 {% set name = "lytekit" %}
-{% set version = "0.15.28" %}
+{% set version = "0.15.27" %}
 {% set python_max = "3.11" %}
 
 package:
@@ -7,8 +7,8 @@ package:
   version: {{ version }}
 
 source:
-  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
-  sha256: 37fb835cc6bb6f3ad9b8e493b6c751c29ba50541cc94f831ba267132556a9da5
+  url: https://files.pythonhosted.org/packages/da/ec/c65f7809bddcc302c86564830f3324ce593cbdb55337210c3f57883ce374/{{ name|lower }}-{{ version }}-py3-none-any.whl
+  sha256: e222de9de2012f5802731ce8032418598075d5e8bca5855f83036d65da0d8c53
 
 build:
   skip: true  # [py<37]
@@ -18,7 +18,7 @@ build:
     - pyflyte-map-execute=flytekit.bin.entrypoint:map_execute_task_cmd
     - pyflyte=flytekit.clis.sdk_in_container.pyflyte:main
     - flyte-cli=flytekit.clis.flyte_cli.main:_flyte_cli
-  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  script: {{ PYTHON }} -m pip install {{ name|lower }}-{{ version }}-py3-none-any.whl --no-deps --no-build-isolation --no-cache-dir -vvv 
   number: 0
 
 requirements:
@@ -61,7 +61,7 @@ requirements:
     - diskcache >=5.2.1
     - cloudpickle >=2.0.0
     - cookiecutter >=1.7.3
-    - latch-persistence ==0.1.5
+    - latch-persistence ==0.1.4
     - memory_profiler ==0.61.0
 
 test:
--- a/recipes/lytekit/meta.yaml
+++ b/recipes/lytekit/meta.yaml
@@ -1,6 +1,5 @@
 {% set name = "lytekit" %}
 {% set version = "0.15.27" %}
-{% set python_max = "3.11" %}
 
 package:
   name: {{ name|lower }}
@@ -27,7 +26,7 @@ requirements:
     - pip
     - setuptools
   run:
-    - python >={{ python_min }},<={{ python_max }}
+    - python >={{ python_min }}
     - lyteidl ==0.2.1
     - wheel >=0.30.0,<1.0.0
     - click >=6.6,<9.0
--- a/recipes/lytekit/meta.yaml
+++ b/recipes/lytekit/meta.yaml
@@ -1,0 +1,86 @@
+{% set name = "lytekit" %}
+{% set version = "0.15.25" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
+  sha256: 3bf5aa3554c7d664e7efc874d90af168becb2566586e897f429370a47bf9dffd
+
+build:
+  skip: true  # [py<37]
+  entry_points:
+    - pyflyte-execute=flytekit.bin.entrypoint:execute_task_cmd
+    - pyflyte-fast-execute=flytekit.bin.entrypoint:fast_execute_task_cmd
+    - pyflyte-map-execute=flytekit.bin.entrypoint:map_execute_task_cmd
+    - pyflyte=flytekit.clis.sdk_in_container.pyflyte:main
+    - flyte-cli=flytekit.clis.flyte_cli.main:_flyte_cli
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - lyteidl ==0.2.1
+    - wheel >=0.30.0,<1.0.0
+    - click >=6.6,<9.0
+    - croniter >=0.3.20,<4.0.0
+    - deprecated >=1.0,<2.0
+    - docker-py >=7.1.0
+    - python-dateutil >=2.1
+    - grpcio >=1.43.0,!=1.45.0,<2.0
+    - grpcio-status >=1.43,!=1.45.0
+    - protobuf >=3.6.1,<4
+    - python-json-logger >=2.0.0
+    - pytimeparse >=1.1.8,<2.0.0
+    - pytz
+    - keyring >=18.0.1
+    - requests >=2.18.4,<3.0.0
+    - responses >=0.10.7
+    - sortedcontainers >=1.5.9,<3.0.0
+    - statsd >=3.0.0,<4.0.0
+    - urllib3 >=1.22,<2.0.0
+    - wrapt >=1.0.0,<2.0.0
+    - retry ==0.9.2
+    - dataclasses-json >=0.5.2,<=0.5.6
+    - jsonschema >=4.5.1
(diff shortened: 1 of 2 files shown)


recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,87 @@
+{% set name = "lytekit" %}
+{% set version = "0.15.25" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.io/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
```
nh13, line 9:
```suggestion
  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
```

nh13:

nh13:

nh13:

nh13:
@conda-forge/help-python Ready for review!

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
+  sha256: bf3d9e127ad02b94e88c9b7f4f8df2e81946104c470c935d00ad99b312f46915
+
+build:
+  skip: true  # [py<37 or win or py>310]
```
ocefpaf, line 13: Looks like this one can be noarch.

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python
+    - pip
+    - setuptools
+  run:
+    - python
+    - lyteidl ==0.2.1 =pyhd8ed1ab_1
```
ocefpaf, line 30: If you need to pin a specific build, something is probably wrong with  the `lyteidl` dependency metadata and it is better to fix it before merging this PR.

nh13, line 30: I was testing this right after release, and the CI wasn't picking up the lates build, so this worked for me.  I'll change it back an see what happens

```suggestion
    - lyteidl ==0.2.1
```

nh13:
@ocefpaf looks like builds fail because some merged recipes haven't been removed, but in the meantime, are the updates are what you're looking for?

nh13:

nh13:
@ocefpaf I think I am going in circles with noarch and then package selectors, I am just not understanding.  Any chance you can make some suggestions to the recipe and I can learn what I should be doing?

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/lytekit-{{ version }}.tar.gz
+  sha256: bf3d9e127ad02b94e88c9b7f4f8df2e81946104c470c935d00ad99b312f46915
+
+build:
+  skip: true  # [py<37 or py>310]
```
nh13, line 14: I could not get it to work with requiring a python version lower than 3.11 AND having it be no-arch.  Suggestions welcome, but I think this is ready.

mfansler: The only error was this in **win-64** build

```python
  File "D:\bld\lytekit_1741026804461\_test_env\Scripts\flyte-cli-script.py", line 5, in <module>
    from flytekit.clis.flyte_cli.main import _flyte_cli
  File "D:\bld\lytekit_1741026804461\_test_env\lib\site-packages\flytekit\clis\flyte_cli\main.py", line 19, in <module>
    from google.protobuf.pyext.cpp_message import GeneratedProtocolMessageType as _GeneratedProtocolMessageType
  File "D:\bld\lytekit_1741026804461\_test_env\lib\site-packages\google\protobuf\pyext\cpp_message.py", line 39, in <module>
    from google.protobuf.pyext import _message
ImportError: cannot import name '_message' from 'google.protobuf.pyext' (D:\bld\lytekit_1741026804461\_test_env\lib\site-packages\google\protobuf\pyext\__init__.py)
```

Since only **linux-64** actually needs to build for **noarch** recipes, and this seem like specifically an internal problem with `protobuf` **win-64** builds, I'd suggest that the error can be ignored and this should be re-reviewed *after converting back to **noarch***.

nh13: > _ after converting back to noarch._

I’ve tried a bunch of times to make it “no arch”but the Python selectors preclude this. I’d welcome some help given two folks have suggested this, so I’m missing something obvious 😅

mfansler, line 14: Drop the selector and instead [use `python_min` variable](https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-python):

```suggestion
```

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - pyflyte-fast-execute=flytekit.bin.entrypoint:fast_execute_task_cmd
+    - pyflyte-map-execute=flytekit.bin.entrypoint:map_execute_task_cmd
+    - pyflyte=flytekit.clis.sdk_in_container.pyflyte:main
+    - flyte-cli=flytekit.clis.flyte_cli.main:_flyte_cli
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python
```
mfansler, line 26:
```suggestion
    - python {{ python_min }}
```

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - pip
+    - setuptools
+  run:
+    - python
```
mfansler, line 30:
```suggestion
    - python >={{ python_min }}
```

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - plugins
+  commands:
+    - pip check
+    - pyflyte-execute --help
+    - pyflyte-fast-execute --help
+    - pyflyte-map-execute --help
+    - pyflyte --help
+    - flyte-cli --help
+  requires:
+    - python
```
mfansler, line 79:
```suggestion
    - python  {{ python_min }}
```

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
+{% set name = "lytekit" %}
+{% set version = "0.15.27" %}
+{% set python_max = "3.11" %}
```
mfansler, line 3: Is this from something concretely in the package code that is incompatible with 3.11+? If so, then use it in the `run:` section.

If this was only because dependencies couldn't solve, then I'd recommend removing this:

```suggestion
```

mfansler (changes requested):
Here are some concrete suggestions for `noarch`. Hope that helps!

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - urllib3 >=1.22,<2.0.0
+    - wrapt >=1.0.0,<2.0.0
+    - retry ==0.9.2
+    - dataclasses-json >=0.5.2,<=0.5.6
+    - jsonschema >=4.5.1
+    - marshmallow-jsonschema >=0.12.0
+    - marshmallow-enum
+    - natsort >=7.0.1
+    - docker-image-py >=0.1.10
+    - singledispatchmethod  # [py<38]
```
nh13, line 58: Here's the selector that I found the linter was complaining about!
```suggestion
    - singledispatchmethod
```

nh13:
@mfansler thank-you for the suggestions, I think this is ready!

recipes/lytekit/meta.yaml
mfansler: The `noarch` strategy only tests `python_min`, so you don't need to block it with `run` requirements. Please remove.

recipes/lytekit/meta.yaml
mfansler: The current `python_min` is 3.9, so this dependency could be altogether removed.

mfansler:
> @mfansler thank-you for the suggestions, I think this is ready!

@nh13 I'm not a regular reviewer for conda-forge/help-python, so I think it would be better if you get a review from them as this is not a trivial package.

FWIW, I did double-check that [the `noarch` artifact](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1190414&view=artifacts&pathAsName=false&type=publishedArtifacts) installs locally for me (`osx-arm64`) and the basic tests in the recipe similarly work fine. It would be good that someone test this similarly on Windows to ensure the entrypoints work.

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - diskcache >=5.2.1
+    - cloudpickle >=2.0.0
+    - cookiecutter >=1.7.3
+    - latch-persistence ==0.1.4
+    - memory_profiler ==0.61.0
+
+test:
+  imports:
+    - flytekit
+    - plugins
```
mfansler, line 70: This strikes me as odd. Is there some special mechanism for packages to deliver into a common `plugins` module? Otherwise, that seems like something super prone to clobber.

nh13: I agree, but I think this is more an issue with the package and less with this recipe.

nh13:
@conda-forge/help-python Ready for review!

nh13:
@ocefpaf bump!

recipes/lytekit/meta.yaml
ocefpaf: I wonder if we should do multiple outputs and create that namespace. Plugins is quite generic though...

recipes/lytekit/meta.yaml
ocefpaf: @nh13 there are still some review actions to be addressed, like this one.

recipes/lytekit/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
```
nh13, line 69: I think this can just be removed, since lytekit doesn't actually install this from the top-level install.  I think it was added by `grayskull`:
```suggestion
```

nh13:
I think we're coming down to the final steps. Thanks @ocefpaf !

recipes/lytekit/meta.yaml
mfansler: But this recipe does install a `plugins` module. Removing the test obscures this fact.

I would suggest either keep the test, or remove the module if it actually isn't necessary.

ocefpaf:
@mfansler do you want to become on "official" staged-recipes reviewer? I used quotes b/c your reviews are better than mine and all you are missing is the commit rights here. Let us know if you do.

nh13:
Thank-you both!

PR #29180, opened 2025-02-16
add tinygrad

h-vetinari:
https://github.com/tinygrad/tinygrad is pretty cool 😎 

conda-forge should be well-positioned to provide this, because we ship C++/CUDA compilers etc.


h-vetinari:
@conda-forge/help-python, this should be ready, PTAL.

This package is somewhat unusual in that it uses a JIT-compiler with many possible backends, so it needs the compilers at runtime. I'm not pinning the version so that users can choose the version they want themselves.

To ensure everything works as expected I wanted to run the test suite, which works well up to a small number of tests which aren't suitable for us at the moment. For now it's easier to just use clang everywhere (their logic for the default backend fails with GCC, and I'd prefer minimal patching for now). Once we've got the full setup, I'll try to work through the open questions with upstream

h-vetinari:
> @conda-forge/help-python, this should be ready, PTAL.

Sorry that this had been a bit premature. Turns out the tests on osx only really run reliably if `pytest-xdist` is active.

recipes/tinygrad/recipe.yaml
```diff
@@ -0,0 +1,137 @@
+context:
+  version: "0.10.1"
+  build_number: 0
+  # see github.com/conda-forge/conda-forge.github.io/issues/1059 for naming discussion
+  torch_proc_type: ${{ "cuda" ~ cuda_compiler_version | version_to_buildstring if cuda_compiler_version != "None" else "cpu" }}
+  tests_to_skip: >
+    _not_a_real_test
+    ${{ "fails `assert Device.DEFAULT in failed_platforms`" if 0 }}
```
jaimergp, line 8: First time seeing this type of comments. Surprised it works o.o
Makes me wonder if we never discussed `{# #}` equivalents for the new recipe format.

recipes/tinygrad/recipe.yaml
```diff
@@ -0,0 +1,137 @@
...
+package:
+  name: tinygrad
+  version: ${{ version }}
+
+source:
+  url: https://github.com/tinygrad/tinygrad/archive/refs/tags/v${{ version }}.tar.gz
+  sha256: a22402a8c08da0930720b13a8a66a9c9f99507b15fca9caf31fde63776529ca8
+  patches:
+    - patches/0001-also-install-tinygrad.viz.patch
+    # clang 19 doesn't seem to support -std=metal3.1 yet
```
jaimergp, line 35: this comment better belongs in the patch itself, I think

recipes/tinygrad/recipe.yaml
```diff
@@ -0,0 +1,137 @@
...
+    - patches/0002-downgrade-default-metal-standard.patch
+
+build:
+  number: ${{ build_number }}
+  string: ${{ torch_proc_type }}_py${{ python | version_to_buildstring }}_h${{ hash }}_${{ build_number }}
+  script: python -m pip install . -vv
+  skip:
+    - match(python, "<3.10")
+    # toolchain for 11.8 is not usable in end-user environment
+    - cuda_compiler_version == "11.8"
```
jaimergp, line 45: Should this use `match()` to make sure you don't run into errors derived of strict string comparison (e.g. 11.8.0 != 11.8, but it should).

recipes/tinygrad/recipe.yaml
```diff
@@ -0,0 +1,137 @@
...
+        - sentencepiece
+        - tabulate
+        - tiktoken
+        - tqdm
+        - transformers
+    files:
+      source:
+        - examples/
+        - extra/
+        - test/
```
jaimergp, line 110: These are adding a buuunch of files to the package. Just mentioning it in case you want to think whether all of them are needed or not.

jaimergp (approved):
Just a couple comments that shouldn't block the merge. I wouldn't run the whole test suite once known it actually passes. Maybe you want to pick only a few? Same for the test files.

h-vetinari:
Thanks for the review!

> I wouldn't run the whole test suite once known it actually passes.

For pure python packages, I'm fine with that. For something as complicated as a cross-platform on-the-fly JIT compiler, I'm going to take all the testing I can get.

recipes/tinygrad/recipe.yaml
h-vetinari: I can split them off into a `tinygrad-tests` package, but I do want to keep running the test suite...

recipes/tinygrad/recipe.yaml
h-vetinari: Good point, thanks

recipes/tinygrad/recipe.yaml
h-vetinari: > Makes me wonder if we never discussed `{# #}` equivalents for the new recipe format.

Yeah, I really don't like the way comments work in this context, but this is what our rattler translation crew came up with in https://github.com/conda-forge/torchvision-feedstock/blob/main/recipe/recipe.yaml, so I didn't explore this further.

CC @wolfv

recipes/tinygrad/recipe.yaml
jaimergp: Yea I think that's reasonable.

PR #29201, opened 2025-02-19
Add fenics-oasis recipe

johannesring:
Checklist
- [x] Build number is 0.


keiyamamo:
I  confirm that I'm willing to be a maintainer for `fenics-oasis1 on conda-forge.

RemDelaporteMathurin:
Shouldn't you skip builds on windows since FEniCS is only available via conda from Linux and osx?

johannesring:

recipes/fenics-oasis/meta.yaml
```diff
@@ -0,0 +1,58 @@
+{% set name = "fenics-oasis" %}
+{% set version = "2019.1" %}
+{% set python_min = "3.7" %}
```
ocefpaf, line 3: This is not supported on conda-forge. While we can merge we won't guarantee it will work on py<39

recipes/fenics-oasis/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+    - pip
+    - setuptools
+  run:
+    - python
+    - fenics
+    - numpy
+    - matplotlib-base
+    - procps-ng  # [linux]
+    - wget  # [linux]
+    - curl  # [osx]
```
ocefpaf, line 32: Are these really needed? We prefer noarch package, if they are really mandatory dependencies you should try OS-noarch: https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

recipes/fenics-oasis/meta.yaml
johannesring: @ocefpaf I have updated the minimum Python version to 3.9 as suggested. Thanks!

recipes/fenics-oasis/meta.yaml
```diff
@@ -1,6 +1,6 @@
 {% set name = "fenics-oasis" %}
 {% set version = "2019.1" %}
-{% set python_min = "3.7" %}
+{% set python_min = "3.9" %}
```
ocefpaf, line 3: You only need to set this one if your min is higher than conda-forge's min. Otherwise, remove it and let the defaults be used.

```suggestion
```

recipes/fenics-oasis/meta.yaml
johannesring: @ocefpaf I have removed `wget` and `curl` from the runtime dependencies and restored `noarch: python`.

These tools were only used to fetch mesh files for example problems, which are not essential for the core functionality of `fenics-oasis`.

Thanks for the review!

recipes/fenics-oasis/meta.yaml
johannesring: @ocefpaf I have removed `python_min` so conda-forge defaults are used. Thanks for the feedback!

recipes/fenics-oasis/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+  host:
+    - python
+    - pip
+    - setuptools
+  run:
+    - python
+    - fenics
+    - numpy
+    - matplotlib-base
+    - procps-ng  # [linux]
```
ocefpaf, line 29: This one is also "arch'ed", is it optional?

johannesring: @ocefpaf Yes, I have removed `procps-ng` now since it is only used for optional memory usage information and is not strictly needed.

johannesring:
@ocefpaf - How can I keep the package `noarch`, while skipping the build on Windows? This package is similar to the tutleFSI package (see [meta.yaml](https://github.com/conda-forge/turtlefsi-feedstock/blob/main/recipe/meta.yaml) for `turtlefsi-feedstock`), which is `norarch` and depends on `fenics`, but does not skip the build on Windows.

synapticarbors:
When a dependency is missing on a particular platform, we tend to just ignore failures on that platform if the error is related to the missing dependency because. In the feedstock you mentioned (turtlefsi-feedstock), since it is `noarch: python` it is never built on a windows CI -- it uses linux to build the conda package and then it can be used anywhere. 

So I think everything is ok here once you switch to using the `python_min` jinja variable as described in:

johannesring:
@synapticarbors Thanks for the clarification! I have updated the recipe to use `python_min` as recommended.

recipes/fenics-oasis/meta.yaml
```diff
@@ -7,14 +7,15 @@ package:
...
-  sha256: 4456d5f042292e30bffda0487b57e282fe6e776cb6cdc832140072983b2c9d8d
+  sha256: a610ddd83dcd514a193ed3dc7933475995ee850582e64e4aec90b27d28241592
 
 build:
   entry_points:
     - oasis=oasis.run_oasis:main
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
+  skip: true  # [win]
```
ocefpaf: It is OK for noarch package to fails on Windows with a missing dependency. We prefer noarch b/c it is faster to build and the package will be ready when the missing dependency is available on Windows.

ocefpaf:
```suggestion
```

johannesring:
@ocefpaf Thanks for the feedback! So it is good to go now without the `skip: true  # [win]`?

recipes/fenics-oasis/meta.yaml
```diff
@@ -15,15 +15,14 @@ build:
   noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
-  skip: true  # [win]
 
 requirements:
   host:
-    - python
+    - python >=3.9
```
ocefpaf:
```suggestion
    - python {{ python_min }}
```

recipes/fenics-oasis/meta.yaml
```diff
@@ -15,15 +15,14 @@ build:
...
     - pip
     - setuptools
   run:
-    - python
+    - python >=3.9
```
ocefpaf:
```suggestion
   - python >={{ python_min }}
```

recipes/fenics-oasis/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+test:
+  imports:
+    - oasis
+  source_files:
+    - tests/*
+  commands:
+    - pytest ./tests/
+  requires:
+    - python {{ python_min }}
+    - pytest
```
ocefpaf, line 39:
```suggestion
  requires:
    - python {{ python_min }}
    - pip
    - pytest
```

johannesring:
It seems that the failures are related to the following PR in the `fenics` package: https://github.com/conda-forge/fenics-feedstock/pull/256

ocefpaf:
> It seems that the failures are related to the following PR in the `fenics` package: [conda-forge/fenics-feedstock#256](https://github.com/conda-forge/fenics-feedstock/pull/256)

Comment the pip-check test out and leave a note in the line above to re-enable this test when they fix it upstream.

johannesring:
@ocefpaf I have commented out `pip check` and added a reference to the upstream issue.

PR #29285, opened 2025-02-26
Add PyHough recipe

andrew-l-miller:
Checklist
- [x ] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [x ] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/5eddbd7fc9d1502169089da06c3688d9759be978/recipes/example/meta.yaml#L64-L73) for an example).
- [x ] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).
- [x ] If static libraries are linked in, the license of the static library is packaged.
- [x ] Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md).
- [x ] Build number is 0.
- [x ] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details).
- [x ] GitHub users listed in the maintainer section have posted a comment confirming they are willing to be listed there.
- [x ] When in trouble, please check our [knowledge base documentation](https://conda-forge.org/docs/maintainer/knowledge_base.html) before pinging a team.


andrew-l-miller:
What do I do now?

andrew-l-miller:
why are the tests failing?

andrew-l-miller:
ok, now what?

andrew-l-miller:

duboism:
Would anybody have time to review this ?

duboism:
Any news ?

recipes/pyhough/meta.yaml
```diff
@@ -0,0 +1,50 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/pyhough-{{ version }}.tar.gz
+  sha256: 75d4635c450088df2c6c7495024223880168c713b48730c5ea2dc1cf16362d8e
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: True  # [win]
```
ocefpaf, line 15: This one can probably be noarch, no?

andrew-l-miller: What exactly should be noarch?

andrew-l-miller: I vaguely remember having noarch in there before but the build was failing. Is this essential?

ocefpaf: Let's see the failure , it may be due to a missing dependency in a platform/arch, which is fine, or due to missing entry-points. It is essential b/c it helps us save CI time by building it once instead of a full OS x Python matrix.

andrew-l-miller: Hi @ocefpaf , I've just updated the meta.yaml file to include the noarch. The builds on each operating system are failing, and I don't know why

andrew-l-miller: Hi @ocefpaf , I think I know the reason why noarch wasn't working

My package depends on pyfstat, and pyfstat depends on lalsuite, which contains native C/C++/Fortran code. Therefore, noarch: python is not allowed, I think

andrew-l-miller: Hello, now it is compiling

ocefpaf: > pyfstat, one of the packages my package depends on, is also not on conda forge

ocefpaf: I need to see the failure.

recipes/pyhough/meta.yaml
```diff
@@ -1,47 +1,46 @@
 {% set name = "pyhough" %}
 {% set version = "0.0.7" %}
+{% set python_min = "3.8" %}
```
ocefpaf, line 3: Only set `python_min` if the minimum required version is higher than the conda-forge default. 

```suggestion
```

recipes/pyhough/meta.yaml
```diff
@@ -1,47 +1,46 @@
...
 source:
   url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/pyhough-{{ version }}.tar.gz
-  sha256: ccebab24362daa826c7ac3aa7db188c041562262ff17cfd6f80e0bfc6d99532e 
+  sha256: ccebab24362daa826c7ac3aa7db188c041562262ff17cfd6f80e0bfc6d99532e
 
 build:
-  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  noarch: python
+  script:
+    - {{ PYTHON }} -m pip install pyfstat
```
ocefpaf, line 16: You have to package that as a conda-forge package.

andrew-l-miller:
I cannot package pyfstat as a conda forge package - it's not my package.
It's just one that I use

andrew-l-miller:
I just tried to put pyfstat on conda forge but it seems it already exists. I got [this error](https://github.com/conda-forge/staged-recipes/actions/runs/15266806936/job/42933846434?pr=29285)

I'm confused. I thought the builds were failing because pyfstat is not on conda forge

andrew-l-miller:
@ocefpaf how can I resolve [this](https://github.com/conda-forge/staged-recipes/actions/runs/15278804283/job/42972167639?pr=29285)?

ocefpaf:
Looks like the `pyfstat` was already in conda-forge. You never had to pip install it to start with. Please remove that recipe from this PR and let's see how it goes.

recipes/pyhough/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - hatchling
+    - numpy
+    - pyfstat
+    - matplotlib-base
+    - versioneer
+  run:
+    - python >={{ python_min }}
+    - lalsuite
+    - lalpulsar
+    - lalapps
```
ocefpaf, line 30: These are note defined in the pyproject.toml.

andrew-l-miller: these are all dependencies of pyfstat. Do they not need to be here?

recipes/pyhough/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - hatchling
+    - numpy
+    - pyfstat
+    - matplotlib-base
```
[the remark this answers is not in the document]
ocefpaf: There are build/host dependencies, like compilers, headers, packaging backends, etc. And run-time dependencies, the ones that you need to run the program.

Those seems to be mixed upstream. They are defining build dependencies instead of run.

recipes/pyhough/meta.yaml
ocefpaf: No. You need only to depend on `pyfstat`. Unless `pyhough` imports them directly, which doesn't seems like.

andrew-l-miller:
I do not understand this [error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1255379&view=logs&s=6884a131-87da-5381-61f3-d7acc3b91d76&j=4cabe686-70ae-553a-7fd0-310379f2cbac)

recipes/pyhough/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - numpy
+    - matplotlib-base
+
+test:
+  imports:
+    - pyhough
+  commands:
+    - pip check
+  requires:
+    - bash
```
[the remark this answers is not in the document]
andrew-l-miller: I don't think so, let's see. Just resubmitted

recipes/pyhough/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+
+build:
+  noarch: python
+  script:
+    - {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python {{ python_min }}
```
ocefpaf, line 20:
```suggestion
    - python {{ python_min }}
    - pip
```

andrew-l-miller: did this, still getting [error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1255461&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff)

It seems to say this

pyfstat 2.2.1 requires lalsuite, which is not installed.
pyfstat 2.2.1 requires versioneer, which is not installed.

But I don't import `lalsuite` or `versioneer` directly. Just `pyfstat`

ocefpaf: This is a problem with the https://github.com/conda-forge/pyfstat-feedstock/blob/main/recipe/meta.yaml recipe. Do you mind sending a PR there adding the missing dependencies?

andrew-l-miller: Ok, what exactly do I do? I have to pull their conda-forge repo, add `lalsuite` and `versioneer` under "run"?

Or can I open an issue and ask them to do it?

andrew-l-miller: @ocefpaf I removed pip check because according to the pyfstat people, it "complains about the dependencies being defined in setup.py but not explicitly satisfied. I think for lalsuite this is unavoidable thanks to the split-up approach in conda but single-package in pypi. "

Now, the build only fails on win_64 for [some out of memory reason](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1256339&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de). Is that ok?

recipes/pyhough/meta.yaml
```diff
@@ -30,8 +30,6 @@ requirements:
 test:
   imports:
     - pyhough
-  commands:
-    - pip check
```
ocefpaf: We should really fix this upstream. Otherwise we risk forcing all downstream package to skip this test.

andrew-l-miller: @ocefpaf Can you clarify where upstream? 

I'm adding @dbkeitel to the loop here, who handles `pyfstat`

I think `pyfstat` doesn't implement `pip check`

dbkeitel: PS: Or would this be something that could be solved on our (PyFstat) end by migrating from declaring dependencies in `setup.py` to doing it in a `pyproject.toml` instead? If so, I'd appreciate some pointers, but would realistically only be able to try this out in autumn.

ocefpaf: @dbkeitel everything you declare as a dependency, setup.py or pyproject.toml, will be caught by pip-check. If your conda package is fully functional, maybe the best way forward if to patch it to remove the full dependency on `lalsuite` as you mention.

PS: Or the full `lalsuite` dep is to build only? If so, declaring a build-system in a pyproject.toml would work. Pip-check examine only the run-time dependencies.

dbkeitel: FYI here is the latest pyfstat recipe build: https://github.com/conda-forge/pyfstat-feedstock/pull/35

andrew-l-miller: @ocefpaf I've added that comment. Are there next steps?

PR #29293, opened 2025-02-26
Added package for node_red_contrib_web_worldmap

phreed:
Checklist
- [x] Build number is 0.

Pending acceptance of the `node-red` feedstock.


phreed:
@conda-forge/help-nodejs is ready for review pending approval of https://github.com/conda-forge/staged-recipes/pull/29292

danielnachun:
If this package does not produce an executable, I'm not sure it's possible for us to package it (yet) in conda-forge as its own package.  It might make sense to add it to your other `node-red` PR instead.  You'd have to add the tarball from here as another file (probably saved to a subdirectory), and then just run the same set of commands to install it so that it ultimately gets bundled with the main `node-red` package.

phreed:
> If this package does not produce an executable, I'm not sure it's possible for us to package it (yet) in conda-forge as its own package. It might make sense to add it to your other `node-red` PR instead. You'd have to add the tarball from here as another file (probably saved to a subdirectory), and then just run the same set of commands to install it so that it ultimately gets bundled with the main `node-red` package.

Yes, I could do that but there are a lot of node-red extensions.
It could get out of hand very fast.
Maybe this is just not the kind of thing that should be done in a conda package.
It is kind of like an extra but there would just be too many of them.

PR #29311, opened 2025-02-27
Adding the Mumble voip package

phreed:
Checklist
- [x] Build number is 0.

In order to address the issue of vendoring other packages this package is waiting on the following packages:
* https://github.com/conda-forge/staged-recipes/pull/31483
* https://github.com/conda-forge/staged-recipes/pull/31484
* https://github.com/conda-forge/staged-recipes/pull/31485
* https://github.com/conda-forge/staged-recipes/pull/31486
* https://github.com/conda-forge/staged-recipes/pull/31004
* https://github.com/conda-forge/staged-recipes/pull/31010


phreed:
I have some known issues for which I would like guidance:
```text
CMake Error at CMakeLists.txt:58 (include):
 │ │   include could not find requested file:
 │ │     FindPythonInterpreter
 ```
 I included `python` as a build dependency, I thought that would fix this issue but it does not.
 
 This package has dependencies which do not have conda-forge packages.

phreed:
This recipe has several patches. I am not the owner of the upstream package. Is is acceptable to go ahead with producing the feedstock and address the upstream issues there?
This recipe has two outputs, I have commented one of them out (mumble-client).
I want to get the upstream issues in the mumble-server addressed before activating the client.

phreed:

danielnachun:
This recipe is currently very complicated and has a lot of unnecessary content that needs to be cleaned up before it can be merged.  

A few changes to start off with:
- You are setting many values in `variants.yaml` which are already the default values in conda-forge-pinning.  We don't want to set these values manually unless they must be a non-default value.  The only value that will probably need to be set manually is `MACOSX_SDK_VERSION`.
- You should not normally need to set `c_stdlib_version` on macOS for C++ programs - instead you need to add `-D_LIBCPP_DISABLE_AVAILABILITY` to the `CXXFLAGS` and make sure `MACOSX_SDK_VERSION` is set to the required OS minimum.
- You are trying skip using conda-forge supplied compilers on macOS, which is not permitted.  Instead of doing this, we need to figure out why the build isn't working with our compilers and fix that instead, even if it requires patching.  Right now I suspect your problems are originating from the typical issues with targeting a newer `libc++` on an older macOS, which can usually be solved with the above fix I have suggested.
- The recipe is too long to have the build script embedded in it.  Please put the build script in a separate file.
- While I appreciate your effort to try to use `nushell` to get a single build script for both OSes, I'm afraid it adds quite a bit of unnecessary complexity to the logic.  Most CMake projects only require a relatively small number of commands to run so it is not difficult to implement these as separate bash and batch scripts.
- You are vendoring package with stable versions in the recipe.  When a stable version can be used, it is preferable to patch the CMake file to use the "system" version rather than rebuilding the same package again.
- For the vendored dependencies which are not from stable versions, their licenses must be included in the recipe.

I'm also happy to try to push some commits for this as I think the whole recipe can be greatly simplified and that will make it much easier to review some of the trickier issues.

phreed:
I was not sure what to do with the vendoring.
The source is working with copies of the dependencies.
As I am not sure if those copies were modified, I included them as well.
I thought I had included the licenses properly but maybe not.

phreed:
I have made a PR for the patches

> I'm also happy to try to push some commits for this as I think the whole recipe can be greatly simplified and that will make it much easier to review some of the trickier issues.

If it is all the same to you I would like to try to get this in shape with your guidance as a learning exercise.

phreed:
This mumble system typically bundles (vendors) several packages. 
This is not the conda-forge way.
To that end I am presenting another set of staged-recipes which are those packages upon which mumble depends.

https://github.com/conda-forge/staged-recipes/pull/31004

Until those are properly assigned feedstocks I will keep this package as a draft.

phreed:
@conda-forge-admin retract review request

PR #29323, opened 2025-02-28
Add py-libnuma

iamthebot:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/py-libnuma/recipe.yaml
+++ b/recipes/py-libnuma/recipe.yaml
@@ -1,0 +1,44 @@
+context:
+  version: '1.2'
+
+package:
+  name: py-libnuma
+  version: ${{ version }}
+
+source:
+  - url: https://pypi.org/packages/source/p/py-libnuma/py-libnuma-${{ version }}.tar.gz
+    sha256: e9d67ad04f274015e9acac497d642ce92f8a96d1d4e7be24e39124b24320b95a
+
+build:
+  number: 0
+  script: ${{ PYTHON }} -m pip install .
+  skip:
+    - win
+    - osx
+
+requirements:
+  host:
+    - python
+    - pip
+  run:
+    - python
+    - libnuma
+
+tests:
+  - python:
+      imports:
+        - numa
+      pip_check: true
+
+about:
+  homepage: https://github.com/eedalong/pynuma
+  summary: Python3 libnuma ctypes wrapper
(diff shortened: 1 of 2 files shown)


recipes/py-libnuma/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+source:
+  - url: https://pypi.org/packages/source/p/py-libnuma/py-libnuma-${{ version }}.tar.gz
+    sha256: e9d67ad04f274015e9acac497d642ce92f8a96d1d4e7be24e39124b24320b95a
+
+build:
+  number: 0
+  script: ${{ PYTHON }} -m pip install .
+  skip:
+    - win
+    - osx
```
ocefpaf, line 17: This one should probably be OS-noarch. No idea on how to do that in the v1 recipe format though.

iamthebot: @ocefpaf we can't use libnuma on windows or OSX. Those OS' must be skipped. And `noarch: generic` cannot be used with OS selectors.

Is this a hard blocker?

ocefpaf: Please see https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

iamthebot: @ocefpaf there's two problems with that approach here it seems:

1. You can't even use this package outside of linux (it actually requires libnuma which is linux specific due to numactl). So even though the build "works" with `noarch: python` + the virtual package `__linux` in a selector conda-forge CI will still try to build+test on windows and OSX (and that will not work when we go to do an import test)
2. The linter does not allow selectors on noarch packages.

How do you suggest proceeding?

recipes/py-libnuma/recipe.yaml
```diff
@@ -0,0 +1,48 @@
...
+  host:
+    - python
+    - pip
+    - setuptools
+    - wheel
+  run:
+    - python
+    - if: linux
+      then:
+        - __linux
```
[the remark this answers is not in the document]
wolfv: To create two packages, you should add a `conda_build_config.yaml` that has an entry that says:

```yaml
os_variant:
- linux
- not_linux
```

And then make sure to use this variant in a condition. So change `if: linux` to `if: os_variant == "linux"  ...` - to make sure that two variants are being built.

PR #29342, opened 2025-03-03
add `ragatouille`

moritzwilksch:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/ragatouille/recipe.yaml
+++ b/recipes/ragatouille/recipe.yaml
@@ -1,6 +1,5 @@
 context:
   version: 0.0.9
-  python_min: 3.9
 
 package:
   name: ragatouille
@@ -17,7 +16,7 @@ build:
 
 requirements:
   host:
-    - python >=${{ python_min }}
+    - python ${{ python_min }}.*
     - pip
     - setuptools
   run:
@@ -36,17 +35,17 @@ requirements:
 
 tests:
   - python:
+      python_version: ${{ python_min }}
       imports:
         - ragatouille
-      pip_check: true
+      pip_check: false
 
 about:
-  homepage: https://github.com/bclavie/RAGatouille
-  documentation: https://github.com/bclavie/RAGatouille
-  summary: Library to facilitate the use of state-of-the-art retrieval models in common RAG contexts.
+  homepage: https://github.com/AnswerDotAI/RAGatouille
+  summary: Easily use and train state of the art late-interaction retrieval methods (ColBERT) in any RAG pipeline. Designed for modularity and ease-of-use, backed by research.
   license: Apache-2.0
   license_file: LICENSE
 
 extra:
   recipe-maintainers:
-    - GITHUB_USERNAME
+    - moritzwilksch
--- a/recipes/ragatouille/recipe.yaml
+++ b/recipes/ragatouille/recipe.yaml
@@ -1,0 +1,52 @@
+context:
+  version: 0.0.9
+  python_min: 3.9
+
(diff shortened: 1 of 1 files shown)


recipes/ragatouille/recipe.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - pytorch >=1.13
+    - fast-pytorch-kmeans
+    - sentence-transformers
+
+tests:
+  - python:
+      python_version: ${{ python_min }}
+      imports:
+        - ragatouille
+      pip_check: false
```
ocefpaf, line 41: Why? This will force all downstream packages to disable it too. If there is a problem that we can fix, let's fix it. If not, we should document the problem and open an  issue.

moritzwilksch: True. The original reason was:

```
 │ │ colbert-ai 0.2.21 requires git-python, which is not installed.
 │ │ colbert-ai 0.2.21 requires ninja, which is not installed.
```

which errors even though I provided them as run dependencies in the `colbert-ai-feedstock`:

https://github.com/conda-forge/colbert-ai-feedstock/blob/a76ce8c4437fdf05cf1e81f36777ead2745f2169/recipe/recipe.yaml#L27-L29

For `git-python`, I'm not sure whether `gitpython` is a drop-in replacement:

https://github.com/stanford-futuredata/ColBERT/blob/7067ef598b5011edaa1f4a731a2c269dbac864e4/setup.py#L25

But the `git-python` on pypi looks sketchy:
https://pypi.org/project/git-python/

ocefpaf: Can you open an issue in https://github.com/conda-forge/colbert-ai-feedstock/blob/main/recipe/recipe.yaml#L41 about that? Also, please add a comment here so we don't forget.

moritzwilksch: I just created that feedstock myself last week. Happy to fix that but do you have an idea how? Ninja is a run dependency and I'm not quite sure what to do about `git-python`

ocefpaf: Fix what you can and open issues/notes for the rest. The git-python vs gitpython probably need some metadata fix upstream, not only the recipe.

moritzwilksch: Alright, I fixed the issues related to `colbert-ai`:
https://github.com/conda-forge/colbert-ai-feedstock/pull/1

- gitpython is incorrectly required by upstream (PR to fix upstream [here](https://github.com/stanford-futuredata/ColBERT/pull/388), I patched it in the feedstock)
- `ninja` is required because `colbert-ai` uses `torch`'s ability to load and JIT cpp modules (see [here](https://pytorch.org/docs/stable/cpp_extension.html)) but since the `ninja` package only contains a ninja binary in `bin/ninja` it fails the pip check

Last problem is `faiss-cpu` as a direct dependency. Could this be a similar problem where the package is not recognized as a python package?

ocefpaf: > Last problem is `faiss-cpu` as a direct dependency. Could this be a similar problem where the package is not recognized as a python package?

No idea. Sorry, my packages are usually quite "vanilla pure python." If that is the case we should add a comment about it and skip pip-check for now.

monaldoj:
updated PR for `git-python` here: (https://github.com/conda-forge/staged-recipes/pull/29263). Hopefully this will help rebuild `colbert-ai` and then fix these errors

PR #29368, opened 2025-03-07
Adding endf-parserpy

tkittel:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/endf-parserpy/meta.yaml
+++ b/recipes/endf-parserpy/meta.yaml
@@ -22,7 +22,6 @@ requirements:
   build:
     - python                                 # [build_platform != target_platform]
     - cross-python_{{ target_platform }}     # [build_platform != target_platform]
-    - numpy                                  # [build_platform != target_platform]
     - pybind11 >=2.10.0                      # [build_platform != target_platform]
     - {{ stdlib('c') }}
     - {{ compiler('c') }}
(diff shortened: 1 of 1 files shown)


tkittel:
@gschnabel I have added you as recipe maintainer. For that to work, you need to add a comment here in this PR to confirm that you are willing to be listed there.

tkittel:
Because until you do, the CI fails with:

```
File-specific lints and/or hints:
21

22
- `recipes/endf-parserpy/meta.yaml`:
23
  - lints:
24
    - The following maintainers have not yet confirmed that they are willing to be listed here: gschnabel. Please ask them to comment on this PR if they are.
```

gschnabel:
The build logs look good. Regarding the installation procedure, please introduce the environment variable `INSTALL_ENDF_PARSERPY_CPP=yes` so that pip install sees it. This will trigger an error if the C++ module compilation fails, otherwise the installation procedure just silently continues without the compiled module.

gschnabel:

tkittel:
Ok, thanks will try to add the env var.

tkittel:
Btw @gschnabel abel ., is the upper bound on setuptools actually needed:

```
setuptools >=58.0.0,<69.0.0
```

tkittel:
Also @gschnabel could you also verify that the various check marks in the description above are checked correctly? I.e. there is no vendering of software, etc.

gschnabel:
> Btw @gschnabel, is the upper bound on setuptools actually needed:
> 
> ```
> setuptools >=58.0.0,<69.0.0
> ```

I believe there was a reason to put it (Maybe I don't remember correctly but I believe that I relied on some poetry-core version to use the undocumented generate-setup-file option and this poetry-core version wasn't the newest one and didn't support setuptools larger than a specific version). Therefore, I would be hesitant to remove it if not strictly necessary without testing the proper working.

tkittel:
Ok, thanks for the answers @gschnabel ! Let us leave the pinning in place then, but if you remove it downstream one day we should probably also remove it here. I guess a build-time pinning is anyway not as harmful as a runtime pinning :-)

So as far as I am concerned, this is now ready for review :-)

xhochy (approved):

PR #29374, opened 2025-03-08
NEW: Add recipe for npp-plus

carterbox:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/npp+/meta.yaml
+++ b/recipes/npp+/meta.yaml
@@ -1,0 +1,154 @@
+{% set version = "0.9.0.0" %}
+
+package:
+  name: libnpp-plus-split
+  version: {{ version.split(".")[:3] | join(".") }}
+
+{% set arm_variant_type = arm_variant_type | default("None") %}
+{% set cuda_compiler_version = cuda_compiler_version | default("None") %}
+
+source:
+  - url: "https://developer.download.nvidia.com/compute/nppplus/redist/libnpp_plus/linux-x86_64/libnpp_plus-linux-x86_64-0.9.0.0_cuda11-archive.tar.xz"  # [linux and x86_64 and (cuda_compiler_version or "").startswith("11")]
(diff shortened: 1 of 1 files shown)


recipes/npp+/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+      license_file: LICENSE
+      description: >-
+        This is a runtime package only. Developers should install libnpp-plus-dev to build
+        with NPP+.
+
+  - name: libnpp-plus-dev
+    build:
+      run_exports:
+        - {{ pin_subpackage("libnpp-plus", max_pin=None) }}
+        - {{ pin_subpackage("libnpp-plus" ~ version.split(".")[0], max_pin=None) }}
```
carterbox, line 94: What is ABI compatibility policy for this library? I assume it uses libtool versioning for the library version. i.e. libraries with the same soname are ABI compatible.

recipes/npp+/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+    files:
+      - lib/stubs/*.so                                  # [linux]
```
carterbox, line 96: Are stubs needed?

recipes/npp+/meta.yaml
```diff
@@ -0,0 +1,157 @@
...
+      - lib/libnpp_plus_*.so                            # [linux]
+      - include/nppPlus                                    # [linux]
```
carterbox, line 98: I didn't include npp headers because there is a separate npp package. Is that correct or npp+ needs its own copy of the npp headers?

recipes/npp+/meta.yaml
jakirkham: Yeah this makes sense for the dev library and we already do this elsewhere. Some builds may look for them

recipes/npp+/meta.yaml
jakirkham: As noted offline, the `0.` is indicating this is unstable

So think we need to pin tightly for now

PR #29389, opened 2025-03-10
Add r-fivethirtyeightdata

IsabelParedes:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/r-fivethirtyeightdata/build.bat
+++ b/recipes/r-fivethirtyeightdata/build.bat
@@ -1,0 +1,2 @@
+"%R%" CMD INSTALL --build .
+IF %ERRORLEVEL% NEQ 0 exit 1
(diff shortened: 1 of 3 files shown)


IsabelParedes:
@conda-forge/help-r, ready for review. Thanks!

mfansler (changes requested):
Thank you for submitting this!

Unfortunately, I do not support adding this to Conda Forge for the following reasons:

- the package is not (and cannot be) on CRAN, and as such it lacks regular upstream checks that it maintains consistency with R version releases;
- there has been no active development on this package since its creation (5 years ago), whereas the upstream data source continues to actively release data;
- the audience is unclear and I find no obvious data to indicate the level of demand;
- the license reporting in the R package seems incorrect, specifically the license on the data from the source is CC-BY-4.0 but this package only reports an MIT license, which technically violates CC-BY-4.0.

I would also note that this is a non-compiled package that only delivers data, so using `install.packages` as the `r-fivethirtyeight` package suggests should work just fine. If a Conda package is entirely necessary, individuals can host Conda packages on personal Anaconda Cloud accounts for free.

All that said, I'm not strongly opposed to this, so if another reviewer from @conda-forge/r wants to let it through, I won't object. However, the license issue should be addressed and the recipe should be changed to `noarch`.

PR #29392, opened 2025-03-11
Add pyschism recipe

mansurjisan:
Checklist
- [x] Build number is 0.


mansurjisan:

recipes/pyschism/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+sed -i -e "7s|.*|# |g" setup.py
+sed -i -e "8s|.*|# |g" setup.py
+sed -i -e "10s|.*|# |g" setup.py
```
ocefpaf, line 4: Why are these necessary? Can you do that change via a patch?

recipes/pyschism/meta.yaml
```diff
@@ -0,0 +1,80 @@
+{% set name = "pyschism" %}
+{% set version = "0.1.15" %}
+{% set python_min = "3.9" %}
```
ocefpaf, line 3:
```suggestion
```

recipes/pyschism/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pyschism/pyschism-{{ version }}.tar.gz
+  sha256: 42bed698d21bd8e24427a8ffba9148f74519da02ec35e1e0e463feacb47d5bac
+
+build:
+  skip: True  # [win]
```
ocefpaf, line 14: Looks like this one can be noarch.

It is OK for noarch package to fails on Windows with a missing dependency. We prefer noarch b/c it is faster to 
build and the package will be ready when the missing dependency is available on Windows.

recipes/pyschism/run_test.py
```diff
@@ -0,0 +1,16 @@
+# run_test.py
+import sys
+import subprocess
+import os
+
+
+subprocess.check_call([sys.executable, "-m", "pip", "install", "typepigeon==1.0.5"])
+subprocess.check_call([sys.executable, "-m", "pip", "install", "wget"])
```
ocefpaf, line 8: You should not pip install packages in a conda environment. Wget is python-wget, you may need to add this older version if that one is really needed.

recipes/pyschism/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
+
```
ocefpaf, line 1: This file is in the tarball. You don't need to add it twice.

recipes/pyschism/meta.yaml
mansurjisan: Thank you for the suggestion! 
I have updated the recipe to use noarch: python and ensured that the dependencies are properly specified.

Since some dependencies are missing on Windows, I'll proceed with noarch, as you mentioned, so the package will be available once those dependencies become available.

Let me know if there's anything else I should adjust!

recipes/pyschism/run_test.py
mansurjisan: Thank you for your feedback. I've updated the recipe to use proper conda dependencies instead of pip installations. 

I've removed the `run_test.py` file that was using pip to install packages, and added typepigeon and python-wget directly as conda dependencies in the run section.

You're right about wget - I've changed it to use the python-wget package from conda-forge. The package does require these dependencies as they're imported directly in the code.

recipes/pyschism/build.sh
mansurjisan: Thank you for the suggestion. You're right that using a patch file is the better approach. I've replaced the sed commands in build.sh with a proper patch file (setup.patch) that comments out the problematic lines in setup.py where it tries to run pip install commands.

recipes/pyschism/LICENSE.txt
mansurjisan: Thank you for pointing that out. You're right - since LICENSE.txt is already included in the source tarball, I don't need to package it separately with the recipe. I've removed it from the recipe files and kept the reference in meta.yaml with just the `license_file: LICENSE.txt` line to indicate where the license is located within the source distribution

recipes/pyschism/meta.yaml
mansurjisan: I tried removing the python_min variable as suggested, but conda-smithy's linter requires that noarch Python packages have a minimum Python version specified.

ocefpaf: Sure, but you don't need this. Please remove it. Let's check the linter message then and fix it accoendingly.

recipes/pyschism/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools >=41.2
+    - setuptools_scm
+    - dunamai
+    - netcdf-flattener >=1.2.0
+
```
ocefpaf, line 29:
```suggestion
```

recipes/pyschism/meta.yaml
```diff
@@ -12,23 +11,21 @@ source:
...
-
 build:
   noarch: python
   number: 0
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
 
 requirements:
   host:
-    - python {{ python_min }}
+    - python >=3.9
```
ocefpaf, line 21: This should stay the same. That variable is defined for you. Please follow only the suggested changes.

drf5n:
You could drop the dependency on pygeos (and therefore shapely) since pyschism doesn't actually use it.

Also, if you set a minimum version on cf-python, it might ease some of the other restrictions.  cf-python>=3.18 worked for me.

PR #29394, opened 2025-03-11
Add FeynHiggs

matthewfeickert:
* Add FeynHiggs ~v2.16.1~ v2.19.0. (given [ATLAS needs](https://gitlab.cern.ch/jwuerzin/Run3ModelGen/-/blob/4ea5195b3ad51081fb7357fa7186529069d01e56/README.md#software-content))
   - c.f. http://www.feynhiggs.de/
* Add static output as stable, given issues with build for shared library, and then adding a manual approach for the shared library.
* Using https://github.com/hep-packaging-coordination/feynhiggs-release-mirror as official source deletes versions without warning.

Checklist
   - Using https://github.com/hep-packaging-coordination/feynhiggs-release-mirror as official source deletes versions without warning.
- [N/A] If static libraries are linked in, the license of the static library is packaged.
   - Using static library output, but following CFEP-18.
- [x] Build number is 0.

### DIFF
--- a/recipes/feynhiggs/recipe.yaml
+++ b/recipes/feynhiggs/recipe.yaml
@@ -17,15 +17,16 @@ source:
 
 build:
   number: 0
+  skip:
+    # FIXME: Get osx builds working
+    - not linux
 
 outputs:
-  - package:
-      name: feynhiggs-static
-
+  # Build once into a shared staging cache, then ship the static archive and the
+  # shared library as two separate packages (FeynHiggs upstream is static only).
+  - staging:
+      name: feynhiggs-build
     build:
-      skip:
-        # FIXME: Get osx builds working
-        - not linux
       script:
         # c.f. https://conda-forge.org/docs/maintainer/knowledge_base/#cross-compilation-examples
         # Get an updated config.sub and config.guess
@@ -35,12 +36,19 @@ outputs:
         - if: linux and (x86_64 or ppc64le)
           then: ./configure --prefix=$PREFIX --enable-full-g-2 --enable-slhapara
           else: ./configure --prefix=$PREFIX --enable-full-g-2 --enable-slhapara CONF_QUAD=0
+        # Build position-independent objects: configure hard-codes -fPIE for the
+        # Fortran sources, which cannot go into a shared library. Forcing -fPIC
+        # lets the same objects feed both libFH.a and libFH.so.
+        - sed -i 's/ -fPIE / -fPIC /g' makefile
         - make --jobs="${CPU_COUNT}"
         - make install
-        - if: linux
-          then:
-            - mv $PREFIX/lib64/libFH.a $PREFIX/lib/
-            - rm -r $PREFIX/lib64
+        - mkdir -p $PREFIX/lib
+        - mv $PREFIX/lib64/libFH.a $PREFIX/lib/
+        - rm -r $PREFIX/lib64
+        # Link a shared library from the same PIC objects (upstream ships none)
+        - if: linux and (x86_64 or ppc64le)
+          then: $FC -shared -Wl,-soname,libFH.so $LDFLAGS -o $PREFIX/lib/libFH.so -Wl,--whole-archive $PREFIX/lib/libFH.a -Wl,--no-whole-archive -lstdc++ -lquadmath -lgfortran
+          else: $FC -shared -Wl,-soname,libFH.so $LDFLAGS -o $PREFIX/lib/libFH.so -Wl,--whole-archive $PREFIX/lib/libFH.a -Wl,--no-whole-archive -lstdc++ -lgfortran
 
     requirements:
       build:
@@ -54,6 +62,19 @@ outputs:
         - sed
         - pkg-config
         - gnuconfig
+
+  - package:
+      name: feynhiggs-static
+    inherit: feynhiggs-build
+    build:
+      skip:
+        - not linux
+      files:
+        - include/*.h
+        - bin/FeynHiggs
+        - bin/fcc
+        - bin/table
+        - lib/libFH.a
 
     tests:
       - package_contents:
@@ -74,6 +95,70 @@ outputs:
             - lib/libFH.a
             # tests
             - etc/conda/test-files/feynhiggs-static/0/example/*
+
+      - files:
+          source:
+            - example/SLHA/TestEXTPAR.spc
+            - example/demo.cc
+            - example/demo.F
+
+        requirements:
+          run:
+            - ${{ compiler("cxx") }}
+            - ${{ compiler("fortran") }}
+
+        script:
+          - FeynHiggs example/SLHA/TestEXTPAR.spc
+
+          - cd example
+          - echo "\nC++ demo"
+          - if: linux and (x86_64 or ppc64le)
+            then: $CXX demo.cc -o demo $CXXFLAGS $LDFLAGS -lFH -lgfortran -lquadmath
+            else: $CXX demo.cc -o demo $CXXFLAGS $LDFLAGS -lFH -lgfortran
+          - ./demo
+          - rm demo
+
+          - echo "\nFortran demo"
+          - $FC demo.F -o demo $FFLAGS $FLDFLAGS -lFH
+          - echo "\n$FC"
+          - ./demo
+
+  - package:
+      name: feynhiggs
+    inherit: feynhiggs-build
+    build:
+      skip:
+        - not linux
+      files:
+        - include/*.h
+        - bin/FeynHiggs
+        - bin/fcc
+        - bin/table
+        - lib/libFH.so
+    requirements:
+      run_exports:
+        # No upstream ABI versioning; pin to the build's minor version.
+        - ${{ pin_subpackage("feynhiggs", upper_bound="x.x") }}
+    tests:
+      - package_contents:
+          strict: true
+          lib:
+            - FH
+          include:
+            - CFeynHiggs.h
+            - CSLHA.h
+            - FHCouplings.h
+            - FHRecord.h
+            - PDG.h
+            - SLHA.h
+            - SLHADefs.h
+          bin:
+            - FeynHiggs
+            - fcc
+            - table
+          files:
+            # tests
+            - etc/conda/test-files/feynhiggs/0/example/*
 
       - files:
           source:
--- a/recipes/feynhiggs/recipe.yaml
+++ b/recipes/feynhiggs/recipe.yaml
@@ -1,0 +1,119 @@
+context:
+  version: "2.19.0"
+
+recipe:
+  name: feynhiggs
+  version: ${{ version }}
+
+source:
+  url: https://github.com/hep-packaging-coordination/feynhiggs-release-mirror/archive/refs/tags/${{ version }}.tar.gz
+  sha256: b0bb1beadb34ea58dbf4bcd12c4e399b1bd917d76abce00551034db39bc719bf
+  patches:
+    - henning-bahl-compliation-patches.patch
+    # Move malloc()/mkdtemp()/pipe()/fork() out of assert() so they are not
+    # skipped when NDEBUG is defined (conda-forge sets -DNDEBUG). Otherwise, the
+    # LoopTools cache node is never allocated -> NULL deref -> runtime SIGSEGV.
+    - fix-side-effects-in-assert.patch
+
+build:
+  number: 0
+
+outputs:
+  - package:
+      name: feynhiggs-static
+
+    build:
+      skip:
+        # FIXME: Get osx builds working
+        - not linux
+      script:
+        # c.f. https://conda-forge.org/docs/maintainer/knowledge_base/#cross-compilation-examples
+        # Get an updated config.sub and config.guess
+        - cp $BUILD_PREFIX/share/gnuconfig/config.* .
+        - ./configure --help
+        # configure appends -ffixed-line-length-none -std=legacy -fallow-argument-mismatch
+        - if: linux and (x86_64 or ppc64le)
+          then: ./configure --prefix=$PREFIX --enable-full-g-2 --enable-slhapara
+          else: ./configure --prefix=$PREFIX --enable-full-g-2 --enable-slhapara CONF_QUAD=0
+        - make --jobs="${CPU_COUNT}"
+        - make install
+        - if: linux
+          then:
+            - mv $PREFIX/lib64/libFH.a $PREFIX/lib/
+            - rm -r $PREFIX/lib64
+
+    requirements:
+      build:
+        - ${{ stdlib("c") }}
+        - ${{ compiler("cxx") }}
+        - ${{ compiler("fortran") }}
+        - automake
+        - libtool
+        - make
+        - gawk
+        - sed
+        - pkg-config
+        - gnuconfig
+
+    tests:
+      - package_contents:
+          strict: true
+          include:
+            - CFeynHiggs.h
+            - CSLHA.h
+            - FHCouplings.h
+            - FHRecord.h
+            - PDG.h
+            - SLHA.h
+            - SLHADefs.h
+          bin:
+            - FeynHiggs
+            - fcc
+            - table
+          files:
+            - lib/libFH.a
+            # tests
+            - etc/conda/test-files/feynhiggs-static/0/example/*
+
+      - files:
+          source:
+            - example/SLHA/TestEXTPAR.spc
+            - example/demo.cc
+            - example/demo.F
+
+        requirements:
+          run:
+            - ${{ compiler("cxx") }}
+            - ${{ compiler("fortran") }}
+
+        script:
+          - FeynHiggs example/SLHA/TestEXTPAR.spc
+
+          - cd example
+          - echo "\nC++ demo"
+          - if: linux and (x86_64 or ppc64le)
+            then: $CXX demo.cc -o demo $CXXFLAGS $LDFLAGS -lFH -lgfortran -lquadmath
+            else: $CXX demo.cc -o demo $CXXFLAGS $LDFLAGS -lFH -lgfortran
+          - ./demo
+          - rm demo
+
+          - echo "\nFortran demo"
+          - $FC demo.F -o demo $FFLAGS $FLDFLAGS -lFH
+          - echo "\n$FC"
+          - ./demo
+
+about:
+  homepage: http://www.feynhiggs.de/
+  summary: 'FeynHiggs'
+  description: |
+    FeynHiggs is a Fortran code for the (diagrammatic/EFT/hybrid) calculation
+    of the masses, mixings and much more of the Higgs bosons in the MSSM with
+    real/complex parameters at the highest level of accuracy.
+  license: GPL-3.0-or-later
+  license_file: COPYING
+  documentation: http://www.feynhiggs.de/
+
+extra:
+  feedstock-name: feynhiggs
+  recipe-maintainers:
+    - matthewfeickert
--- a/recipes/feynhiggs/fix-side-effects-in-assert.patch
+++ b/recipes/feynhiggs/fix-side-effects-in-assert.patch
@@ -1,0 +1,58 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Matthew Feickert <matthew.feickert@cern.ch>
+Date: Wed, 10 Jun 2026 17:30:00 +0000
+Subject: [PATCH] fix: Do not put side effects inside assert()
+
+The LoopTools integral cache (src/LT/cache.c) and the external
+self-energy helper (src/Util/extself.c) place calls with required side
+effects -- malloc(), mkdtemp(), pipe(), fork() -- *inside* assert().
+
+When the C preprocessor macro NDEBUG is defined (which conda-forge does
+by default via CPPFLAGS for release builds), assert(expr) expands to a
+no-op and `expr` is never evaluated. The malloc() in cache.c is then
+never called, leaving `node` NULL, and the immediately following
+`node->succ = NULL;` dereferences a NULL pointer -> SIGSEGV at runtime
+(e.g. during FHHiggsCorr loop-integral evaluation).
+
+This is why FeynHiggs built with distribution toolchains (which do not
+define NDEBUG) runs fine, while a conda-forge build segfaults despite an
+otherwise identical configuration.
+
+Move the side-effecting calls out of assert() so they always execute;
+assert() is kept purely as a success check. Verified to produce
+bit-identical Higgs masses to a build with assert active.
+---
+ src/LT/cache.c       | 3 ++-
+ src/Util/extself.c   | 9 ++++++---
+ 2 files changed, 8 insertions(+), 4 deletions(-)
+
+--- a/src/LT/cache.c
++++ b/src/LT/cache.c
+@@ -129,7 +129,8 @@
+ 	   an integer index into base */
+     size_t mem = sizeof(Node) +
+       npara*sizeof(RealType) + nval*sizeof(ComplexType);
+-    assert( (node = malloc(mem)) );
++    node = malloc(mem);
++    assert( node );
+     node = (Node *)((char *)node +
+       (PtrDiff(base, &node->para[npara]) & (sizeof(ComplexType) - 1)));
+     node->succ = NULL;
+--- a/src/Util/extself.c
++++ b/src/Util/extself.c
+@@ -114,9 +114,12 @@
+     strcat(tmpdir_template, "/fh-extse.XXXXXX");
+   }
+   strcpy(tmpdir, tmpdir_template);
+-  assert( mkdtemp(tmpdir) != NULL &&
+-          pipe(fd) != -1 &&
+-          (pid = fork()) != -1 );
++  char *tmpdir_ok = mkdtemp(tmpdir);
++  int pipe_ok = pipe(fd);
++  pid = fork();
++  assert( tmpdir_ok != NULL &&
++          pipe_ok != -1 &&
++          pid != -1 );
+ 
+   if( pid == 0 ) {
+     close(0);
--- a/recipes/feynhiggs/henning-bahl-compliation-patches.patch
+++ b/recipes/feynhiggs/henning-bahl-compliation-patches.patch
@@ -1,0 +1,306 @@
+From 1103335f8d13f29b694a8ccbec1ca5c8396ff146 Mon Sep 17 00:00:00 2001
+From: Matthew Feickert <matthew.feickert@cern.ch>
+Date: Sun, 1 Mar 2026 09:50:13 +0100
+Subject: [PATCH] fix: Apply Henning Bahl compliation patches
+
+* On 2026-02-27 Henning Bahl provided this series of patches for FeynHiggs
+  v2.19.0 which aim to fix compilation with modern Fortran.
+---
+ src/EFT/Resum4H1.F       |  5 +++--
+ src/EFT/beta4H.F         | 12 ++++++------
+ src/EFT/beta4H4c.F       | 13 +++++++------
+ src/EFT/beta4H4r.F       | 24 ++++++++++++------------
+ src/Util/ODESolve.F      | 19 +++----------------
+ src/Util/ODESolveRK.F    | 16 ++++++++--------
+ src/Util/ODESolveStiff.F |  8 ++++----
+ 7 files changed, 43 insertions(+), 54 deletions(-)
+
+diff --git a/src/EFT/Resum4H1.F b/src/EFT/Resum4H1.F
+index 7ca5ab2..c44b56b 100644
+--- a/src/EFT/Resum4H1.F
++++ b/src/EFT/Resum4H1.F
+@@ -110,9 +110,10 @@
+ 
+ ************************************************************************
+ 
+-	subroutine rge(dydx, y, x)
++	subroutine rge(n, x, y, dydx)
+ 	implicit none
+-	RealType dydx(*), y(*), x
++	integer n
++	RealType x, y(n), dydx(n)
+ 
+ #include "const.h"
+ 
+diff --git a/src/EFT/beta4H.F b/src/EFT/beta4H.F
+index 538c45e..de2931d 100644
+--- a/src/EFT/beta4H.F
++++ b/src/EFT/beta4H.F
+@@ -8,10 +8,10 @@
+ #include "debug.h"
+ 
+ 
+-	subroutine betaSM2L(dy, y, t, userarg)
++	subroutine betaSM2L(n, t, y, dy)
+ 	implicit none
+-	RealType dy(*), y(*), t
+-	external userarg
++	integer n
++	RealType t, y(n), dy(n)
+ 
+ #include "Resum4H.h"
+ 
+@@ -158,10 +158,10 @@
+ 
+ ************************************************************************
+ 
+-	subroutine betaSplit2L(dy, y, t, userarg)
++	subroutine betaSplit2L(n, t, y, dy)
+ 	implicit none
+-	RealType dy(*), y(*), t
+-	external userarg
++	integer n
++	RealType t, y(n), dy(n)
+ 
+ #include "Resum4H.h"
+ 
+diff --git a/src/EFT/beta4H4c.F b/src/EFT/beta4H4c.F
+index dac17f8..16c755c 100644
+--- a/src/EFT/beta4H4c.F
++++ b/src/EFT/beta4H4c.F
+@@ -1,10 +1,11 @@
+ #include "debug.h"
+ #include "types.h"
+ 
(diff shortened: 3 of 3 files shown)


matthewfeickert:
## Working build in Fedora Linux container

```console
$ docker run --rm -ti fedora:latest
# dnf install -y g++ gfortran curl make less
# curl -skLO https://wwwth.mpp.mpg.de/members/heinemey/feynhiggs/newversion/FeynHiggs-2.19.0.tar.gz
# tar -xzf FeynHiggs-2.19.0.tar.gz
# cd FeynHiggs-2.19.0
# ./configure
looking for gcc... /usr/bin/gcc
looking for g++... /usr/bin/g++
looking for fortran... /usr/bin/gfortran
extracting the Fortran libraries... ok
does gfortran append underscores... yes
does gfortran support REAL*16... yes
does gfortran need 'external' for U77 routines... no
are we big-endian... no
do we have MathLink... no
creating makefile

now you must run make

# make
# ./build/FeynHiggs example/SLHA/TestEXTPAR.spc
 ------------------------------------------------------------------------------
 FeynHiggs 2.19.0
 built on Mar 13, 2025
 H. Bahl, T. Hahn, S. Heinemeyer, W. Hollik, S. Passehr, H. Rzehak, G. Weiglein
 http://feynhiggs.de
 ------------------------------------------------------------------------------
mssmpart = 4       # full MSSM [recommended]
higgsmix = 2       # 2x2 (h0-HH) mixing in the neutral Higgs sector
p2approx = 4       # no approx., UHiggs evaluated at p^2 = 0 [recommended]
looplevel = 2      # include 2L corrections
loglevel = 3       # NNLL resummation (for large MCha,MNeu,MGlu,MSUSY)
runningMT = 1      # use running top mass (SM MSbar 2L)
botResum = 2       # resummed 1L-MB in Higgs se, 2L-MB in decays [rec.]
tlCplxApprox = 0   # all corrections evaluated in the rMSSM
 % invAlfa0 defaults to    137.03599908400000
 % GammaW defaults to    2.1179999999999999
 % GammaZ defaults to    2.4952000000000001
 % scalefactor defaults to    1.0000000000000000
 % prodSqrts defaults to    14.000000000000000
 % deltaLLL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaLLL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaLLL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR13 defaults to                (0.0000000000000000,0.0000000000000000)
FHSetPara warning: Qt = sqrt(MStop1*MStop2) assumed in EFT      &                 calculation if MGl <= sqrt(MStop1*MStop2),      &                 Qt = MGl if MGl > sqrt(MStop1*MStop2)
SbottomsDR warning: 2L Deltab corrections not available for DRbar input variables.
 FHHiggsCorr contains code by:
 P. Slavich et al. (2L rMSSM Higgs self-energies)
 H. Rzehak et al. (2L cMSSM asat Higgs self-energies)
 S. Passehr et al. (2L cMSSM atat Higgs self-energies)
FHCouplings warning: Extrapolating GammaSM(H0VV(1,3)) in MHiggs
FHCouplings warning: Extrapolating GammaSM(H0VV(1,4)) in MHiggs
FHCouplings warning: Extrapolating GammaSM(H0VV(2,3)) in MHiggs
FHCouplings warning: Extrapolating GammaSM(H0VV(2,4)) in MHiggs
FHCouplings warning: Extrapolating GammaSM(H0VV(3,3)) in MHiggs
FHCouplings warning: Extrapolating GammaSM(H0VV(3,4)) in MHiggs
 FHHiggsProd contains code by
 - SM XS for VBF, WH, ZH, ttH are taken
   from the LHC Higgs Cross Section Working Group,
   https://twiki.cern.ch/twiki/bin/view/LHCPhysics/CrossSections
   see there for an extensive list of references
 - SM bbH XS: Harlander et al. hep-ph/0304035
 - SM ggH XS: Grazzini et al.
   http://theory.fi.infn.it/grazzini/hcalculators.html
   see there for an extensive list of references
 - 2HDM charged Higgs XS: Plehn et al.
 - heavy charged Higgs XS:
   Dittmaier et al. arXiv:0906.2648
   Flechl et al. arXiv:1307.1347
 error           0  -999.00000000000000       -999.00000000000000             -999
FHHiggsProd warning: Extrapolating tHm2 in MHiggs
FHHiggsProd warning: Extrapolating tHm2lo in MHiggs
FHHiggsProd warning: Extrapolating tHm2hi in MHiggs
 error           0  -999.00000000000000        279.13452931036409             -999
```

## Failing build using conda-forge tools

```console
$ docker run --rm -ti fedora:latest
# curl -fsSL https://pixi.sh/install.sh | bash
# . /root/.bashrc
# mkdir example
# cd example
# pixi init
# pixi add cxx-compiler fortran-compiler make curl sed gawk less tar
# pixi run download
# pixi run build
# pixi shell
# FeynHiggs FeynHiggs-2.19.0/example/SLHA/TestEXTPAR.spc &> out.txt
Segmentation fault (core dumped)
# cat out.txt
 ------------------------------------------------------------------------------
 FeynHiggs 2.19.0
 built on Mar 13, 2025
 H. Bahl, T. Hahn, S. Heinemeyer, W. Hollik, S. Passehr, H. Rzehak, G. Weiglein
 http://feynhiggs.de
 ------------------------------------------------------------------------------
mssmpart = 4       # full MSSM [recommended]
higgsmix = 2       # 2x2 (h0-HH) mixing in the neutral Higgs sector
p2approx = 4       # no approx., UHiggs evaluated at p^2 = 0 [recommended]
looplevel = 2      # include 2L corrections
loglevel = 3       # NNLL resummation (for large MCha,MNeu,MGlu,MSUSY)
runningMT = 1      # use running top mass (SM MSbar 2L)
botResum = 2       # resummed 1L-MB in Higgs se, 2L-MB in decays [rec.]
tlCplxApprox = 0   # all corrections evaluated in the rMSSM
 % invAlfa0 defaults to    137.03599908400000
 % GammaW defaults to    2.1179999999999999
 % GammaZ defaults to    2.4952000000000001
 % scalefactor defaults to    1.0000000000000000
 % prodSqrts defaults to    14.000000000000000
 % deltaLLL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaLLL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaLLL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaELR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaERR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaQLL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaULR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaURR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDLR13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRL13 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR12 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR23 defaults to                (0.0000000000000000,0.0000000000000000)
 % deltaDRR13 defaults to                (0.0000000000000000,0.0000000000000000)
FHSetPara warning: Qt = sqrt(MStop1*MStop2) assumed in EFT      &                 calculation if MGl <= sqrt(MStop1*MStop2),      &                 Qt = MGl if MGl > sqrt(MStop1*MStop2)

Program received signal SIGSEGV: Segmentation fault - invalid memory reference.

Backtrace for this error:
#0  0x76608cc43dbf in ???
#1  0x57ce352c1966 in ???
#2  0x57ce352c19fc in ???
#3  0x57ce3527a065 in ???
#4  0x57ce34a7ecdb in ???
#5  0x57ce34a7ebfd in ???
#6  0x57ce347beb06 in ???
#7  0x57ce347733cc in ???
#8  0x57ce34732b5f in ???
#9  0x57ce34732070 in ???
#10  0x76608cc2d247 in ???
#11  0x76608cc2d30a in ???
#12  0x57ce347320a0 in ???
#13  0xffffffffffffffff in ???

# ldd $(command -v FeynHiggs)
	linux-vdso.so.1 (0x00007ffc42591000)
	libstdc++.so.6 => /example/.pixi/envs/default/lib/libstdc++.so.6 (0x00007b5d11a1b000)
	libgfortran.so.5 => /example/.pixi/envs/default/lib/libgfortran.so.5 (0x00007b5d11876000)
	libm.so.6 => /lib64/libm.so.6 (0x00007b5d1178e000)
	libgcc_s.so.1 => /example/.pixi/envs/default/lib/libgcc_s.so.1 (0x00007b5d1176f000)
	libquadmath.so.0 => /example/.pixi/envs/default/lib/libquadmath.so.0 (0x00007b5d11737000)
	libc.so.6 => /lib64/libc.so.6 (0x00007b5d11543000)
	libdl.so.2 => /lib64/libdl.so.2 (0x00007b5d1153e000)
	/lib64/ld-linux-x86-64.so.2 (0x00007b5d13ee4000)
# cd FeynHiggs-2.19.0/example/
# $CXX demo.cc -o demo $CXXFLAGS $LDFLAGS -lFH -lgfortran -lquadmath
/example/.pixi/envs/default/bin/../lib/gcc/x86_64-conda-linux-gnu/13.3.0/../../../../x86_64-conda-linux-gnu/bin/ld: warning: ODESolve.o: requires executable stack (because the .note.GNU-stack section is executable)
# ./demo 
 ------------------------------------------------------------------------------
 FeynHiggs 2.19.0
 built on Mar 13, 2025
 H. Bahl, T. Hahn, S. Heinemeyer, W. Hollik, S. Passehr, H. Rzehak, G. Weiglein
 http://feynhiggs.de
 ------------------------------------------------------------------------------
MCha  = 188.593	517.065
MNeu  = 169.262	204.206	256.794	83.2567
MStop = 829.823	1169.42
MSbot = 998.762	1003.17
 FHHiggsCorr contains code by:
 P. Slavich et al. (2L rMSSM Higgs self-energies)
 H. Rzehak et al. (2L cMSSM asat Higgs self-energies)
 S. Passehr et al. (2L cMSSM atat Higgs self-energies)
Mh1 = 119.738
Mh2 = 253.595
Mh3 = 250
MHp = 262.118
sin alpha_eff = (-0.28073,0)
```

where the `pixi` manifest `pixi.toml` file is

```toml
[workspace]
channels = ["conda-forge"]
name = "feynhiggs-build"
platforms = ["linux-64"]
version = "0.1.0"

[tasks]
download = """
rm -f FeynHiggs-2.19.0.tar.gz && \
rm -rf FeynHiggs-2.19.0 && \
curl -skLO https://wwwth.mpp.mpg.de/members/heinemey/feynhiggs/newversion/FeynHiggs-2.19.0.tar.gz && \
tar -xzf FeynHiggs-2.19.0.tar.gz
"""

build = """
cd FeynHiggs-2.19.0 && \
./configure --prefix=$CONDA_PREFIX --debug && \
make clean && \
make && \
make install && \
mv $CONDA_PREFIX/lib64/libFH.a $CONDA_PREFIX/lib64/
"""

test = """
FeynHiggs FeynHiggs-2.19.0/example/SLHA/TestEXTPAR.spc
"""

[tasks.start]
depends-on = ["download", "build"]


[dependencies]
cxx-compiler = ">=1.11.0,<2"
fortran-compiler = ">=1.11.0,<2"
make = ">=4.4.1,<5"
curl = ">=8.14.1,<9"
sed = ">=4.9,<5"
gawk = ">=5.3.1,<6"
less = ">=679,<680"
tar = ">=1.35,<2"
gdb = ">=16.3,<17"
```

The conda-forge compilers are fine, so I believe the issue lies with the code. I've additionally manually edited the `makefiles` so that the only relevant difference is what the compiler used is, and so this seems like a pathological issue if errors are compiler dependent.

matthewfeickert:
The FeynHiggs authors had asked me to show them examples with the debug symbols turned on, but it seems that their `configure` is broken only for conda-forge and not for `dnf` packages. :?

```console
$ docker run --rm -ti fedora:latest
# dnf install -y g++ gfortran curl make less gdb
# curl -skLO https://wwwth.mpp.mpg.de/members/heinemey/feynhiggs/newversion/FeynHiggs-2.19.0.tar.gz
# tar -xzf FeynHiggs-2.19.0.tar.gz
# cd FeynHiggs-2.19.0
# ./configure --help
./configure configures FeynHiggs, i.e. determines or guesses the
compiler and flags and writes out a makefile.

./configure understands the following options:

--prefix=DIR
	use DIR as installation directory
--host=HOST
	target host triplet, e.g. i386-pc-linux
--static
	link the executables statically
--quad
	compile with quadruple precision (ifort, gfortran 4.6+, xlf only)
--real10
	use REAL*10 instead of REAL*16 (gfortran 4.6+ only)
--debug
	compile with debug flags and without optimization
--32
	force 32-bit compilation
--64
	force 64-bit compilation
--native
	optimize code for the host machine
--enable-full-g-2
	compile in the full 2-loop corrections to (g_mu - 2)_SUSY
--enable-slhapara
	enable output of sfermion and *ino masses in the SLHA
--disable-vt100
	disable VT100 escape sequences used for colour output

# ./configure --debug
looking for gcc... /usr/bin/gcc
looking for g++... /usr/bin/g++
looking for fortran... /usr/bin/gfortran
extracting the Fortran libraries... ok
does gfortran append underscores... yes
does gfortran support REAL*16... yes
does gfortran need 'external' for U77 routines... no
are we big-endian... no
do we have MathLink... no
creating makefile

now you must run make

# make
...
gfortran -O0 -g -Wall -Wno-unused-dummy-argument -Wno-tabs -Wno-integer-division -Wno-real-q-constant -ffpe-trap=invalid,overflow,zero -ffixed-line-length-none -std=legacy -fPIE -DQUAD=0 -DQUADSIZE=16 -DU77EXT=0 -DVT100=1 -DSLHAPARA=0 -DGM2FULL=0   -c -o SLHAPDGName.o SLHAPDGName.F
...
# gdb ./build/FeynHiggs
(gdb) run example/SLHA/TestEXTPAR.spc
(gdb) bt full
No stack.
```

I'm not sure what's wrong with the `configure` that is seems unable to find things if the conda package environments are activated.

It is confusing though as the `configure` outputs are the same

```console
$ pixi run 'cd FeynHiggs-2.19.0 && ./configure --debug --prefix=$CONDA_PREFIX'
looking for gcc... /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-cc
looking for g++... /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-c++
looking for fortran... /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-gfortran
extracting the Fortran libraries... ok
does /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-gfortran append underscores... yes
does /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-gfortran support REAL*16... yes
does /work/.pixi/envs/default/bin/x86_64-conda-linux-gnu-gfortran need 'external' for U77 routines... no
are we big-endian... no
do we have MathLink... no
creating makefile

now you must run make
```

Though as @chrisburr notes, they aren't actually using `autotools`, so all bets are off for build system logic

```console
$ head -n 5 FeynHiggs-2.19.0/configure 
#! /bin/bash
# configure script for FeynHiggs
# note: has nothing to do with GNU autoconf
# last modified 10 May 21 th

```

kratsg:
Attempt here for `cmake`: https://github.com/kratsg/FeynHiggs-2.19.0/

matthewfeickert:
> Attempt here for `cmake`: https://github.com/kratsg/FeynHiggs-2.19.0/

(To update others following this)

From https://github.com/kratsg/FeynHiggs-2.19.0/pull/1, @kratsg seems to have identified a potential error in the Fortran code, which if [`-fcheck=all`](https://gcc.gnu.org/onlinedocs/gfortran/Code-Gen-Options.html#index-fcheck) is added to the `FFLAGS` (e.g. `export FFLAGS="$FFLAGS -fcheck=all"`) will be clearly raised at runtime

```
Fortran runtime error: Substring out of bounds: lower bound (0) of 'd' is less than one
```

With the CMake build approach it is seen though that if debug symbols are turned on via `-DCMAKE_BUILD_TYPE=Debug` this error goes away and things don't segfault at runtime. So it seems that this bug is only present under certain compiler optimization approaches (which ones aren't yet understood). It also isn't clear if the results under the patch that @kratsg made are correct or not (though given the presence of `FHLoopRecord error` in output, I'm assuming "no"), and the project authors have been contacted to ask for validation output, as the source tarballs give no way to validate builds.

matthewfeickert:
Jonas was able to get a working build in https://gitlab.cern.ch/jwuerzin/Run3ModelGen/-/merge_requests/11 (of just building with Pixi) but he had to unset effectively all environmental variables to get there. That doesn't seem good.

matthewfeickert:
Also interestingly enough, the working Fedora Linux container build (https://github.com/conda-forge/staged-recipes/pull/29394#issuecomment-2722881607) is now failing:

```console
$ docker run --rm -ti fedora:latest
# dnf install -y g++ gfortran curl make less
# gcc --version | head -n 1
gcc (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7)
# gfortran --version | head -n 1
GNU Fortran (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7)
# curl -skLO https://wwwth.mpp.mpg.de/members/heinemey/feynhiggs/newversion/FeynHiggs-2.19.0.tar.gz
# tar -xzf FeynHiggs-2.19.0.tar.gz
# cd FeynHiggs-2.19.0
# ./configure
looking for gcc... /usr/sbin/gcc
looking for g++... /usr/sbin/g++
looking for fortran... /usr/sbin/gfortran
extracting the Fortran libraries... ok
does gfortran append underscores... yes
does gfortran support REAL*16... yes
does gfortran need 'external' for U77 routines... no
are we big-endian... no
do we have MathLink... no
creating makefile

now you must run make

# make
...
gfortran -O3 -ffixed-line-length-none -std=legacy -fPIE -DQUAD=0 -DQUADSIZE=16 -DU77EXT=0 -DVT100=1 -DSLHAPARA=0 -DGM2FULL=0 -o FeynHiggs FeynHiggs.o pager.o libFH.a -lstdc++
/usr/sbin/ld: libFH.a(ODESolve.o): in function `fodepack.0':
ODESolve.F:(.text+0x14): undefined reference to `f_'
make[1]: *** [makefile:461: FeynHiggs] Error 1
make[1]: Leaving directory '/FeynHiggs-2.19.0/build'
make: *** [makefile:76: default] Error 2
```

matthewfeickert:
Claude Code's summary of what happened with the segfaults:

# Root cause found: the runtime segfault is a `malloc()` inside `assert()`, elided by conda-forge's `-DNDEBUG`

After a long hunt, the FeynHiggs runtime `SIGSEGV` is fully diagnosed and fixed. It is **not** the ODE solver, the stack protector, quad precision, an optimization bug, or the Henning Bahl compilation patch. It is a single classic C anti-pattern in the LoopTools integral cache that only bites when the C preprocessor macro `NDEBUG` is defined — which conda-forge does by default and most Linux distribution builds do not.

## TL;DR

`src/LT/cache.c` allocates its cache node **inside** an `assert()`:

```c
assert( (node = malloc(mem)) );   /* the ONLY allocation of the cache node */
node = (Node *)((char *)node + (PtrDiff(base, &node->para[npara]) & (sizeof(ComplexType) - 1)));
node->succ = NULL;                /* cache.c:135 — this is where it crashes */
```

conda-forge's compiler activation puts `-DNDEBUG` in `CPPFLAGS` (standard for release builds). With `NDEBUG` defined, the C standard says `assert(expr)` expands to `((void)0)` **and `expr` is not evaluated**. So `malloc()` is never called, `node` stays `NULL`, and the very next line dereferences a near-`NULL` pointer → segfault, deep inside `FHHiggsCorr` when the first loop integral is cached.

The fix moves the side-effecting calls out of `assert()` so they always run:

```c
node = malloc(mem);
assert( node );
```

`src/Util/extself.c` has the identical bug (`mkdtemp`/`pipe`/`fork` inside an `assert`), which would silently break the external self-energy feature under `NDEBUG`; it is fixed the same way.

## Why it happened

`assert()` is documented to do nothing when `NDEBUG` is defined, and crucially **its argument is not evaluated** in that case. Putting code with required side effects inside `assert()` is therefore a latent bug: the program is only correct in debug builds. This particular cache code has presumably worked for years because the people building it (and the upstream author's own builds) compile **without** `-DNDEBUG`, so the `malloc` always runs.

conda-forge, by design, compiles release packages with `-DNDEBUG` as part of its standard flag set. That single define is the entire difference between "runs fine" and "segfaults," even though the configure output, source, compiler family, and architecture flags are otherwise identical.

## Why it looked like a conda-forge-specific compiler problem (but wasn't)

This is exactly the "builds and runs fine with the system/`apt`/`dnf` toolchain, segfaults only with conda-forge" signature, which strongly *suggests* a compiler or toolchain incompatibility. It is really a portable C bug that the conda-forge flag set is simply strict enough to expose:

- **Distribution builds don't define `NDEBUG`.** FeynHiggs's own `configure`/`makefile` don't add it, and a typical `dnf`/`apt` build doesn't either. So `assert(malloc(...))` runs the `malloc`, the cache works, and there is no crash. The bug is invisible.
- **conda-forge's `CPPFLAGS` carries `-DNDEBUG`.** GNU `make`'s implicit rules append `$(CPPFLAGS)` to the C compile, so `cache.c` is compiled with `-DNDEBUG` and the `malloc` is stripped out.
- The earlier recipe already stripped `-O2` from `CPPFLAGS` to keep the `-Og` debug flags from being overridden. That was a reasonable thing to try, but it could never fix this — the trigger is `-DNDEBUG`, which is completely independent of the optimization level. The crash reproduces at `-Og`, `-O0`, and `-O2` alike.

## Why other tools/flags didn't catch it

The various debugging knobs that were tried all missed it, for understandable reasons:

- **`-fstack-protector-all` → `-fstack-protector-strong`**: harmless and unrelated. The crash is a NULL dereference, not stack-canary corruption, so changing the stack protector does nothing.
- **`-Og` / `-DCMAKE_BUILD_TYPE=Debug`**: in some earlier experiments the crash appeared to "go away" in a debug build — that is the tell-tale fingerprint of an `NDEBUG`-gated bug. A debug build typically does **not** define `NDEBUG`, so `assert` becomes active again, the `malloc` runs, and the symptom vanishes. It looked like an optimization bug; it was the opposite.
- **`-fcheck=all`**: this is a *Fortran* runtime check. It surfaced a genuine-but-unrelated Fortran substring warning and sent the investigation toward the Fortran ODE-solver code, while the actual fault is in the C cache layer that `-fcheck=all` does not instrument.
- **`gfortran` version / EFT RGE / ODE solver**: red herrings. The crash is reached from `fjcput` (`Cget.F`) → `cachecopy_` → `Lookup` (`cache.c`), i.e. the loop-integral memoization, not the Runge–Kutta/ODEPACK path.

The reason none of these helped is that the failure is invisible at the source level unless you look at the *preprocessed* C with `NDEBUG` defined: the line literally compiles down to "skip the `malloc`, then dereference the pointer you didn't get."

## How it was pinned down

1. Reproduced the segfault locally with the **exact conda-forge toolchain** (`gfortran 14.3.0`, `sysroot 2.17`, GCC 14 to match the pinning) using the recipe's precise flag manipulations.
2. The makefile `strip`s the binary on install, so the install artifact has no symbols. Rebuilt with `strip` neutralized to keep debug info, then ran under `gdb`:

   ```
   Program received signal SIGSEGV
   Lookup (...) at cache.c:135
   135       node->succ = NULL;
   ...
   node = 0x0
   #1 fjcachecopy_ ... at cache.c:167
   #2 fjcput ... at Cget.F:53
   #3 fjhgagasm_abbr1s () at hgagaSM-abbr1s.F:19
   ```

   `node = 0x0` on the line immediately after the `assert(malloc(...))` made the cause unambiguous.
3. **Control experiment:** rebuilt the *unmodified* source with `-DNDEBUG` stripped from `CPPFLAGS` (i.e. mimicking a distribution build). No segfault. This isolates `-DNDEBUG` as the sole trigger.
4. **Faithfulness check:** rebuilt the *patched* source **with** `-DNDEBUG` still defined. No segfault, and the computed Higgs masses are **bit-for-bit identical** to the non-`NDEBUG` reference (`Mh0 = 12.8797455`, `MHH = 108.415377`, `MA0 = 15.8113883`, `MHp = 80.9881433` GeV for `TestEXTPAR.spc`). The fix changes *where* `malloc` is called, never *what* is computed.

## The fix

New patch `recipes/feynhiggs/fix-side-effects-in-assert.patch`, added to the recipe's patch list. It edits two files:

- `src/LT/cache.c` — pull `malloc()` out of `assert()`:
  ```diff
  -    assert( (node = malloc(mem)) );
  +    node = malloc(mem);
  +    assert( node );
  ```
- `src/Util/extself.c` — pull `mkdtemp()`, `pipe()`, and `fork()` out of `assert()`:
  ```diff
  -  assert( mkdtemp(tmpdir) != NULL &&
  -          pipe(fd) != -1 &&
  -          (pid = fork()) != -1 );
  +  char *tmpdir_ok = mkdtemp(tmpdir);
  +  int pipe_ok = pipe(fd);
  +  pid = fork();
  +  assert( tmpdir_ok != NULL &&
  +          pipe_ok != -1 &&
  +          pid != -1 );
  ```

`assert()` is retained purely as a success check, so the code is now correct regardless of `NDEBUG`. This is the proper source-level fix; it is preferable to merely stripping `-DNDEBUG` from the build flags, which would leave the latent bug in place for any future release/`NDEBUG` build.

## Verification

The full build the maintainer uses now passes end to end:

```
docker system prune -f && rm -rf build_artifacts/* && pixi run build-linux linux64
```

- `✔ all tests passed!` — no segfault.
- The `FeynHiggs example/SLHA/TestEXTPAR.spc` run completes and writes a full SLHA output block.
- Both the C++ and Fortran demos run and report a sensible light Higgs mass `Mh1 = 119.738 GeV`.
- Package produced: `feynhiggs-static-2.19.0-h3e0115e_0.conda`.

## Notes for posterity / upstream

- There is one more instance of the same pattern in `web/unescape.c` (`assert(*var = realloc(...))`, `assert(*cur++ = strdup(s))`). That file is the CGI web frontend and is **not** compiled by this recipe, so it is intentionally left untouched here.
- This is a genuine upstream bug, not a packaging quirk: it will affect *any* release/`NDEBUG` build of FeynHiggs, not just conda-forge. It is worth reporting to the upstream authors (Thomas Hahn et al.) so the next release ships the side effects outside `assert()`.

matthewfeickert:

matthewfeickert:
cc @chrisburr (for potential review :grin:) @jwuerzinger (for potential interest)

recipes/feynhiggs/recipe.yaml
```diff
@@ -0,0 +1,204 @@
...
+        - sed -i 's/ -fPIE / -fPIC /g' makefile
+        - make --jobs="${CPU_COUNT}"
+        - make install
+        - mkdir -p $PREFIX/lib
+        - mv $PREFIX/lib64/libFH.a $PREFIX/lib/
+        - rm -r $PREFIX/lib64
+        # Link a shared library from the same PIC objects (upstream ships none)
+        - if: linux and (x86_64 or ppc64le)
+          then: $FC -shared -Wl,-soname,libFH.so $LDFLAGS -o $PREFIX/lib/libFH.so -Wl,--whole-archive $PREFIX/lib/libFH.a -Wl,--no-whole-archive -lstdc++ -lquadmath -lgfortran
+          else: $FC -shared -Wl,-soname,libFH.so $LDFLAGS -o $PREFIX/lib/libFH.so -Wl,--whole-archive $PREFIX/lib/libFH.a -Wl,--no-whole-archive -lstdc++ -lgfortran
```
matthewfeickert, line 51: Manual approach of getting a dynamically linked library out given that the upstream does not support this.

recipes/feynhiggs/recipe.yaml
```diff
@@ -0,0 +1,204 @@
...
+        # Fortran sources, which cannot go into a shared library. Forcing -fPIC
+        # lets the same objects feed both libFH.a and libFH.so.
+        - sed -i 's/ -fPIE / -fPIC /g' makefile
+        - make --jobs="${CPU_COUNT}"
+        - make install
+        - mkdir -p $PREFIX/lib
+        - mv $PREFIX/lib64/libFH.a $PREFIX/lib/
+        - rm -r $PREFIX/lib64
+        # Link a shared library from the same PIC objects (upstream ships none)
+        - if: linux and (x86_64 or ppc64le)
```
chrisburr, line 49: Why `x86_64 or ppc64le`?

matthewfeickert: @chrisburr these are the only platforms that support `libquadmath` as quad precision is available natively on `linux-aarch64` through `long double`.

chrisburr: THere is an else, sorry I can't read..

chrisburr (approved):

PR #29400, opened 2025-03-11
Dead batteries packages: standard-aifc, standard-sunau, standard-chunk, and audioop-lts

bmcfee:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/standard-aifc/meta.yaml
+++ b/recipes/standard-aifc/meta.yaml
@@ -1,0 +1,40 @@
+{% set name = "standard-aifc" %}
+{% set version = "3.13.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/standard_aifc-{{ version }}.tar.gz
+  sha256: 64e249c7cb4b3daf2fdba4e95721f811bde8bdfc43ad9f936589b7bb2fae2e43
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python
+    - setuptools >=61.0
+    - pip
+  run:
+    - python
+
+test:
+  imports:
+    - aifc
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  summary: Standard library aifc redistribution. "dead battery".
+  license: BSD-4-Clause
+  license_file: LICENSE
+  home: https://github.com/youknowone/python-deadlib
+
+extra:
+  recipe-maintainers:
+    - bmcfee
--- a/recipes/standard-sunau/meta.yaml
+++ b/recipes/standard-sunau/meta.yaml
@@ -1,0 +1,40 @@
+{% set name = "standard-sunau" %}
+{% set version = "3.13.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/standard_sunau-{{ version }}.tar.gz
+  sha256: b319a1ac95a09a2378a8442f403c66f4fd4b36616d6df6ae82b8e536ee790908
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python
+    - setuptools >=75.0
+    - pip
+  run:
+    - python
+
+test:
+  imports:
+    - sunau
+  commands:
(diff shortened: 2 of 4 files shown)


bmcfee:
@conda-forge/help-python :wave: I could use some input on how this particular set of packages should work in conda-forge.

Summary of the problem:

- This PR includes four "dead batteries" packages that were removed from the standard library in python 3.13, all relating to audio processing.  (aifc, sunau, audioop, and chunk).  These packages are now hosted externally on pypi to provide continued support for downstream libraries, some of which exist in the conda-forge space already (audioread and librosa, notably).
- I've stubbed out the package meta.yaml's according to grayskull and smithy linting, but I suspect the default recommendations are not going to be correct here.

The major problem that I'm anticipating is that these packages collide with standard namespace packages in python <=3.12.  (That is, `standard-aifc` provides the `aifc` package, and so on.)  On PyPI, this is skirted around by adding a `python>='3.13'` constraint on the requirements so that these packages are not installed on older pythons.

To my understanding, we *can* do this in conda-forge as well, but not with noarch packages.  This effectively means that any downstream package that depends on these would have to stop using noarch because we can't use selectors in the requirements spec.

Any input on how to resolve this situation would be much appreciated.

recipes/standard-aifc/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/standard_aifc-{{ version }}.tar.gz
+  sha256: 64e249c7cb4b3daf2fdba4e95721f811bde8bdfc43ad9f936589b7bb2fae2e43
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
ocefpaf, line 13: This one can probably be noarch, right?

recipes/standard-sunau/meta.yaml
```diff
@@ -0,0 +1,40 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/standard_sunau-{{ version }}.tar.gz
+  sha256: b319a1ac95a09a2378a8442f403c66f4fd4b36616d6df6ae82b8e536ee790908
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
ocefpaf, line 13: Maybe this one can be noarch too.

recipes/standard-aifc/meta.yaml
ocefpaf: >  This effectively means that any downstream package that depends on these would have to stop using noarch because we can't use selectors in the requirements spec.


Ah, I misread. OK. If this module is done "right" you probably can install it without clobbering, it will become a stub. If not, you will have to keep this as-is.

bmcfee: > If this module is done "right" you probably can install it without clobbering

To clarify, do you mean the implementation of the module, or the packaging of the module?

My understanding is that the module code itself is a direct extraction from old standard lib code, and doesn't have the smarts to stub itself out.  At the same time, I don't think the deadlib maintainers are open to contributions that would add the stub logic, since they want to keep the code essentially as archival.

bmcfee: > Giving the fact that is clobbers existing files?

It's not clobbering the existing files, but "correct" behavior (ie using the standard lib implementation instead of the external package) depends on the module import search order.  While this is (to my knowledge) well-defined, it does make me a little nervous.



> It is better too sacrifice noarch advantages here.

For these packages, that's fine.  I'm more concerned about the downstream packages that may want to stay noarch and have these deadlibs as dependencies if needed.  This is where I'm stuck, and I don't see a simple path forward with the usual packaging pipeline.

I do wonder if it's possible to do something clever here:  what if these deadlibs are not built as noarch, and we create stub packages for python <=3.12 that do not install anything?  For 3.13+ the packages would install as we currently have them in this PR.

This would make it so that the downstream dependents could still be noarch and depend on these libs, but installing, say, `standard-aifc` in a 3.12 environment would act as a no-op.

Is such a thing possible?

ocefpaf: > While this is (to my knowledge) well-defined, it does make me a little nervous.

Same.

> This is where I'm stuck, and I don't see a simple path forward with the usual packaging pipeline.

Yeah, better this one "arched" and many other noarch.

> Is such a thing possible?

Probably. Not sure if it is worth it though. IMO, we should fix the "semi-arch" concept so we can do things like this as noarch moving on. However, that is a fix on recipe format and conda itself.

PR #29403, opened 2025-03-11
Add NVIDIA TransformerEngine

pstjohn:
Checklist
- [x] Build number is 0.


recipes/transformer-engine/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+  name: transformer_engine
+  version: {{ version }}
+
+source:
+  - url: https://github.com/NVIDIA/TransformerEngine/archive/refs/tags/v{{ version }}.tar.gz
+    sha256: 90dd8f694cba29707c502a4f4d12871cb9d4c1f157d329f22cb3d382de8468ae
+
+  - url: https://github.com/NVIDIA/cudnn-frontend/archive/refs/tags/v1.8.0.tar.gz
+    sha256: 5907bf8baa8f3b74ad430d54027aadd14a0a61166ae4b00c2daece0df384409e
+    folder: 3rdparty/cudnn-frontend
```
weiji14, line 20: Is it ok to vendor `cudnn-frontend` here? I've packaged this at https://github.com/conda-forge/python-cudnn-frontend-feedstock and have been trying to get it to work (unsuccessfully) at https://github.com/conda-forge/staged-recipes/pull/26296#issuecomment-2113409609. But this solution looks much simpler :slightly_smiling_face:

pstjohn: Yeah that's a good question. In conversations with the TE team, they work pretty closely with cudnn-frontend and sometimes have odd tags or versions here. I also don't think the repo is set up to be able to pull these dependencies from a source outside `3rdparty`. That might be something we could try to fix in subsequent versions though. `cudnn-frontend` is a header-only library though, so it should be fairly lightweight.

jakirkham:
cc @gforsyth (for review)

recipes/transformer-engine/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - libcusparse-dev
+    - libcusolver-dev
+    - cudnn
+    - cuda-profiler-api
+    {% endif %}
+
+  run:
+    - python
+    - scipy
+    - pytorch_scatter
```
gforsyth, line 76: I don't see these as requirements upstream -- should they be here?

recipes/transformer-engine/build.sh
```diff
@@ -0,0 +1,37 @@
...
+
+set -euxo pipefail
+
+mkdir -p $PREFIX/etc/conda/activate.d
+cp $RECIPE_DIR/transformer-engine-env.sh $PREFIX/etc/conda/activate.d/
+
+if [[ ${cuda_compiler_version} != "None" && "$target_platform" == linux-64 ]]; then
+    export FORCE_CUDA="1"
+    if [[ ${cuda_compiler_version} == 12.6 ]]; then
+        export TORCH_CUDA_ARCH_LIST="5.0;6.0;6.1;7.0;7.5;8.0;8.6;8.9;9.0+PTX"
```
gforsyth, line 12: And is this env-var used by TransformerEngine?  Or is this for `pytorch`?

gforsyth (commented):
Hi, I have a few questions, largely because I'm not as familiar with this project, so apologies if the questions are off-base.

Overall I think this looks reasonable -- the build links don't appear to be valid, so I haven't been able to review the build logs, which should be looked at before this gets merged in.

Some other questions:
1. This is for version 1.13, but I see upstream there are releases in the 2.x series -- is the plan to get this version in and then also add those later versions?
2. In the `setup.py` upstream, there is support for `jax` and `paddle`, in addition to `pytorch`.  I think it's fine to keep things limited to one of those to start with, but it is worth considering if you want to rename the output to reflect that.  It might be a little odd if there is (in the future) an explicit `transformer_engine_jax`, but `transformer_engine` gets you the `pytorch` version

recipes/transformer-engine/meta.yaml
pstjohn: whoops, probably leftover from another recipe. Removed!

recipes/transformer-engine/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - cuda-cudart-dev
+    - cuda-nvrtc-dev
+    - cuda-nvtx-dev
+    - cuda-nvml-dev
+    - libcublas-dev
+    - libcusparse-dev
+    - libcusolver-dev
+    - cudnn
+    - cuda-profiler-api
+    {% endif %}
```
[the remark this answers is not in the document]
pstjohn: Yup, thanks -- cleaned up some of this logic

recipes/transformer-engine/build.sh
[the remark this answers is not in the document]
pstjohn: i actually don't think we need this... unless I'm mistaken for some reason, this might have been leftover from the torch build recipe. Anyways, deleted and the build still seems to pass

pstjohn: Probably for pytorch; I actually suspect we could just delete this

pstjohn:
> This is for version 1.13, but I see upstream there are releases in the 2.x series -- is the plan to get this version in and then also add those later versions?

Yeah that's right. We may even need to add a 1.12 version; I think I may have overshot the oldest version folks are using. But we'd want to quickly add the 2.x series as well.

> In the setup.py upstream, there is support for jax and paddle, in addition to pytorch. I think it's fine to keep things limited to one of those to start with, but it is worth considering if you want to rename the output to reflect that. It might be a little odd if there is (in the future) an explicit transformer_engine_jax, but transformer_engine gets you the pytorch version

IIUC there's not any paddle support anymore? but I'll double check. Anyways yes; that's a good idea. We're going to hopefully clean up the packaging here to make these targets more obvious, but in python currently you use optional dependencies (i.e.`transformer_engine[pytorch]`) to build the different frontends; which isn't super obvious. It might be easier to move to a transformer_engine_core / transformer_engine_pytorch / transformer_engine_jax type naming structure. Let me check in with the TE team on how they want to handle things here

recipes/transformer-engine/meta.yaml
gforsyth: From https://conda-forge.org/docs/maintainer/guidelines/#generalities

> The source should not contain vendored code. If it does:
>    Package the vendored project separately and specify the needed dependency in the requirements section. Preferred if the vendored code is needed at runtime.
>    Allow the vendored code, but make sure the license files are included in the about.license field. Usually ok if it's only a build-time dependency (e.g. headers-only library)
```

**Comment by pstjohn** [comment]

**Comment by weiji14** [comment]
> > It might be easier to move to a transformer_engine_core / transformer_engine_pytorch / transformer_engine_jax type naming structure. Let me check in with the TE team on how they want to handle things here
> 
> This change doesn't need to be made upstream -- handling it with optional dependencies is the pattern of choice with `pip`, but `conda` doesn't have that option.

FYI, optional dependency groups might be coming to conda soon with this [CEP](https://github.com/conda/ceps/pull/111), xref https://github.com/conda/conda/issues/7502#issuecomment-2644521813. So you could also just do `transformer_engine` (core) here first, and handle the `jax`/`pytorch` extras later.

**Review by gforsyth — APPROVED** [review]
Nice work on this @pstjohn!

File: recipes/transformer-engine/001-cmake.patch [hunk 5]
```diff
@@ -0,0 +1,11 @@
+--- a/build_tools/build_ext.py
++++ b/build_tools/build_ext.py
+@@ -56,7 +56,7 @@ class CMakeExtension(setuptools.Extension):
+             cmake_path,
+             "-B",
+             build_dir,
+-            f"-DPython_EXECUTABLE={sys.executable}",
++            "-DPython_EXECUTABLE=$PYTHON",
```
jakirkham, line 8: Think we can do this in a more Pythonic way

```suggestion
+            "-DPython_EXECUTABLE={os.environ['PYTHON']}",
```

Also what happens when `sys.executable` is here?

recipes/transformer-engine/transformer-engine-env.sh
```diff
@@ -0,0 +1 @@
+export CUDA_HOME=$CONDA_PREFIX
```
jakirkham, line 1: Would check if `CUDA_HOME` is set and if so keep a copy

```suggestion
if [[ ! -z "${CUDA_HOME +x}" ]]
then
    export CUDA_HOME_BACKUP="${CUDA_HOME}"
fi
export CUDA_HOME="${CONDA_PREFIX}"
```

Then would add a separate deactivate script to restore the original value

```
if [[ ! -z "${CUDA_HOME_BACKUP+x}" ]]
then
  export CUDA_HOME="${CUDA_HOME_BACKUP}"
  unset CUDA_HOME_BACKUP
fi
```

jakirkham (commented):
Thanks Peter for the contribution! 🙏

Also thanks Gil and Wei for your reviews!

Had a couple suggestions below

recipes/transformer-engine/001-cmake.patch
gforsyth: `sys.executable` led to this failure during build:

```
      CMake Error at /home/conda/staged-recipes/build_artifacts/transformer-engine_1745610886178/_build_env/share/cmake-4.0/Modules/FindPackageHandleStandardArgs.cmake:227 (message):
        Could NOT find Python (missing: Interpreter Development.Module) (found
        version "3.11.12")
```

recipes/transformer-engine/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+about:
+  home: https://github.com/NVIDIA/TransformerEngine
+  summary: A library for accelerating Transformer models on NVIDIA GPUs.
+  license: Apache-2.0
+  license_file: LICENSE
+  dev_url: https://github.com/NVIDIA/TransformerEngine
+
+extra:
+  recipe-maintainers:
+    - pstjohn
```
carterbox, line 104:
```suggestion
extra:
  feedstock-name: transformer-engine
  recipe-maintainers:
    - pstjohn
```

recipes/transformer-engine/build_te.sh
```diff
@@ -0,0 +1,21 @@
...
+chmod +x $RECIPE_DIR/gcc_shim
+export CC="$RECIPE_DIR/gcc_shim"
+
+# Create a symlink to the nvvm directory where transformer-engine's Cmake expects it.
+ln -s $PREFIX/nvvm $PREFIX/targets/x86_64-linux/nvvm
+cp -n $BUILD_PREFIX/targets/x86_64-linux/include/*.h $PREFIX/targets/x86_64-linux/include
+cp $PREFIX/include/cudnn*.h $PREFIX/targets/x86_64-linux/include
+
+echo "Installing transformer-engine"
+${PYTHON} -m pip install .
```
carterbox, line 21: Logs show that the pip package version metadata contains some kind of hash.

e.g.  transformer_engine-1.13.0+6a31486c93

Is this intentional? Or is there an override variable you can set.



https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1235681&view=logs&j=df9408f7-e795-500f-143d-aa7735a0ed8b&t=47addc80-ccbf-566b-ee3a-66cdb8436937&l=2088

recipes/transformer-engine/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+        - importlib-metadata >=1.0
+        - packaging
+        - cuda-driver-dev
+        - cuda-cudart-dev
+        - cuda-nvrtc-dev
+        - cuda-nvtx-dev
+        - cuda-nvml-dev
+        - libcublas-dev
+        - libcusparse-dev
+        - libcusolver-dev
```
carterbox, line 76: There are various unused dependencies at runtime from these CUDA libraries. Are we using dlopen or statically linking? Or are the deps actually unused?

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1235681&view=logs&j=df9408f7-e795-500f-143d-aa7735a0ed8b&t=47addc80-ccbf-566b-ee3a-66cdb8436937&l=2494

recipes/transformer-engine/meta.yaml
carterbox: Revending headers for compile-time is OK. What's not OK is revending them at runtime. This package is shipping a copy of CUDNN in $PREFIX/targets.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1235681&view=logs&j=df9408f7-e795-500f-143d-aa7735a0ed8b&t=47addc80-ccbf-566b-ee3a-66cdb8436937&l=4381

carterbox (changes requested):

recipes/transformer-engine/meta.yaml
pstjohn: Ah, yes, that is a problem. I moved some headers around so the CMAKE command could find them, but didn't realize they were getting re-distributed. Since we're patching those cmake build flags anyways I'll try to point cmake to the right place to find them in these conda builds

recipes/transformer-engine/meta.yaml
pstjohn: Yeah good question; they might only be needed at build time. Would I move these from `host` to `build` then? I'm pretty sure pydantic just wasn't exercised during the import tests, added an additional import to hopefully hit that.

recipes/transformer-engine/meta.yaml
pstjohn: Well it's not easy to get the cmake command to find these in a different location. The alternative could just be to clean these up post-build, right? Let me try doing that.

recipes/transformer-engine/meta.yaml
pstjohn: done!

recipes/transformer-engine/transformer-engine-env.sh
pstjohn: sure, done

recipes/transformer-engine/001-cmake.patch
pstjohn: > Think we can do this in a more Pythonic way
No oddly, I get the same `missing: Interpreter Development.Module` when I do `os.environ['PYTHON']`; at least locally. Not sure why these would be different?

recipes/transformer-engine/build_te.sh
pstjohn: Good catch; I think I can override this with NVTE_NO_LOCAL_VERSION. Will do
https://github.com/NVIDIA/TransformerEngine/blob/d33525249d08ee6b2c80fded1b326248a2a1cb77/build_tools/te_version.py#L15

recipes/transformer-engine/meta.yaml
carterbox: As discussed offline, we are assuming these packages are dlopen'd and will decide whether to make them optional at a later time.

recipes/transformer-engine/activate.sh
```diff
@@ -0,0 +1,5 @@
+if [[ ! -z "${CUDA_HOME +x}" ]]
```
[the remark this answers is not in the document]
jakirkham: Autocomplete kept adding spaces 🤦‍♂️ 

Thought I removed them all from the suggestion, but this was missed here

```suggestion
if [[ ! -z "${CUDA_HOME+x}" ]]
```

jakirkham:
Thanks all! 🙏

PR #29439, opened 2025-03-17
Add recipe for refrapy app

aureliobarbosa:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/refrapy/meta.yaml
+++ b/recipes/refrapy/meta.yaml
@@ -39,6 +39,7 @@ test:
     - pip check
     - refrapy
   requires:
+    - python {{ python_min }}
     - pip
 
 about:
--- a/recipes/refrapy/meta.yaml
+++ b/recipes/refrapy/meta.yaml
@@ -1,5 +1,7 @@
 {% set name = "refrapy" %}
 {% set version = "0.2.4.1" %}
+{% set python_min = "3.10" %}
+{% set python_max = "3.12" %}
 
 package:
   name: {{ name|lower }}
@@ -13,17 +15,17 @@ build:
   entry_points:
     - refrapy = refrapy.__main__:main
   noarch: python
-  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  script: {{ PYTHON }} -m pip install . -vv
   number: 0
 
 requirements:
   host:
-    - python >=3.10,<3.13
+    - python ={{python_min}},<={{python_max}}
     - hatchling
     - hatch-vcs
     - pip
   run:
-    - python >=3.10,<3.13
+    - python ={{python_min}},<={{python_max}}
     - obspy >=1.3
     - pygimli >=1.5
     - numpy >=1.19
--- a/recipes/refrapy/meta.yaml
+++ b/recipes/refrapy/meta.yaml
@@ -20,7 +20,7 @@ build:
 
 requirements:
   host:
-    - python ={{python_min}},<={{python_max}}
+    - python {{ python_min }}
     - hatchling
     - hatch-vcs
     - pip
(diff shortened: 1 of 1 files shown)


recipes/refrapy/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+requirements:
+  host:
+    - python ={{python_min}},<={{python_max}}
+    - hatchling
+    - hatch-vcs
+    - pip
+  run:
+    - python ={{python_min}},<={{python_max}}
+    - obspy >=1.3
+    - pygimli >=1.5
```
synapticarbors, line 30: It looks like `pygimli` is not available on conda-forge. You'll need to open a PR to add it.

recipes/refrapy/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+build:
+  entry_points:
+    - refrapy = refrapy.__main__:main
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
+  number: 0
+
+requirements:
+  host:
+    - python ={{python_min}},<={{python_max}}
```
synapticarbors, line 23:
```suggestion
    - python {{ python_min }}
```

recipes/refrapy/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - hatchling
+    - hatch-vcs
+    - pip
+  run:
+    - python ={{python_min}},<={{python_max}}
```
synapticarbors, line 28:
```suggestion
    - python >={{ python_min }},<={{ python_max }}
```

recipes/refrapy/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - pmw >=2.0
+
+test:
+  imports:
+    - refrapy
+  commands:
+    - pip check
+    - refrapy
+  requires:
+    - pip
```
synapticarbors, line 42:
```suggestion
    - python {{ python_min }}
    - pip
```

recipes/refrapy/meta.yaml
aureliobarbosa: Thanks for the review. 

I am not affiliated with pygimli and it is hosted on its own conda channel `gimli`, isn't there a way to import it from there? 

For installing it alone we [use](https://www.pygimli.org/installation.html)
```bash
conda create -n pg -c gimli -c conda-forge "pygimli>=1.5.0"
```
Note that it is hosted on the channel gimli but it uses packages from conda-forge. 

I now this is peculiar, but if it is not possible to import from there, I suspect it will be necessary to open an issue on their repo and check whether it is of their interest to move to conda-forge.

aureliobarbosa: I forgot to mention that pygimli is also on PyPI.

synapticarbors: `pygimli` is up on pypi, but unfortunately they don't provide a source distribution (sdist), only wheels. As far as conda-forge goes, we don't allow dependencies from outside channels, so the best way to go is to get `pygimli` up on conda-forge. I can take a shot at doing that. I can ping the maintainers of pygimli to see if they want to co-maintain it, but that's not a strict requirement.

synapticarbors: So it looks like we'd also need `pgcore` on conda-forge, so this is a bit more complicated. I'm not sure if I'll have time to get a PR up, but it may be worth it to ping the gimli developers and see if they'd be interested in getting their packages on conda-forge.

aureliobarbosa:
Hi @synapticarbors,

Thanks for your revision and attention to my submission. I took a look into the pygimli project and, in my perception, it will demand some extra work to generate recipes to include their package on conda-forge. Besides, since they opted for creating a private channel for their package, it looks counter intuitive to ask them to bring their package to conda-forge. Considering this, and also that we have only a few users on refrapy, I decided to drop this submission conda-forge. 

Thanks!

PR #29477, opened 2025-03-20
add libcamera and related recipes

carlodri:
https://libcamera.org/index.html

Checklist
- [x] Build number is 0.


carlodri:
@conda-forge/help-c-cpp can someone help me in landing this? I think this a pretty relevant library. But my c-cpp skills are too low to get it working....

This is also important to have the `picamera2` Python library built in a flexible way and still maintaining the strict link required between the versions of `picamera2`-`libcamera`. As soon as `libcamera` works I would like to add it here to the same feedstock.

carlodri:
CC @hmaarrfk sorry to ping you on this but I had the feeling that you could also be interested in this. Forgive me if not...

carlodri:

recipes/libcamera/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+set -xeuo pipefail
+# export CXXFLAGS="${CXXFLAGS} -Wno-error -D_GNU_SOURCE"
+
+meson setup build
```
hmaarrfk, line 5:
```suggestion
meson setup build ${MESON_ARGS}
```

recipes/libcamera/conda_build_config.yaml
```diff
@@ -0,0 +1,9 @@
+os_version:
+  - linux_aarch64: alma8
```
hmaarrfk, line 2: do you really need this? this seems risky, lets migrate aarch later?

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - pyyaml
+    - ply
+    - jinja2
+    - libudev
+    - libudev1
+    - gstreamer
+    - libevent
+    - libdrm
+    - libtiff
+    - qt6-main
```
hmaarrfk, line 41: these should go in the host section, not in the build section.

Build section should only have build tools.

hmaarrfk:
if you don't personally need osx support, then don't worry about it. let somebody else come in and help you with it "later".

Now you should still move your dependencies to the host section, keep tools like `pkg-config`, `meson`, `cmake`, `compilers` in the build section.

recipes/libcamera/conda_build_config.yaml
carlodri: I actually added this because I was trying to fix an error with `memfd_create`. Which in the latest build seems indeed fixed (but there are other issues...

recipes/libcamera/meta.yaml
carlodri: ok thanks, I will move them right away.

carlodri:
at the moment I only need `aarch64`. Am I understanding you suggestion correctly, to drop the `aarch_64` for now, build only on `linux_64` and then migrate later?

hmaarrfk:
> at the moment I only need aarch64. Am I understanding your suggestion correctly, to drop the aarch_64 for now, build only on linux_64 and then migrate later?

yes. mostly doing cross compilation blind is sure to fail.... especially if you are having trouble organizing the dependencies correctly.

You can just ask the bot to rerender immeimdately and it gives you a PR that you can edit.

carlodri:

carlodri:
> > at the moment I only need aarch64. Am I understanding your suggestion correctly, to drop the aarch_64 for now, build only on linux_64 and then migrate later?
> 
> yes. mostly doing cross compilation blind is sure to fail.... especially if you are having trouble organizing the dependencies correctly.
> 
> You can just ask the bot to rerender immeimdately and it gives you a PR that you can edit.

thanks a lot for the insight and help @hmaarrfk, trying now on `linux_64` only.

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  skip: true  # [not linux]
+  
+requirements:
+  build:
+    - cmake
+    - ninja
+    - meson
+    - pkg-config
+    - python
+    - pybind11
```
hmaarrfk, line 23: I think python and pybind11 should oddly be moved to host.

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+    - libdrm
+    - libjpeg-turbo
+    - sdl2
+    - openssl
+    - libboost-devel
+    - pyyaml
+    - ply
```
hmaarrfk, line 33: all this should be moved to host.

recipes/libcamera/meta.yaml
carlodri: hmm I'm skeptical about this,  but let's try. I think it was ok in build.

hmaarrfk:
There is always something odd with meson stuff.

Basically, meson pulls in python, and then the build python gets found.... if python is actually a build time dpenednecy, then we can move all the python stuff to the build prefix.

But hte "libraries", like libjpegturbo should be in the host section for sure.

recipes/libcamera/build.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -xeuo pipefail
+
```
hmaarrfk, line 3:
```suggestion

rm -f subprojects/gtest.wrap
```

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+build:
+  number: 0
+  skip: true  # [not linux]
+  
+requirements:
+  build:
+    - cmake
+    - ninja
+    - meson
```
hmaarrfk, line 20:
```suggestion
    - meson
    # Needed for some build processes???
    - ply
    - pyyaml
    - jinja2
```

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - pkg-config
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
```
hmaarrfk, line 24:
```suggestion
    - {{ stdlib('c') }}
```

hmaarrfk:
I think you also need a patch like:

```
diff --git a/src/libcamera/base/memfd.cpp b/src/libcamera/base/memfd.cpp
index ed0b299b..10469a7f 100644
--- a/src/libcamera/base/memfd.cpp
+++ b/src/libcamera/base/memfd.cpp
@@ -12,6 +12,7 @@
 #include <sys/mman.h>
 #include <sys/syscall.h>
 #include <unistd.h>
+#include <linux/memfd.h>

 #include <libcamera/base/log.h>

**Comment by hmaarrfk** [comment]
> I think you also need a patch like:

If this patch resolves things, you should submit it upstream.

File: recipes/libcamera/meta.yaml [hunk 9]
```diff
@@ -0,0 +1,64 @@
...
+    # Needed for some build processes???
+    - ply
+    - pyyaml
+    - jinja2
+    - pkg-config
+    - {{ compiler('cxx') }}
+    - {{ compiler('c') }}
+    - {{ stdlib('c') }}
+  host:
+    - python
```
hmaarrfk, line 30:
```suggestion
    - libgl-devel
    - python
```

recipes/libcamera/build.sh
```diff
@@ -0,0 +1,6 @@
...
+rm -f subprojects/gtest.wrap
+meson setup build ${MESON_ARGS}
```
hmaarrfk, line 5:
```suggestion
meson setup build ${MESON_ARGS} \
     -Ddocumentation=disabled \
     -Dtest=false
```

hmaarrfk:
This kinda builds for me locally with

```
conda build -m .ci_support/linux64.yaml recipes/libcamera/
```

you can use this to iterate faster than the CIs.

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "libcamera" %}
+{% set version = "0.4.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  git_url: https://git.libcamera.org/libcamera/libcamera.git
+  git_rev: v{{ version }}
```
hmaarrfk, line 10: make sure to use an archive whe nyou are done.

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "libcamera" %}
+{% set version = "0.4.0" %}
+
+package:
+  name: {{ name|lower }}
```
hmaarrfk, line 5:
```suggestion
  name: libcamera
```

extra variables don't bring too much value tbh..

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "libcamera" %}
```
hmaarrfk, line 1:
```suggestion
```

carlodri:
> This kinda builds for me locally with
> 
> ```
> conda debug -m .ci_support/linux64.yaml recipes/libcamera/
> ```
> 
> you can use this to iterate faster than the CIs.
> 
> edit: `build` -> `debug`

thanks, that is very helpful

hmaarrfk:
> > edit: `build` -> `debug`
> 
> thanks, that is very helpful

These debugging tools are what is stopping me from going all in on the `v1` recipes.... they are "fast" but once you need to debug, you are stuck....

hmaarrfk:
See: https://github.com/conda-forge/staged-recipes/pull/29477#issuecomment-2746488627 and https://github.com/conda-forge/staged-recipes/pull/29477#issuecomment-2746489196

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - jinja2
+    - libudev
+    - libudev1
+    - gstreamer
+    - libevent
+    - libdrm
+    - libtiff
+    - qt6-main
+    - libdrm
+    - libjpeg-turbo
```
hmaarrfk, line 44:
```suggestion
    - libjpeg-turbo
    - yaml
    - libunwind
```

carlodri:
looks like it landed!!! Thanks a ton @hmaarrfk for your guidance!

Now, I have a question: it seems like the raspberry pi people have their own fork for their devices: https://github.com/raspberrypi/libcamera.

should we build also this as libcamera-rpi?

carlodri:
@conda-forge/staged-recipes this is ready for review/merger

carlodri:
sorry pinged the staged recipes team by mistake, this is not ready yet

carlodri:
@hmaarrfk what do you think? I'm not sure about the libcamera-rpi lacking tests. And maybe we could merge this one before adding https://github.com/raspberrypi/picamera2

Let me know your thoughts (and if you want to help maintain this mess I made 😅)

recipes/libcamera-rpi/meta.yaml
```diff
@@ -0,0 +1,78 @@
+{% set version = "0.4.0+rpt20250213" %}
+{% set url_safe_version = version | urlencode %}
+
+package:
+  name: libcamera-rpi
+  version: {{ version }}
+
+source:
+  url: https://github.com/raspberrypi/libcamera/releases/download/v{{ url_safe_version }}/libcamera-{{ url_safe_version }}.tar.xz
```
hmaarrfk, line 9: is this a patched package specifically for the raspeberry pi?

this isn't the "conda-forge way" and we would typically just include the patches in the `arm` build and call it a day. 

what do you think? are the patches major?

the problem is that the symbols and python module name might crash.

recipes/libcamera-rpi/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - openssl
+    - libboost-devel
+    - pyyaml
+    - ply
+  run:
+    - python
+
+test:
+  commands:
+    - echo "Tests not implemented for libcamera-rpi, needs rpicam-apps."
```
hmaarrfk, line 60: you can likely add test for existance of headers and `so` files.

recipes/libcamera-rpi/meta.yaml
carlodri: The only thing I know is that the raspberry pi fork is necessary for the picamera2 library, which is not guaranteed to work with the standard libcamera.

And judging from [this](https://github.com/raspberrypi/libcamera/compare/v0.4.0...v0.4.0+rpt20250127) the differences are quite large...

carlodri: but yes I get your point about the symbols/names clash, not sure how to handle this properly... Is there a way to express mutual exclusivity in conda-forge? Some way to say "package A and package B cannot be be installed together".

hmaarrfk: There are ways to handle exclusive packages, but the easiest by far is to have "different variants".

If things are ABI compatible (not sure if they are) then that is by far the easiest thing to do.

we don't typically pull different sources for variants, but maybe this is an exceptional case that warrants this.

we could add "rpi" to the build string or something.... i typically don't like that but you know....


The easiest way to make a mutex is always putting it all in the same recipe.

carlodri: xref: https://github.com/raspberrypi/libcamera/issues/251

carlodri: ok, the two forks are ABI compatible, we can try the build variants with different sources...

I will try to search for some feedstock where this is already implemented to use as a reference.

hmaarrfk:
See tensorflow or PyTorch for recipes with variants. They have selectors in the meta.yaml. You would have a selector for the source url and the source sha256

hmaarrfk:
don't bother creating different pakcage names, just different build strings.

that will make them exclusive.

carlodri:
> don't bother creating different pakcage names, just different build strings.
> 
> that will make them exclusive.

Sorry I'm not sure I understand, so I will likely ask silly questions 🙃. I would like to install the packages by saying `pixi install libcamera-rpi` or `pixi install libcamera`. If I set the same name how will I distinguish them? Did you mean to distinguish them by architecture for which they are built? Like `libcamera-rpi` will install only on `aarch64` and `libcamera` on `linux-64`?

hmaarrfk:
That’s the thing. There should only be so files and header files in the naked name 

libcamera 

that will create the best mutually exclusive pairs of packages. They can be differentiated by their build string. 

Then you can create a nice shortcut 

libcamera-rpi


That has libcamera too build string as its runtime requirements.

hmaarrfk:
Sorry for pushing untested code to your branch, but I think something like the commit I made.

The package

```
libcamera
```
is always the one with all the so an the python file (though maybe the python stuff can be broken out).

users can
```
mamba install libcamera=*=*_upstream
# or
mamba install libcamera=*=*_rpi
```
to install the one they want.

It seems like you want an other shortcut

```
mamba install libcamera-rpi
```

That is fine, not my favorite way to do it, but should be ok.

carlodri:
> Sorry for pushing untested code to your branch, but I think something like the commit I made.

don't be sorry, thanks a lot for your help in landing this, I'm biting a little more than I can chew here 😬🙂

carlodri:
xref successful log of built files: https://dev.azure.com/conda-forge/84710dde-1620-425b-80d0-4cf5baca359d/_apis/build/builds/1210044/logs/34

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+    - openssl
+    - libboost-devel
+    - pyyaml
+    - ply
+  run:
+    - python
+test:
+  commands:
+    - test -f $PREFIX/lib/libcamera.so
+    - test -f $PREFIX/include/libcamera/libcamera/libcamera.h
```
hmaarrfk, line 64: Is this really their intended installation tree?

seems overly verbose...

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,99 @@
...
+      commands:
+        - cam -l
+  - name: libcamera-rpi
+    build:
+      skip: true  # [variant == "upstream"]
+    requirements:
+      run:
+        # I wish we could pin exact... but I think with the python build matrix
+        # It makes things complicated.
+        - {{ pin_subpackage('libcamera') }}
```
hmaarrfk, line 78:
```suggestion
        - libcamera {{ version }}=*_{{ PKG_BUILDNUM }}_{{ variant }}
```

this should help it be pinned correctly to the build number and the variant while leaving python floating.

recipes/libcamera/meta.yaml
carlodri: yes, seems weird to me too, but that's what their installation script does... at least judging from the ci output linked below...

recipes/libcamera/meta.yaml
carlodri: I'll wait for the ci to complete before committing, thank you

carlodri:
```
conda_build.exceptions.RecipeError: Mismatching hashes in recipe. Exact pins in dependencies that contribute to the hash often cause this. Can you change one or more exact pins to version bound constraints?
Involved packages were:
Mismatching package: libcamera (id py310_h9eeb801_0_rpi_fork); dep: libcamera 0.4.0 *_0_rpi_fork; consumer package: libcamera-rpi
```

hmaarrfk:
> Not sure how to deal with this:

I've tried to push.


I think remains, what is the desired behavior.

why did the fork get created? which one should get installed by default.

There are many ways to "push" one variant to be installed by default.

I prefer bumping its build number by 100. It is simple, and just works..

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+{% set libcamera_rpi_version = "0.4.0+rpt20250213" %}
+{% set rpi_url_safe_version = libcamera_rpi_version | urlencode %}
+
+package:
+  name: libcamera
+  version: {{ libcamera_version }}
+
+source:
+{% if variant == "upstream" %}
+  git_url: https://git.libcamera.org/libcamera/libcamera.git
```
hmaarrfk, line 11:
```suggestion
  # libcamera does not offer tar balls
  git_url: https://git.libcamera.org/libcamera/libcamera.git
```

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,111 @@
...
+  git_rev: v{{ libcamera_version }}
+{% elif variant == "rpi_fork" %}
+  url: https://github.com/raspberrypi/libcamera/releases/download/v{{ rpi_url_safe_version }}/libcamera-{{ rpi_url_safe_version }}.tar.xz
```
hmaarrfk, line 14:
```suggestion
  # In conversations with upstream
  # INSERT LINK TO CONVERSATION
  # We found that they attempt to maintain ABI compatibility
  # Thus we build Raspberry-Pi specific modifications as as a variant instead of a separate package
  url: https://github.com/raspberrypi/libcamera/releases/download/v{{ rpi_url_safe_version }}/libcamera-{{ rpi_url_safe_version }}.tar.xz
```

carlodri:
> 
> why did the fork get created?

not entirely sure, I think the reason was to create a custom release cycle where some features could be added, likely specific to raspberry pi or raspberry cameras, independently of upstream.

> which one should get installed by default.

difficult to say, but if I had to decide I would install upstream by default and the `rpi` variant at user choice, or by specific downstream packages such as `picamera2`.

> 
> There are many ways to "push" one variant to be installed by default.
> 
> I prefer bumping its build number by 100. It is simple, and just works..

so the upstream variant would have `output.build.number: 101`, while the `rpi` variant 1?

hmaarrfk:
> so the upstream variant would have output.build.number: 101, while the rpi variant 1?

something like

```
{% set build = 0 %}

{% if variant == "upstrea" %}
{% set build = build + 100 %}
{% endif %}

build:
   number: {{ build }}
```

useing the variable `build` helps the bot auto update you

recipes/libcamera/meta.yaml
```diff
@@ -0,0 +1,122 @@
...
+  # Thus we build Raspberry-Pi specific modifications as as a variant instead of a separate package
+  url: https://github.com/raspberrypi/libcamera/releases/download/v{{ rpi_url_safe_version }}/libcamera-{{ rpi_url_safe_version }}.tar.xz
+  sha256: 764b9b2ab16dbc4e90cd858cd2019be76e087e093bbdd08048d87d0a704f8e28
+{% endif %}
+  patches:
+    - memfd_symbols.patch
+
+build:
+  number: {{ build }}
+  skip: true  # [not linux]
```
hmaarrfk, line 33:
```suggestion
  skip: true  # [not linux]
  {% if variant == "rpi_fork" %}
  skip: true  # [not aarch64]
  {% endif %}
```

right???? does the rpi fork make sense for other linux?

carlodri: no no, that's perfectly correct, thank you!

carlodri:
something is wrong here, it seems like it's not building or testing anything of the `-rpi` variant....

carlodri:
hmm, don't know why it is still not building anything for `aarch64`...

carlodri:

hmaarrfk:
You can probably do rpi stuff in the feeedstock. 

Do you want to split the recipe in c/python packages?

this is the last thing to do here.

carlodri:
> Do you want to split the recipe in c/python packages?

I don't think it's so much relevant...but if it's against the conda-forge laws we can abide🙂

hmaarrfk:
no, this is fine, just wanted to know how much effort you wanted to put into this.

PR #29480, opened 2025-03-20
Add gotranx to conda-forge

finsberg:
Checklist
- [x] Build number is 0.


finsberg:
@conda-forge/python, ready for review!

recipes/gotranx/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - typer
+    - toml
+test:
+  requires:
+    - python {{ python_min }}
+
+  imports:
+    - gotranx
+
+  source_files:
```
synapticarbors, line 42: Unless you're running the tests, these should be removed

recipes/gotranx/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - tests/
+    - pyproject.toml
+
+  commands:
+    - gotranx --help
```
synapticarbors, line 47:
```suggestion
    - gotranx --help
    - pip check
```

and add `pip` under `test/requires`

recipes/gotranx/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+source:
+  url: "https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/{{ name_underscore }}-{{ version }}.tar.gz"
+  sha256: a00f6a94a6038de21ceee525e8e1870077a4473a4f5e272f6f8696bce4bb28cd
+
+build:
+  number: 0
+  noarch: python
+  entry_points:
+    - "gotranx=gotranx.cli:app"
+  script: python -m pip install -v --no-deps .
```
synapticarbors, line 18:
```suggestion
  script: {{ PYTHON }} -m pip install . -vv --no-build-isolation --no-deps
```

recipes/gotranx/meta.yaml
```diff
@@ -0,0 +1,60 @@
+{% set name = "gotranx" %}
+{% set name_underscore = "gotranx" %}
```
synapticarbors, line 2: I don't understand why you need this separate name. It seems like you could just use `name`. If that's the case, remove this and change it accordingly in `url`.

recipes/gotranx/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+  entry_points:
+    - "gotranx=gotranx.cli:app"
+  script: python -m pip install -v --no-deps .
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools
+  run:
```
synapticarbors, line 25: is `myokit` missing? It appears in the `pyproject.toml`.

recipes/gotranx/meta.yaml
finsberg: I agree with you. I have now removed this line and changed `name_underscore` to `name` in the `url` in https://github.com/conda-forge/staged-recipes/pull/29480/commits/66cc0778f4c1aef19607795d96baf28a4c205908

recipes/gotranx/meta.yaml
finsberg: Done in https://github.com/conda-forge/staged-recipes/pull/29480/commits/d8a7bf90da40cbbb8fc3f39502bb5b57853e7550

recipes/gotranx/meta.yaml
finsberg: The issue is that `myokit` is not on `conda-forge`, which is my I also the reason why I didn't run `pip check`. 
I did update the package to work without `myokit` (see https://github.com/finsberg/gotranx/pull/170), but I think it should be part of the default `pip` dependencies since it is a package that would be needed in most scenarios.

finsberg:
I also tried to run the test suite, but for some reason the tests just errors, see https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1205130&view=logs&jobId=4cabe686-70ae-553a-7fd0-310379f2cbac&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff

Therefore I ended up removing them. It looks like the problem is that `pytest` is unable to find my fixutres which are located `conftest.py`. Don't know if this is a problem with my package or something else (my conclusion was that it is probably not a problem with my package).

recipes/gotranx/meta.yaml
synapticarbors: It's probably advisable to get `myokit` up on conda-forge. Once that's available, make sure to add it and then add back in the `pip check`.

PR #29502, opened 2025-03-22
opensees first recipe

ldallolio:
Checklist
- [X] Build number is 0.


ldallolio:
@conda-forge/help-c-cpp Please review this OpenFOAM recipe. Tests are passing, and it's ready for a final review before merging. Thank you!

recipes/opensees/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+  build:
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('fortran') }}
+    - {{ stdlib('c') }}
+    - cmake
+    - make
+    - openmpi
+  host:
+    - python 3.11
```
xhochy, line 26: Can you use a skip instead of hardcoding the Python version here?

recipes/opensees/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - pip
+    - mumps
+    - eigen 3.4
```
xhochy, line 29: Please omit these pins so that the global pins are used.

recipes/opensees/meta.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - mkl-include
+    - mkl-devel
+  run:
+    - metis
+    - hdf5
+    - libscotch
+    - libaec
+    - libcurl
+    - openssl
+    - scalapack
```
xhochy, line 46:
```suggestion
```
All these should be removed as they come in via run_exports

xhochy (changes requested):

recipes/opensees/meta.yaml
ldallolio: Removed version constraints, thank you for pointing that out

recipes/opensees/meta.yaml
ldallolio: Removed python version, the build and tests are still passing :-)

recipes/opensees/meta.yaml
ldallolio: I could remove hdf5 but I had to leave the other ones, otherwise I get errors for missing runtime dependencies or overlinking when the binaries are analyzed... I do not really understand why though. Would you maybe have any hint to share ?

xhochy: You should add them to `host` instead; then they can be omitted from `run`

ldallolio: It seems to work (thanks for the tip!) for anything but scalapack, would that do it in your opinion ? Thanks once again!

xhochy (approved):

recipes/opensees/meta.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('fortran') }}
+    - {{ stdlib('c') }}
+    - cmake
+    - make
+    - openmpi
+  host:
+    - python
+    - pip
```
xhochy, line 27:
```suggestion
    - pip
    - setuptools
```

xhochy (changes requested):
Small nit

PR #29509, opened 2025-03-23
New recipe for LPub3D

trevorsandy:
Checklist
- [x] Build number is 0.

Cheers,


trevorsandy:

recipes/lpub3d/recipe.yaml
```diff
@@ -0,0 +1,113 @@
...
+  version: ${{ lp3d_version }}
+
+source:
+  - url: https://github.com/trevorsandy/${{ lp3d_name }}/archive/${{ lp3d_sha }}.tar.gz
+    sha256: 875a1da26bda9f8cf191fb4064897aa9196db4f55103eb2daf4d21a843bae729
+    target_directory: ${{ lp3d_name }}
+
+  - url: https://github.com/trevorsandy/${{ lp3d_ldview }}/archive/lpub3d-build.tar.gz
+    sha256: f8d5db81e65d4372b302610f686b50f3ed17797ac724cd968d9a10a942e11db0
+    target_directory: ${{ lp3d_ldview }}
```
carterbox, line 23: Conda claims that the hash of this archives has changed. This is caused by one/both of two problems. GitHub archives hashes are not stable over a period of months OR this url points to a tag that has moved. The first issue is a known issue that Github will not fix. The work around is that you have to attach your own archive to a release because user uploaded files are static unlike GitHub generated archives which are created on demand and then cached for only a few months.

danielnachun:
Unfortunately I don't think we can accept this recipe as it is for several reasons:
- It is building multiple packages that should be separate recipes.
- Some of the packages do not use stable tagged releases.  We cannot accept packages that don't use tagged releases with semantic versioning - please work with the relevant upstream repositories to get tagged releases for those repositories, and please keep in mind that future version updates also need to be tagged releases as well, not git commits.

trevorsandy:
> Unfortunately I don't think we can accept this recipe as it is for several reasons...

Very well. Many thanks for your time.

Cheers,

PR #29525, opened 2025-03-24
add r-easy.utils

theAeon:
Checklist
- [X] Build number is 0.


theAeon:
@conda-forge/help-r, ready for review!

mfansler (changes requested):
Please use `conda_r_skeleton_helper` for CRAN recipes.

theAeon:
> Please use `conda_r_skeleton_helper` for CRAN recipes.

this is that, followed by a run of feedrattler, followed by a few manual tweaks that feedrattler was messing up.

recipes/r-easy.utils/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+tests:
+  - script:
+      - if: win
+        then: Rscript -e \"library('easy.utils')\"
+        else: Rscript -e  \"library\('easy.utils'\)\"
+
+about:
+  summary: "Some utility functions for validation and data manipulation. These functions can be helpful to reduce internal codes everywhere in package development."
+  license: MIT
+  license_file: LICENSE
```
mfansler, line 47: Original license is `License: MIT + file LICENSE`, so the MIT license from `r-base` should be bundled:

```suggestion
  license_file:
    - {{ environ["PREFIX"] }}/lib/R/share/licenses/MIT
    - LICENSE
```

Maybe this can be instead `${{ PREFIX }}/lib/R/share/licenses/MIT`?

recipes/r-easy.utils/recipe.yaml
```diff
@@ -0,0 +1,52 @@
...
+  homepage: https://github.com/ycli1995/easy.utils
+
+extra:
+  recipe-maintainers:
+    - theAeon
```
mfansler, line 52: Please include CF R team:

```suggestion
    - conda-forge/r
    - theAeon
```

mfansler (changes requested):
This is the first I've reviewed a recipe with this schema, but this is what I expect it should be.

AFAIK, there has been no discussion among or with the CF R team regarding how our standard translates to the new schema, so this is a bit *ad hoc*. Hopefully we can work out here what it should look like. Please understand that keeping all R recipes to a similar standard helps make the annual R migrations run more smoothly.

For future recipes, it would be preferable that they continue to be submitted in schema v0 format as outputted by `conda_r_skeleton_helper`. Then we can bulk convert when the converter is mature and implements our expectations.

theAeon:
`invalid operation: Environment variable PREFIX not found` 😵‍💫

theAeon:
Not sure how to deal with the prefix environment thing short of just keeping it in the feedstock

theAeon:
Noted. Will fix. RcppPlanc (which was the other one I converted) will also be converted back when I bump that to 2.x, as I was mostly using that as a testing ground for libplanc.

mfansler (approved):

mfansler:
@theAeon thanks for your patience! Was definitely useful to see what kind of translation problems arise.

PR #29543, opened 2025-03-26
Adding pio r package as r-pio

brandon-hastings:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/r-pio/meta.yaml
+++ b/recipes/r-pio/meta.yaml
@@ -1,89 +1,59 @@
-# Note: there are many handy hints in comments in this example -- remove them when you've finalized your recipe
-# If your package is python based, we recommend using Grayskull to generate it instead:
-# https://github.com/conda-incubator/grayskull
-
-# Jinja variables help maintain the recipe as you'll update the version only here.
-# Using the name variable with the URL in line 16 is convenient
-# when copying and pasting from another recipe, but not really needed.
 {% set name = "r-pio" %}
-{% set version = "1.0.0" %}
+{% set version = "v1.0.0" %}
 
 package:
   name: {{ name|lower }}
   version: {{ version }}
 
 source:
-  # uncomment the line below, and modify as needed. Use releases if available:
-  # url: https://github.com/simplejson/simplejson/releases/download/{{ version }}/simplejson-{{ version }}.tar.gz
-  # and otherwise fall back to archive:
-  url: https://github.com/caravagn/pio/archive/v{{ version }}.tar.gz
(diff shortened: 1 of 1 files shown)


mfansler:
Please update your `conda_r_skeleton_helper` and regenerate the recipe.

recipes/r-pio/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+{% set posix = 'm2-' if win else '' %}
+{% set native = 'm2w64-' if win else '' %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/caravagn/pio/archive/refs/tags/{{ version }}.tar.gz
```
mfansler, line 12: Generally, we prefer to work downstream of CRAN. Is there a reason why this package cannot first be deposited there?

recipes/r-pio/meta.yaml
```diff
@@ -0,0 +1,63 @@
+{% set name = "r-pio" %}
+{% set version = "v1.0.0" %}
```
mfansler, line 2: The tag shows `v1.0.0`, but the DESCRIPTION file shows v0.1.0.

mfansler (changes requested):

recipes/r-pio/meta.yaml
brandon-hastings: Hi, the reason that the pio package is not on CRAN is that there are similar packages now hosted on CRAN that do the same thing since it is input output highlighting. However, we are in the process of submitting a nextflow pipeline to the nf-core repository that is made up of many previous packages from our lab that are dependent on pio, as at one point our pio package was the only one that gave us the output format we prefer. The nf-core maintainers requested that we repackage some of our custom modules in our submitted pipeline to be standard nf-core modules, which require that they are available in conda. In order to create the conda environment for these modules we first need pio to be available on conda-forge, but did not necessarily plan to submit pio to CRAN since nowadays there are similar packages available. Pio is stable, if it helps, with no need for updates in the past 5 years even with all of our other packages being dependent on it.

recipes/r-pio/meta.yaml
brandon-hastings: Of course, I will bump the version number and repackage

recipes/r-pio/meta.yaml
mfansler: I'd recommend submitting this to Bioconda instead. You can also ping me there for review/merge.

CC: @conda-forge/bioconda-recipes

PR #29568, opened 2025-03-29
add gen3, yamlfix

danielnachun:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/gen3/meta.yaml
+++ b/recipes/gen3/meta.yaml
@@ -1,0 +1,68 @@
+{% set version = "4.26.1" %}
+
+package:
+  name: gen3
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/g/gen3/gen3-{{ version }}.tar.gz
+  sha256: 4572b67b499c7660d41fdb0acac6fa10bdf0e3953528710914ce6c7341b14091
+
+build:
+  entry_points:
+    - gen3 = gen3.cli.__main__:main
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python ={{ python_min }}
+    - poetry >=1.4.1
+    - poetry-core
+    - pip
+  run:
+    - python >={{ python_min }}
+    - requests
+    - indexclient >=2.3.0,<3.0.0
+    - drsclient >=0.3.0
+    - aiohttp
+    - backoff
+    - cdislogging >=1.1.0,<2.0.0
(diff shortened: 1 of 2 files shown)


danielnachun:
@conda-forge/staged-recipes @conda-forge/help-python This is ready for review.  `gen3` is slightly odd because the `pip check` and `gen3` commands have a circular dependency on `pypfb`, which itself depends on `gen3`.  Once I've added `pypfb`, I'll be able to re-enable those tests.

recipes/gen3/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - httpx
+    - xmltodict >=0.13.0,<0.14.0
+    - pyyaml >=6.0.1
+    - gen3users
+
+test:
+  imports:
+    - gen3
+  commands:
+    # - pip check
```
ocefpaf, line 52: Why? If it is something upstream this package, we should try to fix it.

recipes/gen3/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - indexclient >=2.3.0,<3.0.0
+    - drsclient >=0.3.0
+    - aiohttp
+    - backoff
+    - cdislogging >=1.1.0,<2.0.0
+    - click
+    - fhirclient
+    - jsonschema
+    - dataclasses-json <=0.5.9
+    # - pypfb >=0.5.30
```
ocefpaf, line 36: Is this one optional?

recipes/gen3/meta.yaml
danielnachun: This will fail because of a circular dependency on `pypfb`, which itself depends on `gen3`.  The only solution I could come up with so far is to leave these commented out and re-enable later.  The `gen3` library itself can be imported without `pypfb`, but the CLI interface tries to import, and `pip check` will complain that `pypfb` isn't installed.  I'm open to other solutions but this was the only thing that worked for me so far.

recipes/gen3/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - xmltodict >=0.13.0,<0.14.0
+    - pyyaml >=6.0.1
+    - gen3users
+
+test:
+  imports:
+    - gen3
+  commands:
+    # - pip check
+    # - gen3 --help
```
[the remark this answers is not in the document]
danielnachun: See above comment.

recipes/gen3/meta.yaml
danielnachun: See above comment.

PR #29591, opened 2025-04-01
add temporal cli (temporalio-cli)

wolfv:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/temporalio-cli/recipe.yaml
+++ b/recipes/temporalio-cli/recipe.yaml
@@ -2,7 +2,7 @@ context:
   version: "1.3.0"
 
 package:
-  name: temporal-cli
+  name: temporalio-cli
   version: ${{ version }}
 
 source:
--- a/recipes/temporalio-cli/recipe.yaml
+++ b/recipes/temporalio-cli/recipe.yaml
@@ -43,6 +43,10 @@ about:
   license_file:
     - LICENSE
     - library_licenses/
+    # these two are not properly collected by go-licenses
+    - go-statsd-client.txt
+    - mathutil-license.txt
+
   summary: Command-line interface for running and interacting with Temporal Server and UI
   description: |
     Temporal CLI is a command-line interface for running and interacting with Temporal Server and UI.
--- a/recipes/temporal-cli/recipe.yaml
+++ b/recipes/temporal-cli/recipe.yaml
@@ -11,6 +11,7 @@ source:
   target_directory: src
 
 build:
+  number: 0
   script:
     - cd src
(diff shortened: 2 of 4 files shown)


recipes/temporalio-cli/recipe.yaml
```diff
@@ -0,0 +1,54 @@
...
+source:
+  url: https://github.com/temporalio/cli/archive/refs/tags/v${{ version }}.tar.gz
+  sha256: 15be9f155cd5114367942568f884969f7ed2d3262ad39bb665cf359735f643b3
+  target_directory: src
+
+build:
+  number: 0
+  script:
+    - cd src
+    - go-licenses save ./cmd/temporal --save_path ../library_licenses --ignore github.com/cactus/go-statsd-client/statsd --ignore modernc.org/mathutil
```
pavelzw, line 17: could you manually add the missing license files into the feedstock (or add them as an additional source)?

recipes/temporalio-cli/recipe.yaml
```diff
@@ -0,0 +1,54 @@
...
+      then: go build -ldflags="-s -w -X github.com/temporalio/cli/temporalcli.Version=${{ version }}" -v -o %LIBRARY_BIN%\temporal.exe ./cmd/temporal
+      else: go build -ldflags="-s -w -X github.com/temporalio/cli/temporalcli.Version=${{ version }}" -v -o $PREFIX/bin/temporal ./cmd/temporal
+    - if: unix
+      then:
+        - mkdir -p $PREFIX/share/bash-completion/completions
+        - temporal completion bash > $PREFIX/share/bash-completion/completions/temporal
+        - mkdir -p $PREFIX/share/zsh/site-functions
+        - temporal completion zsh > $PREFIX/share/zsh/site-functions/_temporal
+        - mkdir -p $PREFIX/share/fish/vendor_completions.d
+        - temporal completion fish > $PREFIX/share/fish/vendor_completions.d/temporal.fish
```
pavelzw, line 28: how will this work when cross-compiling from osx-64 to osx-arm64?

recipes/temporalio-cli/recipe.yaml
wolfv: Done!

recipes/temporalio-cli/recipe.yaml
wolfv: That is a good question. It won't work well, if we are not also building a `osx-64` binary at the same time (only to generate the completions).

The other idea would be to create a `noarch` package that only contains the completions, and skip that when cross compiling but that might also not be the ideal solution.

pavelzw: As discussed, we’ll do native builds for this one so it should be fine

pavelzw (approved):

PR #29623, opened 2025-04-03
Recipe for the fastemriwaveforms package

mpigou:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/fastemriwaveforms/meta.yaml
+++ b/recipes/fastemriwaveforms/meta.yaml
@@ -1,0 +1,224 @@
+{% set name = "FastEMRIWaveforms" %}
+{% set version = "2.0.0" %}
+{% set number = 0 %}
+
+{% set PYTHON = "python" %}
+{% set PYTHON = "$PREFIX/bin/python" %}  # [linux]
+{% set PYTHON = "%PREFIX%\\python" %}  # [win]
+
+{% set RECIPE_DIR = "$RECIPE_DIR/" %}
+{% set RECIPE_DIR = "%RECIPE_DIR%\\" %}  # [win]
+
+package:
+  name: {{ name|lower }}-split
+  version: {{ version }}
+
+source:
+  url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms/archive/refs/tags/v2.0.0.tar.gz
+  sha256: d5c8bcba1ec24cb71be8a293f75cf1491e2e474515c1481559a4ef07378e0c36
+
+build:
+  number: {{ number }}
+  skip: True  # [(py<310)]
+
+outputs:
+  - name: fastemriwaveforms
+    version: {{ version }}
+    build:
+      number: {{ number }}
+      skip: true  # [(py<310) or (cuda_compiler == "nvcc")]
+      script_env:
+        - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+      script:
+        - >-
+          {{ PYTHON }} -m pip install . -vv -Ccmake.define.FEW_LAPACKE_FETCH=OFF -Ccmake.define.FEW_WITH_GPU=OFF -Ccmake.define.FEW_LAPACKE_DETECT_WITH=PKGCONFIG
+        - >-
+          {{ PYTHON }} {{ RECIPE_DIR }}move_tests.py
+      entry_points:
+        - few_citations = few.cmd.citations:main
+        - few_files = few.cmd.files:main
+    requirements:
+      build:
+        - python                              # [build_platform != target_platform]
+        - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+        - numpy                               # [build_platform != target_platform]
+        - scikit-build-core                   # [build_platform != target_platform]
+        - setuptools_scm                      # [build_platform != target_platform]
+        - cython                              # [build_platform != target_platform]
+        - pkg-config                          # [build_platform != target_platform]
+        - liblapacke                          # [build_platform != target_platform]
+        - libblas                             # [build_platform != target_platform]
+        - libtmglib                           # [build_platform != target_platform]
+        - {{ compiler('c') }}
+        - {{ stdlib('c') }}
+        - {{ compiler('cxx') }}
+        - libcxx-devel <19  # [osx]
+        - ninja
+        - cmake
+        - make
+      host:
+        - python
+        - pip
+        - scikit-build-core
+        - setuptools_scm
+        - cython
+        - numpy
+        - pkg-config
+        - liblapacke
+        - libblas
+        - libtmglib
+      run:
+        - python
+        - exceptiongroup  # [py<311]
+        - h5py
+        - jsonschema
+        - liblapacke
+        - lisaconstants
+        - multispline
+        - numba
+        - numpy
+        - nvidia-ml-py
+        - platformdirs
+        - pydantic
+        - pyyaml
+        - requests
+        - rich
+        - scipy
+        - tqdm
+        - wrapt
+    test:
+      imports:
+        - few
+        - few.tests
+        - few_backend_cpu
+      requires:
+        - pip
+        - matplotlib
+      commands:
+        - pip check
+        - python -c "import importlib.metadata; assert importlib.metadata.version('fastemriwaveforms') == '{{ version }}'"
+        - python -c "import few; few.get_backend('cpu')"
+        - python -m few.tests --disable testfile
+    about:
+      home: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms
+      summary: 'Blazingly fast EMRI waveforms'
+      description: |
+        This package contains the highly modular framework for fast and accurate extreme mass ratio
+        inspiral (EMRI) waveforms. This is the core package with only CPU support
+        for waveform generation.
+      license: GPL-3.0-or-later
+      license_family: GPL
+      license_file: LICENSE
+      doc_url: https://bhptoolkit.org/FastEMRIWaveforms/html/index.html
+      dev_url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms
+
+  - name: fastemriwaveforms-cuda12x
+    version: {{ version }}
+    build:
+      number: {{ number }}
+      skip: true  # [(py<310) or (cuda_compiler != "cuda-nvcc") or not linux]
+      script_env:
+        - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+      script:
+        - sed -i'' -e '/@SKIP_PLUGIN@/d' pyproject.toml
+        - >-
+          {{ PYTHON }} {{ RECIPE_DIR }}update_name.py fastemriwaveforms-cuda12x
+        - >-
+          {{ PYTHON }} -m pip install . -vv -Ccmake.define.FEW_LAPACKE_FETCH=OFF -Ccmake.define.FEW_WITH_GPU=ONLY -Ccmake.define.FEW_CUDA_ARCH=all
+      ignore_run_exports_from:
+        - {{ compiler('cuda') }}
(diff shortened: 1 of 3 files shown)


mpigou:

cchapmanbird:
I accept to be a recipe maintainer.

mpigou:

mpigou:

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
+{% set name = "FastEMRIWaveforms" %}
+{% set version = "2.0.0" %}
+{% set number = 0 %}
```
xhochy: Please set this directly.
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+
+{% set RECIPE_DIR = "$RECIPE_DIR/" %}
+{% set RECIPE_DIR = "%RECIPE_DIR%\\" %}  # [win]
+
+package:
+  name: {{ name|lower }}-split
+  version: {{ version }}
+
+source:
+  url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms/archive/refs/tags/v2.0.0.tar.gz
```
xhochy:
```suggestion
  url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms/archive/refs/tags/v{{ version }}.tar.gz
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+  sha256: d5c8bcba1ec24cb71be8a293f75cf1491e2e474515c1481559a4ef07378e0c36
+
+build:
+  number: {{ number }}
```
xhochy:
```suggestion
  number: 0
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
+{% set name = "FastEMRIWaveforms" %}
+{% set version = "2.0.0" %}
+{% set number = 0 %}
+
+{% set PYTHON = "python" %}
+{% set PYTHON = "$PREFIX/bin/python" %}  # [linux]
+{% set PYTHON = "%PREFIX%\\python" %}  # [win]
```
xhochy, line 6: `{{ PYTHON }}` already exists in the local variable scope.
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+
+build:
+  number: {{ number }}
+  skip: True  # [(py<310)]
+
+outputs:
+  - name: fastemriwaveforms
+    version: {{ version }}
+    build:
+      number: {{ number }}
```
xhochy:
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+      entry_points:
+        - few_citations = few.cmd.citations:main
+        - few_files = few.cmd.files:main
+    requirements:
+      build:
+        - python                              # [build_platform != target_platform]
+        - cross-python_{{ target_platform }}  # [build_platform != target_platform]
+        - numpy                               # [build_platform != target_platform]
+        - scikit-build-core                   # [build_platform != target_platform]
+        - setuptools_scm                      # [build_platform != target_platform]
```
xhochy: pure python packages don't need to be listed here
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+    build:
+      number: {{ number }}
+      skip: true  # [(py<310) or (cuda_compiler == "nvcc")]
+      script_env:
+        - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+      script:
+        - >-
+          {{ PYTHON }} -m pip install . -vv -Ccmake.define.FEW_LAPACKE_FETCH=OFF -Ccmake.define.FEW_WITH_GPU=OFF -Ccmake.define.FEW_LAPACKE_DETECT_WITH=PKGCONFIG
+        - >-
+          {{ PYTHON }} {{ RECIPE_DIR }}move_tests.py
```
xhochy: Don't copy the tests into the prefix. They should not be part of the package.
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+        - cmake
+        - make
+      host:
+        - python
+        - pip
+        - scikit-build-core
+        - setuptools_scm
+        - cython
+        - numpy
+        - pkg-config
```
xhochy: `pkg-config` should only be in build
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+        - tqdm
+        - wrapt
+    test:
+      imports:
+        - few
+        - few.tests
+        - few_backend_cpu
+      requires:
+        - pip
+        - matplotlib
```
xhochy:
```suggestion
        - matplotlib-base
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+      license: GPL-3.0-or-later
+      license_family: GPL
+      license_file: LICENSE
+      doc_url: https://bhptoolkit.org/FastEMRIWaveforms/html/index.html
+      dev_url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms
+
+  - name: fastemriwaveforms-cuda12x
+    version: {{ version }}
+    build:
+      number: {{ number }}
```
xhochy:
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,224 @@
...
+        This package contains the highly modular framework for fast and accurate extreme mass ratio
+        inspiral (EMRI) waveforms. This is the core package with only CPU support
+        for waveform generation.
+      license: GPL-3.0-or-later
+      license_family: GPL
+      license_file: LICENSE
+      doc_url: https://bhptoolkit.org/FastEMRIWaveforms/html/index.html
+      dev_url: https://github.com/BlackHolePerturbationToolkit/FastEMRIWaveforms
+
+  - name: fastemriwaveforms-cuda12x
```
xhochy, line 108: Why is this a separate package? It should be a different build with the same name of the first package.

recipes/fastemriwaveforms/meta.yaml
mpigou: In multi-output recipes, `{{ PYTHON }}` does not correctly work so I used the same solution as in `cupy-feedstock` (see [here](https://github.com/conda-forge/cupy-feedstock/blob/8f9d82294db69e3d48756a2c4cd92a0267316acc/recipe/meta.yaml#L17))

recipes/fastemriwaveforms/meta.yaml
mpigou: I'd like to have a way for users to execute tests without having to clone the project sources.
These a functional tests that can be used by users to ensure their setup uses a GPU accelerator when available, test memory consumption and measure performance on multiple systems and measure the speed-up between the CPU and the CUDA computing backends.
What I mean is they are not just here for developers and are a useful (documented) feature also for users.

If we start implementing short unit-test mainly for the purpose of code quality, I agree that these should not be embedded in the package.

Maybe a solution we be to have an other output package (maybe `fastemriwaveforms-test`) which would contain the `few.tests` module?

But for now, we'd like to offer a unified experience to users not matter whether they install the project v2.0 release doing `pip install fastemriwaveforms` or `conda install fastemriwaveforms` so could you please allow embedding tests in the package, at least for this 2.0 release? And we will see how we can improve things for the next release.

recipes/fastemriwaveforms/meta.yaml
mpigou: Thanks I just learnt the difference between matplotlib and matplotlib-base, so this change makes at lot of sense!

However I completely removed the need for matplotlib here, we actually don't need it anymore for test execution and leaving it here was an oversight on my part.

recipes/fastemriwaveforms/meta.yaml
mpigou: The first package contains all python code, and the result of compiling the C++/CUDA sources with a C++ compiler (so targeting execution on a CPU).

The separate package only contains the result of compiling the C++/CUDA sources with a CUDA compiler (this targeting execution on a GPU) and has the first package as dependency (with the python code being the glue selecting whether we use the CPU or the GPU dynamic libraries).

The goal is to have a modular package with multiple available execution backends that we select at runtime.

recipes/fastemriwaveforms/meta.yaml
xhochy: > Maybe a solution we be to have an other output package (maybe fastemriwaveforms-test) which would contain the few.tests module?

That would be the preferred solution.

> But for now, we'd like to offer a unified experience to users not matter whether they install the project v2.0 release doing pip install fastemriwaveforms or conda install fastemriwaveforms so could you please allow embedding tests in the package, at least for this 2.0 release? And we will see how we can improve things for the next release.

But why do you need to call a separate script then after `pip install` if this is also the `pip install experience`?

recipes/fastemriwaveforms/meta.yaml
xhochy: This line should be deleted.

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,219 @@
...
+        - cuda-cudart-dev
+        - libcublas-dev
+        - libcusparse-dev
+
+      run:
+        - python
+        - {{ pin_subpackage('fastemriwaveforms', exact=True) }}
+        - {{ pin_compatible('cuda-version', min_pin='x') }}
+        - libstdcxx
+        - libgcc
```
xhochy:
```suggestion
```

recipes/fastemriwaveforms/meta.yaml
```diff
@@ -0,0 +1,219 @@
...
+        - cupy
+        - cuda-cudart
+        - libcublas
+        - libcusparse
```
xhochy, line 172: Should these come in via `run_export`?

xhochy (changes requested):

recipes/fastemriwaveforms/meta.yaml
mpigou: Actually you are right, this script is not needed. The `few.tests` module is available in all cases (since the build backend, `scikit-build-core`, takes care of packaging the `tests` directory into the submodule `few.tests`.

I honestly cannot remember why I added that script. I did many tests for conda packaging before we actually release the v2 of the FastEMRIWaveforms package and I guess this was an artifact of one of those tests...

recipes/fastemriwaveforms/meta.yaml
mpigou: I do not think so, I followed the guidelines for [CUDA compatibility](https://github.com/conda-forge/cuda-feedstock/blob/main/recipe/doc/recipe_guide.md#cuda-enhanced-compatibility): by ignoring `run_exports` from the `-dev` cuda packages, and then adding them in the `run` section with `pin_compatible('cuda-version', ...)`, the package should be compatible with all CUDA 12.x minor versions.

At least that's how I understood the documentation, though I must admit it is not always clear to me what should go in the different requirements subsections.

xhochy (approved):

PR #29671, opened 2025-04-08
Add floability recipe

saifulislampi:
Checklist
- [x] Build number is 0.


dthain:
I confirm I want to be a recipe maintainer.

saifulislampi:

recipes/floability/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+{% set version = "0.1.1" %}
+{% set sha256 = "0be93de999f0484adbb9c7405eabde3c502593366b5c97ba28d17cf562c96b6b" %}
+
+package:
+  name: {{ name | lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/floability/floability-cli/archive/refs/tags/v{{ version }}.tar.gz
+  sha256: {{ sha256 }}
```
synapticarbors, line 11: Just put the `sha256` here -- no need to use a jinja variable.

recipes/floability/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+build:
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+  skip: true  # [win]
```
synapticarbors, line 16: Can this be `noarch: python`?

recipes/floability/meta.yaml
saifulislampi: Removed jinja variable.

recipes/floability/meta.yaml
saifulislampi: No, because ndcctools is platform-specific.

saifulislampi:

saifulislampi:
@conda-forge-admin, please ping conda-forge/synapticarbors

recipes/floability/meta.yaml
h-vetinari: which aspect of the package is platform specific. From the dependencies and installation it looks like pure python (in which case it should be noarch)

saifulislampi: We rely on `ndcctools`, which is not packaged for Windows. Marking the recipe` noarch: python` would let conda offer Floability to Windows users but then fail to solve, so we intentionally keep per-platform builds `(skip: true # [win])`. If/when `ndcctools` becomces available for Windows,  we can revisit making Floability `no-arch`.

h-vetinari: > Marking the recipe` noarch: python` would let conda offer Floability to Windows users but then fail to solve

Just like it would fail to solve if you try to install it on windows as a per-platform build. Noarch describes a single package, not its full dependency graph - if `floatability` is pure python, it should be noarch. And then you don't have to do surgery later when `ndcctools` becomes available on windows at some point.

btovar:
@h-vetinari  If we try the noarch, the build fails on windows with:

```
\u2514\u2500 ndcctools =* * does not exist (perhaps a typo or a missing channel).
```

floability will only be used on linux clusters, thus the flexibility of noarch is not needed. Could we simply skip the windows build?

h-vetinari:
The windows failure here is irrelevant. Once the recipe is moved to its own feedstock, it will only build on linux (or more generally, the specified [`noarch_platforms`](https://conda-forge.org/docs/maintainer/conda_forge_yml/#noarch-platforms))

h-vetinari (approved):
Thanks. You can ping me on the feedstock if you have further questions.

PR #29758, opened 2025-04-17
fix: remove lint for maintainer teams

### LINKED ISSUE #29583: Turn off linter errors for maintainer teams
When we add a feedstock that's very dependent on something else (e.g. we break out a part), it's often desirable to keep the same maintainers as that original feedstock, and simply add 

```yaml
extra:
  recipe-maintainers:
    - conda-forge/foobar
```
However, the linter in staged-recipes wants a "confirmation" from that maintainer
> `recipes/foobar-child/meta.yaml`:
>   * lints:
>     * The following maintainers have not yet confirmed that they are willing to be listed here: conda-forge/foobar. Please ask them to comment on this PR if they are.

with no way to satisfy that demand (AFAICT). IMO we should disable this lint for teams.

CC @conda-forge/core

### LINKED ISSUE #29626: Linter checking for maintainer comments is case-sensitive, leading to false failures
### Bug Summary

When submitting a new recipe, the linter performs a check that all maintainers listed in the recipe have commented on the Pull Request. However, if the casing differs between the maintainer's GitHub username listed in the recipe and their actual username, the linter raises a false error.

### Example

Here is an example PR comment where this happened: https://github.com/conda-forge/staged-recipes/pull/29613#issuecomment-2773498706

I have also mentioned this in the Zulip chat here: [#general > staged-recipe: anaconda-cli and (maybe?) linter error @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/staged-recipe.3A.20anaconda-cli.20and.20.28maybe.3F.29.20linter.20error/near/509782360)

### Cause

I believe that the comparison is performed [here](https://github.com/conda-forge/staged-recipes/blob/550a5103b268159b71237cb5433276cb0bedd1fb/.github/workflows/scripts/linter.py#L201), where `maintainer not in commenters` check is happening. In this case, `maintainer` is read in from the recipe, while the `commenters` dictionary has a key of the actual GitHub username read from comments. If the casing differs, this will return False.

### Recommended Fix

I'd be happy to help contribute a fix if this is confirmed to be undesired behavior. Since GitHub usernames are considered to be case-sensitive in every case I've personally seen, except the UI, I believe the key of the `commenters` dictionary should be lower-cased, and then the `maintainer` should also be lowercased before this membership check is performed.

beckermr:
This PR removes linting for maintainer teams. We could check team membership, but this would involve token permissions the workflow does not have. So I am going to punt.

cc @h-vetinari 

closes #29583 
closes #29626

### DIFF
--- a/.github/workflows/scripts/linter.py
+++ b/.github/workflows/scripts/linter.py
@@ -192,24 +192,25 @@ def _lint_recipes(gh, pr):
         # 5. Ensure all maintainers have commented that they approve of being listed
         if maintainers:
             # Get PR author, issue comments, and review comments
-            pr_author = pr.user.login
+            pr_author = pr.user.login.lower()
             issue_comments = pr.get_issue_comments()
             review_comments = pr.get_reviews()
 
             # Combine commenters from both issue comments and review comments
-            commenters = {comment.user.login for comment in issue_comments}
-            commenters.update({review.user.login for review in review_comments})
+            commenters = {comment.user.login.lower() for comment in issue_comments}
+            commenters.update({review.user.login.lower() for review in review_comments})
 
             # Check if all maintainers have either commented or are the PR author
             non_participating_maintainers = set()
-            for maintainer in maintainers:
+            for orig_maintainer in maintainers:
+                maintainer = orig_maintainer.lower()
                 if (
                     maintainer not in commenters
                     and maintainer != pr_author
                     and maintainer not in NOCOMMENT_REQ_TEAMS
                     and ("/" not in maintainer or maintainer.split("/", 1)[0] != "conda-forge")
                 ):
-                    non_participating_maintainers.add(maintainer)
+                    non_participating_maintainers.add(orig_maintainer)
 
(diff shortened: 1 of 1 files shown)


beckermr:

.github/workflows/scripts/linter.py
```diff
@@ -192,23 +192,25 @@ def _lint_recipes(gh, pr):
...
             # Check if all maintainers have either commented or are the PR author
             non_participating_maintainers = set()
-            for maintainer in maintainers:
+            for orig_maintainer in maintainers:
+                maintainer = orig_maintainer.lower()
                 if (
                     maintainer not in commenters
                     and maintainer != pr_author
                     and maintainer not in NOCOMMENT_REQ_TEAMS
+                    and ("/" not in maintainer or maintainer.split("/", 1)[0] != "conda-forge")
```
beckermr, line 211:
```suggestion
                    and "/" not in maintainer
```

PR #29853, opened 2025-04-26
Add vtpk-reader recipe

kshklovsky:
Checklist
- [*] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [*] License file is packaged (see [here](https://github.com/conda-forge/staged-recipes/blob/5eddbd7fc9d1502169089da06c3688d9759be978/recipes/example/meta.yaml#L64-L73) for an example).
- [*] Source is from official source.
- [*] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).
- [*] If static libraries are linked in, the license of the static library is packaged.
- [*] Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md).
- [ ] Build number is 0.
- [*] A tarball (`url`) rather than a repo (e.g. `git_url`) is used in your recipe (see [here](https://conda-forge.org/docs/maintainer/adding_pkgs.html#build-from-tarballs-not-repos) for more details).

### DIFF
--- a/recipes/vtpk-reader/meta.yaml
+++ b/recipes/vtpk-reader/meta.yaml
@@ -1,0 +1,43 @@
+{% set name = "vtpk-reader" %}
+{% set version = "0.1.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/vtpk_reader-{{ version }}.tar.gz
+  sha256: e919dd0795923de1b4963f66e101605229aea48931b869444d34d27e70b2292c
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.9,<4.0
+    - poetry-core
+    - pip
+  run:
+    - python >=3.9.0,<4.0.0
+    - shapely >=2.0.0,<3.0.0
+    - pyproj >=3.4.1,<4.0.0
+    - mapbox-vector-tile >=2.1.0,<3.0.0
+
+test:
+  imports:
+    - vtpk_reader
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  summary: Reading ESRI Vector Tile Package (vtpk) files.
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - kirillshklovsky
--- a/recipes/vtpk-reader/meta.yaml
+++ b/recipes/vtpk-reader/meta.yaml
@@ -36,8 +36,9 @@ test:
 about:
   summary: Reading ESRI Vector Tile Package (vtpk) files.
   license: MIT
+  license_family: MIT
   license_file: LICENSE
 
 extra:
   recipe-maintainers:
-    - kirillshklovsky
+    - kshklovsky


kshklovsky:
@conda-forge/help-python Not sure I follow the `home` linter comment, looks like version is pinned to `python_min`. The `<4` portions were autogenerated by grayskull

recipes/vtpk-reader/meta.yaml
```diff
@@ -0,0 +1,47 @@
+{% set name = "vtpk-reader" %}
+{% set version = "0.1.0" %}
+{% set python_min = "3.9" %}
```
shermansiu, line 3:
```suggestion
```
Python 3.9 is currently the minimum Python version already, so maybe we can omit this to future-proof the recipe?

kshklovsky: Done, thank you!

shermansiu:

recipes/vtpk-reader/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+  sha256: e919dd0795923de1b4963f66e101605229aea48931b869444d34d27e70b2292c
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >={{ python_min }},<4.0
```
shermansiu, line 19:
```suggestion
    - python {{ python_min }}
```

shermansiu (commented):
The linter is probably looking for this change.

recipes/vtpk-reader/meta.yaml
kshklovsky: Yup, I think that did it! Ty

shermansiu:
I think my only concern (nitpick) is that the recipe uses Poetry for dependency resolution when it gets built. I think it makes sense to use it for the PyPI package, but maybe we can [patch](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#patches) it out so that it just relies on Conda here?

shermansiu (approved):

shermansiu:
Then again, I don't have write access to `staged-recipes`, so someone else will need to merge this in.

recipes/vtpk-reader/meta.yaml
```diff
@@ -0,0 +1,46 @@
...
+    - pip check
+  requires:
+    - python {{ python_min }}
+    - pip
+
+about:
+  home: https://github.com/kshklovsky/vtpk-reader
+  summary: A library to read ESRI Vector Tile Package (.vtpk) files
+  license: MIT
+  license_family: MIT
```
danielnachun, line 41:
```suggestion
```
This is now redundant.

danielnachun:
> I think my only concern (nitpick) is that the recipe uses Poetry for dependency resolution when it gets built. I think it makes sense to use it for the PyPI package, but maybe we can [patch](https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#patches) it out so that it just relies on Conda here? Then again, I've seen some packages merged that use poetry-core anyway, so don't worry too much about it.

For future reference, to my knowledge there is no specific problem with using `poetry-core` - as far as conda-forge is concerned, it is not really any different from using `setuptools`.

PR #29888, opened 2025-04-30
Add mamba-ssm

weiji14:
Mamba state-space model! Repo at https://github.com/state-spaces/mamba

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mamba-ssm/recipe.yaml
+++ b/recipes/mamba-ssm/recipe.yaml
@@ -1,0 +1,51 @@
+context:
+  name: mamba-ssm
+  version: 2.2.4
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  url: https://pypi.org/packages/source/${{ name[0] }}/${{ name }}/mamba_ssm-${{ version }}.tar.gz
+  sha256: e4114c69302796c91b71e90032c2d974f611608fab331582a80de6eaf075efb9
+
+build:
+  number: 0
+  noarch: python
+  script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  # CUDA < 12 not supported by pytorch anymore
+  skip: cuda_compiler_version == "11.8"
+
+requirements:
+  host:
+    - ninja
+    - packaging
+    - pytorch
+    - pip
+    - python >=3.9
+    - setuptools >=61.0.0
+    - wheel
+  run:
+    - einops
+    - packaging
+    - python >=3.9
+    - pytorch
+    - setuptools >=61.0.0
+    - transformers
+
+tests:
+  - python:
+      imports:
+        - mamba_ssm
+      pip_check: true
+
+about:
+  summary: Mamba state-space model
+  license: Apache-2.0
+  license_file: LICENSE
+  homepage: https://github.com/state-spaces/mamba
+
+extra:
+  recipe-maintainers:
+    - weiji14


weiji14:
There's an RuntimeError on the `import mamba_ssm` test:

```python-traceback
 Installing test environment
 ✔ Successfully updated the test environment
 + python $SRC_DIR/conda_build_script.py
   File "$SRC_DIR/conda_build_script.py", line 1, in <module>
     import mamba_ssm
   File "$PREFIX/lib/python3.12/site-packages/mamba_ssm/__init__.py", line 3, in <module>
     from mamba_ssm.ops.selective_scan_interface import selective_scan_fn, mamba_inner_fn
   File "$PREFIX/lib/python3.12/site-packages/mamba_ssm/ops/selective_scan_interface.py", line 16, in <module>
     from mamba_ssm.ops.triton.layer_norm import _layer_norm_fwd
   File "$PREFIX/lib/python3.12/site-packages/mamba_ssm/ops/triton/layer_norm.py", line 169, in <module>
     @triton.autotune(
      ^^^^^^^^^^^^^^^^
   File "$PREFIX/lib/python3.12/site-packages/triton/runtime/autotuner.py", line 352, in decorator
     return Autotuner(fn, fn.arg_names, configs, key, reset_to_zero, restore_value, pre_hook=pre_hook,
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "$PREFIX/lib/python3.12/site-packages/triton/runtime/autotuner.py", line 127, in __init__
     self.do_bench = driver.active.get_benchmarker()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "$PREFIX/lib/python3.12/site-packages/triton/runtime/driver.py", line 23, in __getattr__
     self._initialize_obj()
   File "$PREFIX/lib/python3.12/site-packages/triton/runtime/driver.py", line 20, in _initialize_obj
     self._obj = self._init_fn()
                 ^^^^^^^^^^^^^^^
   File "$PREFIX/lib/python3.12/site-packages/triton/runtime/driver.py", line 8, in _create_driver
     raise RuntimeError(f"{len(actives)} active drivers ({actives}). There should only be one.")
 RuntimeError: 0 active drivers ([]). There should only be one.
 × error Script failed with status 1
 × error Work directory: '/tmp/.tmpKdXIy2'
 × error To debug the build, run it manually in the work directory (execute the `./conda_build.sh` or `conda_build.bat` script)
```

Should have been fixed in https://github.com/triton-lang/triton/pull/6570 (see also https://github.com/triton-lang/triton/issues/6150). Could either skip the import check (import should technically work on a CUDA-enabled device), or wait for a new release of triton>3.2.0 with that patch.

weiji14:

weiji14:
Gentle ping again @conda-forge/help-python to see if this can be reviewed and merged :pray:

recipes/mamba-ssm/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+  sha256: e4114c69302796c91b71e90032c2d974f611608fab331582a80de6eaf075efb9
+
+build:
+  number: 0
+  script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  skip:
+    # mamba-ssm requires CUDA (or ROCm) GPU
+    - cuda_compiler_version == "None"
+    # CUDA < 12 not supported by pytorch anymore
+    - cuda_compiler_version == "11.8"
```
ocefpaf: I'm not a pytorch cuda user, but can't this be noarch?

weiji14: I think this CUDA-dependent package is Linux only. And actually, the reason I'm packaging this is to get builds working on arm64 which aren't available on the release page at https://github.com/state-spaces/mamba/releases/tag/v2.2.4, so it's probably not a noarch package.

ocefpaf: Shouldn't you need compilers then? @jakirkham or @isuruf you have more experience with cuda recipes. Can you chime in and help us?

recipes/mamba-ssm/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+requirements:
```
danielnachun, line 25:
```suggestion
requirements:
  build:
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
    - {{ compiler('cxx') }}
    - {{ compiler('cuda') }}
```
Since this package appears have CUDA code in https://github.com/state-spaces/mamba/tree/main/csrc/selective_scan, it definitely needs at a minimum these build dependencies, and well as some others in `host`.

recipes/mamba-ssm/recipe.yaml
```diff
@@ -0,0 +1,58 @@
...
+  host:
```
danielnachun, line 31:
```suggestion
  host:
    - cuda-version {{ cuda_compiler_version }}
    - cuda-driver-dev
    - cuda-cudart-dev
```
Probably you will also need some other libraries like `libcublas`, but this is enough to get the basic CUDA dependencies in.

danielnachun (changes requested):
I made some suggestions for what I think are the basic CUDA requirements but there will probably be more needed.

recipes/mamba-ssm/recipe.yaml
weiji14: Hmm, those files under `csrc/selective_scan/*` doesn't seem to be packaged into the sdist at https://pypi.org/packages/source/m/mamba-ssm/mamba_ssm-2.2.4.tar.gz (which is why the build passed without these compilers). I might need to pull in from the GitHub source instead, gimme a few more hours.

danielnachun: That's kind of weird, let's see what happens if you use the GitHub source instead.

weiji14: Actually, I had a look at the logs here - https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1280949&view=logs&j=df9408f7-e795-500f-143d-aa7735a0ed8b&t=47addc80-ccbf-566b-ee3a-66cdb8436937&l=3772, and noticed that the built `linux-64/mamba-ssm-2.2.4-py312h8fadf9f_0.conda` package is 255.6MB in size, and there is this large shared library file:

```
?rwxr-xr-x 0/0  444626864 2025-07-04 01:08:03 lib/python3.12/site-packages/selective_scan_cuda.cpython-312-x86_64-linux-gnu.so 
```

So maybe we don't need to pull from GitHub source, since the `.so` file is already there? It seems that there is some logic at https://github.com/state-spaces/mamba/blob/v2.2.4/setup.py#L39-L42 to pull from the prebuilt wheels on their GitHub releases page.

I can force a fresh build from source if necessary, just depends on what's the proper way to handle this situation.

weiji14: @jakirkham or @carterbox, could you suggest if a rebuild from source is necessary here, or can we just reuse the `.so` file from the PyPI wheel?

danielnachun: Binaries should always be built from source.  This is important for both licensing and dependency tracking.

weiji14: Ok, I've switched to building from source now by setting the `MAMBA_FORCE_BUILD=TRUE` environment variable (following this [line](https://github.com/state-spaces/mamba/blob/e0761ece1db07e0949dd88b4f4cd440420a19fd9/.github/workflows/publish.yaml#L146)). Could you take another look once the CI builds complete?

recipes/mamba-ssm/recipe.yaml
weiji14: I added `${{ compiler('cuda') }}`, `cuda-version` and `triton`, and that seemed to be enough to get it to compile.

recipes/mamba-ssm/recipe.yaml
danielnachun: Thank you for building from source.  This looks good now.

recipes/mamba-ssm/recipe.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - pytorch
+    - setuptools >=61.0.0
+    - transformers
+
+tests:
+  - python:
+      imports:
+        - mamba_ssm
+      # disabled pip check because ninja is not available
+      # see https://github.com/conda-forge/ninja-feedstock/issues/26
```
danielnachun, line 56: Can we add `ninja` as a test-only dependency here?

weiji14: Don't think there's a way to add test-only dependencies with v1 recipes for python tests - https://rattler.build/latest/reference/recipe_file/#python-tests, only able to set the python-version.

weiji14: Gentle nudge again @danielnachun, are you ok with me keeping `pip check` disabled, or would you prefer another solution from https://github.com/conda-forge/ninja-feedstock/issues/26?

danielnachun: At this point you can probably just patch out `ninja` from the `pyproject.toml` - if `pip check` doesn't work for this package, it also won't work for any Python package that ever depends on this, which is more of a concern.   A simple `sed -i 's/"ninja",//g' pyproject.toml` should hopefully let `pip check` pass.

weiji14: Ok, I've added the patch, and Python 3.10 build's `pip check` seems to have passed. Just need to wait for the other matrix CI builds to finish.

weiji14:
Linter error seems to be due to https://github.com/conda-forge/staged-recipes/issues/30936, and windows build error can be ignored since this recipe is for CUDA-enabled Linux builds only.

recipes/mamba-ssm/recipe.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+build:
+  number: 0
+  script:
+    env:
+      MAMBA_FORCE_BUILD: "TRUE"
+    content: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  skip:
+    # mamba-ssm requires CUDA (or ROCm) GPU
+    - cuda_compiler_version == "None"
```
danielnachun, line 23:
```suggestion
  skip:
    # mamba-ssm requires CUDA (or ROCm) GPU
    - cuda_compiler_version == "None"
    - not linux
```
I think we should also just be skipping macOS and Windows altogether here since this requires a CUDA compiler.

danielnachun:
> Linter error seems to be due to #30936, and windows build error can be ignored since this recipe is for CUDA-enabled Linux builds only.

That does appear to be unrelated, but let's give the upstream team a chance to respond since that issue so new just to make sure there isn't some weird cryptic issue with the recipe itself here.

danielnachun:
My apologies but it appears a separate PR for this was already merged: https://github.com/conda-forge/staged-recipes/pull/30861.  In that case we'll have to close this, but please feel free to add yourself as a maintainer to the feedstock at https://github.com/conda-forge/mamba-ssm-feedstock.

weiji14:
Oh great, I'll try and bring the changes from here into https://github.com/conda-forge/mamba-ssm-feedstock then. Thanks again for your time.

PR #29891, opened 2025-04-30
Add multirtc package

forrestfwilliams:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/multirtc/meta.yaml
+++ b/recipes/multirtc/meta.yaml
@@ -1,0 +1,61 @@
+{% set name = "multirtc" %}
+{% set version = "0.1.0" %}
+{% set python_min = "3.9" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/multirtc-{{ version }}.tar.gz
+  sha256: 526487600722b67504fca5339d2d609141c8e08f4dfae5bd1be20bf61a9b201f
+
+build:
+  entry_points:
+    - prep_burst = multirtc.prep_burst:main
+    - multirtc = multirtc.multirtc:main
+  noarch: python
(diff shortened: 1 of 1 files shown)


forrestfwilliams:
I `forrestfwilliams` acknowledge that I will be the maintainer of this package.

forrestfwilliams:
@conda-forge/help-python The recipe has passed all checks and is ready for review - thanks!

I'm not using a `noarch` for this recipe, because the key dependency ([isce3](https://anaconda.org/conda-forge/isce3)) is not available on Windows. I believe I've designed the recipe correctly for this scenario, but let me know if there is anything that needs to be changed.

recipes/multirtc/meta.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/multirtc-{{ version }}.tar.gz
+  sha256: 526487600722b67504fca5339d2d609141c8e08f4dfae5bd1be20bf61a9b201f
+
+build:
+  entry_points:
+    - prep_burst = multirtc.prep_burst:main
+    - multirtc = multirtc.multirtc:main
+  skip: true  # [win or py<39]
```
ocefpaf, line 16: Looks like this one can be noarch, no?

forrestfwilliams: Ok, sounds good - l'll convert to noarch and expect the windows build to fail.

forrestfwilliams:
@conda-forge/help-python The recipe has passed all checks (except windows as discussed above) and is ready for a re-review. Thanks!

recipes/multirtc/meta.yaml
```diff
@@ -0,0 +1,61 @@
+{% set name = "multirtc" %}
+{% set version = "0.1.0" %}
+{% set python_min = "3.9" %}
```
ocefpaf, line 3: We only need to set `python_min` if the min version is higher than conda-forge's min. 

```suggestion
```

PR #29893, opened 2025-04-30
add adam core

akoumjian:
- **Add conda-forge recipe for adam_core**
- **Update with python min, correct license**

### DIFF
--- a/recipes/adam_core/meta.yaml
+++ b/recipes/adam_core/meta.yaml
@@ -11,7 +11,7 @@ source:
   sha256: a1b9cf9048e4cc578b35a3ee42076af52eed35571e6d1db4af5c144445ae3cf6
 
 build:
-  noarch: python
+  skip: true  # [win]
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
 
--- a/recipes/adam_core/meta.yaml
+++ b/recipes/adam_core/meta.yaml
@@ -13,15 +13,15 @@ source:
 build:
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
-  skip: true  # [win]
+  skip: true  # [win or py<311]
 
 requirements:
   host:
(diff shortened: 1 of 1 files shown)


moeyensj:
I agree to help maintain this conda-forge package and related feedstock.

ntellis:
I agree to help maintain this conda-forge package and related feedstock.

akoumjian:
@conda-forge/python Ready for review!

This package skips windows as it appear some of the underlying dependencies are not available. 
My understanding is I had to remove the `noarch` in order to do the windows skip.
Without `noarch`, it appears I am not allowed to set the `python_min` pinning in the same way.
I believe using the `py<311` skip is the correct way to do this in this circumstance when a minimum python version is still required, but I welcome feedback.

recipes/adam_core/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/adam_core-{{ version }}.tar.gz
+  sha256: a1b9cf9048e4cc578b35a3ee42076af52eed35571e6d1db4af5c144445ae3cf6
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: true  # [win or py<311]
```
ocefpaf, line 16: This one can probably be noarch. What is the reason to skip Windows?

akoumjian: From my previous comment:
> This package skips windows as it appear some of the underlying dependencies are not available.
My understanding is I had to remove the noarch in order to do the windows skip.
Without noarch, it appears I am not allowed to set the python_min pinning in the same way.
I believe using the py<311 skip is the correct way to do this in this circumstance when a minimum python version is still required, but I welcome feedback.

I am happy to revert back to noarch and let the windows build fail due to missing dependencies. I didn't think it would get merged in with a failing build.

PR #29899, opened 2025-05-01
Add r-mockr

matthewfeickert:
* Add the mockr R package.
   - c.f. https://github.com/krlmlr/mockr

Recipe generated using https://github.com/bgruening/conda_r_skeleton_helper

Needed to test `r-sandpaper` (c.f. PR https://github.com/conda-forge/staged-recipes/pull/29866)

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/r-mockr/meta.yaml
+++ b/recipes/r-mockr/meta.yaml
@@ -1,0 +1,66 @@
+{% set version = '0.2.2' %}
+{% set posix = 'm2-' if win else '' %}
+
+package:
+  name: r-mockr
+  version: {{ version|replace("-", "_") }}
+
+source:
+  url:
+    - {{ cran_mirror }}/src/contrib/mockr_{{ version }}.tar.gz
+    - {{ cran_mirror }}/src/contrib/Archive/mockr/mockr_{{ version }}.tar.gz
+  sha256: 08199520c6590e19307b5e6cbfa79fbdbe0ea6ecb690957f31222d4bf8d3792d
+
+build:
+  number: 0
+  noarch: generic
+  rpaths:
+    - lib/R/lib/
+    - lib/
+  missing_dso_whitelist:
+    - '*/R.dll'        # [win]
+    - '*/Rblas.dll'    # [win]
+    - '*/Rlapack.dll'  # [win]
+
+requirements:
+  build:
+    - {{ posix }}zip               # [win]
+    - cross-r-base {{ r_base }}    # [build_platform != target_platform]
+  host:
+    - r-base
+    - r-rlang
+    - r-withr
+  run:
+    - r-base
+    - r-rlang
+    - r-withr
+
+test:
+  requires:
+    - r-testthat
+  source_files:
+    - tests/
+  commands:
+    - $R -e "library('mockr')"  # [not win]
+    - $R -e "testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)"  # [not win]
+    - "\"%R%\" -e \"library('mockr')\""  # [win]
+    - "\"%R%\" -e \"testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)\""  # [win]
+
+about:
+  home: https://krlmlr.github.io/mockr/
+  dev_url: https://github.com/krlmlr/mockr
+  doc_url: https://krlmlr.github.io/mockr/
+  license: MIT
+  summary: Drop-in replacement for testthat::with_mock()
+  description: |
+   Provides a means to mock a package function, i.e., temporarily substitute it for testing.
+   Designed as a drop-in replacement for the now deprecated 'testthat::with_mock()'
+   and 'testthat::local_mock()'.
+  license_family: MIT
+  license_file:
+    - '{{ environ["PREFIX"] }}/lib/R/share/licenses/MIT'
+
+extra:
+  recipe-maintainers:
+    - conda-forge/r
+    - matthewfeickert
--- a/recipes/r-mockr/bld.bat
+++ b/recipes/r-mockr/bld.bat
@@ -1,0 +1,2 @@
+"%R%" CMD INSTALL --build . %R_ARGS%
+IF %ERRORLEVEL% NEQ 0 exit /B 1
--- a/recipes/r-mockr/build.sh
+++ b/recipes/r-mockr/build.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+export DISABLE_AUTOBREW=1
+${R} CMD INSTALL --build . ${R_ARGS}


matthewfeickert:
:wave: @krlmlr As this PR will add `mockr` to conda-forge I wanted to check to see if you'd like to be a maintainer on it. There's no expectation of work, but whenever I add other people's packages I like to see if they'd like to be maintainers so that they can retain control over distributions of their software.

matthewfeickert:
@conda-forge/help-r, ready for review!

recipes/r-mockr/meta.yaml
```diff
@@ -0,0 +1,66 @@
...
+    - $R -e "library('mockr')"  # [not win]
+    - $R -e "testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)"  # [not win]
+    - "\"%R%\" -e \"library('mockr')\""  # [win]
+    - "\"%R%\" -e \"testthat::test_file('tests/testthat.R', stop_on_failure=TRUE)\""  # [win]
+
+about:
+  home: https://krlmlr.github.io/mockr/
+  dev_url: https://github.com/krlmlr/mockr
+  doc_url: https://krlmlr.github.io/mockr/
+  license: GPL-3.0-or-later
```
cbrueffer, line 53:
```suggestion
  license: GPL-3.0-only
```

cbrueffer (changes requested):
Looks good, apart from the small license fix.

matthewfeickert:
Thanks @cbrueffer. General question about GPL licenses: How do you tell quickly if it is [GPL-3.0-only](https://spdx.org/licenses/GPL-3.0-only.html) or [GPL-3.0-or-later](https://spdx.org/licenses/GPL-3.0-or-later.html)? Or do you personally just know that `r-base`'s `{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'` is GPL-3.0-only?

cbrueffer:
@matthewfeickert we go by the License string in the Description files. For example `r-mockr` defines `GPL-3` (doesn't say anything about later GPL versions), while e.g. `r-dosearch` explicitly defines `GPL (≥ 3)`. It's a recurring problem with `conda_r_skeleton_helper` expanding the license string wrongly, we should really fix that :-)

matthewfeickert:
> @matthewfeickert we go by the License string in the Description files. For example `r-mockr` defines `GPL-3` (doesn't say anything about later GPL versions), while e.g. `r-dosearch` explicitly defines `GPL (≥ 3)`. It's a recurring problem with `conda_r_skeleton_helper` expanding the license string wrongly, we should really fix that :-)

@cbrueffer In that case, using `'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'` is wrong then as that is `GPL-3.0-or-later`.

cbrueffer:
@matthewfeickert correct me if I'm wrong, but that particular part of the license refers to which language to use when licensing a piece of software under a specific GPL version, not that the piece of software this license file came with can be used with "any later version".

matthewfeickert:
> but that particular part of the license refers to which language to use when licensing a piece of software under a specific GPL version, not that the piece of software this license file came with can be used with "any later version".

Sorry, I think there's too many pronouns here for me to be clear what part in particular you're referring to. Though I appreciate you taking the time to try to help me understand things more.

To clarify my point/opinion, "GPL-3" without "only" or "or later" is ambiguous and should not be used.

From

> go by the License string in the Description files. For example r-mockr defines GPL-3 (doesn't say anything about later GPL versions)

then one would need an explicit `GPL-3.0-only` file in `'{{ environ["PREFIX"] }}/lib/R/share/licenses/GPL-3'`. There is only a `GPL-3` file that contains the license text that exists in both `GPL-3.0-only` and `GPL-3.0-or-later` (they're the same all the way up to the Standard License Header section). SPDX has the (very easy to miss IMO) note in both pages (with the other license name flipped)

> This license was released: 29 June 2007. This license identifier refers to the choice to use the code under GPL-3.0-only, as distinguished from the use of code under GPL-3.0-or-later (i.e., GPL-3.0 or some later version). **The license notice (as seen in the Standard License Header field below) states which of these applies to the code in the file.** The example in the How to Apply These Terms appendix of the license shows the "or later" approach.

So without the Standard License Header, there's no distinction in the text, and this means that people are now going to have to be explicit in every case if they want `GPL-3.0-only` and `GPL-3.0-or-later` in their Description files.

I'm obviously not a copyright lawyer, and so I might be missing something super obvious, but it seems like the license selection is ambiguous and the default is to go to `GPL-3.0-only`. I don't think there should be a default.

cbrueffer:
I'm no lawyer either, so no worries ;-) Certainly agreed that it would be preferable to have CRAN / R Description files use SPDX identifiers directly, so we don't have to interpret things downstream.

I think I understand your issue now. So you'd prefer license versions to be shipped that have the `Standard License Header` bits mentioned on the SPDX sites attached (which are lacking from the `GPL-3` r-base ships), and have two file name versions to distinguish them? Anything is possible, but this discussion would be better placed on the `r-base` feedstock which is kind of the "central place" for R discussions, not here.

For the time being, the approach currently adopted by `conda_r_skeleton_helper` is what we use across the whole conda-forge R ecosystem, so it should be followed here as well. Any change to this, as the one you suggest, would need to be adopted across the whole conda-forge R ecosystem, e.g. through a migration.

krlmlr:
Thanks for the heads-up. Please go ahead, I'm fine with CRAN and GitHub as the canonical source of my packages, with other repos pulling from there as needed.

matthewfeickert:
Thanks @krlmlr. For the purposes of license clarification, can you let us know (either here or on https://github.com/krlmlr/mockr/issues/55) if you intended the use of the GPL-3.0 license text to be [GPL-3.0-only](https://spdx.org/licenses/GPL-3.0-only.html) or [GPL-3.0-or-later](https://spdx.org/licenses/GPL-3.0-or-later.html)?

> Please go ahead, I'm fine with CRAN and GitHub as the canonical source of my packages, with other repos pulling from there as needed.

Great! One last clarification question: does "go ahead" mean "go ahead and add me as a maintainer" or "go ahead and continue on without my involvement in the distribution on conda-forge"?

krlmlr:
Thanks for the follow-up. I'm fine with other repos pulling from CRAN/GitHub as needed, and don't need to be involved.

matthewfeickert:
> Thanks for the follow-up. I'm fine with other repos pulling from CRAN/GitHub as needed, and don't need to be involved.

Thanks for the clarification!

Given that `mockr` relicensed to MIT given this question getting brought up (https://github.com/krlmlr/mockr/issues/55#issuecomment-2845074197) I'll wait for `v0.2.2` to get processed on CRAN and then update this PR to be use that version so that GPL doesn't ever have to enter the `r-mockr` distributions. :+1: 

Thanks for all your patience and explanations, @cbrueffer. I'll follow up in the more appropriate locations as you suggest. :+1:

matthewfeickert:
@conda-forge/help-r, this is now ready for review after a version bump with a corresponding license change.

mfansler (approved):

PR #29943, opened 2025-05-06
VisionWorkbench feedstock

oleg-alexandrov:
Checklist
- [x] Build number is 0.


oleg-alexandrov:
@conda-forge-admin : This is ready for review.

oleg-alexandrov:

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,182 @@
...
+    - tbb-devel
+    - xerces-c >=3.2.4,<3.3.0a0
+    - xorg-kbproto
+    - xorg-libice
+    - xorg-libsm
+    - xorg-libx11
+    - xorg-libxi
+
+  run:
+    - libgdal
```
hobu: If this were trimmed down to `libgdal-core` it might help cut down the number of packages this pulls in. Same for `libpdal-core`...

Feel free to add @hobu as a maintainer here too. I can help with this once it's a feedstock.

oleg-alexandrov: Noted, thank you. I left this one for dead. 

I will circle back and add you as maintainer as well once the overlords review and pull it in.

oleg-alexandrov: Unless you are part of the club and can help.

hobu: I can help with osx-arm64 builds too once it is up.

hobu: > Unless you are part of the club and can help.

No I can't create feedstocks, but I do maintain a bunch of the geo-related ones. @ocefpaf has been extremely invaluable helping shepherd all of us to do that.

oleg-alexandrov: Thanks. I trimmed the dependencies to gdal-core an pdal-core.

ocefpaf:
@oleg-alexandrov there is no re-render in staged recipes. Also, I'm not really a cpp reviewer, but I can a first pass to see if we can get the attention of someone with more experience.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# Fetch visionworkbench
+cd $SRC_DIR
+wget https://github.com/visionworkbench/visionworkbench/archive/refs/tags/3.5.1.tar.gz > /dev/null 2>&1 # this is verbose
```
ocefpaf: Why not use the source fetched from the recipe?
Also, we should not hardcode the version. We should use the one defined in the recipe. Here is would be `$PKG_VERSION` if I 'm not mistaken. Still, that is not recommended. Please try to use conda's fetched source from the recipe.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,56 @@
...
+tar xzf 3.5.1.tar.gz
+
+echo List all files
+ls -d *
```
ocefpaf: This is not necessary and there are many commented out lines here that should be removed/cleaned if they do not serve any purpose.

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,178 @@
...
+build:
+  number: 0
+  skip: true  # [win]
+  
+requirements:
+  build:
+    - cmake >=3.15
+    - make
+    - c-compiler 1.8.0
+    - cxx-compiler 1.8.0
```
ocefpaf: This is not allowed. Please use `{{ compiler('c') }}` instead. Specifying the compiler like that will break the recipe and may render the package incompatible with the rest of conda-forge.

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,178 @@
...
+    - qt-main >=5.15,<5.16
+    - qwt <6.3.0
+    - tbb
+    - tbb-devel
+    - xerces-c >=3.2.4,<3.3.0a0
+    - xorg-kbproto
+    - xorg-libice
+    - xorg-libsm
+    - xorg-libx11
+    - xorg-libxi
```
ocefpaf: Are those really needed?

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,178 @@
...
+
+test:
+  commands:
+    - echo success
```
ocefpaf:
```suggestion
```

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,178 @@
...
+    - test -f $PREFIX/bin/hillshade  # [unix]
```
ocefpaf: Why not test its functionality instead of just existence?

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Thank you for the advice. 

This is the trickiest one. New clang (2020 or so) breaks a lot of code in some dependencies. If I have any confidence conda-forge people will in fact engage on this, I'll find time to take a look. 

For now there are weeks of silence, and when I merge to latest in conda-forge, stuff that used to work just breaks. For example, see the latest run, pdal-core and gdal-core do not get along. https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1262218&view=logs&jobId=4cabe686-70ae-553a-7fd0-310379f2cbac&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff

This is a painful process and I'll wait for folks with authority to merge in to seriously engage before I continue trying to keep up.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Adressed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: There are two recipes here: visionworkbench, and stereo-pipeline. The second depends on the first. For now, in second recipe I do things that way as the first one is not yet a conda package it can be just added as a dependency.

The first recipe, visionworkbench, implements things correctly. So maybe one can defer looking at the second one till the first one is reviewed and accepted.

ocefpaf: Conda knows how to build them in the correct order here. You don't need to do this.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Switched to {{ compiler('cxx') }} for visionworkbench. The bigger problems show up with the second recipe, stereo-pipeline, that I'd like to merge too, but that can wait.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I addressed this by deleting the problematic build.sh logic in the stereo-pipeline recipe. For now I hope the base visionworkbench recipe is adopted, and then it will be used as a prerequisite for the other one.

I spent a few hours dealing with outrageous build problems. The libgdal-core package cannot be reconciled with the proj package. That is quite a feat, I may say. Once I added the larger libgdal package, that conflicts with the hdf5 package. 

All that with no package versions pinned. The conda ecosystem is more terrible than usual. I leave it at that, for now.

hobu: > All that with no package versions pinned. The conda ecosystem is more terrible than usual. I leave it at that, for now.

Careful pins can help. `libgdal-core`, `proj`, and `hdf5` are packages with tons of dependencies and opportunity for conflicts. There's also a PROJ release and GDAL package rendering going on due to a recent bug, so that might explain some of it.

oleg-alexandrov: Good advice. It worked with libgdal 3.9. It is surprising conda is unable to find a compatible set of dependencies without explicitly pinning one of them. Its whole purpose is to help with such problems.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: I went back to existence test.  The test for functionality returns status 1 unless called with an actual dataset to work on. Test for existence may do. These are C++ tools, so they get tested plenty at link time.

oleg-alexandrov:
@conda-forge/help-c-cpp, @conda-forge-admin Any updates? There have been a couple of passes of reviews above, and issues have been addressed. Hoping soembody qualified to merge in can review and offer more thoughts.

oleg-alexandrov:

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+    - libzlib
+    - zlib
+    - libboost-devel
+    - libjpeg-turbo
+    - libtiff
+    - libpng
+    - lz4-c
+    - libstdcxx  # [linux]
+    - libgcc  # [linux]
+    - libcxx  # [osx]
```
ocefpaf: You should not need these.

recipes/visionworkbench/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+  skip: true  # [win]
+
+requirements:
+  build:
+    - cmake >=3.15
+    - make
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - wget
```
ocefpaf:
```suggestion
```

recipes/visionworkbench/build.sh
```diff
@@ -0,0 +1,15 @@
...
+mkdir build
+cd build
+cmake ..                             \
+    -DCMAKE_PREFIX_PATH=${PREFIX}    \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DASP_DEPS_DIR=${PREFIX}         \
+    -DUSE_OPENEXR=OFF                \
+    -DCMAKE_VERBOSE_MAKEFILE=ON
+make -j${CPU_COUNT}
+make install
```
ocefpaf, line 14: Is there a `make check` or `ctest`?

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,15 @@
...
+cd build
+cmake ..                             \
+    -DCMAKE_PREFIX_PATH=${PREFIX}    \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DASP_DEPS_DIR=${PREFIX}         \
+    -DUSE_OPENEXR=OFF                \
+    -DCMAKE_VERBOSE_MAKEFILE=ON
+make -j${CPU_COUNT}
+make install
+
```
ocefpaf: Is there a `make check` or `ctest`?

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+  skip: true  # [win]
+
+requirements:
+  build:
+    - cmake >=3.15
+    - make
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - wget
```
ocefpaf:
```suggestion
```

recipes/visionworkbench/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+    - libzlib
+    - zlib
+    - libboost-devel
+    - libjpeg-turbo
+    - libtiff
+    - libpng
+    - lz4-c
+    - libstdcxx  # [linux]
+    - libgcc  # [linux]
+    - libcxx  # [osx]
```
ocefpaf:
```suggestion
```

recipes/visionworkbench/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+    # - python
+    # - python_abi
+    - qt-main
+    - qwt
+    - tbb
+    - tbb-devel
+    - xerces-c
+
+  run:
+    - libgdal  3.9
```
ocefpaf: If you pin gdal like that it will be impossible to use this package with others in conda-forge. Why not leave the conda-forge default pin?

recipes/visionworkbench/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+    - libzlib
+    - zlib
+    - libboost-devel
+    - libjpeg-turbo
+    - libtiff
+    - libpng
+    - lz4-c
+    - libstdcxx  # [linux]
+    - libgcc  # [linux]
+    - libcxx  # [osx]
```
ocefpaf:
```suggestion
```

recipes/stereo-pipeline/meta.yaml
```diff
@@ -0,0 +1,168 @@
...
+    - qt-main
+    - qwt
+    - tbb
+    - tbb-devel
+    - xerces-c
+
+test:
+  commands:
+    # Do not run this, as it it returns status code 1. Check for existence only.
+    - test -f $PREFIX/bin/hillshade
```
ocefpaf: Both recipes are building the same thing? I'm confused.

oleg-alexandrov:
@ocefpaf: Thank you for the additional input. I would like to confirm if you have the authority to merge this in, if all review issues are addressed. A few days ago I already spent more than half a day on this (mostly due to conda itself having problems), and hoping that if the recipe is improved enough this would be mergeable.

xhochy:
@ocepaf has the authority to merge it, but once you have addressed all concerns, we will review. Only once the recipe is in a good state, we will merge. We cannot make promises on the point of time but @ocefpaf already brings you very close.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Adressed.

recipes/visionworkbench/meta.yaml
oleg-alexandrov: Adressed.

recipes/visionworkbench/build.sh
oleg-alexandrov: I have a very large regression test suite that runs daily (https://github.com/NeoGeographyToolkit/StereoPipelineTest).

I could put a token unit test, I guess, but I hope this is not a show-stopper. The conda-forge build is not a good place for where to test this software.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed above for the other recipe.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Addressed.

recipes/visionworkbench/meta.yaml
oleg-alexandrov: Addressed.

recipes/visionworkbench/meta.yaml
oleg-alexandrov: The reason for this is because recent GDAL fails to find a compatible package with HDF5. See right below (https://github.com/conda-forge/staged-recipes/pull/29943/commits/f7a5f306a43fc5f3bee1b315103d72691845db06), where I tried right today to remove the pin, and it failed again.

I think for now GDAL better be pinned while those folks figure how to release a more recent version that is compatible with other stuff. 

This is beyond my control, and is a temporary measure.

recipes/visionworkbench/meta.yaml
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: This is because you did not like me fetching the code of one package to build the other. This recipe has two proposed conda packages, and one depends on the other. I could not figure out the clever way conda-forge can handle that.

For now, I eliminated the second package (stereo-pipeline), and I hope we can proceed with the first one (visionworkbench), where I addressed your concerns.

oleg-alexandrov:
@xhochy: I expect no promises, of course. Just hoping this will get reviewed and merged in if looks good. I addressed @ocefpaf 's comments.

recipes/stereo-pipeline/meta.yaml
ocefpaf: > This is because you did not like me fetching the code of one package to build the other. This recipe has two proposed conda packages, and one depends on the other. I could not figure out the clever way conda-forge can handle that.

I'm a bit lost b/c of the many changes here. Conda will know which one depend on on the other and build them in topological order for you. Unless there are circular dependencies? If not, you can build both in their own recipe. What I recall when I added this comment is that both recipes where too similar and tested for the same things.

recipes/visionworkbench/meta.yaml
ocefpaf: Let's try that again. There where a few iterations of gdal since then.

recipes/stereo-pipeline/meta.yaml
oleg-alexandrov: Getting even one package to build successfully on conda-forge has been hard enough on its own, without trying the interdependency feature. 

The hope is that the advice you gave here will also carry over to the second package, and perhaps you will be willing to help push that one to the finish line when this one is done. The second package has a handful more dependencies but none bigger than GDAL.

recipes/visionworkbench/meta.yaml
oleg-alexandrov: No luck with the unpin.

hobu: Looks like it is hdf that needs the pin, but in the end this is going to control the GDAL version too.

recipes/stereo-pipeline/meta.yaml
ocefpaf: Just be sure to cross-reference the next PR to this one so we can clarify things a bit. 48 commits is too much for me to inspect at the moment.

recipes/visionworkbench/build.sh
ocefpaf: > The conda-forge build is not a good place for where to test this software.

I disagree. Any third party packaging system, like conda-forge, should run as many tests as we can. Even more so for compiled stuff. With that said, my question was "are there tests?" It seems that the answer is not, right?

ocefpaf:
> @ocepaf has the authority to merge it, but once you have addressed all concerns, we will review. Only once the recipe is in a good state, we will merge. We cannot make promises on the point of time but @ocefpaf already brings you very close.

@xhochy if you have some free cycles I'd love your opinion on this recipe, specially on the use of of libblas, etc.

@oleg-alexandrov at the moment the only test here is if `hillshade` exists. It would be nice if we could add some real tests. maybe a CLI one? I'm it sure what this software does to help you but, if you need some data to run it, see the [fiona feedstock for some inspiration](https://github.com/conda-forge/fiona-feedstock/blob/main/recipe/meta.yaml#L64-L65) on how to do it.

recipes/visionworkbench/build.sh
oleg-alexandrov: Test of running one of the tools added.

oleg-alexandrov:
A real CLI test added. It will load a image that is shipped together with the package and apply processing relevant to this package. 

GDAL pinned back for now as it fails anyway, and this way can see if the newly added test passes.

recipes/visionworkbench/meta.yaml
```diff
@@ -118,8 +118,7 @@ requirements:
 
 test:
   commands:
-    # Do not run this, as it it returns status code 1. Check for existence only.
-    - test -f $PREFIX/bin/hillshade
+    - $PREFIX/bin/hillshade $PREFIX/share/vw/tests/tinyDemAN.tif
```
ocefpaf: If you need to specify `$PREFIX`  this is in the wrong place. The CLI should be in the test path and calling `hillshade` should suffice.

ocefpaf:
> GDAL pinned back for now as it fails anyway, and this way can see if the newly added test passes.

For future reference this is the conflict in the logs:

```
  - package libgdal-3.11.0-h3b705f5_5 requires libgdal-netcdf 3.11.0.*, but none of the providers can be installed

├─ hdf5 =1.14.3 * is requested and can be installed;
└─ libgdal =3.11 * is not installable because it requires
   └─ libgdal-netcdf =3.11.0 *, which requires
      └─ hdf5 >=1.14.6,<1.14.7.0a0 *, which conflicts with any installable versions previously reported.

```

We need to investigate what package is requiring incompatible versions of netcdf/hdf5 here.

oleg-alexandrov:
@ocefpaf: I followed other conda recipes, that suggest there should be a sanity check to validate that the package builds and runs correctly.

Because of your repeated insistence, which likely was a good thing, I have an executable from my package run with a test dataset from my package, and it returns status 0, which means the tool was able to correctly load and process a dataset.

I will propose you do not let the good be the enemy of the perfect. I spent a day in total going through your suggestions, and we have a usable recipe with an executable that runs and validates the software.

ocefpaf:
> I will propose you do not let the good be the enemy of the perfect. I spent a day in total going through your suggestions, and we have a usable recipe with an executable that runs and validates the software.

The amount of dependencies that may not be necessary, hence request for more testing, can create resolution conflicts rendering this package uninstallable outside of a single environment that is identical to the build env. I made a few modifications and will push them here. If you agree, we can merge, if not you can revert them and we can request another reviewer for you here. Mostly I trimmed down the dependencies and update it to v1 to get faster builds.

I could not push to your branch, so I created a new PR in https://github.com/conda-forge/staged-recipes/pull/30354

oleg-alexandrov:
I appreciate the extra effort you put in. 

I would like to note that the test that I have now (hillshade) is not related to the issue of how many dependencies the package has.

I think we can leave the test alone, and work on reducing the number of dependencies. 

Looks like it failed on Mac because of removing libopenblas.

You switched from meta.yaml to recipe.yaml even though that seems not relevant to the problem of fewer dependencies. 

You added bliss package even though I do not use it. 

You added the openexr package even though the software is configured to work without it.

I reduced the number of dependencies, inspired by your proposed fixes. I need all of them for the second package, stereo-pipeline, but that's for another day.

ocefpaf:
> I appreciate the extra effort you put in.
> 
> I would like to note that the test that I have now (hillshade) is not related to the issue of how many dependencies the package has.
> 
> I think we can leave the test alone, and work on reducing the number of dependencies.

Any test should fail if we remove something it needs to pass. That is the idea here.
 
> Looks like it failed on Mac because of removing libopenblas.

I could not see anything linking to it o Linux. I'll try to debug this on macos.

> You switched from meta.yaml to recipe.yaml even though that seems not relevant to the problem of fewer dependencies.

It is. Ready my comment on speed of testing. Also, this use rattler-build instead of conda-build, allowing me to inspect it closer.

> You added bliss package even though I do not use it.

It does link to it. In your case, b/c you have a lot of dependencies, that is happening indirectly.

> You added the openexr package even though the software is configured to work without it.

As I said I followed the dependencies in the README. Is there a reason to not add that fuctionality?

> I reduced the number of dependencies, inspired by your proposed fixes. I need all of them for the second package, stereo-pipeline, but that's for another day.

Can you please share that recipe with me? I would like to understand why those cannot be in that package.

oleg-alexandrov:
I added you as a maintainer to this fork. Given the recent interactions, I think you will feel more comfortable doing things as you see fit. As long as this builds properly and is merged in, all is good with me. 

I will have you as maintainer once this is an official recipe as well.

>Any test should fail if we remove something it needs to pass. That is the idea here.

Great. You can make it pass without $PREFIX.

>I could not see anything linking to it o Linux. I'll try to debug this on macos.

I think you can allow for now all "blas" packages even they look redundant. There is "libblas", "openblas", etc. These are small packages, we can remove some later.

> Is there a reason to not add that fuctionality?

Recent OpenEXR changed many internal APIs and my stuff no longer builds with it. We don't use OpenEXR much, as OpenCV and GDAL are enough, so I turned it off.

>Can you please share that recipe with me? I would like to understand why those cannot be in that package.

The extra dependencies can be in the extra package and are not needed here. I wanted to see if conda can build an env with all of them. That is not guaranteed, as you saw even with gdal and hdf5.

Thank you for your help. You can push directly to this repo, and let us see how it works.

ocefpaf:
> That is not guaranteed, as you saw even with gdal and hdf5.

What I've been trying to say is that those dependencies should be in the other package and both should be in this PR. Then we can access what can or cannot be solved correctly and using conda-forge's pins.

oleg-alexandrov:
Indeed, eventually both of these need to be on conda-forge. It looks to me that even making the first one is turning out to be months of work and too many iterations.

I hope we can get this done. Then we'll see.

hobu:
> those dependencies should be in the other package and both should be in this PR

So it is ok to do multiple packages in the same staged-recipe PR? I wasn't aware that's possible. So, we should have both a `visionworkbench` and a `stereopipeline` package provided by this PR

oleg-alexandrov:
This all good with me. The problem is that the process is highly onerous. The other package, stereopipeline, has a lot more dependencies, and is much larger as well. 

My time is limited, the response from reviewers is rare, the list of things to do keeps on growing, and conda-forge is very bad at solving for dependencies.

The visionworkbench package is a self-contained image library. Having this in the channel would already go a long way towards me hoping this is going anywhere at all. 

Then, one could focus all effort on second package, rather than trying to solve two problems at once.

ocefpaf:
> I added you as a maintainer to this fork. Given the recent interactions, I think you will feel more comfortable doing things as you see fit. As long as this builds properly and is merged in, all is good with me.

I do not need access to your fork. Also, I don't feel comfortable doing things as a I see fit. I'm trying to help b/c you are a bit resistant to implement the changes requested from the review and it is just easier for me to do them in another PR.

> So it is ok to do multiple packages in the same staged-recipe PR? I wasn't aware that's possible. So, we should have both a `visionworkbench` and a `stereopipeline` package provided by this PR

Possible and desired when the packages are closely connected and their interaction not well known. We even have some docs about this, see https://conda-forge.org/docs/maintainer/adding_pkgs/#adding-multiple-packages-at-once

> Then, one could focus all effort on second package, rather than trying to solve two problems at once.

The reason I do not want to approve this is b/c the recipe I saw had intertwined, complex, and pinned dependencies that are a recipe for disaster when they hit the channel. 

If you can provide that draft I can try to make it work here.

oleg-alexandrov:
> I'm trying to help b/c you are a bit resistant to implement the changes requested from the review and it is just easier for me to do them in another PR.

If I merge that in, can we call it a day? It has been almost two months.

ocefpaf:
> If I merge that in, can we call it a day? It has been almost two months.

These recipes are not simple. You are putting quite a burden on the reviewer by not acting on what we ask and by constantly complaining about our free labor to you.

I confess that those comments make me close the tab and avoid this for a few days to avoid burnout. To be honest, the only reason I keep coming back here is b/c @hobu is a great community person and his interest here give me some energy to get back.

If you do what the reviewers are asking instead of fighting us all the way, sure, it will get merged.

TL;DR I'm willing to help you finish this if you can add the next recipe here for inspection. I cannot merge something that overlinks b/c it is "needed for another recipe."

oleg-alexandrov:
This recipe will not overlink. It will have the absolute minimum of dependencies. It is a standalone library. 

I will fix the test as you want. I will merge all you proposed so far. 

If you would like to help more, you can help when this is stand-alone package is online. The other package is much more work. I am not optimistic we will finish that as it goes so far.

oleg-alexandrov:
>The reason I do not want to approve this is b/c the recipe I saw had intertwined, complex, and pinned dependencies that are a recipe for disaster when they hit the channel.

I actually implemented all these things that you asked.

Just as you, I want a good recipe that will help the community.

The problem is that you keep on expanding the scope with things that have nothing to do with this.

Now we have, as you want, a single, not intertwined, not pinned (apart from GDAL as I have no choice) recipe, with a test. Your requirements are satisficed. I will go the extra step of merging your extra particular changes that you like. 

If you are willing to get this over the finish line, can work on the second recipe.

recipes/visionworkbench/meta.yaml
ocefpaf: You keep saying you addressed everything and that I'm expanding the scope of the review. In fact you take a a few pings and convincing to address things, or ignore them as this one, and fails to see that the review is an iterative process.

oleg-alexandrov: In any iterative work, it is good to focus on more important things first.

I tried to merge your pull request. Even after extra work I did trying to fix them, the switch from meta.yaml to "v1" breaks GDAL again.

Because of that, I have to go back to my own way of doing things, and implementing your high-level requests. 

Let us keep things at the high level, and ensure a quality recipe with minimal dependencies.

It is fine with me not to use $PREFIX/bin/hillshade above, if it works without PREFIX.  Maybe for later. Trying to get the thing to build.

ocefpaf:
I will merge https://github.com/conda-forge/staged-recipes/pull/30354 if the author comments there that he is OK. It has the author original commits and the correct fixes necessary to make this recipe OK for conda-forge publishing.

oleg-alexandrov:
That is fine with me. Thank you very much.

PR #30009, opened 2025-05-13
Add recipe for nputop

youyve:
Checklist
- [x] Build number is 0.


recipes/nputop/LICENSE
```diff
@@ -0,0 +1,211 @@
+This project, nputop, is based on the original project 'nvitop' by Xuehai Pan.
```
synapticarbors, line 1: You don't need to include `COPYING`, `LICENSE`, or `NOTICE` since they are included in the source distribution

recipes/nputop/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set name = "nputop" %}
+{% set version = "0.0.4" %}
+{% set python_min = "3.7" %}
```
synapticarbors, line 3: conda-forge doesn't officially support python < 3.9. I would just drop this and use the global default of 3.9

recipes/nputop/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/youyve/nputop/archive/v{{ version }}.tar.gz
+  sha256: c0650244af08d5006936738147f900c5de653cccd0972199306fd372f8e5b1d2
+
+build:
+  number: 0
+  script: {{ PYTHON }} setup.py install --single-version-externally-managed --record=record.txt
```
synapticarbors, line 15: Is there a reason not to use:

```
script: {{ PYTHON }} -m pip install . -vv --no-build-isolation --no-deps
```

recipes/nputop/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  entry_points:
+    - nputop = nputop.cli:main
+  skip: True  # [not linux]  # 仅 Linux 构建
+
+requirements:
+  build:
+    - python {{ python_min }}
+    - setuptools
+  run:
+    - python  # 移除版本约束
```
synapticarbors, line 25:
```suggestion
    - python  >= {{ python_min }}
```

If you are using `noarch: python`, otherwise just `python`

recipes/nputop/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/youyve/nputop/archive/v{{ version }}.tar.gz
+  sha256: c0650244af08d5006936738147f900c5de653cccd0972199306fd372f8e5b1d2
+
+build:
```
synapticarbors, line 13: You are treating this as a `noarch: python` recipe:

as you are using `python_min`, but you haven't defined that. It looks like this could be built on windows with a few extra dependencies. See:

recipes/nputop/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - termcolor >=1.0.0
+
+test:
+  imports:
+    - nputop
+  commands:
+    - nputop --version || echo "Skipping hardware-dependent test"
+  requires:
+    - python {{ python_min }}
+    - psutil >=5.6.6
```
synapticarbors, line 37: you shouldn't repeat the runtime dependencies here (Except for python if you're doing a `noarch: python` recipe).

synapticarbors (changes requested):

recipes/nputop/meta.yaml
youyve: "Non noarch packages should have python requirement without any version constraints.", so I removed python version restrict.

recipes/nputop/meta.yaml
youyve: Thank you for your suggestion！ I changed this.

recipes/nputop/meta.yaml
youyve: I'm using non noarch packages, so I just removed the python version restrict.

recipes/nputop/meta.yaml
youyve: Developed for Linux only, so changed to non-noarch package and removed python_min.

recipes/nputop/meta.yaml
youyve: Thank you for your suggestion! I removed the runtime dependencies.

recipes/nputop/LICENSE
youyve: I deleted COPYING, LICENSE, and NOTICE from [recipes/nputop/]

PR #30031, opened 2025-05-15
[draft] Add bun

whophil:
This continues the work started on https://github.com/conda-forge/staged-recipes/pull/27937

Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/bun/recipe.yaml
+++ b/recipes/bun/recipe.yaml
@@ -30,8 +30,14 @@ source:
 build:
   number: 0
   script:
-    - mkdir -p $PREFIX/bin
-    - mv */bun $PREFIX/bin
+    - if: unix
+      then:
+        - mkdir -p $PREFIX/bin
+        - mv bun $PREFIX/bin
+    - if: win
+      then:
+        - if not exist "%PREFIX%\bin" mkdir "%PREFIX%\bin"
+        - move bun "%PREFIX%\bin"
 
 tests:
   - script:
--- a/recipes/bun/meta.yml
+++ b/recipes/bun/meta.yml
@@ -1,0 +1,48 @@
+context:
+  version: 1.1.31
+
+package:
+  name: bun
+  version: ${{ version }}
+
+source:
+  - if: target_platform == "osx-64"
+    then:
+      url: https://github.com/oven-sh/bun/releases/download/bun-v${{ version }}/bun-darwin-x64.zip
+      sha256: 0041317ccc54a7bb256baa425f3ebd6badae8f42ab25ed1adc4f8043760e87bd  
+  - if: target_platform == "osx-arm64"
+    then:
+      url: https://github.com/oven-sh/bun/releases/download/bun-v${{ version }}/bun-darwin-aarch64.zip
+      sha256: 74e4057e4c4288e16602bd750a349d48a36988b11191b5566b002ecbc0129091
+  - if: target_platform == "linux-aarch64"
+    then:
+      url: https://github.com/oven-sh/bun/releases/download/bun-v${{ version }}/bun-linux-aarch64.zip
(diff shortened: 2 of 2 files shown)


whophil:
This is ready for review @conda-forge/help-nodejs @conda-forge/staged-recipes !

conda-forge-linter is currently faililng as `license_file` is not provided, and no license file is available in the release tarballs.

danielnachun:
In general our policy for open source software packages is to build them from source.  We have the `zig` compiler available in conda-forge and the other dependencies are available as well.

whophil:
Thanks @danielnachun - the recommended process to build `bun` requires `bun` (see https://bun.sh/docs/project/contributing#building-bun).

After this is converted to a feedstock, the feedstock could be changed to build bun from source. Merging this PR could be seen as a sort of bootstrap to make that possible (i.e., in a future iteration, `bun` itself would be a build requirement of `bun`)

Thoughts on this? Is there any precedent for this?

whophil:
I can give it a shot @danielnachun 

I am curious - in this case, what is the value of compiling as compared to using pre-built binaries? To my understanding, all of `bun` is a single executable, with no dynamically linked libraries or other assets.

danielnachun:
> I can give it a shot @danielnachun
> 
> I am curious - in this case, what is the value of compiling as compared to using pre-built binaries? To my understanding, all of `bun` is a single executable, with no dynamically linked libraries or other assets.

There are a number of reasons we prefer to build everything from source when possible:
- Static linking is actually a bad practice from a package management perspective - it make security updates much more complicated because rather than updating one shared library, every package that is statically linked to that library must be rebuilt.  It also leads to wasted compute time spent recompiling the same coded, and wasted disk space on extra statically linked copies of a library.  I do not know if `bun` allows any dynamic linking of its dependencies, but we should attempt to do so if possible.
- License management is also complex with statically linked libraries, as the licenses of all statically linked dependencies must be distributed with the binary.  I've dealt with this extensively for Go, Rust, Haskell, Ocaml and the JDK and .NET ecosystems and it is a lot of work to deal with.
- Conda-forge also seeks to maximize portability of executables across OS versions, at least on Linux.  What this means in practice is that we try to target the oldest C library and kernel header versions we can support.  This is unfortunately often not a priority for distributed pre-compiled binaries, so they are less likely to work on older OSes.

So for that reason we really need to try to get the source build for `bun` figured out, even if it takes significantly more work up front.  I have been interested in getting this onto `conda-forge` for some time so I'll try to help out here as my time permits!

recipes/bun/recipe.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - lld
+    - llvm
+    - llvm-tools
+    - clang 19.*
+    - clangxx 19.*
+    - clang_impl_${{ build_platform }}
+    - clangxx_impl_${{ build_platform }}
+    - zig
+    - ${{ compiler('rust') }}
+    - ${{ compiler('go') }}
```
xhochy, line 50: Please use either `go-cgo` or `go-nocgo`
```suggestion
    - ${{ compiler('go-nocgo') }}
```

recipes/bun/recipe.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - ruby
+    - xz
+    - unzip
+    - lld
+    - llvm
+    - llvm-tools
+    - clang 19.*
+    - clangxx 19.*
+    - clang_impl_${{ build_platform }}
+    - clangxx_impl_${{ build_platform }}
```
xhochy, line 47: See https://conda-forge.org/news/2024/04/30/clang-everywhere/
```suggestion
```

whophil: @xhochy do you know if `conda_build_config.yaml` is supposed to be respected in `staged-recipes`?

xhochy: Yes, it will be respected.

PR #30129, opened 2025-05-23
Adding python/rust module dttlib

evonreis:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+export LIBCLANG_PATH=$CONDA_PREFIX/lib
+export C_INCLUDE_PATH=$CONDA_PREFIX/include
+export CPLUS_INCLUDE_PATH=$CONDA_PREFIX/include
+
+maturin develop --features python,python-pipe,nds --release
+
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -1,7 +1,12 @@
 #!/bin/sh
 
-export LIBCLANG_PATH=$CONDA_PREFIX/lib
-export C_INCLUDE_PATH=$CONDA_PREFIX/include
-export CPLUS_INCLUDE_PATH=$CONDA_PREFIX/include
+export LIBCLANG_PATH=${PREFIX}/lib
+export C_INCLUDE_PATH=${PREFIX}/include
+export CPLUS_INCLUDE_PATH=${PREFIX}/include
+export LIBRARY_PATH=${PREFIX}/lib
+
+if [[ ${target_platform} =~ .*linux.* ]]; then
+	ln -sf ${PREFIX}/lib/libglib-2.0.so.0 ${PREFIX}/lib/libglib-2.0.so
+fi
 ./gen_stub
 $PYTHON -m pip install . -vv
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -6,10 +6,12 @@ export CPLUS_INCLUDE_PATH=${PREFIX}/include
 export LIBRARY_PATH=${PREFIX}/lib
 
 if [[ ${target_platform} =~ .*linux.* ]]; then
-	ln -sf ${PREFIX}/lib/libglib-2.0.so.0 ${PREFIX}/lib/libglib-2.0.so
+    ln -sf ${PREFIX}/lib/libglib-2.0.so.0 ${PREFIX}/lib/libglib-2.0.so
 fi
 ./gen_stub
 $PYTHON -m pip install . -vv
+
+cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
 
 if [[ ${target_platform} =~ .*linux.* ]]; then
     rm ${PREFIX}/lib/libglib-2.0.so
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -3,6 +3,6 @@
 export LIBCLANG_PATH=$CONDA_PREFIX/lib
 export C_INCLUDE_PATH=$CONDA_PREFIX/include
 export CPLUS_INCLUDE_PATH=$CONDA_PREFIX/include
-
-maturin develop --features python,python-pipe,nds --release
-
+./gen_stub
+#maturin develop --features python,python-pipe,nds --release 
+$PYTHON -m pip install . -vv
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -10,3 +10,7 @@ if [[ ${target_platform} =~ .*linux.* ]]; then
 fi
 ./gen_stub
 $PYTHON -m pip install . -vv
+
+if [[ ${target_platform} =~ .*linux.* ]]; then
+    rm ${PREFIX}/lib/libglib-2.0.so
+fi
--- a/recipes/dttlib/build.sh
+++ b/recipes/dttlib/build.sh
@@ -4,5 +4,4 @@ export LIBCLANG_PATH=$CONDA_PREFIX/lib
 export C_INCLUDE_PATH=$CONDA_PREFIX/include
 export CPLUS_INCLUDE_PATH=$CONDA_PREFIX/include
 ./gen_stub
-#maturin develop --features python,python-pipe,nds --release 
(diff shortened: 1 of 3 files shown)


recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+package:
+  name: {{ name }}
+  version: {{ version }}
+
+source:
+  url: https://git.ligo.org/cds/software/cr_tools/dttlib/-/archive/{{ version }}/{{ version }}.tar.gz
+  sha256: 3071c249c45383a7bda9e3de9cfa0f8b185283dec3ffa6d2c43fb91f5f24b424
+
+build:
+  skip: true  # [win or osx]
```
danielnachun:
```suggestion
  skip: true  # [win or osx]
```
Did you disable OSX and Windows because they were failing?  If so, can you please re-enable them so we can see the errors?  It would be better if we can see what is failing for them and fix the issues if possible.

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+   - nds2-client 
+   - libcds 
+   - gds-sigp 
+   - fftw 
+   - zstd 
+   - rapidjson
+   - libclang
+   - gcc
+   - clang
+
```
danielnachun: There are many duplicate entries in the requirements lists, and many of those requirements are probably run time only unless they are being dynamically linked at build time.  And even then, they would only be needed in host.

danielnachun (changes requested):

recipes/dttlib/meta.yaml
evonreis: I would also need to get some dependencies working on windows and OSX, such as libcds and gds-sigp.

recipes/dttlib/meta.yaml
evonreis: libcds, gds-sigp, nds2-client, rapidjson are all dynamically linked.  I'll try to reduce the redundancy in a new commit.

evonreis:
@danielnachun I've pushed a commit that eliminates some dependencies.  I need to have some of the linked libraries in both build and host for the local build to finish.  I'm not sure why.  It goes against my understanding of how conda build works.

recipes/dttlib/meta.yaml
danielnachun: I see, in that case skipping those OSes is fine for the moment but it would be a good idea to add OSX support at least to both the dependencies and this package.  Windows can obviously be much more difficult so it's usually less of a priority.

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+requirements:
+  build:
+   - {{ compiler('c') }}
+   - {{ stdlib("c") }}
+   - {{ compiler('cxx') }}
+   - python
+   - clang
+   - fftw 
+   - gcc
```
danielnachun: It should never be necessary to explicitly depend on `gcc` or `clang`.  If you need to use `clang` as the compiler on Linux, you can follow these instructions: https://conda-forge.org/news/2024/04/30/clang-everywhere/.  If some part of your build process is hard-coded to use `gcc` executable, we prefer to do this instead:
```
ln -sf ${CC} ${BUILD_PREFIX}/bin/gcc
```

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+   - gds-sigp 
+   - glib
+   - libcds 
+   - libclang
```
danielnachun: Once you've specified `clang` as you compiler as described, you should hopefully not need this dependency any more.

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+   - make  # [not win]
+   - pip
+   - pkg-config  # [not win]
```
danielnachun:
```suggestion
   - make 
   - pip
   - pkg-config
```
`make` and `pkg-config` now support windows, so there is no reason to place this restriction on them.

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,71 @@
...
+   - ruff
+   - rust
```
danielnachun:
```suggestion
   - {{ compiler('rust') }}
```
Part of the reason you may be needing to add so many libraries to the `build` section is because you need to use the compiler function to get a `rust` compiler.

recipes/dttlib/meta.yaml
evonreis: Yes we plan to support OSX soon.   Our goal is also to support Windows but that's going to be much farther down the road.

recipes/dttlib/meta.yaml
evonreis: I tried adding conda_build_config.yaml while removing clang as a explicit dependency, but mamba wasn't able to solve for the environment, so I abandoned that approach.  

I've re-learned some things I learned a few weeks ago: clang is explicitly included in order to get the cxx crate a stdio.h that it knows how to find.

recipes/dttlib/meta.yaml
evonreis: I went back to the explicit clang specification (explained in another comment), so needed to keep the explicit libclang specification.

recipes/dttlib/meta.yaml
evonreis: The OS tags are removed in a recent commit.

recipes/dttlib/meta.yaml
evonreis: the dependency is now `{{ compiler('rust') }}` but was still unable to build without the libraries under 'build'.  The cxx crate needs these library packages for some header files.

evonreis:
@danielnachun thank you for reviewing the PR.  I've responded your comments and am ready for another review.

danielnachun:
I did some local testing in a container and figured out how to do two things:
- I got it to use clang properly with `conda_build_config.yaml`
- I was able to get it to search for dynamic libraries correctly from `${PREFIX}` - the main issue was needing to set `LIBRARY_PATH`, but there was also a weird problem with `glib` where the linker was expecting to find `libglib-2.0.so` but there is only `libglib-2.0.so.0`.  I set up a hack around this but we should hopefully find a better solution for this in the future.

recipes/dttlib/build.sh
```diff
@@ -0,0 +1,19 @@
...
+export LIBCLANG_PATH=${PREFIX}/lib
+export C_INCLUDE_PATH=${PREFIX}/include
+export CPLUS_INCLUDE_PATH=${PREFIX}/include
+export LIBRARY_PATH=${PREFIX}/lib
+
+if [[ ${target_platform} =~ .*linux.* ]]; then
+    ln -sf ${PREFIX}/lib/libglib-2.0.so.0 ${PREFIX}/lib/libglib-2.0.so
+fi
+./gen_stub
+#maturin develop --features python,python-pipe,nds --release
```
danielnachun: Is this needed anymore?

recipes/dttlib/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  run:
+    - python
+
+test:
+  requires:
+    - pytest
+  source_files:
+    - test_dttlib.py
+  commands:
+    #- python -c 'import dttlib; dttlib.DTT(lambda: None)'
```
danielnachun: I'm assuming the `pytest` tests you are running will already test this, so I think we can remove this line too.

danielnachun (changes requested):
I just had a few questions about some commented out lines and then I think this will be ready to merge.

recipes/dttlib/build.sh
evonreis: No, I'll delete that

recipes/dttlib/meta.yaml
evonreis: I'll delete that.  That was just the first test before there were pytests.

danielnachun:
I want to rebase and merge here but it's claiming there are conflicts.  Can you update your main brain and then rebase this one?  I'm not sure what's going on here!

evonreis:
git reported a conflict with `recipe/dttlib/build.sh` on the commented out line, conflicting with no line, but of course there was no dttlib directory in main.  

I've rebased the branch and pushed.

PR #29975, opened 2025-05-09
Add azure-cli

janjagusch:
Follow-up to https://github.com/conda-forge/staged-recipes/pull/29966.

Depends on:
- azure-batch >=15.0.0b1 ✅ 
  - https://github.com/conda-forge/azure-batch-feedstock/pull/8
  - https://github.com/conda-forge/azure-batch-feedstock/pull/9
- azure-cli-core >=2.72.0 ✅ 
  - https://github.com/conda-forge/azure-cli-core-feedstock/pull/29 
  - https://github.com/conda-forge/azure-cli-core-feedstock/pull/32
- azure-datalake-store-feedstock >= v1.0.0a0 ✅ 
  - https://github.com/conda-forge/azure-datalake-store-feedstock/pull/14
  - https://github.com/conda-forge/azure-datalake-store-feedstock/pull/16
- azure-mgmt-batchai >=7.0.0b1 ✅ 
- azure-mgmt-batch >=17.3.0 ✅ 
  - https://github.com/conda-forge/azure-mgmt-batch-feedstock/pull/6
- azure-mgmt-compute >=34.1.0 ✅ 
  - https://github.com/conda-forge/azure-mgmt-compute-feedstock/pull/40 
- azure-mgmt-containerinstance >=10.2.0b1 ✅ 
  -  https://github.com/conda-forge/azure-mgmt-containerinstance-feedstock/pull/9
  - https://github.com/conda-forge/azure-mgmt-containerinstance-feedstock/pull/10
- azure-mgmt-datalake-store >=1.1.0b1 ✅ 
  - https://github.com/conda-forge/azure-mgmt-datalake-store-feedstock/pull/4 
  - https://github.com/conda-forge/azure-mgmt-datalake-store-feedstock/pull/5
  - https://github.com/conda-forge/azure-datalake-store-feedstock/pull/17
- azure-mgmt-dns >=8.0.0 ✅ 
  - https://github.com/conda-forge/azure-mgmt-dns-feedstock/pull/5 
- azure-mgmt-iotcentral >=10.0.0b1 ✅ 
  - https://github.com/conda-forge/azure-mgmt-iotcentral-feedstock/pull/1 
- azure-mgmt-keyvault >=11.0.0 ✅ 
  - https://github.com/conda-forge/azure-mgmt-keyvault-feedstock/pull/10 
- azure-mgmt-loganalytics >=13.0.0b4 ✅ 
[...]

PR #30180, opened 2025-05-28
Set channel_sources in combined_spec manually

pavelzw:
The channel sources don't get propagated from `config.exclusive_config_files` so i manually added channel_sources instead.

needed for #30179 and #29975

follow-up of #29895

closes https://github.com/prefix-dev/rattler-build/issues/1688


pavelzw:
@conda-forge/core @wolfv this is ready for review

PR #30240, opened 2025-06-04
FEAT: Broke out tardigrade_error_tools build updates

NateAM:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/tardigrade_error_tools/conda_build_config.yaml
+++ b/recipes/tardigrade_error_tools/conda_build_config.yaml
@@ -1,0 +1,4 @@
+libstdcxx:  # [linux]
+#  - 12      # [linux]
+  - 13      # [linux]
+  - 14      # [linux]
--- a/recipes/tardigrade_error_tools/meta.yaml
+++ b/recipes/tardigrade_error_tools/meta.yaml
@@ -1,0 +1,100 @@
+{% set name = "tardigrade_error_tools" %}
+{% set version = "0.9.7" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/UCBoulder/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
+  path: '..'
+  sha256: af55feef37f99979c3ccfb20c2f563b5e10175cb7599264b441c9608fdb68652
+
+build:
+  number: 0
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script:
+    - set -x
+    # Build and install c++ libraries
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=lib
+    - cmake --build build --target tardigrade_error_tools tardigrade_error_tools_PYTHON
+    - cmake --install build --prefix ${PREFIX}
+    # Build and install Conda package
+    - $PYTHON -m pip install build/src/python/dist/$PKG_NAME-$PKG_VERSION.tar.gz --no-deps --ignore-installed -v --no-build-isolation
+    - $PYTHON -m pip show $PKG_NAME
+  missing_dso_whitelist:
+      - /lib64/librt.so.1
+      - /lib64/libc.so.6
+      - /lib64/libpthread.so.0
+  skip: true # [osx or win]
+
+requirements:
+  build:
+    - cython
+    - python {{ python }}
+    - breathe >=4.3.0
+    - sphinx >=3.0.4
+    - sphinx_rtd_theme >=0.4.3
+    - sphinxcontrib-bibtex
+    - setuptools_scm[toml] >=6.2
+  host:
+    - {{ compiler('cxx') }}
+    - libgcc {{ libstdcxx }}             # [linux]
+    - libstdcxx {{ libstdcxx }}          # [linux]
+    - python {{ python }}
+    - python-build
+    - pip
+    - cython
+    - setuptools >=62
+    - setuptools_scm >=6.2
+    - make
+    - cmake >=3.18
+    - eigen >=3.3.7
+    - boost >=1.59
+    - doxygen
+  run:
+    - python
+    - eigen >=3.3.7
+
+test:
+  requires:
+    - {{ compiler('cxx') }}
+    - {{ pin_compatible('libgcc') }}     # [linux]
+    - {{ pin_compatible('libstdcxx') }}  # [linux]
+    - pytest
+    - pytest-xdist
+    - make
+    - cmake >=3.18
+    - boost >=1.59
+    - doxygen
+    - sphinx >=3.0.4
+  source_files:
+    - CMakeLists.txt
+    - src/cpp/tests/CMakeLists.txt
+    - src/cpp/tests/test_tardigrade_error_tools.cpp
+  commands:
+    - echo running tests
+    - set -x
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=conda-test -DCMAKE_INSTALL_LIBDIR=lib
+    - cmake --build build --target all --verbose --parallel 4
+    - ctest --test-dir build --verbose --parallel 4
+    - ctest --test-dir build --output-on-failure -R 'test_' --parallel 4
+    - cd $SP_DIR/$PKG_NAME
+    - pytest
+
+about:
+  home: https://github.com/UCBoulder/tardigrade_error_tools
+  license: BSD-3-Clause
+  license_file: LICENSE
+  summary: 'A collection of tools used for error handling. '
+  description: |
+    tardigrade_error_tools A collection of tools used for error handling. These tools provide a
+    linked list class which can help a user both understand where an error happened and also help the developer to
+    correct for the errors.
+  dev_url: https://github.com/UCBoulder/tardigrade_error_tools
+
+extra:
+  recipe-maintainers:
+    - NateAM
+    - kbrindley


NateAM:
@kbrindley will you comment on this PR to say you are willing to be a maintainer?

kbrindley:
I am willing to be a maintainer

NateAM:
@conda-forge/help-c-cpp This is ready for review

NateAM:

Just in case I hadn't signaled this was ready for review properly

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+  sha256: af55feef37f99979c3ccfb20c2f563b5e10175cb7599264b441c9608fdb68652
+
+build:
+  number: 0
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script:
+    - set -x
+    # Build and install c++ libraries
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=lib
```
danielnachun:
```suggestion
    - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release ${CMAKE_ARGS}
```
Please add this variable at the end of the `cmake` invocation to make sure that the necessary arguments are added, and note that `-DCMAKE_INSTALL_LIBDIR=lib` is already provided by `${CMAKE_ARGS}` so it doesn't need to be added explicitly.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
+{% set name = "tardigrade_error_tools" %}
+{% set version = "0.9.7" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/UCBoulder/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
+  path: '..'
```
danielnachun:
```suggestion
```
Usually this isn't needed, as you are already specifying a subdirectory for CMake to avoid building in tree and I don't think this will matter for the python part either.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - cmake --build build --target tardigrade_error_tools tardigrade_error_tools_PYTHON
+    - cmake --install build --prefix ${PREFIX}
+    # Build and install Conda package
+    - $PYTHON -m pip install build/src/python/dist/$PKG_NAME-$PKG_VERSION.tar.gz --no-deps --ignore-installed -v --no-build-isolation
+    - $PYTHON -m pip show $PKG_NAME
+  missing_dso_whitelist:
+      - /lib64/librt.so.1
+      - /lib64/libc.so.6
+      - /lib64/libpthread.so.0
+  skip: true  # [osx or win]
```
danielnachun:
```suggestion
  skip: true  # [osx or win]
```
This software does not seem to be Linux-only, so we should at least enable OSX and see if there are any fixable errors.  Windows may be trickier, although since this is using CMake and doesn't appear to have any dependency that are not available on Windows, we should also enable it there.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+
+requirements:
+  build:
+    - cython
+    - python {{ python }}
```
danielnachun, line 35:
```suggestion
    - python
```
The extra Jinja is not needed for compiled recipes.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - breathe >=4.3.0
+    - sphinx >=3.0.4
+    - sphinx_rtd_theme >=0.4.3
+    - sphinxcontrib-bibtex
+    - setuptools_scm[toml] >=6.2
+  host:
+    - {{ compiler('cxx') }}
+    - libgcc {{ libstdcxx }}             # [linux]
+    - libstdcxx {{ libstdcxx }}          # [linux]
```
danielnachun:
```suggestion
    - {{ compiler('cxx') }}
```
`libgcc` and `libstdcxx` do not need to be added to the host section explicitly.  Also please move `{{ compiler('cxx') }}` to the build section.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,100 @@
...
+    - doxygen
+  run:
+    - python
+    - eigen >=3.3.7
+
+test:
+  requires:
+    - {{ compiler('cxx') }}
+    - {{ pin_compatible('libgcc') }}     # [linux]
+    - {{ pin_compatible('libstdcxx') }}  # [linux]
```
danielnachun:
```suggestion
```
This should not be needed when `{{ compiler('cxx')}}` is used.

danielnachun (changes requested):

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility.

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility

NateAM:
@danielnachun Thanks for the review! I made the changes and using the local build utility they seem to work well. I've requested that you re-review.

recipes/tardigrade_error_tools/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+libstdcxx:  # [linux]
+#  - 12      # [linux]
```
kbrindley:
```suggestion
```

@NateAM commented out build variant. We should delete or uncomment.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,94 @@
...
+  source_files:
+    - CMakeLists.txt
+    - src/cpp/tests/CMakeLists.txt
+    - src/cpp/tests/test_tardigrade_error_tools.cpp
+  commands:
+    - echo running tests
+    - set -x
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=conda-test -DCMAKE_INSTALL_LIBDIR=lib
+    - cmake --build build --target all --verbose --parallel 4
+    - ctest --test-dir build --verbose --parallel 4
```
kbrindley: I think this line is redundant. This will run all tests while line below will run a subset matching the ``test_`` regular expression. It shouldn't make a difference for this library (no Abaqus system tests), but the downstream libraries use the regular expression to avoid running Abaqus system tests on systems that don't have Abaqus installed.

Probably we want to keep the ``-R 'test_'`` version of the ``ctest`` command for consistency with the downstream libraries?

```suggestion
```

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - doxygen
+    - sphinx >=3.0.4
+  source_files:
+    - CMakeLists.txt
+    - src/cpp/tests/CMakeLists.txt
+    - src/cpp/tests/test_tardigrade_error_tools.cpp
+  commands:
+    - echo running tests
+    - set -x
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=conda-test -DCMAKE_INSTALL_LIBDIR=lib
```
kbrindley, line 68:
```suggestion
    - cmake -S . -B build -DCMAKE_BUILD_TYPE=conda-test ${CMAKE_ARGS}
```

Probably the test build doesn't need the ``CMAKE_INSTALL_LIBDIR`` option at all. However, I think the ``CMAKE_ARGS`` contains options that should be consistent between the release build and conda-test build.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - cmake --build build --target all --verbose --parallel 4
```
kbrindley, line 69: I'm surprised this doesn't throw errors. Maybe the CMake configuration is smartly skipping the main library build?

In the past, the conda-build had to specify that we only wanted to build the test library: https://github.com/UCBoulder/tardigrade_error_tools/blob/dev/recipe/meta.yaml#L68.

```suggestion
    - cmake --build build --target test_tardigrade_error_tools --verbose --parallel 4
```

recipes/tardigrade_error_tools/meta.yaml
kbrindley: c++ tests are indeed running twice: https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1286144&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff

```
+ ctest --test-dir build --verbose --parallel 4
UpdateCTestConfiguration  from :$SRC_DIR/build/DartConfiguration.tcl
Test project $SRC_DIR/build
Constructing a list of tests
Done constructing a list of tests
Updating test list for fixtures
Added 0 tests to meet fixture requirements
Checking test dependency graph...
Checking test dependency graph end
test 1
    Start 1: test_tardigrade_error_tools

1: Test command: $SRC_DIR/build/src/cpp/tests/test_tardigrade_error_tools
1: Working Directory: $SRC_DIR/build/src/cpp/tests
1: Test timeout computed to be: 9999879
1: Running 7 test cases...
1: 
1: *** No errors detected
1: 
1/1 Test #1: test_tardigrade_error_tools ......   Passed    0.00 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.00 sec
+ ctest --test-dir build --output-on-failure -R test_ --parallel 4
Test project $SRC_DIR/build
    Start 1: test_tardigrade_error_tools
1/1 Test #1: test_tardigrade_error_tools ......   Passed    0.00 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.01 sec
```

recipes/tardigrade_error_tools/meta.yaml
NateAM: Yeah, we can get away with it because the library build doesn't happen because of the `CMAKE_BUILD_TYPE=conda-test` specification. I've been using `all` for consistency with the rest of the stack so I would prefer to leave it this way for future development.

recipes/tardigrade_error_tools/meta.yaml
NateAM: Yes, I can verify removing this still passes in my local builds.

recipes/tardigrade_error_tools/meta.yaml
NateAM: This seems to cause an issue with the test build. I think that `CMAKE_ARGS` must contain something about this being a release build and it overrides the `-DCMAKE_BUILD_TYPE=conda-test`. I'm going to leave this as is.

danielnachun: You don't need to set `CMAKE_ARGS` in the `test` block - these tests will be skipped when cross-compiling anyway, and we're not shipping whatever you build in the tests to users.  I do agree that `-DCMAKE_INSTALL_LIBDIR=lib` can be dropped since nothing gets installed while testing.

recipes/tardigrade_error_tools/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - setuptools >=62
+    - setuptools_scm >=6.2
+    - make
+    - cmake >=3.18
+    - eigen >=3.3.7
+    - boost >=1.59
+    - doxygen
+  run:
+    - python
+    - eigen >=3.3.7
```
danielnachun:
```suggestion
requirements:
  build:
    - {{ compiler('c') }}
    - {{ stdlib('c') }}
    - {{ compiler('cxx') }}
    - make
    - cmake >=3.18
  host:
    - python {{ python }}
    - breathe >=4.3.0
    - sphinx >=3.0.4
    - sphinx_rtd_theme >=0.4.3
    - sphinxcontrib-bibtex
    - python-build
    - pip
    - cython
    - setuptools >=62
    - setuptools_scm >=6.2
    - eigen >=3.3.7
    - boost >=1.59
    - doxygen
  run:
    - python
```
I've cleaned up and rearranged the dependencies here to better match what we expect for native packages.  `build` dependencies must match the architecture of whatever machine we are building on, while `host` packages must match the target architecture.  Some more work will probably be needed when cross-compilation is enabled in the feedstock but this should get us closer to what we want.

danielnachun (changes requested):
I think we should consider splitting this into something like `python-tardigrade_error_tools` for the Python package and `tardigrade_error_tools` for the shared library, with `python-tardigrade_error_tools` depending on `tardigrade_error_tools` to provide the shared library.  This would make it easier for end users who only need the C++ shared library to not have to depend on Python.

kbrindley:
@danielnachun 

> I think we should consider splitting this into something like `python-tardigrade_error_tools` for the Python package and `tardigrade_error_tools` for the shared library, with `python-tardigrade_error_tools` depending on `tardigrade_error_tools` to provide the shared library. This would make it easier for end users who only need the C++ shared library to not have to depend on Python.

This may be a good long-term idea. In the short term, we don't have enough (or any) c++-only users to justify splitting the package like this. There is a larger stack of ``tardigrade_`` prefixed libraries downstream of this package and all existing users will need both the c++ and Python together.

This is more obvious when we get to the more user-centric packages, e.g. https://github.com/conda-forge/staged-recipes/pull/30248, where users will use the Python API to perform material model calibrations and then the c++ library to run finite element simulations against those same calibrations.

recipes/tardigrade_error_tools/meta.yaml
NateAM: This is required in the current cmake implementation because we use it to specify where the required installed project library is using a `find_file` command. I'll see if I can't make this go away.

kbrindley: I don't think I would chase a build system refactor for this PR. If the build system uses this CMake variable to configure the test, then we should keep it.

NateAM: > I don't think I would chase a build system refactor for this PR. If the build system uses this CMake variable to configure the test, then we should keep it.

Cool, then I'll leave it alone.

recipes/tardigrade_error_tools/meta.yaml
NateAM: I've verified this works using the local build utility.

PR #30247, opened 2025-06-04
Feat/add tardigrade_balance_equations

NateAM:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/tardigrade_balance_equations/meta.yaml
+++ b/recipes/tardigrade_balance_equations/meta.yaml
@@ -25,7 +25,7 @@ build:
       - /lib64/libc.so.6
       - /lib64/libpthread.so.0
       - /lib64/libm.so.6
-  skip: true # [osx or win]
+  skip: true  # [osx or win]
 
 requirements:
   build:
--- a/recipes/tardigrade_balance_equations/meta.yaml
+++ b/recipes/tardigrade_balance_equations/meta.yaml
@@ -1,0 +1,120 @@
+{% set name = "tardigrade_balance_equations" %}
+{% set version = "0.2.1" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/UCBoulder/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
+  path: '..'
+  sha256: 611f9adbea2dd74e5a099e3e97120f3e40a512f5bc1e5afe56a19fc05512f9b3
+
+build:
+  number: 0
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script:
+    - set -x
+    # Build and install c++ libraries
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_LIBDIR=lib -DTARDIGRADE_CMML_BUILD_AS_SHARED=ON
+    - cmake --build build --target tardigrade_balance_equations
+    - cmake --install build --prefix ${PREFIX}
+  missing_dso_whitelist:
+      - /lib64/librt.so.1
+      - /lib64/libc.so.6
+      - /lib64/libpthread.so.0
+      - /lib64/libm.so.6
+  skip: true # [osx or win]
+
+requirements:
+  build:
+    - cython
+    - python {{ python }}
+    - numpy
+    - breathe >=4.3.0
+    - sphinx >=3.0.4
+    - sphinx_rtd_theme >=0.4.3
+    - sphinxcontrib-bibtex
+    - setuptools_scm[toml] >=6.2
+    - tardigrade_error_tools >=0.9.7
+    - tardigrade_vector_tools >=0.8.7
+  host:
+    - {{ compiler('cxx') }}
+    - libgcc {{ libstdcxx }}             # [linux]
+    - libstdcxx {{ libstdcxx }}          # [linux]
+    - python {{ python }}
+    - python-build
+    - pip
+    - numpy
+    - cython
+    - setuptools >=62
+    - setuptools_scm >=6.2
+    - make
+    - cmake >=3.18
+    - eigen >=3.3.7
+    - boost >=1.59
+    - doxygen
+    - tardigrade_error_tools >=0.9.7
+    - tardigrade_vector_tools >=0.8.7
+    - tardigrade_constitutive_tools >=0.8.2
+    - tardigrade_stress_tools >=0.6.1
+    - tardigrade_micromorphic_tools >=1.4.2
+    - tardigrade_abaqus_tools >=0.6.3
+    - tardigrade_hydra >=0.4.5
+  run:
+    - python
+    - eigen >=3.3.7
+    - tardigrade_error_tools >=0.9.7
+    - tardigrade_constitutive_tools >=0.8.2
+    - tardigrade_stress_tools >=0.6.1
+    - tardigrade_micromorphic_tools >=1.4.2
+    - tardigrade_abaqus_tools >=0.6.3
+    - tardigrade_hydra >=0.4.5
+
+test:
+  requires:
+    - {{ compiler('cxx') }}
+    - {{ pin_compatible('libgcc') }}     # [linux]
+    - {{ pin_compatible('libstdcxx') }}  # [linux]
+    - pytest
+    - pytest-xdist
+    - numpy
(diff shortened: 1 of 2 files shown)


NateAM:
@kbrindley will you comment if you are willing to be a maintainer?

NateAM:
Blocked by #30240 , #30241 , #30242 , #30243 , #30244 , #30245 , and #30246

kbrindley:
I am willing to be a maintainer

NateAM:
I want to keep this open. We're waiting for the conda-forge c++ team to review some other pull requests.

NateAM:
@conda-forge/help-c-cpp This is ready for review

NateAM:
@traversaro at the risk of being annoying, this one also should be very close to what you just approved.

recipes/tardigrade_balance_equations/meta.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - eigen >=3.3.7
+    - tardigrade_error_tools >=0.9.8
+    - tardigrade_vector_tools >=0.8.7
+    - tardigrade_constitutive_tools >=0.8.5
+    - tardigrade_hydra >=0.8.0
+  run:
+    - eigen >=3.3.7
+    - tardigrade_error_tools >=0.9.8
+    - tardigrade_vector_tools >=0.8.7
+    - tardigrade_constitutive_tools >=0.8.5
```
danielnachun: Are these shared libraries?  If so do they have `run_exports`?  If they do, then they don't have to explicitly added to run.

recipes/tardigrade_balance_equations/meta.yaml
```diff
@@ -0,0 +1,90 @@
+{% set name = "tardigrade_balance_equations" %}
+{% set version = "0.2.2" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/UCBoulder/{{ name }}/archive/refs/tags/{{ version }}.tar.gz
```
danielnachun:
```suggestion
{% set version = "0.2.2" %}

package:
  name: tardigrade_balance_equations
  version: {{ version }}

source:
  url: https://github.com/UCBoulder/tardigrade_balance_equations/archive/refs/tags/{{ version }}.tar.gz
```
The `name` key is not needed here - the package name cannot be changed after the feedstock is made.

recipes/tardigrade_balance_equations/meta.yaml
kbrindley: We didn't get the run exports into the early libraries. That will have to be a future improvement.

recipes/tardigrade_balance_equations/meta.yaml
NateAM: > The `name` key is not needed here - the package name cannot be changed after the feedstock is made.

We're using this formatting because it matches the rest of the upstream stack and allows us to be consistent. If possible, we would prefer to keep this as is with the understanding that we can't change it.

NateAM:
@danielnachun I believe we have addressed your comments or at least provided explanations for why we would like to keep things the way they are. Can you re-review?

recipes/tardigrade_balance_equations/meta.yaml
danielnachun: This is a pretty easy fix - it would be better if we can get the PRs made in those other packages first and see if the run exports work as expected here.  Please let me know if you have questions.

recipes/tardigrade_balance_equations/meta.yaml
danielnachun: Sorry to have to push back here, but those recipes should not have used the `name` key here - see https://github.com/conda-forge/conda-forge.github.io/issues/2712 for our policy change.  Please instead change the other recipes not to use `name` for consistency with this one.

recipes/tardigrade_balance_equations/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script:
+    - set -x
+    # Build and install c++ libraries
+    - cmake -S . -B build -DCMAKE_BUILD_TYPE=Release ${CMAKE_ARGS}
+    - cmake --build build --target tardigrade_balance_equations
+    - cmake --install build --prefix ${PREFIX}
+  skip: true  # [win]
+  run_exports:
+    - {{ pin_subpackage(name|lower, max_pin='x.x') }}
```
danielnachun:
```suggestion
    - {{ pin_subpackage('tardigrade_balance_equations', max_pin='x.x') }}
```

danielnachun:
One other question that I'm confused about - the build logs only seem to show installation of C++ `.h` and `.cpp` files.  What is the reason for all of the Python dependencies?  And why would `.h` and `.cpp` files depend on other C++ packages at run time?  If you need all of this for testing, that is totally fine, just move the dependencies to testing.  Sorry if this is obvious and I've missed it but I don't seen anything about this in the previous discussion.

NateAM:
> One other question that I'm confused about - the build logs only seem to show installation of C++ `.h` and `.cpp` files. What is the reason for all of the Python dependencies? And why would `.h` and `.cpp` files depend on other C++ packages at run time? If you need all of this for testing, that is totally fine, just move the dependencies to testing. Sorry if this is obvious and I've missed it but I don't seen anything about this in the previous discussion.

@danielnachun We've been building up this larger computational stack from smaller components. Some of those components had python bindings to them. I think the python dependencies are cruft that got carried through. I can try cleaning that up if it's desirable.

NateAM:
> > One other question that I'm confused about - the build logs only seem to show installation of C++ `.h` and `.cpp` files. What is the reason for all of the Python dependencies? And why would `.h` and `.cpp` files depend on other C++ packages at run time? If you need all of this for testing, that is totally fine, just move the dependencies to testing. Sorry if this is obvious and I've missed it but I don't seen anything about this in the previous discussion.
> 
> @danielnachun We've been building up this larger computational stack from smaller components. Some of those components had python bindings to them. I think the python dependencies are cruft that got carried through. I can try cleaning that up if it's desirable.

@danielnachun so I think removing some of those dependencies would be a problem for the underlying cmake logic rather than just the conda packaging because it has hooks for python bindings. Because this is what we intend, I would like to leave those in place for now.

recipes/tardigrade_balance_equations/meta.yaml
NateAM: I've made the requested change

kbrindley:
@danielnachun  @NateAM 

> One other question that I'm confused about - the build logs only seem to show installation of C++ `.h` and `.cpp` files. What is the reason for all of the Python dependencies? And why would `.h` and `.cpp` files depend on other C++ packages at run time? If you need all of this for testing, that is totally fine, just move the dependencies to testing. Sorry if this is obvious and I've missed it but I don't seen anything about this in the previous discussion.

> What is the reason for all of the Python dependencies? 

The build uses setuptools_scm to determine the version number. Most of those Python packages are used in the CMake configuration and must be present in the build environment.

> And why would `.h` and `.cpp` files depend on other C++ packages at run time?

This c++ project is built on the larger tardigrade suite that we've been releasing. This project compiles and links against those upstream libraries. It look like @NateAM has been updating the run exports of the upstream packages, but previously our strategy was to specify the runtime requirements explicitly. This was convenient while the suite was changing rapidly and helpful to track the minimum requirements per-package.

NateAM:
@danielnachun can we get you to review the changes we made?

danielnachun:
> The build uses setuptools_scm to determine the version number. Most of those Python packages are used in the CMake configuration and must be present in the build environment.

This is totally fine, I just wanted to make sure the reason for the inclusion of the Python packages was clear.

> This c++ project is built on the larger tardigrade suite that we've been releasing. This project compiles and links against those upstream libraries. It look like @NateAM has been updating the run exports of the upstream packages, but previously our strategy was to specify the runtime requirements explicitly. This was convenient while the suite was changing rapidly and helpful to track the minimum requirements per-package.

For future reference, while the explicit run time version specification is the norm for Python/Perl/Ruby/Lua, for packages with shared libraries, we generally prefer in conda-forge to use the run exports mechanism because it makes clear what the ABI versioning of the library is as opposed to the API version.  I can see that the run exports have been added so the issue is resolved here.

NateAM:
@danielnachun thank you!

PR #30273, opened 2025-06-06
Add Blurb package

CAM-Gerlach:
Checklist
- [x] Build number is 0.
- [x] ~When in trouble, please check our [knowledge base documentation](https://conda-forge.org/docs/maintainer/knowledge_base.html) before pinging a team.~

### DIFF
--- a/recipes/blurb/meta.yaml
+++ b/recipes/blurb/meta.yaml
@@ -1,0 +1,45 @@
+{% set name = "blurb" %}
+{% set version = "2.0.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/blurb-{{ version }}.tar.gz
+  sha256: c78d8114294225a4f7a2eabba6e05d36a6a50e45ba9f5a41afabc198350038e0
+
+build:
+  entry_points:
+    - blurb = blurb.blurb:main
+  noarch: python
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:

recipes/blurb/meta.yaml
```diff
@@ -0,0 +1,70 @@
+{% set name = "blurb" %}
+{% set import = "blurb" %}
```
synapticarbors, line 2: Let's get rid of all of the jinja variables that aren't `name` and `version` and then just be explicit in the rest of the recipe.

synapticarbors (changes requested):

recipes/blurb/meta.yaml
CAM-Gerlach: Hey, thanks for the feedback. Actually, this usage of named placeholders is standard across the numerous feedstocks we maintain (and at least `sha256` widely used elsewhere), single-sourcing the common variables in one place rather than repeating them scattered throughout the business logic, and makes workflows more resuable, syncable and consistent between feedstocks, decreasing overall maintainer burden while increasing quality. Furthermore, at least from our POV as maintainers it actually makes it more clear and explicit which of the several similar or identical names each value actually represents, and allows trivially changing any of them while avoiding mistakes.

Is there a rationale we're missing here for why `name` and `version` are allowed to be variables, but not `import`, `entrypoint` or `sha256` in the packages we maintain? `import` and `entrypoint` are repeated more times than `name` (5 each vs. 4), and `sha256` changes every release like `version` (only repeated twice) putting both in the same place, so there didn't seem to be a clear justification in that regard.

PR #30275, opened 2025-06-07
Add Python-Docs-Theme package

CAM-Gerlach:
Checklist
- [x] Build number is 0.
- [x] ~When in trouble, please check our [knowledge base documentation](https://conda-forge.org/docs/maintainer/knowledge_base.html) before pinging a team.~

This is the theme used for the official Python documentation, as well as other sites. As a core dev of Python focused on the docs, I want to make sure this is available on Conda-Forge.

Note that to be compatible with and support building the docs on the full range of Python versions currently supported upstream and by Conda-Forge (>3.9+, and in fact back to 3.7-3.8), I've set the initial version in the recipe to the somewhat older 2023.3.1. Once the feedstock is created, I'll sequentially update it for `2025.2` (required for several Python versions, and the last version supporting Python 3.8-3.11 and Sphinx 3.4-7.3) and then the latest `2025.5`.


CAM-Gerlach:

recipes/python-docs-theme/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set name = "python-docs-theme" %}
```
synapticarbors, line 1: Please remove all of the jinja variables except `name` and `version` and then just put them in explicitly in the recipe.

recipes/python-docs-theme/LICENSE
```diff
@@ -0,0 +1,48 @@
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
```
synapticarbors, line 1: The license file is included in the source distribution so you shouldn't include a copy here.

recipes/python-docs-theme/meta.yaml
CAM-Gerlach: Hey, thanks for the feedback. Actually, this usage of named placeholders is standard across the numerous feedstocks we maintain (and at least `sha256` widely used elsewhere), single-sourcing the common variables in one place rather than repeating them scattered throughout the business logic, and makes workflows more resuable, syncable and consistent between feedstocks, decreasing overall maintainer burden while increasing quality. Furthermore, at least from our POV as maintainers it actually makes it more clear and explicit which of the several similar or identical names each value actually represents, and allows trivially changing any of them while avoiding mistakes.

Is there a rationale we're missing here for why `name` and `version` are allowed to be variables, but not `import`, `entrypoint` or `sha256` in the packages we maintain?

recipes/python-docs-theme/LICENSE
CAM-Gerlach: Actually, as can be [verified above](https://github.com/conda-forge/staged-recipes/runs/43652446049), the CIs failed with an error that the license was missing in the initial added version here, which I'd confirmed was indeed not included in the [relevant 2023.3.1 sdist](https://files.pythonhosted.org/packages/9e/d9/2c7e63a47b6787ef4762e596fbb0a7ac1270c08739c83d2851e16e43fc72/python-docs-theme-2023.3.1.tar.gz), and adding it fixed the issue.

synapticarbors: You're correct. I was looking at the most recent release up on pypi. When you eventually update the feedstock, please be sure to remove the local copy once it reaches a version that includes it in the sdist.

recipes/python-docs-theme/meta.yaml
synapticarbors: Mostly it's just a convention that we try to follow and is shown in https://github.com/conda-forge/staged-recipes/blob/main/recipes/example/meta.yaml

It's subjective whether it makes it easier to maintain/read, but it typically we ask people to be explicit. Looking through a number of the feedstocks where you've adopted this convention, it looks like they were updated to add the jinja variables after coming out the stage-recipes review. Also as a note, only about 1700 of the ~20k recipes in conda-forge appear to set `sha256` as a jinja variable. It was common historically to do so, but has become less so as it's not what's shown in the example `meta.yaml` or used in the recipes generated by `grayskull`

CAM-Gerlach: Noted, thanks for the explanation and perspective. I'll keep that in mind going forward, and while I and our Spyder team do see named variables as more explicit vs. repeated hardcoding (akin to named constants vs. magic numbers), I'll be more conservative about adding more, or to repos maintained outside our team (e.g. two I just got added to).

recipes/python-docs-theme/LICENSE
CAM-Gerlach: Yup, I figured that might have been it. And thanks for the reminder—doing that update right now!

PR #30283, opened 2025-06-09
Add sqlite-tools feedstock recipe

flaviomartins:
@conda-forge/sqlite

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/sqlite-tools/meta.yaml
+++ b/recipes/sqlite-tools/meta.yaml
@@ -1,0 +1,100 @@
+{% set version = "3.50.1" %}
+{% set year = "2025" %}
+{% set version_split = version.split(".") %}
+{% set major = version_split[0] %}
+{% set minor = version_split[1]|int * 10 %}
+{% set bugfix = version_split[2]|int * 100 %}
+{% set version_coded=(major ~ (("%03d" % minor)|string) ~ (("%03d" % bugfix)|string)) %}
+
+package:
+  name: sqlite-tools
+  version: {{ version }}
+
+source:
+  url: https://www.sqlite.org/{{ year }}/sqlite-src-{{ version_coded }}.zip
+  sha256: 9090597773c60a49caebb3c1ac57db626fac4d97cb51890815a8b529a4d9c3dc
+
+build:
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+    - make  # [not win]
+    - tk
+  host:
+    - libsqlite >={{ version }}  # [not win]
+    - zlib                       # [not win]
+    - tk
+
+outputs:
+  - name: sqldiff
+    files:
+      - bin/sqldiff              # [unix]
+      - Library\bin\sqldiff.exe  # [win]
+    requirements:
+      build:
+        - {{ compiler('c') }}
+        - {{ stdlib("c") }}
+        - make  # [not win]
+      host:
+        - libsqlite >={{ version }}  # [not win]
+        - zlib                       # [not win]
+
+    test:
+      commands:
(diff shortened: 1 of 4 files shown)


flaviomartins:
@conda-forge/sqlite I was under the impression that `tk` was not a runtime dependency but I only got this to build with `tk` requirement in all sections (for sqlite3_analyzer).

PS: investigating, tk-feedstock includes a sqlite3_analyzer binary which clobbers this

flaviomartins:
@conda-forge/sqlite currently installing this sqlite3_analyzer as sqlite3_analyze for lack of a better idea

flaviomartins:
The --dynlink-tools option shaves many bytes from sqldiff and sqlite3_analyze by linking with libsqlite from the sqlite-feedstock

flaviomartins:
@ocefpaf could I ask you to help me on this?

I'm suggesting the tools are a separate recipe/feedstock from sqlite but I'm not sure it makes sense to dynamic link in that scenario but the requirement is like so:

```
host:
  - libsqlite {{ version }}
```

Edit: An alternative is to build them along with the existing sqlite-feedstock. Requires switching from the amalgamation/autoconf source to the canonical source in that recipe, which does not work on all archs due to missing tcl.

flaviomartins:
Building these tools in sqlite-feedstock would require tcl/tk for building and because some architectures are missing it, so that seems to be a dead end (see: https://github.com/conda-forge/sqlite-feedstock/pull/137)

recipes/sqlite-tools/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+    public-domain, SQL database engine.It is the most used database engine
+    in the world.
+  doc_url: http://www.sqlite.org/docs.html
+  doc_source_url: https://github.com/mackyle/sqlite/tree/master/doc
+  dev_url: https://github.com/mackyle/sqlite
+
+extra:
+  feedstock-name: sqlite-tools
+  recipe-maintainers:
+    - conda-forge/sqlite
```
ocefpaf: I don't want to maintain more software and I'm not sure the rest of the sqlite team there wants too. Maybe it is better to add just yourself?

Also, this looks good to me. I'll merge. Does it makes sense to add that to the sqlite as multiple outputs? I'm not sure b/c I don't really use this. I package it a long time ago b/c I needed for another software.

flaviomartins: > I don't want to maintain more software and I'm not sure the rest of the sqlite team there wants too. Maybe it is better to add just yourself?

I set myself as the sole maintainer for this recipe, at least for now.

> 
> Also, this looks good to me. I'll merge. Does it makes sense to add that to the sqlite as multiple outputs? I'm not sure b/c I don't really use this. I package it a long time ago b/c I needed for another software.

In contrast to the sqlite amalgamation/autoconf source tarball in sqlite-feedstock, the tarball used here is missing some preprocessing and thus requires TCL for all build targets. The win_arm64 and osx_arm64 platforms do not have TCL available so we cannot do that :-( We would maybe have to cross-compile, or something else, which just complicates things further.

I cleaned up and squashed this recipe. I am not sure if the libsqlite requirement should be for the exact version 3.50.1 or >=3.50.1.

flaviomartins:
I cleaned up and squashed this recipe. I am not sure if the libsqlite requirement should be for the exact version 3.50.1 or >=3.50.1.

ocefpaf:
> I cleaned up and squashed this recipe. I am not sure if the libsqlite requirement should be for the exact version 3.50.1 or >=3.50.1.

Not sure but we can try >= and patch it later if we find it doesn't work.

xhochy (approved):

PR #30361, opened 2025-06-19
Add recipe for visisipy

crnh:
Checklist
- [x] Build number is 0.


jwmbeenakker:
I confirm

crnh:

recipes/visisipy/conda-forge.yml
```diff
@@ -0,0 +1,2 @@
+noarch_platforms:
```
synapticarbors: Why isn't this just a regular `noarch: python` package? I don't see any platform specific dependencies.

crnh: ZOSPy is a Windows-only package; visisipy won't work without this. A future release of visisipy will also support other OSes, but that is not yet the case.

synapticarbors: Our general policy is that we just leave packages that are themselves `noarch: python` as such even if a dependency might have platform restrictions/is missing on a specific platform. It's ok if the CI fails for those platforms as long as that is the only cause. If you could remove the `conda-forge.yml` and the `__win`, that would be great.

crnh: Okay, I made the requested changes. However, when adding a recipe for ZOSPy, we were asked to add a `conda-forge.yml`, which is why I added it here as well: https://github.com/conda-forge/staged-recipes/pull/24999
Should I adjust ZOSPy's recipe to meet the current policies?

crnh:

recipes/visisipy/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+  number: 0
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - hatchling  >=1.27,<2
+    - pip
+  run:
+    - python >={{ python_min }},<3.14
+    - zospy ==1.2.1
```
danielnachun, line 24:
```suggestion
    - zospy ==1.2.1
```
Strict versioning like this can make future maintenance a challenge. Can this be changed to a softer bound like `>=1.2,<1.3`?

crnh: @danielnachun thanks for your comment. For this version, the bound cannot be changed. We recently released a new version of Visisipy that allows for softer bounds. Should I upgrade the recipe to that new version, or is it better to still finish this PR with Visisipy 0.0.1 and upgrade the recipe to version 0.1.0 after the PR has been merged?

crnh: @danielnachun can you please answer the question above?

PR #30370, opened 2025-06-20
Added package kineticstoolkit_extensions

felixchenier:
Checklist
- [x] Build number is 0.


felixchenier:
@conda-forge/help-python My package is ready to be included

recipes/kineticstoolkit_extensions/meta.yaml
```diff
@@ -0,0 +1,41 @@
+{% set name = "kineticstoolkit_extensions" %}
```
synapticarbors, line 1: Let's change this to `kineticstoolkit-extensions` to match the naming convention on `pypi`.

recipes/kineticstoolkit_extensions/LICENSE.txt
```diff
@@ -0,0 +1 @@
...
      and charge a fee for, acceptance of support, warranty, indemnity,
      or other liability obligations and/or rights consistent with this
      License. However, in accepting such obligations, You may act only
      on Your own behalf and on Your sole responsibility, not on behalf
      of any other Contributor, and only if You agree to indemnify,
      defend, and hold each Contributor harmless for any liability
      incurred by, or claims asserted against, such Contributor by reason
      of your accepting any such warranty or additional liability.

   END OF TERMS AND CONDITIONS
```
synapticarbors: For a future release, can you try to include the license file in the source distribution so that the copy here can be removed?

Also, the formatting of this license file is pretty terrible. Isn't there a canonical version of this that doesn't have the weird spacing?

recipes/kineticstoolkit_extensions/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+requirements:
+  host:
+    - pip
+    - python {{ python_min }}
+    - setuptools
+  run:
+    - python >={{ python_min }}
+    - kineticstoolkit
+
+test:
```
synapticarbors, line 28: It's recommended to add a `pip check`. See:
https://github.com/conda-forge/staged-recipes/blob/main/recipes/example/meta.yaml#L64-L67

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: Hi @synapticarbors I don't know, I couldn't find that convention on pypi's naming conventions, and when I look for package names on packages.anaconda, I find both conventions, and a big one (jupyter) also use underscore (jupyter_core, jupyter_client, jupyterlab_widgets, etc.).

If it's something that can be discussed, I have a strong preference for the underscore instead of the hyphen, for these reasons:

1. On pypi, the package is already published as kineticstoolkit_extensions, and I would like to avoid having a different name between pypi and conda (ok I could change it on pypi too)
2. We cannot `import kineticstoolkit-extensions` (the hyphen is an invalid character for a module/package name). Therefore the package itself needs an underscore. It creates IMO a possible confusion if a user receives a "ModuleNotFound error: kineticstoolkit_extensions" and that they couldn't solve it using `conda install kineticstoolkit_extensions` since the correct package name is not the same as the package they want to import.

Could we keep the underscore?

recipes/kineticstoolkit_extensions/LICENSE.txt
felixchenier: It's already included in the package itself, so I remove it and its reference from the recipe.

recipes/kineticstoolkit_extensions/meta.yaml
synapticarbors: What I was referring to was the name on `pypi`:
https://pypi.org/project/kineticstoolkit-extensions/

Basically we want the same name you would use for `pip install`. Looking into this a bit further, it seems like pypi normalizes underscores to dashes. See:
https://discuss.python.org/t/are-there-any-naming-conventions-for-package-names/47746

It doesn't seem like there is a hard and fast rule, so we can keep the underscore if you prefer.

recipes/kineticstoolkit_extensions/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+  imports:
+    - kineticstoolkit_extensions
+  requires:
+    - python {{ python_min }}
+    - kineticstoolkit
```
synapticarbors: You shouldn't list this here since it comes in through the `run` requirements.

recipes/kineticstoolkit_extensions/meta.yaml
```diff
@@ -0,0 +1,45 @@
...
+    - pip
+  commands:
+    - pip check
+
+
+about:
+  home: https://kineticstoolkit.uqam.ca
+  summary: Additional modules and development of new features for Kinetics Toolkit.
+  license: Apache-2.0
```
synapticarbors, line 40: You need to add back the `license_file` entry here.

recipes/kineticstoolkit_extensions/LICENSE.txt
synapticarbors: I don't see the license file in the source release that is up on `pypi`. Also as noted in another comment (and by the linter) the `license_file` entry is always required.

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: Yeah, sorry I'm doing the changes at the moments, trying to fix a dependency-of-a-dependency problem. I wasn't yet ready.

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: OK I'll do it.

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: Oh I see, I didn't realize that the distribution name had been changed. But `pip install kineticstoolkit_extensions` works, and therefore yes I'd like to keep the underscore so that `conda install kineticstoolkit_extensions` also works.

recipes/kineticstoolkit_extensions/LICENSE.txt
felixchenier: Oh, it's been lost in the packaging. I just pushed a new release on pypi with the license in it. For the formatting, I agree with you that the formatting if bad, however it's the file provided on the official site (https://www.apache.org/licenses/LICENSE-2.0.html), which is unfortunately full of hard line wraps.

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: I tried to add one but there's a problem with the package ezc3d, which is installed by conda as expected, but that is not recognized as being installed by pip even if it is. Therefore I think the import test may be sufficient, unless we hack around it and I'm a bit clueless on how to fix this mismatch...

recipes/kineticstoolkit_extensions/meta.yaml
felixchenier: Done, and this time it's been found in the package since it really is.

felixchenier:
Hi @synapticarbors Thanks for the review, I think I have addressed those issues. Maybe not the pip check element because of the issue with ezc3d (which is discussed in the recipe).

recipes/kineticstoolkit_extensions/LICENSE.txt
synapticarbors: Now that the file is in the `0.0.2` release sdist, you should remove this copy of LICENSE.txt

synapticarbors:
Now that the license file is in the 0.0.2 release sdist, you should remove this copy of `LICENSE.txt`

felixchenier:
Tx, it's done now.

PR #30403, opened 2025-06-24
Stereo Pipeline package

oleg-alexandrov:
Checklist
- [] If static libraries are linked in, the license of the static library is packaged.
- [x] Build number is 0.

This not yet ready for review.

### DIFF
--- a/recipes/stereo-pipeline/build.sh
+++ b/recipes/stereo-pipeline/build.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+mkdir build && cd build
+cmake ${SRC_DIR}                             \
+    -DCMAKE_PREFIX_PATH=${PREFIX}    \
+    -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+    -DASP_DEPS_DIR=${PREFIX}         \
+    -DUSE_OPENEXR=OFF                \
+    -DCMAKE_VERBOSE_MAKEFILE=ON
+make -j${CPU_COUNT}
+make install
+
--- a/recipes/stereo-pipeline/recipe.yaml
+++ b/recipes/stereo-pipeline/recipe.yaml
@@ -1,0 +1,54 @@
+context:
+  name: stereo-pipeline
+  version: "3.5.0"
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  - folder: .
+    url: https://github.com/NeoGeographyToolkit/StereoPipeline/archive/refs/tags/{{ version }}.zip
+    sha256: 4f1fcfff3a2ae058496b2b78437195dc073423a57d1a9db05a24782961278a2f
+
+build:
+  number: 0
+  skip: win
+  script: build.sh
+
+requirements:
+  build:
+    - ${{ compiler("c") }}
+    - ${{ compiler("cxx") }}
+    - ${{ stdlib("c") }}
+    - cmake >=3.15
+    - make
+  host:
+    - flann
+    - libboost-devel
+    - libgdal
+    - libblas * *openblas
+    - libopencv
+    - libjpeg-turbo
+    - libpng
+    - libtiff
+
+tests:
+  - script:
+    - hillshade $PREFIX/share/vw/tests/tinyDemAN.tif
+
+about:
+  homepage: https://github.com/NeoGeographyToolkit/StereoPipeline
+  summary: 'A suite of tools for producing terrain models published by NASA'
+  description: |
+    The NASA Ames Stereo Pipeline is a collection of tools for producing 3D terrain models from images acquired by satellites of planets and of Earth. It also supports alignment of terrains, correction of errors in camera positions and orientations, and structure from motion.
+  license: Apache-2.0
+  license_family: Apache
+  license_file: COPYING
+  documentation: https://stereopipeline.readthedocs.io/en/latest/index.html
+  repository: https://github.com/NeoGeographyToolkit/StereoPipeline
+
+extra:
+  recipe-maintainers:
+    - oleg-alexandrov
+


oleg-alexandrov:

oleg-alexandrov:
Hi @ocefpaf and @hobu.

Thank you very much for your patience and effort in reviewing the visionworkbench feedstock.

This recipe is for stereo-pipeline. I started with @ocefpaf 's recipe and added more dependencies, so hopefully it starts in a good place.

Your thoughts are very welcome. I understand that this is volunteer work, so glad to hear from you when there is time.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+
+# Build dependencies first
+
+# Multiview
+cd $SRC_DIR/MultiView
+mkdir -p build && cd build
+cmake ..                                                              \
+    -DCMAKE_BUILD_TYPE=Release                                        \
+    -DMULTIVIEW_DEPS_DIR=${PREFIX}                                    \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13                               \
```
hobu: This should be conditionalized on whether or not we're OSX, not set for every `unix` build.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
+#!/bin/bash
+
+# Build dependencies first
+
+# Multiview
+cd $SRC_DIR/MultiView
```
hobu: Wait, I thought we were building stereopipeline?

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+mkdir -p build && cd build
+cmake ..                                                              \
+    -DCMAKE_BUILD_TYPE=Release                                        \
+    -DMULTIVIEW_DEPS_DIR=${PREFIX}                                    \
+    -DCMAKE_OSX_DEPLOYMENT_TARGET=10.13                               \
+    -DCMAKE_MODULE_PATH=$PREFIX/share/pcl-1.13/Modules                \
```
hobu: This seems like a hack. Use `CMAKE_ARGS` to get the `CMAKE_PREFIX_PATH`?

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+    -DCMAKE_VERBOSE_MAKEFILE=ON                                       \
+    -DCMAKE_CXX_FLAGS="-O3 -std=c++11 -Wno-error -I${PREFIX}/include" \
```
hobu: These flags should come from `CMAKE_ARGS`, why are we setting them ourselves?

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+    -DCMAKE_C_FLAGS='-O3 -Wno-error'                                  \
```
hobu: same here.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+  -DCMAKE_INSTALL_PREFIX=${PREFIX}             \
+  -DCMAKE_VERBOSE_MAKEFILE=ON                  \
+  -DCMAKE_PREFIX_PATH=${PREFIX}                \
+  -DCMAKE_VERBOSE_MAKEFILE=ON                  \
+  -DBUILD_SHARED_LIBS=ON                       \
+  -DEIGEN_INCLUDE_DIR=${PREFIX}/include/eigen3 \
+  -DBoost_DIR=${PREFIX}/lib                    \
+  -DBoost_INCLUDE_DIR=${PREFIX}/include        \
+  -DBoost_NO_BOOST_CMAKE=OFF                   \
+  -DBoost_DEBUG=ON                             \
```
hobu: IDK if we want this

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+  -DCMAKE_BUILD_TYPE=Release                 \
+  -DCMAKE_CXX_FLAGS="${INC_FLAGS}"           \
+  -DCMAKE_INSTALL_PREFIX:PATH=${PREFIX}      \
+  -DCMAKE_PREFIX_PATH=${PREFIX}              \
+  -DCMAKE_VERBOSE_MAKEFILE=ON                \
+  -DFastGlobalRegistration_LINK_MODE=SHARED  \
+  ${FGR_SOURCE_DIR}
+make -j${CPU_COUNT}
+# Install
+FGR_INC_DIR=${PREFIX}/include/FastGlobalRegistration
```
hobu: Why doesn't `FastGlobalRegistration` have a proper `install` target?

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+export LDFLAGS="-L$PREFIX/lib"
+# Extension
+if [ "$(uname)" = "Darwin" ]; then
+    EXT='.dylib'
+else
+    EXT='.so'
+fi
+# Build the desired programs
+cd 3rdparty/mgm
+perl -pi -e "s#CFLAGS=#CFLAGS=$CFLAGS #g" Makefile
```
hobu: if you're going to use perl to munge variables, `perl` needs to be a `build` dependency.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,193 @@
...
+    ${BIN_DIR}/msmw
+BIN_DIR=${PREFIX}/plugins/stereo/msmw2/bin
+mkdir -p ${BIN_DIR}
+/bin/cp -fv \
+    3rdparty/msmw2/build/libstereo_newversion/iip_stereo_correlation_multi_win2_newversion \
+    ${BIN_DIR}/msmw2
+
+# libelas
+if [ "$(uname -m | grep -i arm)" != "" ]; then 
+    echo Libelas does not build on Arm
```
hobu: So `osx-arm64` is not going to be a possible build configuration for StereoPipeline?

hobu (changes requested):

recipes/stereo-pipeline/build.sh
oleg-alexandrov: The libelas algorithm is a very minor algorithm in StereoPipeline, in a third-party add-on list. This one makes heavy use of x86 assembly so can't be ported easily.

The main StereoPipeline will run on all platforms, including Arm64, but without this option. (For the record, here's our full  [algorithm list](https://stereopipeline.readthedocs.io/en/latest/next_steps.html#stereo-algorithms).)

recipes/stereo-pipeline/build.sh
oleg-alexandrov: This is a third-party package that we adopted as-is. It is from an algorithm paper. I guess we could do further work on it.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: StereoPipeline has many code-level dependencies from other repositories. These are not of general-enough interest to have their own conda packages, so are built as prerequisites to the main thing.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Perl added as a build dependency (will push in a bit). This is another third-party research package that we make use of. Integrating these in the main code base does not make much sense, but some tidying each individual one at least a bit could be done.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Taken out.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I took care of "-I${PREFIX}/include", "$PREFIX/share/pcl-1.13/Modules", and "-DCMAKE_OSX_DEPLOYMENT_TARGET=10.13". 

It is hard to tell precisely if there is something else here that should be addressed. Let us see how this is is building first.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/build.sh
hobu: I guess the question is whether or not anyone needs to dynamically link this after it is installed in a Conda environment. If the answer is no, maybe just statically bury it in libstereopipeline.so?

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Adressed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: We do not have a single  libstereopipeline.so as the code base is big. There are many individual libAsp libraries for various functions. I guess I could rename it to libAspFgrAlign.so or something, but likely having it dynamic and not merged into an unrelated library would be preferable.

oleg-alexandrov:
@hobu: I took a first pass and the build made it.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: You can let me know what you think here. This lib is not something we build daily, and will need a code update and regeneration of our dependency tarballs for the 3 architectures we support (Mac Arm, Mac Intel, and Linux) and while happy to do it, it is something I hope to take on once.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+if [ "$(uname)" = "Darwin" ]; then
+    LIB_FLAG='-dynamiclib'
+    EXT='.dylib'
+    deployFlag="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.13"
+else
+    LIB_FLAG='-shared'
+    EXT='.so'
+    deployFlag=""
+fi
+
```
ocefpaf: ou can use conda's `${SHLIB_EXT}` here instead.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+# Build dependencies first
+
+# Multiview
+cd $SRC_DIR/MultiView
+mkdir -p build && cd build
+cmake ..                                          \
+    -DCMAKE_BUILD_TYPE=Release                    \
+    -DMULTIVIEW_DEPS_DIR=${PREFIX}                \
+    -DCMAKE_VERBOSE_MAKEFILE=ON                   \
+    -DCMAKE_CXX_FLAGS="-O3 -std=c++11 -Wno-error" \
```
ocefpaf: Conda set these these in `${CXXFLAGS}` you should modify those only in extreme case. The default setting ensure compatibility with our ecosystem of packages.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+    -DCMAKE_C_FLAGS='-O3 -Wno-error'              \
```
ocefpaf: Same as above. The default here is `${CFLAGS}`.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+    -DCMAKE_INSTALL_PREFIX=${PREFIX}              \
+    $deployFlag
+make -j${CPU_COUNT} install
```
ocefpaf: Is there a make check or ctest that we can run here?
(Same question for all the other make calls below.)

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+cd $SRC_DIR/s2p
+baseDir=$(pwd)
+export CFLAGS="-I$PREFIX/include -O3 -DNDEBUG -march=native"
+export LDFLAGS="-L$PREFIX/lib"
+# Extension
+if [ "$(uname)" = "Darwin" ]; then
+    EXT='.dylib'
+else
+    EXT='.so'
+fi
```
ocefpaf: You are redefining these from above. Also, you don't need them. See my other comment.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+mkdir -p ${FGR_INC_DIR}
+/bin/cp -fv ${FGR_SOURCE_DIR}/FastGlobalRegistration/app.h ${FGR_INC_DIR}
+FGR_LIB_DIR=${PREFIX}/lib
+mkdir -p ${FGR_LIB_DIR}
+/bin/cp -fv FastGlobalRegistration/libFastGlobalRegistrationLib* ${FGR_LIB_DIR}
+
+# s2p
+cd $SRC_DIR/s2p
+baseDir=$(pwd)
+export CFLAGS="-I$PREFIX/include -O3 -DNDEBUG -march=native"
```
ocefpaf: Will this ship debug symbols?

recipes/stereo-pipeline/build.sh
ocefpaf: You can use `sed`, that is present and avoids a `perl` installation, or better yet a patch to make this change more transparent for someone reading the recipe.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,186 @@
...
+else
+    cd $SRC_DIR/libelas
+    # Set the env
+    export CFLAGS="-I$PREFIX/include -O3 -DNDEBUG -ffast-math -march=native"
+    export LDFLAGS="-L$PREFIX/lib"
+    if [ "$(uname)" = "Darwin" ]; then
+        EXT='.dylib'
+    else
+        EXT='.so'
+    fi
```
ocefpaf: Remove.

PS: this build script is quite convoluted and hard to review and maintain. Consider refactoring it.

recipes/stereo-pipeline/recipe.yaml
```diff
@@ -0,0 +1,154 @@
...
+      sha256: 3bd48acc05af13fc5dfbf4ca397f9ffa28fd53c83a28346358dff0726e4d2996
+      target_directory: FastGlobalRegistration
+      
+    - url: https://github.com/NeoGeographyToolkit/s2p/releases/download/3.5.0/s2p.tar.gz
+      sha256: 8752edac97f9596266d2712123ea1d14e44e7c201c8ea4062223708886ae5237
+      target_directory: s2p
+      
+    - url: https://github.com/NeoGeographyToolkit/libelas/archive/refs/tags/3.5.0.zip
+      sha256: d160bbfd020aefac4782a5b269d9c0f8d6af5f43cb8951d27204d2068315521e
+      target_directory: libelas
```
ocefpaf: All the hardcoded versions here will make this impossible for the bot and our infrastructure to handle this recipe.

Why not separate packages in separate recipes?

recipes/stereo-pipeline/recipe.yaml
```diff
@@ -0,0 +1,154 @@
...
+    - rapidjson
+    - rocksdb
+    - tbb
+    - tbb-devel
+    - suitesparse
+    - superlu
+    - usgscsm
+    - visionworkbench
+    - vlfeat
+    - xerces-c
```
ocefpaf: Most of these have `run_exports` and should not be listed in `run`. See the cleanup done in visionworkbench.

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: These are very small auxiliary software. 

I can make the versions be not hard-coded. 

I do not plan to make these separate packages. I tried this. The maintenance burden on me is too big that way, it always takes weeks to deal with compiler, architecture, and conda issues.

I am very happy to implement all your suggestions except this one. I hope this won't be a deal breaker.

oleg-alexandrov: I can prepare one single code archive, with a single sha256, packaging these into subdirectories. conda-forge won't care then.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: No. NDEBUG means don't use debug symbols.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: The -std=c++11 is expected by some of the third-party software. As a maintainer of this thing, I am aware of long-term issues with keeping old-style code, it will cause problems for me first, as the code won't compile or won't link.

So, this will affect my package before it affects conda. I do not have time now for the massive work needed to review that code and make the needed changes.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: These are many third-party dependencies. StereoPipeline itself has many tools. I have an external test suite of 293 tests that tests all this. See https://github.com/NeoGeographyToolkit/StereoPipelineTest.

Given your preference, I added one test for one tool, in meta.yml. I think it is not reasonable to expect conda to ensure the correctness of all the code I ship.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: It has many third-party packages. I can put these all in an internal script, that conda can run. Would that work?

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: Addressed for most of these. I added this because the build was failing. Let us see how it goes now.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I removed -DCMAKE_C_FLAGS. I need to keep the C++ flags until I find time to review and overhaul the internals of these.

oleg-alexandrov:
@ocefpaf : I put fixes for most of these. I can refactor the code if need be.

As before what I cannot do is split this into multiple conda packages. That puts undue burden on me as the maintainer, and I spend at least two weeks each year dealing with conda and and platform issues. I can abstract away some of these, in one single package.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed.

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: The build failed when I took most of these out. Now I keep only a handful.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I missed this one. Looks like I need perl anyway, for the "parallel" program which is a prerequisite.

hobu: `parallel` should then be used as the dependency, not general perl https://anaconda.org/conda-forge/parallel

recipes/stereo-pipeline/build.sh
hobu: > The -std=c++11 is expected by some of the third-party software

I know this is a long(er) term thing but you should be enforcing the linkage of these things with [CXX_STANDARD](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html) and [C_STANDARD](https://cmake.org/cmake/help/latest/prop_tgt/C_STANDARD.html) on your targets. Deferring this to CMake was annoying at first but definitely tightened up behaviors.

recipes/stereo-pipeline/build.sh
hobu: > NDEBUG means don't use debug symbols.

Another longer term project thing, but I've found it much easier to conditionalize DEBUG/NDEBUG with the `CMAKE_BUILD_TYPE` variable and having that be the trigger for various flag settings and such. Too easy for people compiling to do the wrong thing. As you have things configured, debug symbols must be defaulted to ON, right?

recipes/stereo-pipeline/build.sh
hobu: > I can put these all in an internal script

Or even better, CMake submodules or something. All of the specificity in this build script has me concerned that doing the same on windows is going to be 2x the effort. Maybe Windows builds aren't a thing though.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Yes, parallel is a dependency in the recipe.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Thanks. Indeed. This third-party code is itself made of several components, each with its own build system. Next time around I should replace all that with a cmake build, but this will be some time commitment.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Yes, cmake modules will work. I did that for another package with many loose pieces. To be done. 

For Windows, the users now employ WSL. A native build so far has not been asked for.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Good point. I put CXX_STANDARD in the CMakeLists.txt, and took out the external flag. This way any options conda-forge wants to add on will be appended to the build.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I took out the CXX_FLAGS too, per hobu.

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: I addressed I think the biggest sticking point here. I have now a single versioned source package.

recipes/stereo-pipeline/build.sh
ocefpaf: > Given your preference, I added one test for one tool, in meta.yml. I think it is not reasonable to expect conda to ensure the correctness of all the code I ship.

It is not my preference. It is expected from all compiled Software to be tested in the different architectures and build systems they are shipped. Something that passes and is correct in the CIs upstream may fail and be broken here.

ocefpaf:
@xhochy, @jaimergp, or @isuruf. Apologies for the direct ping, but if any of you have some free cycles do you mind taking over this review? I'm taking a break from stage-recipes.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,167 @@
+#!/bin/bash
+
+# Architecture-dependent flags
+if [ "$(uname)" = "Darwin" ]; then
+    LIB_FLAG='-dynamiclib'
+    deployFlag="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.13"
```
isuruf: This is unneeded.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,167 @@
...
+else
+    LIB_FLAG='-shared'
+    deployFlag=""
+fi
+
+CMAKE_ARGS+=("-DCMAKE_PREFIX_PATH=${PREFIX}")
```
isuruf: this is unnecessary too.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,167 @@
...
+
+# Build dependencies first
+
+# Multiview
+cd $SRC_DIR/MultiView
+mkdir -p build && cd build
+cmake ..                                          \
```
isuruf:
```suggestion
cmake ${CMAKE_ARGS} ..                                          \
```

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,167 @@
...
+make -j${CPU_COUNT} install
+
+# Geoid
+cd $SRC_DIR/geoids
+# Build
+${FC} ${FFLAGS} -fPIC -O3 -c interp_2p5min.f
+${FC} ${LDFLAGS} ${LIB_FLAG} -o libegm2008${SHLIB_EXT} interp_2p5min.o
+# Install
+mkdir -p ${PREFIX}/lib
+/bin/cp -fv libegm2008.* ${PREFIX}/lib
```
isuruf:
```suggestion
cp -fv libegm2008.* ${PREFIX}/lib
```

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,167 @@
...
+    3rdparty/msmw/build/libstereo/iip_stereo_correlation_multi_win2 \
+    ${BIN_DIR}/msmw
+BIN_DIR=${PREFIX}/plugins/stereo/msmw2/bin
+mkdir -p ${BIN_DIR}
+/bin/cp -fv \
+    3rdparty/msmw2/build/libstereo_newversion/iip_stereo_correlation_multi_win2_newversion \
+    ${BIN_DIR}/msmw2
+
+# libelas
+if [ "$(uname -m | grep -i arm)" != "" ]; then
```
isuruf:
```suggestion
if [[ "$target_platform" != "aarch64" && "$target_platform" != "arm64" ]]; then
```

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Thank you for you help. 

We have dozens of tools and lots of modes in which they are used. Regressions (and also documentation) are a very big part of my job. 

My testsuite takes 30GB of disk space and runs for more than 1 hour on 16 cores. This cannot happen on conda-forge servers.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I took out the CMAKE_OSX_DEPLOYMENT_TARGET. The -dynamiclib seems to be needed.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Taken out.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Added to all cmake calls.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Done.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: I had to tweak this a bit, as the if statement applies to it being Arm, rather than not Arm. Done. 

Thank you for reviewing.

recipes/stereo-pipeline/recipe.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - ale
+    - csm
+    - parallel
+    - usgscsm
+    - visionworkbench
+
+tests:
+  - script:
+    # Test loading a camera and doing image-to-ground and ground-to-image operations
+    - cam_test --image $PREFIX/share/tests/ctx.json --cam1 $PREFIX/share/tests/ctx.json --cam2 $PREFIX/share/tests/ctx.json
```
jaimergp, line 72: That `tests/` subdirectory is a bit too general, might be worth renaming to something more package-specific, like `$PREFIX/share/stereo-pipeline/tests/*`.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,166 @@
...
+mkdir -p ${BIN_DIR}
+cp 3rdparty/msmw/build/libstereo/iip_stereo_correlation_multi_win2 \
+   ${BIN_DIR}/msmw
+BIN_DIR=${PREFIX}/plugins/stereo/msmw2/bin
+mkdir -p ${BIN_DIR}
+cp 3rdparty/msmw2/build/libstereo_newversion/iip_stereo_correlation_multi_win2_newversion \
+   ${BIN_DIR}/msmw2
+
+# libelas
+if [[ "$target_platform" == "aarch64" || "$target_platform" == "arm64" ]]; then
```
jaimergp: I don't think these will ever be true, since `$target_platform` takes the form `{OS}-{ARCH}`. You might need a glob or something like that.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,166 @@
...
+make -j${CPU_COUNT} install
+
+# fgr
+cd $SRC_DIR/FastGlobalRegistration
+FGR_SOURCE_DIR=$(pwd)/source
+mkdir -p build && cd build
+INC_FLAGS="-I${PREFIX}/include/eigen3 -I${PREFIX}/include -O3 -L${PREFIX}/lib -lflann_cpp -llz4 -O3 -std=c++11"
+cmake                                        \
+  ${CMAKE_ARGS}                              \
+  -DCMAKE_BUILD_TYPE=Release                 \
```
jaimergp: This is [redundant](https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/8a72807d9d65c6e5170495415332c509e9e6537e/recipe/activate-gcc.sh#L147) with `$CMAKE_ARGS`.

recipes/stereo-pipeline/build.sh
```diff
@@ -0,0 +1,166 @@
...
+  -DCMAKE_CXX_FLAGS="${INC_FLAGS}"           \
+  -DCMAKE_INSTALL_PREFIX:PATH=${PREFIX}      \
```
jaimergp: [Redundant](https://github.com/conda-forge/ctng-compiler-activation-feedstock/blob/8a72807d9d65c6e5170495415332c509e9e6537e/recipe/activate-gcc.sh#L154) with `$CMAKE_ARGS`.

recipes/stereo-pipeline/build.sh
jaimergp: (here and in the other calls too)

recipes/stereo-pipeline/recipe.yaml
```diff
@@ -0,0 +1,90 @@
...
+
+about:
+  homepage: https://github.com/NeoGeographyToolkit/StereoPipeline
+  summary: 'A suite of tools for producing terrain models published by NASA'
+  description: |
+    The NASA Ames Stereo Pipeline is a collection of tools for producing 3D terrain models from images acquired by satellites of planets and of Earth. It also supports alignment of terrains, correction of errors in camera positions and orientations, and structure from motion.
+  license: Apache-2.0
+  license_family: Apache
+  license_file: StereoPipeline/LICENSE
```
jaimergp: Given the amount of projects being built, I assume they are vendored from other sources, and in that case their licenses must appear here.

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: This is the test directory of the stereo-pipeline package that I am shipping. Libraries go to lib, binaries to bin, and tests go tests. The test has a unique enough name, and it is rare tests are stored there, so I'd think the potential for conflict is not big.

recipes/stereo-pipeline/recipe.yaml
oleg-alexandrov: We have third party licenses. I don't know how to include them for conda. Here's our list: https://stereopipeline.readthedocs.io/en/latest/thirdparty.html

oleg-alexandrov: @jaimergp Will you be able to help me push this over the finish line, after further review? I have had much feedback so far, but it seems that that's where things stop.

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed

recipes/stereo-pipeline/build.sh
oleg-alexandrov: Addressed

recipes/stereo-pipeline/recipe.yaml
jaimergp: I don't know, `ctx.json` doesn't sound super unique to me. You don't need to address this now but I'd include as an issue upstream as a good practice to adopt in the future.

recipes/stereo-pipeline/recipe.yaml
jaimergp: You can add them as sources too, but it will be tricky to keep them up to date. Looks like you can list [this file](https://github.com/NeoGeographyToolkit/StereoPipeline/blob/master/THIRDPARTYLICENSES.rst) as one more entry in the `license_file` list as a low-effort change, but some of those licenses require the redistribution of the license itself.

You will need to identify which of those are actually vendored as part of the build, because those are the ones you need to include. The linked libraries carry their own license as part of the conda package, so you don't need to list those.

I might miss replies in this thread (my Github notifications are a mess!) so please feel free to reach me out in a thread in our [Zulip instance](https://conda-forge.zulipchat.com/) and ping me there to make sure I see it.

imincik:
> Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).

@oleg-alexandrov , I recommend to split this package to components based on their original source code repository ([stereo-pipeline](https://github.com/NeoGeographyToolkit/StereoPipeline), [fgr](https://github.com/NeoGeographyToolkit/FastGlobalRegistration), [multiview](https://github.com/NeoGeographyToolkit/MultiView) , ...  ) to avoid vendoring other packages as `StereoPackage.tar.gz` does.

BTW, what is the original source code repository for `geoid` package ?

oleg-alexandrov:
Thank you for this advice. I will adress this in a few weeks.

danielnachun:
@oleg-alexandrov Thanks for being willing to work on devendoring this - vendored dependencies are a real pain for both licensing and so-called "phantom dependencies" so we try as much as possible to avoid vendoring.

oleg-alexandrov:
This takes a lot of lengthy debug cycles to make work. New clang compilers keep on breaking stuff that is in dependencies. Closing this till there's time to work on it.

PR #30420, opened 2025-06-26
Add spade

Prabhu-LANL:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/spade/meta.yaml
+++ b/recipes/spade/meta.yaml
@@ -50,6 +50,7 @@ requirements:
 
 test:
   requires:
+    - python {{ python_min }}
     - pytest
     - pytest-xdist
   imports:


kbrindley:
I am wiling to be a maintainer

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - pytest
+    - pytest-xdist
+  imports:
+    - spade
+  commands:
+    # Test as-installed project unit tests
+    - working_directory=${PWD}
+    - cd $SP_DIR/$PKG_NAME
+    - pytest -n 4 -vvv -m "not require_third_party"
+    - pytest -n 4 -v --no-showlocals -m "systemtest and require_third_party" --tb=short --system-test-dir=${working_directory} --abaqus-command=/apps/abaqus/Commands/abq2024
```
kbrindley: @Prabhu-LANL the conda-forge ci servers won't have the necessary software (Abaqus) to run the ``systemtest and require_third_party`` tests. This entire line should be removed.

Prabhu-LANL: Thanks for catching this @kbrindley :-)

kbrindley:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1332385&view=logs&jobId=4cabe686-70ae-553a-7fd0-310379f2cbac&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff

```
python -m pip install $SRC_DIR/build/dist/spade-0.4.1.tar.gz --prefix $PREFIX --log $SRC_DIR/build/install.log --verbose --no-input --no-cache-dir --disable-pip-version-check --no-deps --ignore-installed --no-build-isolation --no-warn-script-location --no-index
/home/conda/staged-recipes/build_artifacts/spade_1757372080266/_build_env/bin/python: No module named pip
```

kbrindley:
> https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1332385&view=logs&jobId=4cabe686-70ae-553a-7fd0-310379f2cbac&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff
> 
> ```
> python -m pip install $SRC_DIR/build/dist/spade-0.4.1.tar.gz --prefix $PREFIX --log $SRC_DIR/build/install.log --verbose --no-input --no-cache-dir --disable-pip-version-check --no-deps --ignore-installed --no-build-isolation --no-warn-script-location --no-index
> /home/conda/staged-recipes/build_artifacts/spade_1757372080266/_build_env/bin/python: No module named pip
> ```

Pretty sure it failed in the build stage. The SCons "install" alias uses ``python -m pip``, which is probably picking up python from the "build" environment of the conda recipe. ``python -m pip`` forces use of the pip package in the same environment as the python executable.

kbrindley:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1333142&view=logs&jobId=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de

windows failure looks like a bash vs. cmd syntax error

```
[vcvarsall.bat] Environment initialized for: 'x64'
import: 'spade'
import: 'spade'
'working_directory' is not recognized as an internal or external command,
operable program or batch file.
WARNING: Tests failed for spade-0.4.1-py310h401fb6d_0.conda - moving package to D:\bld\broken
```

stemming from the test commands here: https://github.com/conda-forge/staged-recipes/pull/30420/files#diff-b7330e0833a4b184713c911e18167dca4c454a47b7a76d7fac5132034c230bdbR72, which use bash syntax.

Fix is either disable Windows support/builds or move the test commands to shell specific files ``run_test.sh`` and ``run_test.bat``. Probably worth attempting the Windows batch file first. If we can make the unit tests run for low effort, that's a good first step towards real Windows support.

Prabhu-LANL:

Prabhu-LANL:

kbrindley:
@Prabhu-LANL there's a "new" solution for glibc requirements documented here: https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-glibc-versions. New meaning, I don't remember this documentation when we were fighting with glibc several years ago.

edit: I think SPADE will require a newer glibc spec using a ``conda_build_config.yaml`` file

```
c_stdlib_version:          # [linux]
  - "2.28"                 # [linux]
```

the actual spec in the recipe ``meta.yaml`` file looks like it matches the current conda-forge documentation.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,80 @@
+{% set name = "spade" %}
+{% set version = "0.4.6" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/lanl-aea/spade/archive/refs/tags/0.4.6.tar.gz
```
kbrindley: @Prabhu-LANL The url needs to use the version

```suggestion
  url: https://github.com/lanl-aea/spade/archive/refs/tags/{{ version }}.tar.gz
```

kbrindley:
```
ValueError: variant config in /home/conda/staged-recipes-copy/recipes/spade/conda_build_config.yaml is ambiguous because it does not fully implement all zipped keys (missing ['is_python_min']) or specifies a subspace that is not fully implemented (we did not find ['3.10', '3.11', '3.12'] from {'libstdcxx': ['12', '13'], 'hdf5': ['1.12.2', '1.14.3'], 'python': ['3.10', '3.11', '3.12'], 'numpy': ['1.26', '2.0'], 'c_stdlib': ['sysroot'], 'c_stdlib_version': ['2.28']} in python:['3.10.* *_cpython', '3.11.* *_cpython', '3.12.* *_cpython', '3.13.* *_cp313']).
```

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,17 @@
+libstdcxx:  # [linux]
+  - 12      # [linux]
+  - 13      # [linux]
+hdf5:
+  - 1.12.2
+  - 1.14.3
+python:
+  - 3.10
+  - 3.11
+  - 3.12
```
kbrindley: conda-forge wants to manage the python spec. we should remove these python specs.

this is one of the differences between most conda recipes and a conda-forge recipe. conda-forge adds the python variants as multiple Azure pipelines, so when the conda-forge recipe has python variants in the ``conda_build_config.yaml`` file the python specs are provided twice and in conflicting ways.

```suggestion
```

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,17 @@
+libstdcxx:  # [linux]
+  - 12      # [linux]
+  - 13      # [linux]
```
kbrindley: These don't appear to be used anywhere? They also follow our old internal convention, which we've replaced with ``stdlib('c')`` in this conda-forge feedstock.

```suggestion
```

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,80 @@
...
+    - setuptools >=77
+    - setuptools_scm >=8
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - hdf5
+    - scons >=4.6
+    - ld_impl_linux-64         # [linux]
+  run:
+    - python
+    - {{ pin_compatible('numpy') }}
```
kbrindley: https://conda-forge.org/docs/maintainer/knowledge_base/#building-against-numpy

conda-forge prefers to rely on the emitted runtime dependency that comes from putting ``numpy`` in the ``host`` section. If SPADE needs a specific numpy minimum, we're recommended to do something like the following, but *only* if the numpy run-exports explained in the above URL won't work for some reason.

```
host:
  # leave this unpinned!
  - numpy
run:
  - numpy >={{ the_lower_bound_your_package_needs }}
```

```suggestion
```

recipes/spade/run_test.sh
```diff
@@ -0,0 +1,4 @@
+system_test_directory=$PWD
+cd $SP_DIR/$PKG_NAME
+pytest -n 4 -vvv -m "not require_third_party" --system-test-dir=${system_test_directory}
+
```
kbrindley:
```suggestion
```

recipes/spade/conda_build_config.yaml
Prabhu-LANL: Thanks, I didn't know that.

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,10 @@
+hdf5:
+  - 1.12.2
+  - 1.14.3
+numpy:
+  - 1.26
+  - 2.0  #Lowest compatible version
```
kbrindley: I think something in the numpy spec conflicts with the python variants? I didn't study the log completely. I also don't see any direct numpy imports in spade. Is numpy a requirement at all?

If we keep numpy as a requirement, maybe drop to the conda-forge recommended spec?

```suggestion
numpy:
  - 2.0
```

kbrindley:

kbrindley:
I don't understand the macos failure. We only have a ``meta.yaml`` file. I suppose I can try updating the branch once again?

```
+ python .ci_support/build_all.py --arch 64
Building all recipes
/Users/runner/work/1/s/.ci_support/compute_build_graph.py:37: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
  import pkg_resources
  File "/Users/runner/work/1/s/.ci_support/build_all.py", line 399, in <module>
    build_all(os.path.join(root_dir, "recipes"), args.arch)
  File "/Users/runner/work/1/s/.ci_support/build_all.py", line 109, in build_all
    raise ValueError("Mixing meta.yaml and recipe.yaml recipes is not supported")
ValueError: Mixing meta.yaml and recipe.yaml recipes is not supported
```

kbrindley:
@conda-forge/staged-recipes this is ready for review now.

this is a bit of a weird package. It can build like a "noarch" package, but it must perform runtime compilation so it requires an architecture specific cxx compiler.

kbrindley:
> conda-forge/staged-recipes this is ready for review now.
> 
> this is a bit of a weird package. It can build like a "noarch" package, but it must perform runtime compilation so it requires an architecture specific cxx compiler.

@conda-forge/help-python @conda-forge/help-python-c

traversaro:
Interesting, I wonder if it could make sense to just add a run dependency on `cxx-compiler` in this case, even if typically we do not have use them as a regular dependency on conda packages. Perhaps @conda-forge/staged-recipes @conda-forge/help-c-cpp may provide an input on this.

kbrindley:
> Interesting, I wonder if it could make sense to just add a run dependency on `cxx-compiler` in this case, even if typically we do not have use them as a regular dependency on conda packages. Perhaps conda-forge/staged-recipes conda-forge/help-c-cpp may provide an input on this.

@traversaro the package also needs a minimum runtime version of ``glibc >=2.28`` up from the normal ``glibc ==2.17`` Is it possible to set the minimum without ``{{ stdlib('c') }}``?

https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-glibc-versions.

kbrindley:
> > Interesting, I wonder if it could make sense to just add a run dependency on `cxx-compiler` in this case, even if typically we do not have use them as a regular dependency on conda packages. Perhaps conda-forge/staged-recipes conda-forge/help-c-cpp may provide an input on this.
> 
> @traversaro the package also needs a minimum runtime version of `glibc >=2.28` up from the normal `glibc ==2.17` Is it possible to set the minimum without `{{ stdlib('c') }}`?
> 
> https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-glibc-versions.

@traversaro @Prabhu-LANL I've been chasing the Windows build to see if we can get away with ``cxx-compiler`` as a runtime dependency. Long story short, we will definitely need architecture specific builds. The Windows build will need a few additional Windows-specific runtime packages beyond the compilers.

@traversaro If the recipe looks ok as written, I'd like to merge and create the feedstock. It will still be useful to end users to have the functional Python bits of the package while we chase bugs in the Windows runtime compilation behavior.

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_stdlib:
+  - sysroot                    # [linux]
```
traversaro:
```suggestion
```

is this necessary?

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
...
+c_stdlib_version:              # [unix]
+  - 2.28                       # [linux and x86_64]
```
traversaro: I guess the preprocessor directives should be consistent here?

traversaro:
Ok, thanks for the update. However, I do not think that `{{ compiler('cxx') }}` and ` {{ stdlib('c') }}` make sense as run dependency, as they would leak aspects of the build platform to the final package, and I guess this is not what you want.

Can you please:
* Explain where the `glibc` dependency emerges from if this does not contain any compiled code?
* You can have platform-specific dependencies also in noarch packages, see https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies .

danielnachun:
> > Interesting, I wonder if it could make sense to just add a run dependency on `cxx-compiler` in this case, even if typically we do not have use them as a regular dependency on conda packages. Perhaps conda-forge/staged-recipes conda-forge/help-c-cpp may provide an input on this.
> 
> @traversaro the package also needs a minimum runtime version of `glibc >=2.28` up from the normal `glibc ==2.17` Is it possible to set the minimum without `{{ stdlib('c') }}`?
> 
> [conda-forge.org/docs/maintainer/knowledge_base#requiring-newer-glibc-versions](https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-glibc-versions).

If you are dropping the compiled components for now, this answer is not as important, but I'm very curious as to why this would need a newer `glibc` at all.  Often this is a sign of vendored binaries - there are not many packages that actually have an API dependency on new features in `glibc` or the Linux headers.

recipes/spade/conda_build_config.yaml
kbrindley: I think so. The third-party software we link against at runtime has a minimum ``glibc`` requirement.

https://conda-forge.org/docs/maintainer/knowledge_base/#requiring-newer-glibc-versions

I see ``c_stdlib_version`` in the example, but not the ``c_stdlib: sysroot``. I remember seeing them together, but can't find the reference.

recipes/spade/conda_build_config.yaml
kbrindley: @Prabhu-LANL do you remember why these aren't just ``# [linux]``?

kbrindley:
@traversaro 

> Can you please:
> 
> * Explain where the glibc dependency emerges from if this does not contain any compiled code?

@danielnachun 

> If you are dropping the compiled components for now, this answer is not as important, but I'm very curious as to why this would need a newer glibc at all. Often this is a sign of vendored binaries - there are not many packages that actually have an API dependency on new features in glibc or the Linux headers.

We perform runtime compilation and linking against a user's installation of a third-party software, Abaqus. Abaqus requires a minimum glibc version. This package does not include any Abaqus binaries, which makes the runtime behavior more complicated.

The runtime compilation and linking works for the linux platform. I'm just chasing the Windows behavior.

@traversaro 

> You can have platform-specific dependencies also in noarch packages, see https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies.

I'll take a stab at this.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - setuptools >=77
+    - setuptools_scm >=8
+    - sphinx >=7.1
+    - sphinx-argparse
+    - sphinx-book-theme
+    - sphinx-copybutton >=0.5.1
+    - sphinx-design
+    - sphinxcontrib-bibtex
+  run:
+    - cxx-compiler  # [linux or win]
```
kbrindley, line 41: Third-party package that requires runtime compilation, Abaqus, does not support osx. The Python portion of this package is still useful on osx.

recipes/spade/conda_build_config.yaml
kbrindley: no longer relevant. I converted to the multi-os noarch style recipe: https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

(edit: spelling)

recipes/spade/conda_build_config.yaml
kbrindley: no longer relevant. I converted to the multi-os noarch style recipe: https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

(edit: spelling)

danielnachun:
Thank you for clarifying some of the details about how this package works.  Based on what you've said I think the strategy we are taking now makes sense.  I am going to make some suggested changes to simplify things a bit and make the linter happy.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  # Use command: curl -sL <URL above> | openssl sha256
+
+build:
+  number: 0
+  noarch: python
+  script_env:
+    - SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+  script:
+    - scons install --prefix=${PREFIX}  # [unix]
+    - scons install --prefix=%PREFIX%  # [win]
```
danielnachun, line 20:
```suggestion
  script:
    - scons install --prefix=${PREFIX}
```
Can you check if this works on Windows without having to specify `%PREFIX%` instead of `${PREFIX}`.  I think it is smart enough to translate it correctly here, but I could be mistaken.

recipes/spade/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+noarch_platforms:
+  - linux_64
+  - osx_64
+  - win_64
```
danielnachun, line 4:
```suggestion
noarch_platforms:
  - linux_64
  - win_64
c_stdlib_version:  # [linux]
  - 2.28
```
Can you try moving the `glibc` requirement here like this, and then removing it from the recipe?  Let's see if the linter accepts that.  Also I've made some other suggestions that should allows to build on Linux for macOS.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - sphinxcontrib-bibtex
+  run:
+    - python >={{ python_min }}
+    - cxx-compiler  # [linux or win]
+    - hdf5  # [linux or win]
+    - scons >=4.6
+    - __linux  # [linux]
+    - __glibc >=2.28  # [linux]
+    - __osx  # [osx]
+    - __win  # [win]
```
danielnachun, line 46:
```suggestion
    - {{ compiler('cxx') }}
    - hdf5
    - scons >=4.6
    - __linux  # [linux]
    - __osx  # [osx]
    - __win  # [win]
```
I think we can just make `cxx_compiler` and `hdf5` unconditional run time dependencies.  It's not worth the trouble of trying to make a separate noarch variant just to skip these dependencies.

danielnachun (changes requested):
I've made some suggestions here to see if we can get the recipe to pass the linter.  If things fail, just leave the commits as they are and don't revert so we can see what happened and try to fix it.

recipes/spade/meta.yaml
kbrindley: Do the conda-forge runners use a ``cmd`` shell? Some snippets from the log make it look like a ``cmd`` shell. (edit: So I think we need a ``cmd`` specific variable ``%PREFIX%`` for the windows build.)

I also tried some local builds in Powershell 7 with ``${PREFIX}`` and ``$PREFIX``. Both completed without errors, but neither produced a populated package. I think the package files installed to a ``%SRC_DIR%/${PREFIX}`` directory, where the ``${PREFIX}`` portion of the path was treated as literal characters instead of expanding as a shell variable.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1418400&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de

shell start
```
"C:\Windows\system32\cmd.exe" /D /E:ON /V:OFF /S /C "CALL "D:\a\_temp\8cce99da-78a9-431d-8b08-36fd79e59069.cmd""
```

mamba activation output
```
To automatically initialize all future (cmd.exe) shells, run:
    $ mamba shell init --shell cmd.exe --root-prefix=~/.local/share/mamba
If your shell was already initialized, reinitialize your shell with:
    $ mamba shell reinit --shell cmd.exe
```

build script environment reporting
```
(base) %SRC_DIR%>set "PATH=%BUILD_PREFIX%;%BUILD_PREFIX%\Library\mingw-w64 ..." 
```

recipes/spade/meta.yaml
kbrindley: If  we drop the ``noarch_platforms: osx_64``, but keep the platform virtual package ``__linux``, I don't think there will be an osx-64 compatible package at all?

If I follow the documentation to use the linux/osx platform virtual package ``__unix``, but switch to `` {{ compiler('cxx') }}`` won't the linux-64 built runtime compiler dependency use the linux specific compiler, and still be incompatible for osx?

The cxx-compiler and hdf5 libraries bring in a long list of second and third order dependencies. I'm not opposed to supporting an osx-64 noarch variant. Does a third variant cause a significant increase on the overhead of conda-forge compute resources?

recipes/spade/conda_build_config.yaml
kbrindley: When I remove the ``__glibc`` virtual package from the ``meta.yaml`` file and add this ``c_stdlib_version`` to the ``conda_build_config.yaml`` file

* using ``cxx-compiler`` there is no ``glibc`` runtime dependency of the built package
* using ``{{ compiler('cxx') }}`` the linux-64 built package depends on a linux specific compiler ``gxx_linux-64`` and there is still no ``glibc`` runtime dependence.

I think we would need the corresponding ``{{ stdlib('c') }}`` in meta.yaml to get the ``glibc`` runtime dependence. We would end up requiring an osx-64 variant, anyway because both the ``{{ compiler('cxx') }}`` and ``{{ stdlib('c') }}`` for linux-64 and osx-64 resolve to different, platform specific runtime requirements.

kbrindley:
> I've made some suggestions here to see if we can get the recipe to pass the linter. If things fail, just leave the commits as they are and don't revert so we can see what happened and try to fix it.

@danielnachun 

I tried the closest thing to your suggestions that should result in three packages with system specific compilers (running in the upstream's recipe draft, which is nearly identical, but not one-to-one with conda-forge infrastructure). 

```
$ git diff
diff --git a/recipe/conda_build_config.yaml b/recipe/conda_build_config.yaml
index 9d57aa2..8cc9430 100644
--- a/recipe/conda_build_config.yaml
+++ b/recipe/conda_build_config.yaml
@@ -2,3 +2,5 @@ noarch_platforms:  # Conda-forge specific build behavior. Builds one package per
   - linux_64
   - osx_64
   - win_64
+c_stdlib_version:  # [linux]
+  - 2.28  # [linux]
diff --git a/recipe/meta.yaml b/recipe/meta.yaml
index 5309b3c..8710f47 100644
--- a/recipe/meta.yaml
+++ b/recipe/meta.yaml
@@ -36,15 +36,14 @@ requirements:
     - sphinxcontrib-bibtex
   run:
     - python >={{ python_min }}
-    - cxx-compiler  # [linux or win]
-    - hdf5  # [linux or win]
     - scons >=4.6
+    - {{ compiler('cxx') }}  # [linux or win]
+    - {{ stdlib('c') }}  # [linux]
+    - hdf5  # [linux or win]
     - __linux  # [linux]
-    - __glibc >=2.28  # [linux]
     - __osx  # [osx]
     - __win  # [win]
     - getopt-win32  # [win]
-    - vs2022_win-64  # [win]
```

The package runtime dependencies don't look quite like I expected. I expected to see the ``__glibc >=2.28`` instead of ``c_linux-64 2.28.*``.

```
$ conda search -c ./conda-bld --override-channels --info
Loading channels: done
spade 6.6.6 pyheba2bb5_0
------------------------
file name   : spade-6.6.6-pyheba2bb5_0.tar.bz2
name        : spade
version     : 6.6.6
build       : pyheba2bb5_0
build number: 0
size        : 1.8 MB
license     : BSD-3-Clause
subdir      : noarch
url         : file:///projects/kbrindley/repos/spade/conda-bld/noarch/spade-6.6.6-pyheba2bb5_0.tar.bz2
md5         : f7456890fb519ffc8d579d862d8f391a
timestamp   : 2025-12-16 21:05:30 UTC
dependencies: 
  - __linux
  - c_linux-64 2.28.*
  - gxx_linux-64
  - hdf5
  - python >=3.10
  - scons >=4.6
```

kbrindley:
@traversaro @danielnachun I think I've arrived at a set of runtime requirements that does the closest thing to your joint reviews.

1. Minimizes the number of package builds to 3 platform specific noarch packages
2. Does not leak aspects of the build platform to the final package
3. Imitates the os-specific runtime specs of a compiled package (and correctly matches the glibc and sysroot dependencies on linux)

I don't think we can satisfy the linter, which complains about the platform specific, noarch solution documented here: https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies, where both a ``build: noarch: python`` and platform selectors are used together.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - scons >=4.6
+    - cxx-compiler  # [linux or win]
+    - hdf5  # [linux or win]
+    - __linux  # [linux]
+    - __glibc >=2.28,<3  # [linux]
+    - sysroot_linux-64 >=2.28,<3  # [linux]
+    - __win  # [win]
+    - getopt-win32  # [win]
+    - vs2022_win-64  # [win]
+    - __osx  # [osx]
```
danielnachun, line 49:
```suggestion
    - cxx-compiler
    - hdf5
```
I don't know why I didn't suggest this earlier, but I think the above solution can make the linter happy.  It does bring in `cxx-compiler` and the `hdf5` library as unnecessary dependencies on macOS, but that is a small price to pay for avoiding unnecessary builds or linter issues.  

We don't need to worry about matching `glibc`, `vs2022` etc. I think, because we aren't actually shipping any binaries here.  If a user tries to run the proprietary binary on a version of Linux or Windows that is too old, that isn't a problem this package needs to solve.

recipes/spade/meta.yaml
danielnachun: If the solution I suggested below is right, we don't need any `noarch` variants any more, and we can just use the suggestion I made since we'll only need to build on Linux.

recipes/spade/meta.yaml
kbrindley: >  If a user tries to run the proprietary binary on a version of Linux or Windows that is too old, that isn't a problem this package needs to solve.

We still end up using the conda environment's glibc/sysroot because we have to use the propietary software to compile and link the runtime executable from within the conda environment. Under the hood everything is still compiled, linked, and executed from the conda environment compilers, stdlib, glibc, sysroot, etc. It's a somewhat convoluted process imposed by the proprietary third-party software. Matching the minimum glibc requirements of the proprietary software in the conda environment is the best solution we've found so far.

So while this would be simpler, I would like to keep the glibc/sysroot minimum spec. From experience with the conda environments targeted by our user base, it's important to force the newer glibc/sysroot version. As a specific example, the current pytensor recipe overpins sysroot to ``2.17.*`` and we caught the conflict specifically when SPADE set the minimum glibc/sysroot. Prior to setting the minimum sysroot runtime requirement, the conda environment happily installed pytensor and SPADE side-by-side, but SPADE was broken. I'm working with the pytensor feedstock maintainers to relax their runtime compiler requirements here: https://github.com/conda-forge/pytensor-suite-feedstock/pull/194.

kbrindley: >  but that is a small price to pay for avoiding unnecessary builds or linter issues.

The conda-forge knowledge base mixes ``noarch: python`` with os selectors specifically for the this use case where a noarch package has os-specific runtime requirements. Is the documentation out-of-date and this practice is discouraged or is the linter simply missing this edge case?

kbrindley: > > but that is a small price to pay for avoiding unnecessary builds or linter issues.
> 
> The conda-forge knowledge base mixes `noarch: python` with os selectors specifically for the this use case where a noarch package has os-specific runtime requirements. Is the documentation out-of-date and this practice is discouraged or is the linter simply missing this edge case?
> 
> https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

Stumbled on this blog post: https://conda-forge.org/blog/2024/10/15/python-noarch-variants/

> We have been adding noarch: python variants for some feedstocks so that the compiled extension has higher priority and the pure Python extension has lower priority, which makes the conda solver use the noarch: python variant if no suitable compiled variant is available. One issue is that the linter might not like selectors on noarch recipes. We added an option
> ```
> linter:
>   skip:
>     - lint_noarch_selectors
> ```
> 
> to conda-forge.yml that will make the linter skip this warning/error.

Blog post is about selectors for packages that have a pure python and compiled variant, but it looks like it implements a relevant linter skip for noarch packages with os specific dependencies.

Would it be appropriate to satisfy the linter with this skip?

danielnachun: A couple of things:
- It looks like we are close to getting https://github.com/conda-forge/pytensor-suite-feedstock/pull/194 merged, and it would be good to get that done first.
- I guess based on your above comments we do need separate `noarch` platforms for Linux, macOS and Windows.
- I think you might be able to set `__linux >= 4.18` which would have the same effect as setting the `c_stdlib` value to 2.28, because the kernel version and `glibc` version are tied together.  Let's see if that works or not.

kbrindley: TL;DR I'd like to merge the current recipe to put the package in user's hands with the understanding that we can update compiler handling as conda-forge rolls it out. The ``cxx-compiler`` vs ``gxx`` changes were only completed yesterday, work is ongoing.

-----

> It looks like we are close to getting https://github.com/conda-forge/pytensor-suite-feedstock/pull/194 merged, and it would be good to get that done first.

We don't need to wait for an updated pytensor feedstock before users can use SPADE.

This package does not use or interact with pytensor directly. I mention pytensor because they are doing similar runtime compilation and the glibc spec conflicts with SPADE's ``>=2.28`` requirement. Some of SPADE's user commonly install both, which prompted the deeper dive into glibc/sysroot and the PR with relaxed compiler/glibc requirements.

Regarding the recent ``cxx-compiler``/``gxx_{{ target_platform}}`` to ``gxx`` changes discussed in the pytensor PR, SPADE is not quite ready to switch from ``cxx-compiler`` to ``gxx``. The build system assumes that it can read the ``CC`` and ``CXX`` environment variables set by ``cxx-compiler``, but left empty by ``gxx``

```
$ conda create --name gxx-test gxx -c conda-forge --yes
$ conda activate gxx-test
(gxx-test) $ echo $CC

(gxx-test) $ echo $CXX

(gxx-test) $ which cc
/home/kbrindley/miniconda3/envs/gxx-test/bin/cc
(gxx-test) $ which c++
/home/kbrindley/miniconda3/envs/gxx-test/bin/c++
```

It's probably not a difficult update, but it will probably require a new release of SPADE.

Since the ``gxx`` package behavior was only marked complete yesterday and compiler package behavior updates are ongoing, we also can't guarantee that end users will pick up the newest ``gxx`` release without overly restricting the compiler versions available to end users.

> I think you might be able to set __linux >= 4.18 which would have the same effect as setting the c_stdlib value to 2.28, because the kernel version and glibc version are tied together. Let's see if that works or not.

I'm struggling to find documentation that discusses virtual package versions and relationships. I can't find anything that ties ``__linux >=4.18`` to ``__glibc >=2.28``. I would prefer to keep the known, direct ``glibc >=2.28`` spec in the recipe as an explicit runtime requirement because it's easier to maintain a runtime requirement spec when the error message and the runtime spec use the same language. We can read glibc versions straight off the error message when the third-party library updates it's minimum requirement and go directly to the feedstock and put the new minimum version.

recipes/spade/meta.yaml
kbrindley: checked locally and the shell specific syntax is required. we're converging on using os selectors in runtime requirements in a different thread, so I think this discussion is complete.

recipes/spade/meta.yaml
danielnachun: Thanks for digging into this more.  I also did some checking and I was assuming a tighter connection between `__linux` and `sysroot` and I guess there really is not.  I actually think the kernel version doesn't matter for running the binary.

recipes/spade/meta.yaml
```diff
@@ -0,0 +1,67 @@
...
+    - python >={{ python_min }}
+    - scons >=4.6
+    - cxx-compiler  # [linux or win]
+    - hdf5  # [linux or win]
+    - __linux  # [linux]
+    - __glibc >=2.28,<3  # [linux]
+    - sysroot_linux-64 >=2.28,<3  # [linux]
+    - __win  # [win]
+    - getopt-win32  # [win]
+    - vs2022_win-64  # [win]
```
danielnachun:
```suggestion
    - sysroot_linux-64 >=2.28,<3  # [linux and x86_64]
    - __win  # [win]
    - getopt-win32  # [win]
    - vs2022_win-64  # [win and x86_64]
```
The only small detail to resolve here now is that these sysroot dependencies need to be gated by the CPU architecture as well.  Do you know if the proprietary binary supports ARM CPUs on Linux on Windows?  If not, then we can just make the change I suggested and we should be done.  The effect of this change is that it will allow users on non x86 architectures to still use the Python package.

If it does support ARM, all you have to do is add one or both of:
```
  - sysroot_linux-aarch64 >=2.28,<3  # [linux and aarch64]
  - vs2022_win-arm64  # [win and arm64]
```

danielnachun:
Also please go ahead and add the `conda-forge.yml` file as instructed by https://conda-forge.org/blog/2024/10/15/python-noarch-variants/.  I'm not sure if the linter will pass here, but hopefully it does in the feedstock!

recipes/spade/meta.yaml
kbrindley: the third party doesn't support ARM yet, no. I will make these changes.

kbrindley:
> Also please go ahead and add the `conda-forge.yml` file as instructed by https://conda-forge.org/blog/2024/10/15/python-noarch-variants/. I'm not sure if the linter will pass here, but hopefully it does in the feedstock!

@danielnachun

Following the example here: https://github.com/conda-forge/staged-recipes/pull/31581/changes#diff-69cb7952d9175b7d28e9f226ba684fd4657b319224b88ad607dd02a3b4858c4c to put a package specific ``staged-recipes/recipes/spade/conda-forge.yml`` instead of editing the ``staged-recipes/conda-forge.yml`` file. It looks like this resulted in the expected changes for the feedstock here: https://github.com/conda-forge/flamegraph-pl-feedstock/blob/main/conda-forge.yml#L2-L4

kbrindley:
@danielnachun @traversaro all passing. I think this is ready to merge now. Thanks for the reviews! I think the package recipe is in much better shape thanks to your reviews.

PR #30426, opened 2025-06-28
Recipe for diffpy.srfit package

sbillinge:
Checklist
- [x] Build number is 0.


sbillinge:
@cadenmyers13, @Tieqiong and @dragonyanglong please can you comment on this PR if you are willing to listed as a conda-forge maintainer for this package.  Thanks.

sbillinge:
@tieqiong @cadenmyers13 this is ready for review

sbillinge:
@tieqiong @cadenmyers13 I am still working on this please hold off review for the time being

sbillinge:
@Tieqiong, @cadenmyers13, this is ready for review now.  Passing everything except it needs comments from maintainers.

@dragonyanglong please can you also comment if you want to be listed as a a maintainer, so that the staged-recipes linter will pass?

cadenmyers13:
I agree to be listed as a maintainer

Tieqiong:
@sbillinge Thanks for adding me as maintainer. By the way is this doc url correct?

sbillinge:
> By the way is this doc url correct?

it should be but this is a decision we have to make.  Currently `diffpy.github.io` redirects to `diffpy.org`.  Whether we want this to be the case in the future we should think about it....

sbillinge:

recipes/diffpy.srfit/meta.yaml
```diff
@@ -0,0 +1,64 @@
+{% set name = "diffpy.srfit" %}
+{% set version = "3.1.0rc0" %}
```
ocefpaf: This one falls under https://github.com/conda-forge/cfep/blob/main/cfep-05.md

recipes/diffpy.srfit/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+  requires:
+    - python {{ python_min }} 
+    - pip
+    - pytest
+    - sympy
+  source_files:
+    - tests
+  commands:
+    - pip check
+    - pip list
```
ocefpaf:
```suggestion
```

recipes/diffpy.srfit/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - pytest
+
+about:
+  home: https://github.com/diffpy/diffpy.srfit/
+  license: Free To Use But Restricted
+  license_file: LICENSE.rst
```
ocefpaf: Custom license make it very hard for us to accept b/c we are not lawyers and we cannot evaluate all the possible interpretations and consequences. Can ask upstream to if they can use a well-known OSI approved license instead?

recipes/diffpy.srfit/meta.yaml
sbillinge: thanks for the link.  Very helpful.

recipes/diffpy.srfit/meta.yaml
sbillinge: this line has been removed in the latest push.

recipes/diffpy.srfit/meta.yaml
sbillinge: It is not completely clear to me how we do this labelling in the staged-recipe.   To make things easier I think we could wait for our full release to appear on PyPI and then change this to a non-rc release.   I guess merging to an `rc` branch as described in the article by Eric is possible from the feedstock but not hte staged recipe?  Anyway, this is easily fixed if we skip the rc release to conda-forge this time.

however, please see below for the licensing issue.....

recipes/diffpy.srfit/meta.yaml
sbillinge: I agree and can start the process to get this to a more standard license.  However, this involves multiple parties (at least two, possibly three) including Universities and Government labs who own copyright, so clearing this is likely to be a long and arduous process.  

The terms of the license are essentially the same as BSD-3 but with some extra protections for some of hte copyright holders.  We also have other packages on conda-forge with these same licenses.   Is this a policy that has changed recently?  Is there any way we can get our packages with these non-standard licenses onto CF?

Thanks so much @ocefpaf

recipes/diffpy.srfit/meta.yaml
ocefpaf: All you need is to create the `conda_build_config.yaml` in the recipe directory. However, if you can make a stable release upstream, that is better.

sbillinge: I think this is better also.  We will work on that and loop back when it is done.  

We wanted to do it this way as we wanted to conda install this into another package and wanted to test everything including the conda imports before it is public, but since noone is using it as it doesn't exist on CF yet, I think it is just easier to push up a full release.

sbillinge: sorry, that was ambiguous I see when I reread, what I mean is that we will make a stable release instead of the rc

sbillinge:
@conda-forge/help-python, @ocefpaf, this is ready for review!

We have addressed all the comments from the previous reviews.

The one last question is about the non-standard license.  We would like to request that this gets merged in this form.  We note that the license has all the conditions of the standard BSD-3 clause license, on which it is based, but there is simply an additional condition that the US government reserves a no-cost non exclusive right to use the software.   Actually, this is ensured by the BSD-3 clause so does not change the terms from there.  We also note that we have a number of packages already on conda-forge with this exact license: diffpy.utils, diffpy.pdfgui and diffpy.structure.

We will work in the future with Brookhaven National Lab and see if we can return the license to a standard BSD-3, but this could take some time and if we can't get this package (`diffpy.srfit`) and another package we are working on, `diffpy.srreal` that has the same license, it will be very difficult to release the whole materials structure package `diffpy.cmi`., which is top priority for us.

Please could you consider merging this (and diffpy.srreal when it is ready) with the current version of the open-source license?

Thanks very much for the consideration.

recipes/diffpy.srfit/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - sympy
+  source_files:
+    - tests
+  commands:
+    - pip check
+    - pytest
+
+about:
+  home: https://github.com/diffpy/diffpy.srfit/
+  license: Free To Use But Restricted
```
ocefpaf: Can you use `LicenseRef-diffpy` instead? It would be nice to standardized the others onez that are already in conda-forge as well to make then easily discoverable.

recipes/diffpy.srfit/meta.yaml
ocefpaf: I'm not a lawyer,  and sometimes recipes get merged without the proper license care. If you write "this is 100% compatible with bsd3" I would not mind merging it as-is. I will refer to your comment though if any questions arise. Is that ok?

PS: if you can get upstream to move to bsd3 it would be awesome!

recipes/diffpy.srfit/meta.yaml
sbillinge: I have changed this to `LicenseRef-diffpy` as requested and added a statement that the license file is 100% compatible with BSD-3 as you requested.

sbillinge:
btw, @ocefpaf I will take your advice and make these changes in each of our affected packages so they are all consistent.

PR #30430, opened 2025-06-29
Adding memory_graph package

bterwijn:
This PR adds the `memory_graph` package to conda-forge.

- PyPI: https://pypi.org/project/memory_graph/
- Source: https://github.com/bterwijn/memory_graph
- License: BSD-2-Clause
- Maintainer: @bterwijn

`memory_graph` is a teaching and debugging tool that visualizes Python variable bindings and object references using Graphviz. It depends on both the `graphviz` CLI and the `python-graphviz` API wrapper, both of which are already available on conda-forge.

The package is `noarch: python` and supports Python >=3.7.

Checklist
- [ ] Build number is 0.


recipes/memory_graph/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  noarch: python
+  number: 0
+  script: "{{ PYTHON }} -m pip install . --no-deps -vv"
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools
+    - graphviz
```
synapticarbors: You likely don't need `graphviz` and `python-graphviz` under `host`.

recipes/memory_graph/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - python-graphviz
+  run:
+    - python >={{ python_min }}
+    - graphviz
```
synapticarbors: You don't need to specify `graphviz` since it is already a dependency of `python-graphviz`

recipes/memory_graph/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "memory_graph" %}
+{% set version = "0.3.46" %}
+{% set python_min = "3.7" %}
```
synapticarbors: conda-forge only supports python >=3.9. Typically we would prefer you not setting `python_min` explicitly unless it is greater than our global pin (currently 3.9).

recipes/memory_graph/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - python >={{ python_min }}
+    - graphviz
+    - python-graphviz
+
+test:
+  imports:
+    - memory_graph
+    - graphviz
+  commands:
+    - dot -V
```
synapticarbors, line 31:
```suggestion
    - dot -V
    - pip check
```

recipes/memory_graph/meta.yaml
```diff
@@ -0,0 +1,53 @@
+{% set name = "memory_graph" %}
+{% set version = "0.3.46" %}
+{% set python_min = "3.9" %}
```
synapticarbors: I apologize if I was unclear in my previous comment. I just meant that you had to remove this line `{% set python_min = "3.9" %}`

bterwijn: No problem, but I'm a bit confused now, the linter tells me to use 'python_min', so I need to define that variable, or do you want me to write something like?:
```
requirements:
  host:
    - python =3.9
    - pip
    - setuptools
  run:
    - python >=3.9
    - python-graphviz
```

synapticarbors: `python_min` is defined in our global pinnings:
https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml#L823-L826

So if you don't define it, the recipe uses the cond-forge wide definition. So you should leave `python {{ python_min }}`, but you don't have to set it (unless your package requires a value greater than the global pin).

bterwijn: Thanks, I understand

PR #30452, opened 2025-07-02
Add multiformats and multiformats-config

lkluft:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/multiformats-config/meta.yaml
+++ b/recipes/multiformats-config/meta.yaml
@@ -23,16 +23,18 @@ requirements:
     - pip
   run:
     - python >={{ python_min }}
-    - multiformats
+    # Break circular dependency
+    # - multiformats
 
-test:
-  imports:
-    - multiformats_config
-  commands:
-    - pip check
-  requires:
-    - python {{ python_min }}
-    - pip
+# Temporarily disable test to break circular dependency
+# test:
+#   imports:
+#     - multiformats_config
+#   commands:
+#     - pip check
+#   requires:
+#     - python {{ python_min }}
+#     - pip
 
 about:
   home: https://github.com/hashberg-io/multiformats-config
--- a/recipes/multiformats-config/meta.yaml
+++ b/recipes/multiformats-config/meta.yaml
@@ -1,0 +1,46 @@
+{% set name = "multiformats-config" %}
+{% set version = "0.3.1" %}
+
+package:
+  name: {{ name|lower }}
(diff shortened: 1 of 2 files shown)


lkluft:
@conda-forge/help-python The pipelines are failing due to a circular dependency between the two packages that I am trying to add. Is there a mechanism in conda-forge that allows this to be circumvented cleanly?

d70-t:
Hey, I'd help maintaining.

danielnachun:
> @conda-forge/help-python The pipelines are failing due to a circular dependency between the two packages that I am trying to add. Is there a mechanism in conda-forge that allows this to be circumvented cleanly?

In this case, you should just skip the test for `mulitformats-config` until the feedstocks are made, then re-enable it once the `multiformats` package exists.  One tricky thing though with this kind of circular dependency is that it might fail if `pip check` always expects the version `multiformats` to be same as `multiformats-config` when testing `multiformats-config`.  If that's the case the version requirements for `multiformats-config` may need to be relaxed a bit.

lkluft:
> In this case, you should just skip the test for `mulitformats-config` until the feedstocks are made, then re-enable it once the `multiformats` package exists. One tricky thing though with this kind of circular dependency is that it might fail if `pip check` always expects the version `multiformats` to be same as `multiformats-config` when testing `multiformats-config`. If that's the case the version requirements for `multiformats-config` may need to be relaxed a bit.

Thanks for the hint! I think it's safe to temporarily remove the dependency and re-enable it immediately after the merge. In the original package sources, `multiformats-config` only depends on **any** `multiformats` version.

All the tests are passing now, except for the linter. However, I am not sure if we can bypass the failing check or if there is a proper way to skip the testing stage without the linter complaining. _(I tried `skip: True` and and an empty `test:` section, but the linter catches both)_

danielnachun (approved):
> Thanks for the hint! I think it's safe to temporarily remove the dependency and re-enable it immediately after the merge. In the original package sources, `multiformats-config` only depends on **any** `multiformats` version.
> 
> All the tests are passing now, except for the linter. However, I am not sure if we can bypass the failing check or if there is a proper way to skip the testing stage without the linter complaining. _(I tried `skip: True` and and an empty `test:` section, but the linter catches both)_

In that case I am going to merge with the linting error since the design of `multiformats-config` won't leave you stuck in an endless dependency loop after every update.

lkluft:
Thanks for the help!

PR #30472, opened 2025-07-05
Adding mxml v2.0.0

eunos-1128:
This pull request introduces changes to the `mxml` recipe, including updates to dependencies, build scripts, and testing procedures, as well as a minor code adjustment in the `doctype.cpp` file. The most important changes are grouped into recipe enhancements and code inclusion.

### Recipe Enhancements:

* **Addition of `recipe.yaml`:** A detailed recipe file was added to define the build, source, requirements, and testing for the `mxml` library. It includes platform-specific build scripts, dependency management, and metadata about the library.
* **Addition of `variants.yaml`:** A variant configuration was added to specify the `MACOSX_SDK_VERSION` for macOS builds targeting `x86_64` architecture.

### Code Inclusion:

* **Update to `doctype.cpp`:** The `<algorithm>` header was included to support additional functionality in the `mxml` library.

---

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mxml/recipe.yaml
+++ b/recipes/mxml/recipe.yaml
@@ -27,6 +27,7 @@ build:
         - cmake --install build
       else:
         - echo on
+        - sed -i 's/mxml STATIC/mxml SHARED/' CMakeLists.txt
         - cmake -S . -B build %CMAKE_ARGS%
         - if errorlevel 1 exit 1
         - cmake --build build --config Release --parallel %CPU_COUNT%
--- a/recipes/mxml/recipe.yaml
+++ b/recipes/mxml/recipe.yaml
@@ -20,6 +20,7 @@ build:
       then:
         - set -exo pipefail
         - |
+          # TODO: Remove it after mxml-feedstock is created
           if [[ "${OSX_ARCH}" == "x86_64" ]]; then
             export EXTRA_CMAKE_ARGS="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.15"
           fi
--- a/recipes/mxml/recipe.yaml
+++ b/recipes/mxml/recipe.yaml
@@ -24,15 +24,17 @@ build:
           if [[ "${OSX_ARCH}" == "x86_64" ]]; then
             export EXTRA_CMAKE_ARGS="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.15"
           fi
-        - cmake -S . -B build ${CMAKE_ARGS} -DBUILD_SHARED_LIBS=ON --BUILD_STATIC_LIBS=OFF ${EXTRA_CMAKE_ARGS}
+        - cmake -S . -B build ${CMAKE_ARGS} -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=OFF ${EXTRA_CMAKE_ARGS}
         - cmake --build build --config Release --parallel ${CPU_COUNT}
         - ctest -V --test-dir build
         - cmake --install build
       else:
         - echo on
-        - cmake -S . -B build %CMAKE_ARGS% -DBUILD_SHARED_LIBS=ON --BUILD_STATIC_LIBS=OFF
+        - cmake -S . -B build %CMAKE_ARGS% -DBUILD_SHARED_LIBS=ON -DBUILD_STATIC_LIBS=OFF
         - if errorlevel 1 exit 1
         - cmake --build build --config Release --parallel %CPU_COUNT%
+        - if errorlevel 1 exit 1
+        - ctest -V --test-dir build
         - if errorlevel 1 exit 1
         - cmake --install build
         - if errorlevel 1 exit 1
--- a/recipes/mxml/recipe.yaml
+++ b/recipes/mxml/recipe.yaml
@@ -24,13 +24,13 @@ build:
           if [[ "${OSX_ARCH}" == "x86_64" ]]; then
             export EXTRA_CMAKE_ARGS="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.15"
           fi
-        - cmake -S . -B build ${CMAKE_ARGS} -DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF ${EXTRA_CMAKE_ARGS}
+        - cmake -S . -B build ${CMAKE_ARGS} -DBUILD_SHARED_LIBS=ON --BUILD_STATIC_LIBS=OFF ${EXTRA_CMAKE_ARGS}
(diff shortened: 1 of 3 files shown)


eunos-1128:
@conda-forge/help-some-language, ready for review!

eunos-1128:

recipes/mxml/recipe.yaml
```diff
@@ -0,0 +1,84 @@
...
+  number: ${{ build_number }}
+  script:
+    - if: unix
+      then:
+        - set -exo pipefail
+        - |
+          # TODO: Remove it after mxml-feedstock is created
+          if [[ "${OSX_ARCH}" == "x86_64" ]]; then
+            export EXTRA_CMAKE_ARGS="-DCMAKE_OSX_DEPLOYMENT_TARGET=10.15"
+          fi
```
danielnachun:
```suggestion
```
Since you've already set `c_stdlib` to 10.15 in `variants.yaml`, you shouldn't need to set this as well.  If the setting `c_stdlib` isn't sufficient to get `MACOSX_DEPLOYMENT_TARGET` to the right version, there may be a more complicated issue.

recipes/mxml/recipe.yaml
```diff
@@ -0,0 +1,84 @@
...
+        - cmake -S . -B build ${CMAKE_ARGS} -DBUILD_SHARED_LIBS=ON -DBUILD_TESTING=OFF ${EXTRA_CMAKE_ARGS}
+        - cmake --build build --config Release --parallel ${CPU_COUNT}
+        - cmake --install build
```
danielnachun, line 27:
```suggestion
        - ctest -V --test-dir build
        - cmake --install build
```
We should try to `ctest` just to see if there are any unit tests.  You may need to remove `-DBUILD_TESTING=OFF` for the tests to run.  This will also apply to the Windows build too.

recipes/mxml/recipe.yaml
```diff
@@ -0,0 +1,84 @@
...
+    - howardhinnant_date
+  run_exports:
+    - ${{ pin_subpackage(name|lower, upper_bound="x") }}
+
+tests:
+  - script:
+      - if: unix
+        then:
+          - test -d ${PREFIX}/include/mxml
+          - test -f ${PREFIX}/lib/libmxml.a
```
danielnachun:
```suggestion
          - test -f ${PREFIX}/lib/libmxml${SHLIB_EXT}
```
You have enabled building shared libs but are testing for the existence of a static library.  You should test for the shared library and either disabled the static library with `--BUILD_STATIC_LIBS=OFF` or delete `libmxml.a` if it can't be disabled.

recipes/mxml/recipe.yaml
```diff
@@ -0,0 +1,84 @@
...
+        else:
+          - if not exist "%PREFIX%\Library\include\mxml\" exit /b 1
+          - if not exist "%PREFIX%\Library\lib\mxml.lib" exit /b 1
```
danielnachun:
```suggestion
          - if not exist "%LIBRARY_PREFIX%\include\mxml\" exit /b 1
          - if not exist "%LIBRARY_PREFIX%\lib\mxml.lib" exit /b 1
          - if not exist "%LIBRARY_PREFIX%\bin\mxml.dll" exit /b 1
```
You can use `%LIBRARY_PREFIX%` instead of `%PREFIX%\Library`.  Also we want to check for the `mxml.dll` shared library object in `%LIBRARY_PREFIX%\bin` as that is what is actually needed at run time - `mxml.lib` is only used at link time.

danielnachun (changes requested):

recipes/mxml/recipe.yaml
eunos-1128: [As reported](https://github.com/prefix-dev/rattler-build/issues/1774) in the rattler-build repository issue, it seems that the `c_stdlib_version` in variants.yaml is not reflected in the CI configuration for staged-recipe.
(In fact, removing it causes the build to fail.)

I intend to remove this part after this recipe becomes a feedstock.

recipes/mxml/recipe.yaml
eunos-1128: This library does not seem to support dynamic linking.

CMakeLists.txt explicitly says `add_library(mxml STATIC)` and no shared library is created.

Therefore, I reverted the setting.

recipes/mxml/recipe.yaml
eunos-1128: This library does not seem to support dynamic linking.

CMakeLists.txt explicitly says `add_library(mxml STATIC)` and no shared library is created.

Therefore, I reverted the setting.

recipes/mxml/recipe.yaml
danielnachun: Thank you for reporting this.  I suppose the reason this may have been missed is because it is usually possible to get things to work just by setting `MACOSX_SDK_VERSION`.   

Just to confirm since we always need to verify this anyway, did you already try setting `MACOSX_SDK_VERSION` to `10.15` and adding `-D_LIBCPP_DISABLE_AVAILABILITY` to `CXXFLAGS` as described in the knowledge base: https://conda-forge.org/docs/maintainer/knowledge_base/#newer-c-features-with-old-sdk?   In most cases that is enough to convince the compiler that the newer symbols are available from the conda-forge `libcxx`.  Where `c_stdlib` typically needs to be raised is if the software actually needs newer symbols from C standard library in `libSystem`.  Sorry to nitpick on this, but it's a confusing point that we try to get right to maximize compatibility.

eunos-1128: Thank you. This worked!

recipes/mxml/variants.yaml
```diff
@@ -0,0 +1,11 @@
+cxx_compiler_version:
+  - if: linux
+    then: 15
+
+c_stdlib_version:
+  - if: osx and x86_64
+    then: 10.15
+
```
danielnachun:
```suggestion
```
You can remove this now because it doesn't actually need a newer `c_stdlib` on macOS.

danielnachun:
I was looking at the CMakeLists.txt for this package and saw that it seemed quite trivial to used `sed` to build a shared library.  As I had hoped it appears to work, the only downside being that some of the tests don't build any more.  I think this is not a concern for now - getting a shared library is more important in my view, especially given our recent discussions about phantom dependencies.  Let me know if that sounds good to you and I'll merge.

eunos-1128:
> getting a shared library is more important in my view,

I agree totally. Thank you for your commitment!

Could you merge it?

danielnachun (approved):

danielnachun:
Thanks for your contribution!

PR #30502, opened 2025-07-10
Add recipe for `acpica-tools`

hameerabbasi:
Checklist
- [x] Build number is 0.


hameerabbasi:

recipes/acpia-tools/recipe.yaml
```diff
@@ -0,0 +1,46 @@
...
+source:
+  url: https://github.com/acpica/acpica/archive/refs/tags/R${{ version }}.tar.gz
+  sha256: 9991ec103b3660d17715780406ee7409f705cf87ac55e9a32374affe1a6f275a
+
+build:
+  number: 0
+  script: |
+    make PREFIX=$PREFIX OPT_CFLAGS="-DACPI_PACKED_POINTERS_NOT_SUPPORTED"
+    make install PREFIX=$PREFIX
+  skip: not unix
```
hameerabbasi, line 20: Windows support requires Cygwin and dependencies are also missing: https://github.com/acpica/acpica/blob/cbc243e46f577425838ef61a88d00e118a8928ef/generate/release/build.txt#L43-L56

recipes/acpia-tools/patches/buggy-acpicf.h.patch
hameerabbasi, line 1: See https://github.com/acpica/acpica/discussions/1010#discussioncomment-12794730

recipes/acpia-tools/recipe.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ stdlib("c") }}
+    - make
+    - bison
+    - flex
+    - m4
```
danielnachun:
```suggestion
  build:
    - if: unix
      then:
        - ${{ compiler('c') }}
        - ${{ stdlib("c") }}
        - bison
        - flex
        - m4
      else:
        - ${{ compiler('m2w64_c') }}
        - ${{ stdlib("m2w64_c") }}
        - m2-bison
        - m2-flex
        - m2-m4
    - make

```

recipes/acpia-tools/recipe.yaml
```diff
@@ -0,0 +1,48 @@
...
+  sha256: 9991ec103b3660d17715780406ee7409f705cf87ac55e9a32374affe1a6f275a
+  patches:
+    - patches/buggy-acpicf.h.patch
+
+build:
+  number: 0
+  script: |
+    make PREFIX=$PREFIX OPT_CFLAGS="-DACPI_PACKED_POINTERS_NOT_SUPPORTED"
+    make install PREFIX=$PREFIX
+  skip: not unix
```
danielnachun, line 20:
```suggestion
  script:
    if: unix
    then: |
      make PREFIX=$PREFIX OPT_CFLAGS="-DACPI_PACKED_POINTERS_NOT_SUPPORTED"
      make install PREFIX=$PREFIX
    else: |
      make PREFIX=%LIBRARY_PREFIX% OPT_CFLAGS="-DACPI_PACKED_POINTERS_NOT_SUPPORTED"
      make install PREFIX=%LIBRARY_PREFIX%      
```

danielnachun (changes requested):
I made suggested changes that to see if we can get Windows to build.  Some of the dependencies have different names on Windows but are actually available, and our MinGW stack can normally substitute for CygWin.

recipes/acpia-tools/recipe.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - if: unix
+      then:
+        - ${{ compiler('c') }}
+        - ${{ stdlib('c') }}
+        - bison
+        - flex
+        - m4
+      else:
+        - ${{ compiler('m2w64_c') }}
+        - ${{ stdlib('m2w64_c') }}
```
dhirschfeld: Results in:
```
gcc_win-64 14.* cannot be installed because there are no viable options:
```

I wonder if pinning to the newer 15.1.0 would help.

https://prefix.dev/channels/conda-forge/packages/gcc_win-64

hameerabbasi: Probably not, `gcc` isn't specified manually but as part of the MinGW compiler toolchain.

dhirschfeld: Yeah - the pinnings seem to want an older `gcc` which appears problematic because of it's dependemncy on an old `libgomp` which is incompatible with the `m2-flex` it wants to insall.

I'm wondering if forcing `gcc_win-64 ` to 15.1.0 woould pull in a newer `libgomp` which is compatible with the `m2-flex`.

dhirschfeld: Probably clutching at straws 😔

hameerabbasi: No dice, tried locally with a `pixi.toml`.

hameerabbasi: Looks like the solution is in https://github.com/conda-forge/m2-binary-packages-feedstock/issues/12#issuecomment-2664238921.

hameerabbasi:
@danielnachun Happy to accept any advice, you can see a few attempts to build Windows in the last 3 commits.

dhirschfeld:
```
%SRC_DIR%\source\include\platform\accygwin.h(171): fatal error C1083:
Cannot open include file: 'unistd.h': No such file or directory
 ```

**Comment by danielnachun** [comment]
I think this error means we should try to use the `m2w64` compilers on Windows instead.  While it's nice to be able to use MSVC, there are cases where it doesn't work and this may be one of them.  The `m2w64` sysroot that gets added automatically by `stdlib('m2w64_c')` should contain that missing header.  From what I can see, `winflexbison` shouldn't have a conflict with the `m2w64` compiler, but we can see once you change it.

**Comment by hameerabbasi** [comment]
> I think this error means we should try to use the `m2w64` compilers on Windows instead. While it's nice to be able to use MSVC, there are cases where it doesn't work and this may be one of them. The `m2w64` sysroot that gets added automatically by `stdlib('m2w64_c')` should contain that missing header. From what I can see, `winflexbison` shouldn't have a conflict with the `m2w64` compiler, but we can see once you change it.

@danielnachun I tried that initially in https://github.com/conda-forge/staged-recipes/pull/30502/commits/f760cf9696e0d9818b60a4b9b1c06f76193e022b, but ran into [this error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1291137&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de&l=865). I also tried replacing `make` with `m2-make` in https://github.com/conda-forge/staged-recipes/pull/30502/commits/e8789d7ba7efeba8e2615e5ebb59d843ef39046d, but got [yet another error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1291144&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de&l=868). It looks like certain things aren't added to `$PATH` on Windows, although I'm not sure why.

**Comment by hameerabbasi** [comment]
Okay, this is weird. No matter what I do I end up running up against either the compiler or `make` not being found in Windows. Something with `$PATH` I can't quite figure out.

**Comment by danielnachun** [comment]
@hameerabbasi @dhirschfeld I think our efforts to support Windows here so far were valiant, but having read the Makefile more carefully, it seems like quite a bit more effort will be needed to get it work.  The Linux Makefile invokes `bash` commands which means it's **not** portable across operating systems.  I can also see an MSVC folder which might be directly invokable with VS2022.  Regardless, I would say we should move forward with Windows support skipped and try to resolve this in the feedstock.

In the future I'm going to check Makefile projects more carefully before suggesting adding Windows support.  If the Makefile contains no shell commands, it's relatively simple.  If it does have shell commands, it can be as hard as a full autotools package and should probably be handled in the feedstock.

**Comment by hameerabbasi** [comment]
@danielnachun This is good to go I believe.

**Review by danielnachun — APPROVED** [review]

**Comment by synapticarbors** [comment]
It looks like something went wrong here. The feedstock is empty (https://github.com/conda-forge/acpica-tools-feedstock) and the recipe never got cleaned out of `recipes`. I believe that's causing some issues for the CI since the osx builds complain about mixing `meta.yaml` and `recipe.yaml`, and I think it's due to this.

@conda-forge/staged-recipes -- what is the course of action for cleaning this up?

File: recipes/acpia-tools/recipe.yaml [hunk 5]
```diff
@@ -0,0 +1,48 @@
...
+  url: https://github.com/acpica/acpica/archive/refs/tags/R${{ version }}.tar.gz
+  sha256: 9991ec103b3660d17715780406ee7409f705cf87ac55e9a32374affe1a6f275a
+  patches:
+    - patches/buggy-acpicf.h.patch
+
+build:
+  number: 0
+  script: |
+    make PREFIX=$PREFIX OPT_CFLAGS="-DACPI_PACKED_POINTERS_NOT_SUPPORTED"
+    make install PREFIX=$PREFIX
```
hameerabbasi, line 19: Should `$PREFIX` be `$CONDA_PREFIX`?

dhirschfeld: I think `$PREFIX` is correct.
https://docs.conda.io/projects/conda-build/en/stable/user-guide/environment-variables.html

PR #30542, opened 2025-07-15
Add SPZ

hobu:
SPZ from [Niantic](https://github.com/nianticlabs/spz/) does not currently provide a shared library, and this package has both `spz` and a `spz-static` outputs. I hope to upstream fixes to that in the future.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/spz/bld.bat
+++ b/recipes/spz/bld.bat
@@ -2,9 +2,10 @@
 @setlocal EnableDelayedExpansion
 
 cmake -S %SRC_DIR% -B build -G "Ninja" ^
-    -D CMAKE_BUILD_TYPE=Release% ^
-    -D CMAKE_VERBOSE_MAKEFILE=%CMAKE_VERBOSE_MAKEFILE% ^
-    -D BUILD_SHARED_LIBS=ON ^
+    -D CMAKE_BUILD_TYPE:STRING=Release ^
+    -D CMAKE_INSTALL_PREFIX:PATH=%LIBRARY_PREFIX% ^
+    -D CMAKE_VERBOSE_MAKEFILE:BOOL=%CMAKE_VERBOSE_MAKEFILE% ^
+    -D BUILD_SHARED_LIBS:BOOL=ON ^
     -D CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON ^
      %CMAKE_ARGS%
 if errorlevel 1 exit 1
@@ -14,5 +15,6 @@ if errorlevel 1 exit 1
 
 cmake --install build
 if errorlevel 1 exit 1
+
 
 
--- a/recipes/spz/meta.yaml
+++ b/recipes/spz/meta.yaml
@@ -29,9 +29,9 @@ test:
     - test -f ${PREFIX}/include/load-spz.h  # [unix]
     - test -f ${PREFIX}/lib/libspz${SHLIB_EXT}  # [unix]
 
-    - if not exist %LIBRARY_BIN%\\spz.dll exit 1  # [win]
-    - if not exist %LIBRARY_INC%\\load-spz.h exit 1  # [win]
-    - if not exist %LIBRARY_LIB%\\spz.lib exit 1  # [win]
+    - if not exist %LIBRARY_BIN%\spz.dll exit 1  # [win]
+    - if not exist %LIBRARY_INC%\load-spz.h exit 1  # [win]
+    - if not exist %LIBRARY_LIB%\spz.lib exit 1  # [win]
 
 about:
   home: https://github.com/nianticlabs/spz/


hobu:
@conda-forge/help-c-cpp this should be ready for review. 

I don't quite understand why the linter is complaining about no test for the `spz-static` output.

recipes/spz/meta.yaml
```diff
@@ -0,0 +1,58 @@
+{% set version = "v2.0.0" %}
+
+package:
+  name: spz
```
xhochy, line 4:
```suggestion
  name: spz-split
```

recipes/spz/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+  version: {{ version }}
+
+source:
+  url: https://github.com/nianticlabs/spz/archive/{{ version }}.tar.gz
+  sha256: fa14626b531c1dee2c29c29b43de8eaa8b773c3b0ef434ee8372c9712dfb6773
+
+build:
+  number: 0
+  run_exports:
+    - {{ pin_subpackage('spz', max_pin='x.x.x') }}
```
xhochy, line 16: This should be set on the output.

recipes/spz/meta.yaml
```diff
@@ -0,0 +1,58 @@
...
+outputs:
+  - name: spz
+    test:
+      commands:
+        - test -f ${PREFIX}/include/load-spz.h
+        - conda inspect linkages -p $PREFIX spz  # [not win]
+
+  - name: spz-static
+    script: build_static.sh  # [unix]
+    script: bld.bat  # [win]
```
xhochy: This would build a shared version, right?

hobu: There is no shared version available to build with their stock release. I hope to upstream patches to do so and fix the [include path pollution](https://github.com/nianticlabs/spz/issues/43) at some point as well.

danielnachun: I think we could easily patch `CmakeLists.txt` to build a share library.  I believe if you just do something like `sed -i 's/add_library(spz ${spz_sources})/add_library(spz SHARED ${spz_sources})/' CMakeLists.txt` that might be enough to convince it to build a shared library.  

@xhochy in general I think we might want to eventually have some documentation for how to "persuade" CMake and Meson to build shared libraries - I think in most cases a simple `sed` edit is enough.  Autotools and manually created Makefiles can vary a lot in difficulty but some pointers could also be given for those too.

hobu: While it's easy enough to tell CMake to build `SHARED` with a small patch on unix builds, it's a very different thing on Windows which requires going through and decorating the required classes and functions with `__declspec(dllexport)` to export them. Maintaining *that* set of patches in a packaging system is a giant hassle.

The SPZ project, format, and library is new but has a lot of momentum behind it. Having it in Conda Forge will make it more convenient for people to build with it (and for us to use it in `pdal`, where we can unvendor it).

recipes/spz/meta.yaml
hobu: I think the goal is to not have a split recipe that produces a static/shared outputs – we would just have a normal shared library. To get to that point requires upstreaming a bunch of PRs and getting them released. I'll keep that process moving forward. I'd rather not name it `-split` at this point.

recipes/spz/meta.yaml
xhochy: On Windows, you can use `cmake -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON` to export everything.

hobu: > On Windows, you can use `cmake -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON` to export everything.

Fair enough. Is this what you want done for this feedstock to be activated?

danielnachun: Yes what @xhochy suggested has worked in combination the patching I suggested.  I will try this today if I have time - we recently had a success with this for another package so I'd like to see if it can work here.

recipes/spz/meta.yaml
danielnachun: If we can patch this library not to produce a static library at all, I don't think we need multiple outputs anymore?

recipes/spz/meta.yaml
hobu: I updated the recipe to be a single `spz` recipe that emits a shared library for both windows and unix. `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON` is a 🔨 that could potentially have some awful consequences if the library authors aren't careful about their variable naming and such, but I we can live with it until we get proper exports upstreamed.

danielnachun: @hobu I will say from the many cases we've used that, we have yet to encounter any issues so far.  I think things could get messy if authors aren't careful about duplicate symbol names, but when that happens GCC and Clang nowadays tend to throw errors as well.

recipes/spz/bld.bat
```diff
@@ -0,0 +1,20 @@
...
+    -D BUILD_SHARED_LIBS=ON ^
+    -D CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON ^
+    %SRC_DIR%
+if errorlevel 1 exit 1
+
+ninja
+if errorlevel 1 exit 1
+
+ninja install
+if errorlevel 1 exit 1
```
danielnachun:
```suggestion
@echo on
@setlocal EnableDelayedExpansion

cmake -S %SRC_DIR% -B build -G "Ninja" ^
    -D CMAKE_BUILD_TYPE=Release% ^
    -D CMAKE_VERBOSE_MAKEFILE=%CMAKE_VERBOSE_MAKEFILE% ^
    -D BUILD_SHARED_LIBS=ON ^
    -D CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON ^
     %CMAKE_ARGS%
if errorlevel 1 exit 1

cmake --build build -j%CPU_COUNT%
if errorlevel 1 exit 1

cmake --install build
if errorlevel 1 exit 1
```

recipes/spz/build.sh
```diff
@@ -0,0 +1,12 @@
...
+
+mkdir -p build && cd build
+
+cmake ${CMAKE_ARGS} -G Ninja \
+      -DCMAKE_INSTALL_PREFIX=${PREFIX} \
+      -DBUILD_SHARED_LIBS=ON \
+      ${SRC_DIR}
+
+ninja
+ninja install
```
danielnachun:
```suggestion
#!/usr/bin/env bash

set -o xtrace -o nounset -o pipefail -o errexit

cmake -S ${SRC_DIR} -B build -G Ninja \
      -DBUILD_SHARED_LIBS=ON \
      ${CMAKE_ARGS}

cmake --build build -j${CPU_COUNT}
cmake --install build
```

recipes/spz/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+    - ninja
+  host:
+    - zlib
+
+test:
+  commands:
+    - test -f ${PREFIX}/include/load-spz.h  # [unix]
+    - test -f ${PREFIX}/lib/libspz${SHLIB_EXT}  # [unix]
+    - test ! -f ${PREFIX}/lib/libspz.a          # [unix]
+    - if not exist %LIBRARY_BIN%\\spz.dll exit 1  # [win]
```
danielnachun:
```suggestion
    - if not exist %LIBRARY_BIN%\spz.dll exit 1  # [win]
    - if not exist %LIBRARY_LIB%\spz.lib exit 1  # [win]
```
We don't need to test for the static library since we aren't building it.  Instead we also need to check for the `.lib` file on Windows.

danielnachun (changes requested):
I've made some suggestions to the build scripts to make them more consistent with other CMake recipes.  If an argument I suggested removing needs to be added back, let's make sure we have a comment in `bld.bat` or `build.sh` explaining why it's needed.  Normally `CMAKE_ARGS` is supposed to handle things like `CMAKE_INSTALL_PREFIX`, `LIBRARY_PATH`, etc. but I know there can be non-standard configurations where these need to be overwritten.

The patch is acceptable but if it becomes a pain to update you can probably achieve the same thing using `sed` as I suggested earlier.  This is more of a feedstock maintainer preference though and I think at the moment it's okay.

danielnachun:
Also the linting issue is just an extra blank line at the end of `meta.yaml` - should be a quick fix!

hobu:
I am willing to be a maintainer on this feedstock.

danielnachun (approved):

PR #30548, opened 2025-07-16
feat: Add fastframes from ATLAS Experiment

kratsg:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/fastframes/recipe.yaml
+++ b/recipes/fastframes/recipe.yaml
@@ -33,7 +33,6 @@ requirements:
     - ${{ stdlib("c") }}
     - cmake
     - make
-    - rsync
     - if: build_platform != target_platform
       then:
         - cross-python_${{ target_platform }}


kratsg:
@matthewfeickert could use some help here :)

recipes/fastframes/recipe.yaml
```diff
@@ -0,0 +1,90 @@
...
+    - cmake
+    - make
+    - rsync
+    - if: build_platform != target_platform
+      then:
+        - cross-python_${{ target_platform }}
+        - python
+        - root_base >=6.36
+  host:
+    - root_base >=6.36
```
matthewfeickert:
```suggestion
        - root_base
  host:
    - root_base
```

@chrisburr can comment on this (and I'd suggest asking on the conda-forge Zulip) but in general you don't want to set lower bounds on `build` and `host` requirements as if they don't resolve to the latest version already there is a global pinning that is keeping them from doing so and that should be addressed. If you place lower bounds that will override global pinnings and can cause breakage.

kratsg: We need lower bounds. ROOT 6.36 introduces new headers that FastFrames relies on for RDF work -- but these headers don't exist in 6.34. What's the suggestion here? (And ROOT 6.36 doesn't exist on conda-forge yet)

matthewfeickert: > What's the suggestion here?

Don't have lower bounds for `build` and `host` requirements unless the _absence_ of it is causing the solve during build to be unable to get the correct versions.

That shouldn't happen here as ROOT `6.36` should get installed as it is a newer higher release version build (or _will_ be once it is out).

Lower bounds on the `run` requirements are fine.

kratsg: > Don't have lower bounds for build and host requirements unless the absence of it is causing the solve during build to be unable to get the correct versions.

The build will fail because of missing headers. So how are you supposed to define the minimum version of ROOT for building against?

matthewfeickert: [#general > When are lower bounds on &#96;host&#96; requirements approrpiate? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/When.20are.20lower.20bounds.20on.20.60host.60.20requirements.20approrpiate.3F/near/530682353)

matthewfeickert: > So how are you supposed to define the minimum version of ROOT for building against?

Ah, I see your point RE: the fact that [`root_base` is in the global pinning as](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/a0d9655affb597ae3de9468bbffcfa9f8686728f/recipe/conda_build_config.yaml#L868-L870)

```yaml
root_base:
  - 6.32.10
  - 6.34.4
```

So I'm not sure, other than changing the global pinning when `root_base` `6.36` comes out(?). I think we should follow up with `conda-forge/core` on this on the Zulip chat above.

matthewfeickert: Got a recommendation: [#general > When are lower bounds on &#96;host&#96; requirements approrpiate? @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/When.20are.20lower.20bounds.20on.20.60host.60.20requirements.20approrpiate.3F/near/530684388)

> The set of versions in the global pinnings depends on what the end state is after the migration. I'm not sure why there are currently two versions; they're both severely out of date. But let's assume that we update both 6.34.x as well as 6.36.x, then the end result after the migration in the pinning might be
> 
> ```yaml
> root_base:
>   - 6.34.8
>   - 6.36.2
> ```
> 
> For a recipe requiring root 6.36, you should simply add something like
> ```yaml+jinja
> {% if root_base.split(".")[1]|int < 36%}
> build:
>   skip: true
> {% endif %}
> ```
> That ensures the recipe does not attempt to build variants that can never succeed

recipes/fastframes/recipe.yaml
```diff
@@ -0,0 +1,91 @@
+schema_version: 1
+
+context:
+  name: fastframes
+  version: "5.5.0"
+
+package:
+  name: ${{ name|lower }}
```
matthewfeickert:
```suggestion
context:
  version: "5.5.0"

package:
  name: fastframes
```

c.f. https://github.com/conda-forge/conda-forge.github.io/issues/2712

recipes/fastframes/recipe.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+build:
+  number: 0
+  skip:
+    - win
+    - match(root_base, "<6.36")
+  script:
+    - cmake ${CMAKE_ARGS} -DPython3_EXECUTABLE=$PYTHON -DPython3_INCLUDE_DIR=$PREFIX/include/python${PY_VER} -DFASTFRAMES_PYTHON_INSTALL_DIR=$SP_DIR -S . -B build
+    - cmake --build build --clean-first --parallel="${CPU_COUNT}" --verbose
+    - cmake --install build
```
chrisburr, line 27: At the moment this fils site-packages with various generic named modules. I wonder if it would be better to patch this to go under `site-packages/fastframes/...`

recipes/fastframes/recipe.yaml
```diff
@@ -0,0 +1,79 @@
...
+tests:
+  - package_contents:
+      files:
+        - bin/fast-frames.exe
+  - requirements:
+      run:
+        - root_base
+        - python
+    script:
+      - FastFrames.py -h
```
chrisburr:
```suggestion
  - script:
      - FastFrames.py -h
```

recipes/fastframes/recipe.yaml
kratsg: I was thinking that as well. but it requires a lot of rewriting and making it so that we use like scikit-build-core.

recipes/fastframes/recipe.yaml
kratsg: why drop these reqs?

matthewfeickert: I think for this reason: https://github.com/conda-forge/rivet-feedstock/pull/10#discussion_r2884904179

i.e. testing that the `run` requirements by themselves are sufficient to use the packaged binaries.

recipes/fastframes/recipe.yaml
kratsg: This will be followed up with FastFrames developers in https://gitlab.cern.ch/atlas-amglab/fastframes/-/issues/184

kratsg:

recipes/fastframes/recipe.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - cmake --install build
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ compiler("cxx") }}
+    - ${{ stdlib("c") }}
+    - cmake
+    - make
+    - rsync
```
chrisburr:
```suggestion
```

PR #30567, opened 2025-07-18
Add pvxs

lucmaga:


recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - libevent
+  run:
+    - libevent
+
+test:
+  commands:
+    - pvxget -V
+
+about:
+  home: https://github.com/mdavidsaver/pvxs
```
anderslindho: Home nowadays is https://github.com/epics-base/pvxs/

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  license: BSD-4-Clause-UC
+  license_file: LICENSE
+  summary: Library and a set of CLI utilities acting as PVAccess protocol client and/or server
+
```
anderslindho, line 52: You could add `doc_url` or `doc_src_url` (or both) as https://epics-base.github.io/pvxs/ 🔧

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,39 @@
...
+  run:
+    - libevent
+
+test:
+  commands:
+    - pvxget -V
+
+about:
+  home: https://github.com/mdavidsaver/pvxs
+  license: BSD-4-Clause-UC
```
anderslindho, line 47: Could also set `license_family: BSD` 🔧

recipes/pvxs/build.sh
```diff
@@ -0,0 +1,13 @@
...
+
+cat <<EOF > configure/RELEASE.local
+EPICS_BASE=${EPICS_BASE}
+EOF
+
+make install
+
+install -d ${PREFIX}/include/pvxs ${PREFIX}/bin/ ${PREFIX}/lib/ ${PREFIX}/epics/modules/pvxs/dbd
+install bin/${EPICS_HOST_ARCH}/* ${PREFIX}/bin/
+install lib/${EPICS_HOST_ARCH}/libpvxs.* ${PREFIX}/lib/
```
aharrisson: You probably need to install  libpvxsIoc libary as well to fix the overlinking issue in the build. So,either install the files explicitly or move the wildcard slightly to capture "libpvxs*" instead of "libpvx.*".

I am not sure why we need to run explicit "install" commands on these files. I would expect "make install" to install into $PREFIX. :thinking:

recipes/pvxs/meta.yaml
lucmaga: I'm adding `doc_url`, both pointing to the same link would be too repetitive.

recipes/pvxs/build.sh
lucmaga: `make install` would install pvxs, the epics way, inside `O.{ARCH}` directories. So we need to copy to standard directories if we need them there.

aharrisson:
LGTM

lucmaga:
@conda-forge-admin please ping @conda-forge/help-c-cpp

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+  build:
+    - make
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}
+  host:
+    - epics-base
+    - libevent
+  run:
+    - libevent
```
xhochy: `libevent` has a `run_exports` so it will be added here automatically with the right constraints.
```suggestion
```

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+test:
+  commands:
```
xhochy, line 30: Please check for the existence of the shared library.

xhochy (changes requested):
As this packages a shared library, it should also have a matching `run_exports` statement.

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,43 @@
...
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ stdlib("c") }}
+  host:
+    - epics-base
+    - libevent
+
+test:
+  commands:
+    - pvxget -V
```
simon-ess: Comparing to epics-base-feedstock, maybe test each of the generated executables? https://github.com/conda-forge/epics-base-feedstock/blob/main/recipe/meta.yaml#L39

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+test:
+  commands:
+    - pvxcall -V
+    - pvxget -V
+    - pvxinfo -V
+    - pvxlist -V
+    - pvxmonitor -V
+    - pvxmshim -V
+    - pvxput -V
+    - pvxvct -V
```
simon-ess, line 38: This is missing softIocPVX

simon-ess (approved):

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+package:
+  name: pvxs
+  version: "{{ version }}"
+
+source:
+  url: https://github.com/epics-base/pvxs/archive/refs/tags/{{ version }}.tar.gz
+  sha256: 944d16687fbdd31f43c617824048891756ff9edd93d209df260b17d2adc0b446
+
+build:
+  skip: True  # [win or osx]
```
simon-ess, line 14: Do we actually need to skip these? It's worth noting that when switching to epics-base/pvxs for p4p on the p4p feedstock (https://github.com/conda-forge/p4p-feedstock/pull/46), this means we go from p4p is supported for several OSs to only linux.

lucmaga: @simon-ess I don't have resources to test on the other OS, only Linux.
We could add support to them later.

lucmaga:

recipes/pvxs/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  number: 0
+  run_exports:
+    - libevent
```
xhochy: This is not necessary
```suggestion
```

xhochy (changes requested):

recipes/pvxs/build.sh
```diff
@@ -0,0 +1,13 @@
...
+EPICS_BASE=${EPICS_BASE}
+EOF
+
+make install
+
+install -d ${PREFIX}/include/pvxs ${PREFIX}/bin/ ${PREFIX}/lib/ ${PREFIX}/epics/modules/pvxs/dbd
+install bin/${EPICS_HOST_ARCH}/* ${PREFIX}/bin/
+install lib/${EPICS_HOST_ARCH}/*.so* ${PREFIX}/lib/
+install -m 0664 include/pvxs/* ${PREFIX}/include/pvxs/
+install -m 0664 dbd/* ${PREFIX}/epics/modules/pvxs/dbd
```
simon-ess: Looking at this, there is a bit of a complication here: softIocPVX by default looks for softIocPVX.dbd in the relative path `../../dbd/`. So if we want softIocPVX to run correctly and find the dbd file as located here we need to patch softIocPVX...

simon-ess: (It is of course possible to run softIocPVX with the `-D`option to specify a different location. But it won't run without that)

recipes/pvxs/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+cat <<EOF > configure/RELEASE.local
+EPICS_BASE=${EPICS_BASE}
+EOF
+
+make install
+
+install -d ${PREFIX}/include/pvxs ${PREFIX}/bin/ ${PREFIX}/lib/ ${PREFIX}/epics/modules/pvxs/dbd
```
simon-ess: Actually, one other point here. Why are we installing the .dbd files into epics/modules/pvxs/dbd and not just pvxs/dbd?

recipes/pvxs/build.sh
lucmaga: I've added a commit for fixing this, which I think its a quite simple solution. Please let me know what you think.

recipes/pvxs/build.sh
lucmaga: Agreed.

recipes/pvxs/build.sh
lucmaga: Here is what it looks like running:
```
(test_pvxs) [lamm@archlinux pvxs-test]$ export | grep PVXS
declare -x PVXS_SOFTIOCPVX_DBD="/home/lamm/.local/share/mamba/envs/test_pvxs/pvxs/dbd/softIocPVX.dbd"
(test_pvxs) [lamm@archlinux pvxs-test]$ softIocPVX st.cmd
INFO: PVXS QSRV2 is loaded, permitted, and ENABLED.
dbLoadRecords("test.db")
epics> dbl
TEST:AI:1
TEST:AO:1
TEST:BI:1
TEST:BO:1
TEST:CALC:1
TEST:FANOUT:1
TEST:LONGIN:1
TEST:LONGOUT:1
TEST:SEQ:1
TEST:STR:1
TEST:STR:2
TEST:WAVEFORM:1
epics> exit
```

simon-ess: Looks fine, although I might suggest that the env var should be a path and we should try overwrite both `dbd_file` and `exit_file` based on that path, since otherwise the `-x` option breaks as well.

recipes/pvxs/build.sh
```diff
@@ -0,0 +1,24 @@
...
+install -d ${PREFIX}/include/pvxs ${PREFIX}/bin/ ${PREFIX}/lib/ ${PREFIX}/pvxs/dbd ${PREFIX}/pvxs/db
+install bin/${EPICS_HOST_ARCH}/* ${PREFIX}/bin/
+install lib/${EPICS_HOST_ARCH}/*.so* ${PREFIX}/lib/
+install -m 0664 include/pvxs/* ${PREFIX}/include/pvxs/
+install -m 0664 dbd/* ${PREFIX}/pvxs/dbd
+install -m 0664 db/* ${PREFIX}/pvxs/db
+
+mkdir -p $PREFIX/etc/conda/activate.d
+cat <<EOF > $PREFIX/etc/conda/activate.d/pvxs_activate.sh
+export PVXS_PATH="${PREFIX}/pvxs/"
```
simon-ess: Final comment: To be more EPICS-standard-like, this variable should _probably_ be named simply `PVXS`. But that's a pretty minor comment, and could be ignored.

simon-ess (approved):

lucmaga:

PR #30651, opened 2025-07-27
Add recipe for gprofiler

awvwgk:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/gprofiler/build_base.sh
+++ b/recipes/gprofiler/build_base.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -ex
+
+$PYTHON -m pip install . -vv --no-deps --no-build-isolation
--- a/recipes/gprofiler/meta.yaml
+++ b/recipes/gprofiler/meta.yaml
@@ -1,0 +1,57 @@
+{% set name = "gprofiler" %}
+{% set version = "1.0.0" %}
+
+package:
+  name: {{ name|lower }}-pkg
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/g/gprofiler-official/gprofiler-official-{{ version }}.tar.gz
+  sha256: 5015b47f10fbdcb59c57e342e815c9c07afbe57cd3984154f75b845ddef2445d
+
+build:
+  number: 0
+
+outputs:
+  - name: {{ name }}
+    script: build_base.sh
+    build:
+      noarch: python
+    requirements:
+      host:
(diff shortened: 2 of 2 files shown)


awvwgk:
@conda-forge/help-python This recipe is ready for review

recipes/gprofiler/meta.yaml
```diff
@@ -0,0 +1,57 @@
+{% set name = "gprofiler" %}
```
danielnachun, line 1:
```suggestion
```
Please remove this jinja variable and replace all instances of `{{ name }}` with the actual package name.  This name will never change so there is no need for it to be a variable, unlike `python_min` and `version`, which will change over time.

recipes/gprofiler/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+source:
+  url: https://pypi.org/packages/source/g/gprofiler-official/gprofiler-official-{{ version }}.tar.gz
+  sha256: 5015b47f10fbdcb59c57e342e815c9c07afbe57cd3984154f75b845ddef2445d
+
+build:
+  number: 0
+
+outputs:
+  - name: {{ name }}
+    script: build_base.sh
```
danielnachun, line 17:
```suggestion
    script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
You don't need a separate build script in a recipe that uses such a simple build process.

recipes/gprofiler/meta.yaml
```diff
@@ -0,0 +1,57 @@
...
+        - pip
+        - python {{ python_min }}
+
+  # Name used for distribution on PyPI
+  - name: {{ name }}-official
+    build:
+      noarch: generic
+    requirements:
+      run:
+        - {{ pin_subpackage(name, max_pin="x.x.x") }}
```
danielnachun, line 43: Can you explain the reason for this pinning?  Most of the this is needed to preserver ABI compatibility for shared native libraries, which this package is not.

recipes/gprofiler/build_base.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -ex
+
+$PYTHON -m pip install . -vv --no-deps --no-build-isolation
```
danielnachun, line 5:
```suggestion
```
See my other comment for why this probably isn't needed.

danielnachun (changes requested):

PR #30667, opened 2025-07-28
Adding Panic, and alarm and notification toolkit for Tango Control Systems

sergirubio:
The Tango Control System is a collaboration between several scientific institutes for an open source control system, most of them using conda-forge. Panic is used by many of these institutes to configure alarms and notifications for their Tango Control Systems.

@conda-forge/help-python, @benjee, @ovallcorba


ovallcorba:
Hi @sergirubio 
The guidata problem is because we allow in conda newer versions for numpy2 compatibility, see https://github.com/conda-forge/guiqwt-feedstock/commit/981d0c5320cf206f6f42e40a44062b34239388e4
Since in setup.py the version is pinned to 3.1, the pip check fails. You can remove it from tests.
Regards!

ocefpaf:
pip-check is failing with:

```
guiqwt 4.4.5 has requirement guidata==3.1, but you have guidata 3.9.0.
```

This should probably be fixed in `guiqwt`.

ocefpaf:
See https://github.com/conda-forge/guiqwt-feedstock/pull/45


Pinging @beenje here for a fix upstream.

beenje:
> See [conda-forge/guiqwt-feedstock#45](https://github.com/conda-forge/guiqwt-feedstock/pull/45)
> 
> Pinging @beenje here for a fix upstream.

Sorry about that. I made a PR to fix guiqwt: https://github.com/conda-forge/guiqwt-feedstock/pull/46

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools
+  run:
+    - python >={{ python_min }}
+    - fandango
+    - pytango
+    - taurus
```
beenje: I wonder if it wouldn't be worth splitting this package in 2 outputs:
- `pyalarm` that depends on `fandango`, `pytango` and `taurus-core`
- `panic` that depends on `pyalarm` and `taurus-qt`.

`PyAlarm` doesn't require Qt to run. It would be nice to be able to install it without.

You can look at https://github.com/conda-forge/matplotlib-feedstock/blob/main/recipe/meta.yaml to see how to split a recipe.

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  imports:
+    - panic
+  commands:
+    - pip check
+    - PyAlarm --help
+  requires:
+    - python {{ python_min }}  
+    - pip
+    - fandango
+    - pytango
```
beenje: You don't need fandango and pytango here as they are in the run requirements

recipes/panic/meta.yaml
ocefpaf: Agreed. Qt is a huge package and having the possibility of avoiding it makes users life much easier.

sergirubio: thanks, I'll have a look to the split recipe

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/panic-{{ version }}.tar.gz
+  sha256: f2772142ac60f69a8927a98317213804abb5becff565a852438288f97353827b
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  # hint from cibuildwheel
+  script_env:
+    - MACOSX_DEPLOYMENT_TARGET=10.9
```
beenje: You shouldn't need that

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+requirements:
+  build:
+    - {{ compiler("c") }}
+    - {{ stdlib("c") }}
```
beenje: This package is pure python - you don't need any compiler or stdlib("c") to build it
Probably why you have this error about c_stdlib_version and MACOSX_DEPLOYMENT_TARGET

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+  name: panic-suite
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/panic-{{ version }}.tar.gz
+  sha256: f2772142ac60f69a8927a98317213804abb5becff565a852438288f97353827b
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
beenje: You don't want to create a package for panic-suite so the script shouldn't be here.

If you check the logs, panic and panic-gui are empty.

I think that the way multiple outputs work in conda-build is a bit confusing.
I'd personally recommend to use the new v1 format (with rattler-build). See https://github.com/conda-forge/taurus-feedstock/blob/main/recipe/recipe.yaml

Anyway, if you want to keep the meta.yaml, you need to move this script to the panic output.
You don't need any requirement and test at the global level.

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+test:
+  imports:
+    - panic    
+
+outputs:
+  - name: panic
+    build:
+      entry_points:
+        - PyAlarm = panic.ds.PyAlarm:main
```
beenje: You need the script to build the package here.

You also need all entry points here. You can't split the python package itself in different outputs.
For python multiple outputs, you always have one package with the full pyton code. Other outputs only have extra run requirements.

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+          - PyAlarm --help
+        requires:
+          - python {{ python_min }}  
+          - pip
+  # optional panic user interface, built on Taurus (PyQt5)
+  - name: panic-gui
+    build:
+      noarch: generic
+      entry_points:
+        - panic = panic.gui.gui:main
```
beenje: this entry point should be in the panic package - it can't be split. This output is only to have extra run requirements.

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,104 @@
...
+    requirements:
+      host:
+        - python {{ python_min }}
+        - pip
+        - setuptools    
+      run:
+        - python >={{ python_min }}
+        - panic
```
beenje: You want the exact same version: `{{ pin_subpackage('panic', exact=True) }}`

beenje (changes requested):

recipes/panic/meta.yaml
sergirubio: Hi @benjee, I added those requirements and test due to conda-forge linter comments

recipes/panic/meta.yaml
sergirubio: Hi @benjee, I added those requirements and test due to conda-forge linter comments

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+  name: panic
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/panic-{{ version }}.tar.gz
+  sha256: 70a035f7618680a6bbf64dcaa56c3d25de510137eccd9fc89b1763dc65148cad
+  
+build:
+  number: 0
+  skip: true
```
beenje: This makes skipping the whole recipe - nothing is built

sergirubio: I am trying to find out what is still causing that c_stdlib error

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools
+  run:
+    - python >={{ python_min }}
+    - fandango
+    - pytango
+
+test:
```
beenje: If you look at the logs, no tests are run. I think it's because of this entry that you should remove.

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+  imports:
+    - panic
+  requires:
+    - python {{ python_min }}    
+
+outputs:
+  - name: panic-server
```
beenje: I think `panic` was better for this output? I let you decide.
But it shouldn't match the top level package name (which won't be used in this recipe).

recipes/panic/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/panic-{{ version }}.tar.gz
+  sha256: 70a035f7618680a6bbf64dcaa56c3d25de510137eccd9fc89b1763dc65148cad
+  
+build:
+  number: 0
+  noarch: python
+
+requirements:
```
beenje: We are not building any top level package, so I'd remove those requirements, maybe except (according to matplotlib recipe):

```
requirements:
  host:
    - python
```

And let the linter complain.

Or switch to recipe.yaml, because the format is much more explicit for multiple outputs :-)

beenje (commented):
Better but still some issues as the tests aren't run...

beenje:
Recipe looks good!

@ovallcorba can you please just confirm you want to be a maintainer?

ovallcorba:
Hi! Yes, I think this is ready to be merged.

Thanks, Benjamin, for your help with the recipe :)

It’s fine to be a maintainer (I added myself to the `recipe-maintainers` section in the `recipe.yaml`). Should I do anything else?

beenje:
> It’s fine to be a maintainer (I added myself to the `recipe-maintainers` section in the `recipe.yaml`). Should I do anything else?

Nope. Just had to comment here you wanted to be a maintainer. Merging!

PR #30679, opened 2025-07-30
Add cityhash

daneos:
Checklist
- [ ] Build number is 0.

### DIFF
--- a/recipes/cityhash/meta.yaml
+++ b/recipes/cityhash/meta.yaml
@@ -1,0 +1,41 @@
+{% set name = "cityhash" %}
+{% set version = "1.1.1" %}
+{% set git_rev = "f5dc54147fcce12cefd16548c8e760d68ac04226" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://github.com/google/cityhash/archive/{{ git_rev }}.zip
+  sha256: 20ab6da9929826c7c81ea3b7348190538a23f823a8b749c2da9715ecb7a6b545
+
+build:
+  number: 0
+  run_exports: {{ pin_subpackage('cityhash', max_pin='x.x.x') }}
+
+requirements:
+  build:
+    - {{ compiler('cxx') }}
+    - {{ stdlib('c') }}
+    - make
+
+test:
+  commands:
+    - test -f $PREFIX/lib/libcityhash.a         # [linux]
+    - test -f $PREFIX/lib/libcityhash.so.0.0.0  # [linux]
+    - test -L $PREFIX/lib/libcityhash.so.0      # [linux]
+    - test -L $PREFIX/lib/libcityhash.so        # [linux]
+
+about:
+  home: https://github.com/google/cityhash
+  summary: 'CityHash, a family of hash functions for strings'
+  description: |
+    CityHash provides hash functions for strings. The functions mix the
+    input bits thoroughly but are not suitable for cryptography.
+  license: MIT
+  license_file: COPYING
+
+extra:
+  recipe-maintainers:
+    - daneos
--- a/recipes/cityhash/meta.yaml
+++ b/recipes/cityhash/meta.yaml
@@ -12,7 +12,8 @@ source:
 
 build:
   number: 0
-  run_exports: {{ pin_subpackage('cityhash', max_pin='x.x.x') }}
+  run_exports:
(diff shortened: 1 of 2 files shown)


recipes/cityhash/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+outputs:
+  - name: cityhash
+    files:
+      include:
+        - 'include/city.h'
+        - 'lib/*.so*'                               # [linux]
+        - 'lib/*.dylib'                             # [osx]
+    test:
+      commands:
```
anderslindho, line 33: You can use `files` instead and not have to write out as much - same down for the other output

daneos: Are you sure about this?
The documentation for `files` says only that this specifies files to be copied from build env to test env.
https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#test-files
Using `test -f` or `test -L` seems more explicit and in line with the examples
https://conda-forge.org/docs/maintainer/adding_pkgs/#simple-existence-tests

anderslindho: That's odd. It seemed acceptable for Lucas' PR adding pvxs: https://github.com/conda-forge/staged-recipes/pull/30567

Note especially the comment by the core maintainer that was resolved

anderslindho: FYI @lucmaga

daneos: I suppose `files` is fine because it will still fail if the file doesn't exisit. With `test` however I can check for symlinks as well.

daneos:
@conda-forge-admin please ping @conda-forge/help-c-cpp

I have two questions that I'm not sure how to handle here:
- There are no tagged releases, but there are version numbers in NEWS file. I've taken the version number from there, but since this version has been added, some changes were made. Should I change the commit I'm downloading to the one where the version has been added to NEWS? What about version number for the current master? (There is a feature request for tagged releases, but since the project doesn't seem to be actively developed I wouldn't count on it being done https://github.com/google/cityhash/issues/39)
- There are tests that can be run with `make check` in the build env, but conda tests run in separate environment without the build tree. Should I run the test in the build script then?

daneos:
@conda-forge-admin please ping @conda-forge/staged-recipes

recipes/cityhash/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+        - test -f $PREFIX/include/city.h
+        - test -f $PREFIX/lib/libcityhash.so.0.0.0  # [linux]
+        - test -L $PREFIX/lib/libcityhash.so.0      # [linux]
+        - test -L $PREFIX/lib/libcityhash.so        # [linux]
+        - test -f $PREFIX/lib/libcityhash.0.dylib   # [osx]
+        - test -L $PREFIX/lib/libcityhash.dylib     # [osx]
```
danielnachun:
```suggestion
        - test -f $PREFIX/lib/libcityhash${SHLIB_EXT}
```
This is all we need to test for the shared libraries.

danielnachun:
> @conda-forge-admin please ping @conda-forge/help-c-cpp @conda-forge/help-c-cpp, ready for review!
> 
> I have two questions that I'm not sure how to handle here:
> 
> * There are no tagged releases, but there are version numbers in NEWS file. I've taken the version number from there, but since this version has been added, some changes were made. Should I change the commit I'm downloading to the one where the version has been added to NEWS? What about version number for the current master? (There is a feature request for tagged releases, but since the project doesn't seem to be actively developed I wouldn't count on it being done [[Feature request] please make a release google/cityhash#39](https://github.com/google/cityhash/issues/39))

Since this project seems to be inactive, I suppose we can allow an exception here so that this library is available for "archival" purposes.  The only caveat here is that if accumulates too much bit rot to build at some point in the future, it will have to be archived.

> * There are tests that can be run with `make check` in the build env, but conda tests run in separate environment without the build tree. Should I run the test in the build script then?

Yes those tests should always be run in the build script.  For most packages that ship shared libraries we just test for the existence of the shared library and header files, as you have already done here.

daneos:
Thanks, I've now added running tests to the build script

recipes/cityhash/meta.yaml
daneos: Done :white_check_mark:

danielnachun (approved):

PR #30710, opened 2025-08-03
Add PLAnT (`plant`) to conda-forge

gshiroma:
Checklist
- [x] Build number is 0.


recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
+{% set name = "plant" %}
+{% set version = "0.5.5" %}
+{% set python_min = "3.9" %}
```
synapticarbors: You should remove this since `python_min` only needs to be specified if it is greater than the current global conda-forge pin, which is 3.9.

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  path: https://gitlab.com/plant/plant/-/archive/v0.5.5/plant-v{{ version }}.tar.gz
```
synapticarbors:
```suggestion
  url: https://gitlab.com/plant/plant/-/archive/v{{ version }}/plant-v{{ version }}.tar.gz
```

recipes/plant/LICENSE.txt
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env python3
```
synapticarbors: This file is already in the source distribution, so you don't need to add this local copy here.

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+
+build:
+  noarch: python
+  entry_points:
+    - plant_info = plant_info:main
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
```
synapticarbors: I believe you can delete this `build` section.

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - gdal
+    - matplotlib-base
+    - scipy
+    - pillow
+    - h5py
+    - python-dateutil
+    - numexpr
+
+  host:
+    - python >=3.9
```
synapticarbors:
```suggestion
    - python {{ python_min }}
```

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - setuptools >=64.0
+    - setuptools-scm >=6.2
+    - pip >=21.3
+
+  run:
+    - python >=3.9
```
synapticarbors:
```suggestion
    - python >={{ python_min }}
```

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,72 @@
...
+    - numexpr
+
+test:
+  imports:
+    - plant
+  commands:
+    - pip check
+    - plant_info --help
+  requires:
+    - pip
```
synapticarbors, line 60:
```suggestion
    - pip
    - python {{ python_min }}
```

gshiroma:
Thank you so much, @synapticarbors , for taking the time to review this PR. This is my first contribution to conda-forge, and I'm still learning how things work here.

recipes/plant/meta.yaml
```diff
@@ -0,0 +1,60 @@
+{% set name = "plant" %}
+{% set version = "0.5.5" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  path: https://gitlab.com/plant/plant/-/archive/v{{ version }}/plant-v{{ version }}.tar.gz
```
synapticarbors:
```suggestion
  url: https://gitlab.com/plant/plant/-/archive/v{{ version }}/plant-v{{ version }}.tar.gz
```

recipes/plant/LICENSE.txt
```diff
@@ -0,0 +1,16 @@
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
synapticarbors: This whole file can be removed since the build machinery will pick it up from the copy in the source distribution (the  tarball from gitlab).

synapticarbors:
It looks like the error from the CI is due to using a feature of f-strings that isn't available before python 3.12 (see https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings). Has this package been explicitly tested with python<3.12 previously?

gshiroma:
> It looks like the error from the CI is due to using a feature of f-strings that isn't available before python 3.12 (see https://docs.python.org/3/whatsnew/3.12.html#pep-701-syntactic-formalization-of-f-strings). Has this package been explicitly tested with python<3.12 previously?

@synapticarbors , we ran some tests, but they didn't catch that error. Thank you for pointing it out.

synapticarbors:
If the only issue is that `ncurses` is not available on windows, the way we usually handle this is to keep it a `noarch: python` build and then just ignore the failure on windows. Additionally, when I grep through the source for this package, I don't actually see an direct dependency on `ncurses`.  This package looks ostensibly like it's platform independent, so it's better to use `noarch: python` to minimize resource usage of our CI. When it goes to the feedstock, it will just build on linux for the `noarch: python` packages.

gshiroma:
> If the only issue is that `ncurses` is not available on windows, the way we usually handle this is to keep it a `noarch: python` build and then just ignore the failure on windows. Additionally, when I grep through the source for this package, I don't actually see an direct dependency on `ncurses`. This package looks ostensibly like it's platform independent, so it's better to use `noarch: python` to minimize resource usage of our CI. When it goes to the feedstock, it will just build on linux for the `noarch: python` packages.

Sounds good, @synapticarbors , thanks! Indeed, it looks that we can remove `ncurses` from our list of dependencies. I added `noarch: python` back into `meta.yaml`. There's a good chance the CI will fail on Windows, so I'm glad we can likely ignore that failure.

gshiroma:

PR #30760, opened 2025-08-11
Add starlette compress

dennis-wey:
Checklist
- [x] Build number is 0.


recipes/starlette-compress/meta.yaml
```diff
@@ -0,0 +1,44 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/starlette_compress-{{ version }}.tar.gz
+  sha256: dd1fa8aba94e255a3f7d4e8281c65ec1d6bd02de1af7500f1db1d611ac54071d
+
+build:
+  skip: true  # [py<39]
```
synapticarbors: This recipe looks like it should be `noarch: python`:

dennis-wey: HI @synapticarbors according to the docs you link, one of the requirements to use noarch is

> No python version specific requirements

However we do have zstandard which is only present for python<3.14.

Another question: 
the original repo also has this: https://github.com/Zaczero/starlette-compress/blob/e6baaeeadb2a54e0ec1cb7e4af89a2e0913ed438/pyproject.toml#L6
To my understanding that is not important because conda-forge relies on cpython but I want to be sure.

synapticarbors: I'm not quite sure what the best course of action is. You are correct that the `pyproject.toml` does specify `zstandard` for python<3.14 so under strict reading this shouldn't be `noarch: python`. On the other hand, looking at the code including `zstandard` doesn't look like it will break anything for python >=3.14 -- it would just cause the inclusion of an unnecessary package in the environment. We tend to strongly prefer `noarch: python` packages when possible since they are less resource intensive. 

@conda-forge/help-python -- any opinions on how to proceed with this one?

dennis-wey: Thanks @synapticarbors, personally I don't mind both ways.

@conda-forge/help-python can you help us out on this?

dennis-wey: @synapticarbors as we are not getting any feedback here and I don't have a strong opinion about this: Should I go forward in turning this into a noarch package?

dennis-wey:
@synapticarbors In order to get this unblocked, I changed the recipe to `noarch`. Can you take another look?

PR #30823, opened 2025-08-19
add libtommath

zelosleone:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -11,6 +11,8 @@ source:
 
 build:
   number: 0
+  skip:
+    - osx
   script:
     - mkdir build
     - cd build
@@ -39,7 +41,6 @@ tests:
             - lib/libtommath.so*
         - if: win
           then:
-            - Library/lib/tommath.lib
             - Library/bin/tommath.dll
 
 about:
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -30,7 +30,7 @@ requirements:
     - cmake
     - ninja
   run_exports:
-    strong:
+    weak:
       - ${{ pin_subpackage('libtommath', upper_bound='x.x') }}
 
 tests:
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -31,7 +31,7 @@ requirements:
     - ninja
   run_exports:
     strong:
-      - ${{ pin_subpackage('libtommath', max_pin='x.x') }}
+      - ${{ pin_subpackage('libtommath', upper_bound='x.x') }}
 
 tests:
   - package_contents:
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -31,8 +31,8 @@ build:
 
 requirements:
   build:
-    - ${{ compiler('c') }}
-    - ${{ stdlib('c') }}
+    - ${{ compiler("c") }}
+    - ${{ stdlib("c") }}
     - cmake
     - if: win
       then:
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -33,9 +33,6 @@ requirements:
   build:
     - ${{ compiler('c') }}
     - cmake
-    - if: unix
-      then:
-        - ${{ stdlib('c') }}
     - if: win
       then:
         - ninja
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -29,6 +29,9 @@ requirements:
     - ${{ stdlib('c') }}
     - cmake
     - ninja
+  run_exports:
+    strong:
+      - ${{ pin_subpackage('libtommath', max_pin='x.x') }}
 
 tests:
   - package_contents:
--- a/recipes/libtommath/recipe.yaml
+++ b/recipes/libtommath/recipe.yaml
@@ -45,8 +45,8 @@ tests:
             - Library/lib/pkgconfig/libtommath.pc
         - if: unix
           then:
-            - lib/cmake/libtommath/libtommath-*.cmake
-            - lib/pkgconfig/libtommath.pc
+            - **/lib/cmake/libtommath/libtommath-*.cmake
+            - **/lib/pkgconfig/libtommath.pc
 
 about:
   homepage: https://www.libtom.net/LibTomMath/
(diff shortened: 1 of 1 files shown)


zelosleone:

zelosleone:
@conda-forge-admin please rerender.

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+build:
+  number: 0
+  skip:
+    - osx
+  script:
+    - if: unix
+      then:
+        - mkdir build
+        - cd build
+        - cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON ..
```
remimimimimi: I think we actually should prefer dynamic libraries. If you really want to package only static, then read [CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md)

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+tests:
+  - script:
+      - if: win
+        then:
+          - if not exist "%LIBRARY_LIB%\tommath.lib" exit 1
+          - if not exist "%LIBRARY_INC%\tommath.h" exit 1
+      - if: unix
+        then:
+          - test -f $PREFIX/lib/libtommath.a || test -f $PREFIX/lib/libtommath.so
+          - test -f $PREFIX/include/tommath.h
```
remimimimimi: This could be replaced with built-in checks for header files and library files. Check out this section in documentation https://rattler.build/latest/reference/recipe_file/#check-for-package-contents

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ stdlib('c') }}
+    - cmake
+    - make
+    - if: win
+      then:
+        - ninja
```
remimimimimi: I don't see ninja being used. Maybe we should use it on all platforms instead of platform-specific make?

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+      then:
+        - mkdir build
+        - cd build
+        - cmake -DCMAKE_INSTALL_PREFIX=$PREFIX -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DCMAKE_POSITION_INDEPENDENT_CODE=ON ..
+        - make -j$CPU_COUNT
+        - make install
+    - if: win
+      then:
+        - mkdir build
+        - cd build
```
remimimimimi: I think we could move this part out of the conditional to avoid duplicating code.

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - cmake -G "NMake Makefiles" -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF ..
+        - nmake
+        - nmake install
```
remimimimimi: Doesn't it work if you write make instead of nmake? If you can use make as a name, then you can deduplicate code here as well.

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+  version: ${{ version }}
+
+source:
+  url: https://github.com/libtom/libtommath/archive/v${{ version }}.tar.gz
+  sha256: 6d099e93ff00fa9b18346f4bcd97dcc48c3e91286f7e16c4ac5515a7171c3149
+
+build:
+  number: 0
+  skip:
+    - osx
```
remimimimimi: Mac OS and Linux builds should be very similar, so I think you can remove this skip.

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+          - if not exist "%LIBRARY_INC%\tommath.h" exit 1
+      - if: unix
+        then:
+          - test -f $PREFIX/lib/libtommath.a || test -f $PREFIX/lib/libtommath.so
+          - test -f $PREFIX/include/tommath.h
+
+about:
+  homepage: https://www.libtom.net/LibTomMath/
+  repository: https://github.com/libtom/libtommath
+  documentation: https://www.libtom.net/LibTomMath/
```
remimimimimi: The documentation is actually in the PDF file here https://github.com/libtom/libtommath/releases/download/v1.3.0/bn-1.3.0.pdf

remimimimimi (changes requested):
Just a couple of comments :)

recipes/libtommath/recipe.yaml
zelosleone: yup doesnt work on my local machine without nmake

recipes/libtommath/recipe.yaml
zelosleone: ups that stayed before i switched to nmake, thanks!

remimimimimi (approved):

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+        - cmake -G Ninja -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON ..
+    - ninja
+    - ninja install
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ stdlib('c') }}
+    - cmake
+    - ninja
```
wolfv, line 31: Since this is a shared library we should probably add `run_exports`, so that packages that link `libtommath` will also get it as a run-dependency. 

For this it would be useful to estimate how much the ABI changes.

zelosleone: https://abi-laboratory.pro/?view=timeline&l=libtommath it seems like minor versions are fine but major versions x.x introduce breaking changes

zelosleone: added `strong` since this package will definitely be used to build other packages. also made it `x.x` for upperbound since i dont think minor versions will cause problems

recipes/libtommath/recipe.yaml
```diff
@@ -0,0 +1,72 @@
...
+  run_exports:
+    strong:
```
wolfv: this should be `weak` because it only has to go from `host -> run` (not `build -> run`).

PR #30843, opened 2025-08-22
Add at_cascade a public health modeling program.

bradbell:
Checklist
- [x ] Title of this PR is meaningful: e.g. "Adding my_nifty_package", not "updated meta.yaml".
- [x ] Source is from official source.
- [x ] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).
- [ x] If static libraries are linked in, the license of the static library is packaged.
- [x ] Package does not ship static libraries. If static libraries are needed, [follow CFEP-18](https://github.com/conda-forge/cfep/blob/main/cfep-18.md).
- [ x] Build number is 0.

### DIFF
--- a/recipes/at-cascade/recipe.yaml
+++ b/recipes/at-cascade/recipe.yaml
@@ -38,7 +38,6 @@ tests:
         - at_cascade
       pip_check: true
       python_version: ${{ python_min }}.*
-  #
   # Run one of the sandbox tests using the installed version of at_cascade.
   - script:
       # python example/csv/sim_fit_pred.py


bradbell:
@conda-forge/help-python : All the tests pass, waiting for review.

1. All the upstream tests are run in the source directory during the build; see build.sh and build.bat
2. One upstream test is run outside the source directory after the build; see tests section of recipe.yaml.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,73 @@
...
+  version: "2025.0.3"
+  python_min: "3.8"
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/bradbell/at_cascade/archive/${{ version }}.tar.gz
+  sha256: f2c70fa39464f825134da09ab14fc671ffb5c74218809433ff962f92151aed6e
```
dhirschfeld, line 13: I see this is a tag in your repo. IIUC the automatic updates won't work if you don't also create a corresponding GitHub Release.

Just an observation/FYI - I don't think it needs to block merging.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - python >=${{ python_min }}
+    - dismod-at >=2025.0.0
+    - numpy
+    - scipy
+
+tests:
+  - python:
+      imports:
+        - at_cascade
+      pip_check: true
```
dhirschfeld, line 39:
```suggestion
  - python:
      imports:
        - at_cascade
      pip_check: true
      python_version: ${{ python_min }}.*
```

dhirschfeld: It is recommended to test against the `python_min` version

recipes/at-cascade/recipe.yaml
bradbell: I think the comment above means that the automatic updates will work if a later release  is added to the
https://github.com/bradbell/at_cascade/releases/tag
directory ?

dhirschfeld: On the Releases page you can clici on `Draft a new release`:

image

...then choose the tag you want to create the release for:

image

recipes/at-cascade/recipe.yaml
bradbell: It seems that python.3.8 is no longer supported on Fedora 43; see
https://developer.fedoraproject.org/tech/languages/python/multiple-pythons.html

Below is what I get when I try to test the change above locally on my machine:
(If I change ${{ python_min }}.* to 3.9.* it works.)

```  
rattler-build build \
      --test native \
      --keep-build \
      -m .ci_support/linux64_cuda126.yaml \
      --log-style plain \
      --recipe "recipes/at-cascade/recipe.yaml"
.
.
.
   - pip
Error:   × Test failed: failed to setup test environment: Cannot solve the request because of: The following packages are incompatible
  │ ├─ python 3.8.* can be installed with any of the following options:
  ...
 ```

Comment (dhirschfeld) [reply recipes/at-cascade/recipe.yaml:39]: Python 3.9 is EOL this month, so it doesn't really make sense to have anything lower than 3.10 for `python_min`.

I'd actually remove it altogether and just use the `conda-forge` default unless you have a very good reason to override the default (which is 3.10 now)

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/28301a6e396fb24ffa7a6a50bb0c1528e3d5d265/recipe/conda_build_config.yaml#L847-L850

File: recipes/at-cascade/recipe.yaml (hunk 1 above)
Comment (bradbell) [reply recipes/at-cascade/recipe.yaml:13]: Thanks for confirming that conda-forge automatic updates will pick up those releases,

File: recipes/at-cascade/recipe.yaml [hunk 3]
```diff
@@ -0,0 +1,77 @@
+schema_version: 1
+
+context:
+  # name of this package
+  name: at-cascade
+  # latest release for this year's stable version of at-cascade
+  version: "2025.0.3"
+  # Minimum python version that has not reached end of life.
+  python_min: "3.10"
```
dhirschfeld:
```suggestion
```
This variable gets injected automatically. Unless there's a good reason for you to do so (e.g. you require a `python_min` *greater* than the default) you probably shouldn't specify it at all.

bradbell: I tried the suggestion above and it did not work with the command below (if I specify python_min in recipe.yaml it does work with the command below).
```
(build) staged-recipes.git>rattler-build build \
      --test native \
      --keep-build \
      -m .ci_support/linux64_cuda126.yaml \
      --log-style plain \
      --recipe "recipes/at-cascade/recipe.yaml"
Loading variant config file: ".ci_support/linux64_cuda126.yaml"
Error:   × Failed to parse recipe

Error: 
  × failed to parse match spec: unable to parse version spec: >=
    ╭─[recipes/at-cascade/recipe.yaml:31:7]
 30 │   host:
 31 │     - python >=${{ python_min }}
    ·       ─────────────┬────────────
    ·                    ╰── error parsing `python >=` as a match spec
 32 │     - hatchling
    ╰────
.
.
.
```

bradbell: This seems very similar to the following issue
https://github.com/conda-forge/conda-smithy/issues/2178
It was supposedly fixed in version  0.32.0  of rattler-build and I am using version 0.46.0.

Also see the zulip topic
[#general > v1 not picking up python_min](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/v1.20not.20picking.20up.20python_min/with/544347804)


You can test for this problem on your machine by commenting out the setting of python_min in recipe.yaml.

You can speed up testing for this problem by removing the sandbox check of at-cascade; i.e. comment out the following call in build.sh:
`"${PYTHON}" bin/check_py_test.py`

or in build.bat:
`%PYTHON% bin/check_py_test.py`

dhirschfeld:
I don't have merge rights, so if this is ready to go your best bet is to post a message to the Zulip chat to ask for a final review/merge

bradbell:
@dhirschfeld 
> I don't have merge rights, so if this is ready to go your best bet is to post a message to the Zulip chat to ask for a final review/merge

I just did that yesterday; see
https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Pull.20request.20stuck.20between.20review.20and.20merge/with/554914618

Thanks your help with this !!

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,77 @@
...
+  #   - ${{ version }}.patch
+
+# build.sh and build.bat run all the at_cascade tests
+# using the code in its source directory (sandbox).
+#
+# They then install at_cascade using pip as in commented out scipt entry below.
+build:
+  number: 0
+  noarch: python
+  # script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
danielnachun:
```suggestion
  noarch: python
  script:
    - ${{ PYTHON }} bin/check_py_test.py 
    - ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```
I'm not sure why you aren't just using this to build the package.  There is no need to print the version or run `pip show` at the end of the build as the relevant information is already provided by `rattler-build`..  As a `noarch: python` package this shouldn't even have platform specific build scripts either, as it will only be built on `linux-64`.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,77 @@
...
+source:
+  url: https://github.com/bradbell/at_cascade/archive/${{ version }}.tar.gz
+  sha256: f2c70fa39464f825134da09ab14fc671ffb5c74218809433ff962f92151aed6e
+  # patches:
+  #   - ${{ version }}.patch
+
+# build.sh and build.bat run all the at_cascade tests
+# using the code in its source directory (sandbox).
+#
+# They then install at_cascade using pip as in commented out scipt entry below.
```
danielnachun:
```suggestion
```
These comments are not needed as these steps are explained in the `ratter-build` documentation and the patch is no longer used.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,77 @@
...
+  # name of this package
+  name: at-cascade
+  # latest release for this year's stable version of at-cascade
+  version: "2025.0.3"
+  # Minimum python version that has not reached end of life.
+  python_min: "3.10"
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
```
danielnachun:
```suggestion
  version: "2025.0.3"

package:
  name: at-cascade
  version: ${{ version }}
```

The `name` context key is unneeded here as we don't do anything with it afterward that would require Jinja.  It is also not necessary to set `python_min` explicitly unless it is a version higher than our default, which is already set to 3.10.  You may need to set `python_min` in your recipe to get it to build locally, but it will be set automatically in the CI system, so please make sure to remove it when you push to GitHub.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,77 @@
...
+#
+# They then install at_cascade using pip as in commented out scipt entry below.
+build:
+  number: 0
+  noarch: python
+  # script: ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
+  host:
+    - python >=${{ python_min }}
```
danielnachun:
```suggestion
    - python =${{ python_min }}
```
This needs to be set at the exact version, not using it as a lower bound.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,77 @@
...
+tests:
+  - python:
+      imports:
+        - at_cascade
+      pip_check: true
+      python_version: ${{ python_min }}.*
+  #
+  # Run one of the sandbox tests using the installed version of at_cascade.
+  - script:
+      # python example/csv/sim_fit_pred.py
```
danielnachun:
```suggestion
  - script:
```
These comments are superfluous.

bradbell:
@danielnachun 
I think I have made your changes with the following differences:

1. build.bat and build.sh have been removed because they are no longer used.
2. version is still set in the context section because it is used two places.
3. python_min is still in the context section, but it commented out. This makes it easier for me to build and test locally.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,73 @@
...
+requirements:
+  host:
+    - python =${{ python_min }}
+    - hatchling
+    - pip
+    - dismod-at >=2025.0.0
+    - numpy
+    - scipy
+  run:
+    - python =${{ python_min }}
```
danielnachun, line 30:
```suggestion
    - python >=${{ python_min }}
```
For this we want to allow any version of python at or above the minimum, not just the minimum.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,73 @@
+schema_version: 1
+
+context:
+  # Version of package that this release corresponds to.
+  version: "2025.0.3"
+
+  # You may have to set python_min to get a local build to work but
+  # it is preferred to have python_min commented out when you commit this file.
+  # python_min: "3.10"
```
danielnachun: Just make sure to delete this in your last commit before this is merged - it's fine to keep in until you are done with local testing.

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,73 @@
...
+    - ${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+
+requirements:
+  host:
+    - python =${{ python_min }}
+    - hatchling
+    - pip
+    - dismod-at >=2025.0.0
+    - numpy
+    - scipy
```
danielnachun, line 28:
```suggestion
```
Most of the time these packages are not needed on host, only run.  Can you try removing these and see if it fails?  It's okay if we do ultimately need some, we just want to limit this set only to those which are actually needed at build time.

danielnachun (changes requested):
There are few more small changes here and then this will be ready.

recipes/at-cascade/recipe.yaml
bradbell: > Just make sure to delete this in your last commit before this is merged - it's fine to keep in until you are done with local testing.

I am just making sure that  I should remove the following from recipe.yaml ?
```
schema_version: 1

context:
  # Version of package that this release corresponds to.
  version: "2025.0.3"

  # You may have to set python_min to get a local build to work but
  # it is preferred to have python_min commented out when you commit this file.
  # python_min: "3.10"
```
and should replace the uses of ${{ version }} by 2025.0.3 ?

danielnachun: No, I just meant to remove the commented out code once local testing is done.  The `version: "2025.0.3"` should stay.  I guess the way I annotated that was confusing!

bradbell:
The following checks that scipy and numpy are actually used by the installed at-cascade, and not just for examples and testing.
```
at_cascade.git>git grep -l 'import scipy' at_cascade
at_cascade/bilinear.py
at_cascade.git>git grep -l 'import numpy' at_cascade
at_cascade/bilinear.py
at_cascade/create_shift_db.py
at_cascade/csv/pre_one_process.py
at_cascade/csv/pre_parallel.py
at_cascade/csv/simulate.py
at_cascade/fit_one_process.py
at_cascade/fit_parallel.py
```
Note that the at_cascade directory, and its subdirectory at_cascade/csv, end up in the installation. The corresponding repository 'at_cascade.git' can be found at:
https://github.com/bradbell/at_cascade

dhirschfeld:
> The following checks that scipy and numpy are actually used by the installed at-cascade,

If they're used by the installed at-cascade they should be in the run dependencies. I think Daniel was suggesting to remove them from the host dependencies to check if they're not needed at build time.

danielnachun:
Yes having them in `run` makes total sense, but it usually not necessary to also have them in `host` unless your build process is quite complex, which is not the case here.

bradbell:
> Yes having them in `run` makes total sense, but it usually not necessary to also have them in `host` unless your build process is quite complex, which is not the case here.

at-cascade runs all of its tests as part of the build process (to make sure that everything works on the target system). These are run on the sandbox version of the package; i.e., the one in the original repository structure.

Then after the build, during the test section, one of the tests is copied over and run using the installed version. To be specific:
```
    files:
      source:
        - example/csv/sim_fit_pred.py
```

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,69 @@
...
+      # python example/csv/sim_fit_pred.py
+      interpreter: python
+      content:  exec( open('example/csv/sim_fit_pred.py').read() )
+    requirements:
+      run:
+        - python =${{ python_min }}
+        - dismod-at >=2025.0.0
+        - numpy
+        - matplotlib
+        - scipy
```
dhirschfeld: run dependencies will automatically be installed in the test env so shouldn't be duplicated here

bradbell:
I think I have made the requesed changes above. Am I missing anything ?

dhirschfeld:
> I think I have made the requesed changes above. Am I missing anything ?

I don't really think you should be doing your testing at build time. Installing test dependencies in host is a bit of an abuse of the build system. In general, the dependencies in host are for packages you link against and are used to determine `run_exports` constraints for your package.

In this specific case, because your package is noarch you can probably get away with it, but that's a decision you're forcing us to make - whether or not to allow it in this specific case.

It appears the tests being run are a bunch of scripts which expect to be run against a source checkout so refactoring them would be a non-zero amount of work.

I don't have merge privileges, so I'll leave it to Daniel to decide if this is ok or not...

bradbell:
Would it be OK for me to comment out the tests out of of the build, so that I could do it during may testing, but it would not be in the final build ?

dhirschfeld:
IMO that would be fine as many packages on CF only do import tests under the assumption that the published release code has already been tested upstream.

I'll have to leave it to Daniel to decide either way though.

bradbell:
I am worried that the upstream testing does not cover the conda forge environment on all the target systems. I have commented it out so that it can be used when testing a new version, but will not slow down the final version that gets installed.
@danielnachun is this OK ?

bradbell:
@danielnachun I removed the testing comments mentioned above. Is there any thing else you would like changed for this package ?

recipes/at-cascade/recipe.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - numpy
+    - scipy
+
+tests:
+  - python:
+      imports:
+        - at_cascade
+      pip_check: true
+      python_version: ${{ python_min }}.*
+  #
```
danielnachun:
```suggestion
```
This looks like a left over.  Once this is removed I will merge!

danielnachun (approved):

PR #30847, opened 2025-08-22
add robo-gym-server-modules

zelosleone:
Checklist
- [X] Build number is 0.

### DIFF
--- a/recipes/robo-gym-server-modules/recipe.yaml
+++ b/recipes/robo-gym-server-modules/recipe.yaml
@@ -1,0 +1,67 @@
+context:
+  version: "0.3.0.1"
+  python_min: "3.8"
+
+package:
+  name: robo-gym-server-modules
+  version: ${{ version }}
+
+source:
+  url: https://pypi.org/packages/source/r/robo-gym-server-modules/robo-gym-server-modules-${{ version }}.tar.gz
+  sha256: 6e4cbd45a10be9afc29d9eaba167e207d30877da252a0a816125043461ca6a0f
+
+build:
+  noarch: python
+  number: 0
+  script: pip install . -v
+  files:
+    exclude:
+      - robo_gym_server_modules/logs
+
+requirements:
+  host:
+    - python ${{ python_min }}.*
+    - pip
+    - setuptools
+    - wheel
+  run:
+    - python >=${{ python_min }}
+    - grpcio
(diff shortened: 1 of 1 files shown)


zelosleone:
libtmux is a hard requirement https://github.com/jr-robotics/robo-gym-server-modules/blob/master/setup.py on build so skipped windows builds, removed noarch and changed the variables for python versioning.

zelosleone:

recipes/robo-gym-server-modules/recipe.yaml
```diff
@@ -0,0 +1,68 @@
...
+    - python ${{ python_min }}.*
+    - pip
+    - setuptools
+    - wheel
+  run:
+    - python >=${{ python_min }}
+    - grpcio
+    - protobuf >=3.20,<3.21
+    - pyyaml
+    - ${{ 'libtmux' if target_platform.family == 'unix' else null }} # tricking conda-linter here, but we need both noarch and skipping windows for this so :/
```
synapticarbors: If a package should qualify for `noarch: python` and does not support windows, we typically just let it fail in `staged-recipes` on windows since it's going to only be built on linux anyways when the feedstock is created. 

So I would just put the `libtmux` dependency, and go back to `pip_check: true` below.

But just to be clear, this package does not support windows?

recipes/robo-gym-server-modules/recipe.yaml
```diff
@@ -0,0 +1,68 @@
+context:
+  version: "0.3.0.1"
+  python_min: "3.8"
```
synapticarbors: We typically only set `python_min` if it is greater than the globally pinned `python_min`, which is 3.10. Conda-forge does not support python less than the global python_min, so my recommendation would be to remove this line.

recipes/robo-gym-server-modules/recipe.yaml
zelosleone: mainly because of libtmux, this package does not support windows indeed. i can go back to noarch approach without skipping then

recipes/robo-gym-server-modules/recipe.yaml
zelosleone: will do!

zelosleone:
@synapticarbors addressed the comments

PR #30855, opened 2025-08-24
Add pyLabDD recipe

AHartmaier:
Checklist
- [x] Build number is 0.


AHartmaier:
@conda-forge/help-python All tests pass. PR is ready for review. Thanks!

recipes/pylabdd/README.md
```diff
@@ -0,0 +1,100 @@
+[![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gh/AHartmaier/pyLabDD.git/main)
```
synapticarbors: We do not typically include a separate README. Please delete this file.

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - matplotlib-base
+
+test:
+  imports:
+    - pylabdd
+  requires:
+    - pytest
+    - python
+  commands:
+    - python -c "import pylabdd; print('pylabdd imported OK')"
```
synapticarbors: This is redundant with the `imports` test.

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+  run:
+    - python
+    - numpy {{ pin_compatible('numpy') }}
+    - matplotlib-base
+
+test:
+  imports:
+    - pylabdd
+  requires:
+    - pytest
```
synapticarbors: Do not include `pytest` unless you are running the test suite of the package. Also for the `pip check`, you should include `pip` under `requires`

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+    - {{ stdlib("c") }}
+  host:
+    - python
+    - pip
+    - setuptools
+    - numpy
+    - fmodpy
+  run:
+    - python
+    - numpy {{ pin_compatible('numpy') }}
```
synapticarbors:
```suggestion
    - numpy
```

https://conda-forge.org/docs/maintainer/knowledge_base/#building-against-numpy

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,54 @@
...
+requirements:
+  build:
+    - {{ compiler('fortran') }}
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+  host:
+    - python
+    - pip
+    - setuptools
+    - numpy
```
synapticarbors: Does this package actually use numpy in the compilation of the fortran routines? If not, then you probably don't need `numpy` here. I don't see it in the requirements for the `build-system`.

AHartmaier:
@synapticarbors Thx for the revision. I updated the PR according to your suggestions.

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - fmodpy
+  run:
+    - python
+    - numpy
+    - matplotlib-base
+
+test:
+  requires:
+    - pip
+    - python
```
synapticarbors:
```suggestion
```

You don't need to specify `python` here. Anything listed under `requirements` > `run` are automatically there in the test environment.

synapticarbors (commented):
Once final change and then I think we should be good to merge.

recipes/pylabdd/meta.yaml
AHartmaier: Done. Thx for your advice!

AHartmaier:
@synapticarbors Made all suggested changes + redirected source to pypi. Should be good for final check now.

recipes/pylabdd/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  name: pylabdd
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pylabdd/pylabdd-{{ version }}.tar.gz
+  sha256: 5430c839a29c7685788378469ae303d8ef63c268eef3b3421cade2d82595a2ae
+
+build:
+  number: 0
+  skip: true
```
synapticarbors:
```suggestion
```

Not sure how I missed this. This skips all builds, so nothing was actually being built.

AHartmaier: @synapticarbors Sorry, my bad. Now, the build fails because fmodpy is only on pypi but not on conda-forge. I guess there is no way to use in my recipe then?

synapticarbors: I've added a PR for `fmodpy` here:
https://github.com/conda-forge/staged-recipes/pull/30877

Once it is merged we should be good to go to try to run the CI again.

synapticarbors: That PR has been merged now. It will take probably around 30-45 minutes for it to hit the CDN, so after that I will re-start the CI on this PR and we'll see if that fixes everything.

AHartmaier: Very cool. Thanks a lot for your support. I hope it will build cleanly then.

synapticarbors:
So now `fmodpy` is there so it is getting to the build script, but it isn't finding `gfortran`. I'm not 100% sure, but I think that `gfortran` might not be available on path as `setup_fortran.py` expects, but I can see it defined here:
https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1320751&view=logs&j=2f458c3a-f375-5b50-7e4d-ed6fb77e81d3&t=a8e63dc4-a870-591b-5082-e8c52ba4cb9e&l=970

```
INFO: activate-gfortran_linux-64.sh made the following environmental changes:
+DEBUG_FFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -I$BUILD_PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/pylabdd-1.2.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fcheck=all -fbacktrace -fimplicit-none -fvar-tracking-assignments -ffunction-sections -pipe
+DEBUG_FORTRANFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -I$BUILD_PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/pylabdd-1.2.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix -march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-all -fno-plt -Og -g -Wall -Wextra -fcheck=all -fbacktrace -fimplicit-none -fvar-tracking-assignments -ffunction-sections -pipe
+F77=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran
+F90=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran
+F95=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-f95
+FC_FOR_BUILD=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran
+FC=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran
+FFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -I$BUILD_PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/pylabdd-1.2.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix
+FORTRANFLAGS=-march=nocona -mtune=haswell -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O2 -ffunction-sections -pipe -isystem $PREFIX/include -I$BUILD_PREFIX/include -fdebug-prefix-map=$SRC_DIR=/usr/local/src/conda/pylabdd-1.2.2 -fdebug-prefix-map=$PREFIX=/usr/local/src/conda-prefix
+GFORTRAN=$BUILD_PREFIX/bin/x86_64-conda-linux-gnu-gfortran
```

AHartmaier:
I am just trying to get gfortran as compiler directly. If that doesn't work will check, how it can be added to the path.

AHartmaier:
That didn't fix it. It could be a problem of subprocess in setup_fortran.py I will build a version w/o these tests and invoke fmodpy directly. Then we will see if fmodpy can access the compilers correctly.

synapticarbors:
You might need to configure `fmodpy` to use the correct `gfortran`:
https://github.com/tchlux/fmodpy/blob/master/fmodpy/config.py#L25

It looks like the conda-forge fortran compiler directive sets most of the usual environment names that people use to locate the compiler binary.

synapticarbors:
So now we've gotten to the `pip check` failing, and the issue is that the upstream `pyproject.toml` lists `fmodpy` as a run time dependency:
https://github.com/AHartmaier/pyLabDD/blob/main/pyproject.toml#L32

I'm not sure that it is, but to satisfy `pip check` we would need to add that to the `run` section.

AHartmaier:
@synapticarbors I could import FC from the environment and the build works fine now, except for windows. Thx for your guidance! Under Windows, "FC" doesn't seem to be set. ChatGPT suggests to add
    {% if win %}
    set "FC=%BUILD_PREFIX%\Library\mingw-w64\bin\x86_64-w64-mingw32-gfortran.exe"
    {% endif %}
to the build:script. Is this a proper solution to get the path to the fortran compiler under Windows or do you know better ways?

AHartmaier:
Concerning your previous comment: I had to add fmodpy in requirements/run, because it is needed as requirement for the PyPI version, where we cannot build a wheel for all archs. Hence some users might need a compilation during import if they install via PyPI. I guess for conda-forge it would not be strictly required, but I would like to keep one version of the package if it is not too harmful to the conda-forge recipe. If it is not there, pip check throws an error.

AHartmaier:
Back to the windows problem: set "FC=%BUILD_PREFIX%\Library\mingw-w64\bin\x86_64-w64-mingw32-gfortran.exe" does not seem to point to the gfortran compiler. I have no experience with Windows, do you have any suggestions concerning the compiler path or can we exclude windows systems for the package?

synapticarbors:
You might be able to use a `#[win]` selector with `script_env`:
https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#use-environment-variables

Alternatively you could use a `bld.bat` file for windows that sets it and then calls `pip`, and then a `build.sh` for unix that calls the single `script` line for linux/mac.

AHartmaier:
@synapticarbors Big thanks, that worked, but I had to specify a specific Fortran compiler for win, which was the suggested solution for python/fortran packages. I hope this will be acceptable. Otherwise, I can do some more research. I greatly appreciate all your support with the package!

synapticarbors:
@conda-forge/staged-recipes Does anyone have experience with building python packages with a fortran dependency on windows? I'm not quite sure what to recommend in this case. Explicitly specifying the fortran compiler gets `gfortran` on the path on windows, but I'm not sure if that's best practice.

AHartmaier:
@synapticarbors I think I found a proper solution now, I am referring to the [win] fortran compiler in the conda-forge-pinning [conda_build_config.yaml](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/master/recipe/conda_build_config.yaml), which also sets the environment variable FC correctly. Thus, it should be a stable recipe.

AHartmaier:
@synapticarbors Big thanks for your support with this recipe!

PR #30961, opened 2025-09-05
Add GPJax recipe

thomaspinder:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/gpjax/meta.yaml
+++ b/recipes/gpjax/meta.yaml
@@ -1,0 +1,52 @@
+{% set name = "gpjax" %}
+{% set version = "0.12.2" %}
+{% set python_min = "3.10" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/g/gpjax/gpjax-0.12.2.tar.gz
+  sha256: 00a64b802e4a6218f2ab5d54f415b0ceacf78897aa2ab8c7ca118b2ae928b8fe
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >={{ python_min }}
+    - hatchling
+    - pip
+  run:
+    - python >={{ python_min }}
+    - jax >=0.5.0
+    - jaxlib >=0.5.0
+    - optax >0.2.1
+    - numpyro
+    - jaxtyping >0.2.10
+    - tqdm >4.66.2
+    - beartype >0.16.1
+    - flax >=0.10.0
+    - numpy >=2.0.0
+
+test:
+  imports:
+    - gpjax
+  commands:
+    - pip check
+  requires:
+    - pip
+
+about:
+  home: https://github.com/JaxGaussianProcesses/GPJax
+  summary: Gaussian processes in JAX.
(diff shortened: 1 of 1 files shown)


recipes/gpjax/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  host:
+    - python >=3.10,<=3.13
+    - hatchling
+    - pip
+  run:
+    - python >=3.10,<=3.13
```
matthewfeickert:
```suggestion
{% set version = "0.12.0" %}
{% set python_min = "3.10" %}

package:
  name: {{ name|lower }}
  version: {{ version }}

source:
  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/gpjax-{{ version }}.tar.gz
  sha256: f59a19f71f488a06182cbb4cf6bf51fcb8d97aef011f77fb6d1507c808269532

build:
  noarch: python
  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
  number: 0

requirements:
  host:
    - python >={{ python_min }}
    - hatchling
    - pip
  run:
    - python >={{ python_min }}
```

Using `{{ python_min }}` is easier to maintain. You have upper bounds on the Python version, which is coming from the project's `pyproject.toml` but as there aren't any compiled dependencies that can be specific to the Python release version this should get removed in general in favor of the known lower bounds only of Python 3.10.

matthewfeickert (changes requested):

matthewfeickert:
The Windows failure is expected I think (which is okay) so I just have a suggestion about using the `{{ python_min }}` variable.

recipes/gpjax/meta.yaml
thomaspinder: Whilst I agree with the principle of these changes, they do appear have broken the Linux and OSX builds. Unfortunately, I cannot see the logs to understand why. Do you have any ideas?

matthewfeickert:
I'm currently on my phone on the way to the airport but I can take a look at what is happening in a bit. It is possible to look at the logs but you need to do it in a round about way of accessing it from the failed GitHub Action marker.

matthewfeickert:
Ah right

```
Will try a different candidate, due to conflict:
    gpjax 0.12.0 depends on Python <=3.13,>=3.10
ERROR: Package 'gpjax' requires a different Python: 3.13.7 not in '<=3.13,>=3.10'
Exception information:
...
```

The build backed you have set is going to check if the Python version is compatible or not. While that can be easily patched, and then addressed upstream, for now it is better to just get a build into conda-forge. So keep the `{{ python_min }}` but then add back the upper bound for now: `>={{ python_min }},<=3.13`.

thomaspinder:
Hi @matthewfeickert , seems like all builds except windows are now successful. How does it look to you?

matthewfeickert:
@thomaspinder looks good I think. I'd suggest maybe doing a rebase off of upstream/main and then squashing everything into a single commit to make it easier for the reviewers, but I think you can follow the suggestions that are comments in the PR body to ask the python-help team for review now.

thomaspinder:
May I request a review please @conda-forge/help-python ?

hmaarrfk:
jaxlib doensn't exist for windows.

PR #30964, opened 2025-09-05
Add recipe for easchersim

j-eser:
Checklist
- [x] Build number is 0.


aucummin:
I am willing to be a recipe maintainer!

j-eser:
@conda-forge-admin, please ping team @conda-forge/help-python this PR is ready for review

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,68 @@
...
+build:
+  number: 0
+  script: |
+    {% if win %}
+    set SETUPTOOLS_SCM_PRETEND_VERSION={{ version }} && {{ PYTHON }} -m pip install --no-deps . -vv
+    {% else %}
+    export SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+    {{ PYTHON }} -m pip install --no-deps . -vv
+    {% endif %}
+  skip: true  # [win]
```
ocefpaf: Why are you adding a windows if-clause and then skipping it? Also, it looks like this recipe can be noarch, no?

j-eser: The windows if was there for a windows build but an error that I could not track down occurred when building this package for windows and as the target audience is mainly linux/macOS based I decided to ignore for win for now. Happy to remove the if statement if that is better.
I am a novice and hence not sure about the second question, but I think it could be noarch indeed. What would I have to do to change that? Thanks.

ocefpaf: Remove that skip and make it noarch. If the failures is a missing dependency on Windows, it is OK to make it noarch.

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+    - numpy >=1.20
+    - wheel
+
+  run:
+    - numpy >=1.20
+    - python >={{ python_min }}
+    - scipy
+    - matplotlib
+    - importlib_resources
+    - matplotlib
```
ocefpaf:
```suggestion
    - matplotlib-base
    - importlib_resources
```

recipes/easchersim/meta.yaml
j-eser: I made it noarch and the windows build fails with the following error

> LookupError: setuptools-scm was unable to detect version for D:\bld\easchersim_1757429979648\work.
> 
>   Make sure you're either building from a fully intact git repository or PyPI tarballs. Most other sources (such as GitHub's tarballs, a git checkout without the .git folder) don't contain the necessary metadata and will not work.
> 
>   For example, if you're using pip, instead of https://github.com/user/proj/archive/master.zip use git+https://github.com/user/proj.git#egg=proj
> 
>   Alternatively, set the version with the environment variable SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME} as described in https://setuptools-scm.readthedocs.io/en/latest/config/
>   error: subprocess-exited-with-error
>   
>   Preparing metadata (pyproject.toml) did not run successfully.

I assume it is that the win build can not read the environment set for the version with the export command. As it also didn't work when I added the windows if statement in the original submission, I think removing the windows build would be a good solution.

recipes/easchersim/meta.yaml
j-eser: I will make these changes and run to see if it leads to any issues and then commit them here

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,65 @@
+{% set name = "easchersim" %}
+{% set version = "v1.2.0" %}
+{% set python_min = "3.10" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://gitlab.com/c4341/easchersim/-/archive/v1.2.0/easchersim-v1.2.0.tar.gz
```
ocefpaf:
```suggestion
  url: https://gitlab.com/c4341/easchersim/-/archive/v{{ version }}/easchersim-v{{ version }}.tar.gz
```

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools >=46
+    - setuptools_scm[toml] >=3.4
+    - numpy >=1.20
+    - wheel
+
```
ocefpaf:
```suggestion
```

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,65 @@
...
+  script: |
+    export SETUPTOOLS_SCM_PRETEND_VERSION={{ version }}
+    {{ PYTHON }} -m pip install --no-deps . -vv
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools >=46
+    - setuptools_scm[toml] >=3.4
```
ocefpaf: This syntax is Python specific, conda is Python agnostic.

```suggestion
    - setuptools_scm >=3.4
```

recipes/easchersim/meta.yaml
j-eser: suggestion added

recipes/easchersim/meta.yaml
j-eser: suggestion added

recipes/easchersim/meta.yaml
j-eser: suggestion added

recipes/easchersim/meta.yaml
j-eser: suggestion added

j-eser:
The windows build is removed due to different issue arising on windows.
noarch build leads to binary in package as this is default of setuptools for entry points in my understanding
platform specific builds fail on test on windows, hence no windows build foreseen for now

ocefpaf:
> The windows build is removed due to different issue arising on windows. noarch build leads to binary in package as this is default of setuptools for entry points in my understanding platform specific builds fail on test on windows, hence no windows build foreseen for now

Please make it noarch. The Windows failure here is not relevant and the noarch package will ve installable on Windows.

recipes/easchersim/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - matplotlib-label-lines
+    - rich
+
+test:
+  commands:
+    - easchersim -h
+    - easchersim make-config --help
+    - easchersim run -h
+  imports:
+    - easchersim
```
ocefpaf:
```suggestion
  commands:
    - pip check
    - easchersim -h
    - easchersim make-config --help
    - easchersim run -h
  imports:
    - easchersim
  requires:
    - python {{ python_min }}
    - pip
```

j-eser:
I moved back to noarch and added your suggestions for the test as well. As expected the win build fails.

PR #31016, opened 2025-09-11
Add recipe for mdbtools

JulianWachterQC:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/mdbtools/recipe.yaml
+++ b/recipes/mdbtools/recipe.yaml
@@ -16,7 +16,7 @@ build:
     - win
   script:
     content: |
-      autoreconf -fis
+      autoreconf -fis -I "${BUILD_PREFIX}/share/gettext/m4"
       ./configure --prefix="${PREFIX}"
       make -j"${CPU_COUNT}"
       make install
@@ -28,7 +28,7 @@ requirements:
     - autoconf
     - automake
     - gawk
-    - gettext <0.24
+    - gettext
     - libiconv
     - libtool
     - make


JulianWachterQC:
@moritzwilksch What do you think :) Are you up for the job?

moritzwilksch:
sign me up 🫡

JulianWachterQC:
@conda-forge/help-c-cpp ready for review 🐞

recipes/mdbtools/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+      make install
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ stdlib('c') }}
+    - autoconf
+    - automake
+    - gawk
+    - gettext <0.24
```
xhochy: What's the reason for this pin? In `build`, one should only add `gettext-tools` and `if: osx; then: libintl-devel` in `host`

I-Al-Istannen: `gettext` ≥ 24 lacks the m4 macro definition the build needs. This leads to leftover unexpanded macros in the configure script, which does not like it. At least that's what I gathered from dissecting the arch package archives and stracing the build...

It does not use the tools, so depending on `gettext-tools` does not help, I think? It needs the m4 macro definition at compile time.

JulianWachterQC: gettext-24 and higher seem to have removed the `/usr/share/aclocal/po.m4` macro definitions the build depends on. Therefore, I pinned it to <0.24. I can add a comment explaining that.

I don't think gettext-tools helps with that macro file.

ETA: Maybe there is *some* way to get the build to understand that the files now seem to live in `/usr/share/gettext` and not `/usr/share/aclocal`, but this feels like more of an upstream problem.

JulianWachterQC: Okay, I worked around that, but now I need to know the location of the gettext dir. I can hardcode the path to /usr/share/gettext and remove it from build, but that does not really feel much better?

xhochy:
Looks OK, let's finetune on the feedstock.

PR #31023, opened 2025-09-12
Add MAQAO recipe - Modular Assembly Quality Analyzer and Optimizer

toprinse:
The maqao recipe aims to provide a conda package for the binary installation of the MAQAO software analysis tool. This package is only available on linux-64 and linux-aarch64 architectures.

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/maqao/recipe.yaml
+++ b/recipes/maqao/recipe.yaml
@@ -27,8 +27,7 @@ requirements:
   build:
     - ${{ compiler('c') }}
     - ${{ compiler('cxx') }}
-    # - ${{ stdlib('c') }}
-    - sysroot_linux-64
+    - ${{ stdlib('c') }}
     - cmake >=3.3.0
     - make
     - pkg-config


toprinse:
@cvalensi Here is the proposal for the MAQAO package.
Can you confirm that you agree to be listed as a maintainer of the conda-forge package, in the role of official maintainer of MAQAO.

cvalensi:
Hello, thanks for creating this package. I confirm that I agree to be listed as a maintainer.

toprinse:
@conda-forge/staged-recipes. The recipe is ready to be merged. Could you review it please ? Thanks !

toprinse:
This package is implemented in LUA, but the way we made the conda package is just an install of the binaries of the package, so I don't know which team should be ping. Thanks

cvalensi:
The package seems good to me. One additional test I could suggest would be `maqao --detect-proc`, which checks that maqao supports the processor on which it was installed.
We are providing MAQAO as a standalone binary executable for ease of distribution, so having the script installing the binaries in the archive is the right way to deploy.

toprinse:
@conda-forge/help-c-cpp, Hello, so we did not know which team to ping for the review of this PR as the code is only a pull from binaries. Could you help us with it or tell us which team to ping for the review ? Thanks !

toprinse:
@conda-forge/help-c-cpp, Hello, sorry for the spam, but as I did not receive any answer I was wondering if you have seen the PR and if you have any recommendations about the package we propose. Thanks !

traversaro:
Thanks @toprinse, thanks for the contribution. Unless specific corner cases, conda-forge does to support recipe that do a "binary repackaging" of existing binaries (see https://conda-forge.org/docs/maintainer/guidelines/#compiled-objects), but the packages should be build from source using conda-forge compilers. Did you consider doing that here? Thanks in advance!

toprinse:
Hello @traversaro ! Thanks for the answer. So I made a new version of the recipe file which builds the package form the sources. The only thing is that for now the sources could not be build with gcc15 which is taken by default for the build, so i had to specifically use gcc 13. Is this new version ok to be merged ? Thanks !

recipes/maqao/recipe.yaml
```diff
@@ -0,0 +1,72 @@
...
+      ln -s "${CC}" gcc
+      export PATH="$(pwd):$PATH"
+    fi
+    
+    mkdir build
+    cmake -S . -B build -DCMAKE_POLICY_VERSION_MINIMUM=3.5
+    make -C build -j${CPU_COUNT}
+
+    mkdir -p ${PREFIX}/bin
+    cp bin/maqao ${PREFIX}/bin/
```
traversaro: Isn't there an install target in the CMake build system?

recipes/maqao/recipe.yaml
```diff
@@ -0,0 +1,72 @@
...
+build:
+  skip: 
+    - if: not (linux and (x86_64 or aarch64))
+      then: true
+  number: 0
+  script: |
+    if ! command -v gcc >/dev/null; then
+      ln -s "${CC}" gcc
+      export PATH="$(pwd):$PATH"
+    fi
```
traversaro: Why is this necessary? The build system should use `CC` env variable, that by default is not `gcc`, but a different name.

recipes/maqao/recipe.yaml
toprinse: In fact, I cannot modify the CMakeLists.txt of the MAQAO project directly, as I am not a maintainer of the project. Moreover, since the project only has one release per year, I'm not sure it would be necessary to wait for the next release to request this change in the source code. So in the recipe, I'm not sure there really is an alternative for this...

recipes/maqao/recipe.yaml
toprinse: If I remove the symbolic link, I got the same issue than in [this pipeline](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1368810&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff). I think that there is some issue with the compiler which is take in account during the build and I get the error `make[4]: gcc: No such file or directory`. I don't know if the solution is to add `gcc` as a dependency of the package directly in the recipe ? If you have any suggestions it could be great !

recipes/maqao/recipe.yaml
traversaro: But if I am not wrong, the executable is already installed by CMake, see https://gitlab.liparad.uvsq.fr/maqao_release/MAQAO/-/blob/5463caa1666a6d86c068420aff58ac1ef0666978/src/CMakeLists.txt#L294-298 ?

traversaro:
I see that you ticked in the checklist:

>  Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged).

and 

>  If static libraries are linked in, the license of the static library is packaged.

Can you double check, it seems that a lot of other packages are indeed vendored.

recipes/maqao/recipe.yaml
traversaro: The compiler meta package install the gcc compiler under the `x86_64-conda-linux-gnu-c++` name (or different based on the target). However, the exact name should not matter, as most build system observe the `CC` and `CXX` environment variables to get the actual compiler to use. However, the problem here seems that you are trying to build a vendored luajit, instead of reusing the existing package https://github.com/conda-forge/luajit-feedstock .

toprinse:
@traversaro If I understood the conda-forge documentation correctly, is it acceptable to include the licenses of all third-party components used or vendored in MAQAO within the main LICENSE file, as done [here](https://gitlab.liparad.uvsq.fr/maqao_release/MAQAO/-/blob/master/LICENSE?ref_type=heads)? Or is it required to provide each license as a separate file?


MAQAO is a monolithic project with deeply integrated vendored dependencies. As you can see in the CMakeLists.txt files, the build system uses hardcoded paths and direct make calls:

```cmake
SET (LOCAL_LUA_LIB "${CMAKE_SOURCE_DIR}/src/lua-5.1.5/src/liblua.a")
SET (LOCAL_LUA_BIN "${CMAKE_SOURCE_DIR}/src/lua-5.1.5/src/lua")
COMMAND ${CMAKE_COMMAND} -E chdir ${CMAKE_CURRENT_SOURCE_DIR} make linux
COMMAND ${CMAKE_COMMAND} -E copy ${LOCAL_LUA_LIB} ${LIBRARY_OUTPUT_PATH}
```

I don't have rights to modify the upstream build system, and refactoring the entire build system to use external dependencies would likely not be feasible given the extensive integration of vendored components.

Would you recommend:
 1. Accept the vendoring and document it properly (as currently done with all licenses in the LICENSE file), or
 2. Require extensive upstream refactoring (which may not be realistic for this type of monolithic project)?

This seems similar to how other large monolithic packages are handled in conda-forge when unbundling isn't practical. such as for [VTK](https://github.com/conda-forge/vtk-feedstock/blob/main/recipe/meta.yaml) ?

recipes/maqao/recipe.yaml
toprinse: Yes, so I fixed this by adding `-DCMAKE_INSTALL_PREFIX=${PREFIX}` to install the binary at the proper place, is this okay ?

recipes/maqao/recipe.yaml
toprinse: So if we can keep the inside build of luajit as explained in other comments, I don't really know if there is another option than using a symlink to use the conda compiler provided during the build. I'm open to any suggestions but I already tried to export the env variables and it was also not working.

traversaro:
> I don't have rights to modify the upstream build system

Any feedstock can carry on patches to upstream projects, no special rights are necessary.

recipes/maqao/recipe.yaml
traversaro: Yes, but in theory that options should be already passed by `$CMAKE_ARGS`.

recipes/maqao/recipe.yaml
traversaro: Ideally the build system should be patched to actually observer `CC` and `CXX` env variable instead of hardcoding the compiler name. If that is not possible, you can just add an (additional) dependency on `gcc` or `gxx` packages that indeed install the required symlink.

traversaro:
> @traversaro If I understood the conda-forge documentation correctly, is it acceptable to include the licenses of all third-party components used or vendored in MAQAO within the main LICENSE file, as done [here](https://gitlab.liparad.uvsq.fr/maqao_release/MAQAO/-/blob/master/LICENSE?ref_type=heads)?

It is acceptable, but it his not recommended. Introducing such a package with vendored dependencies may require more people from staged-recipes or core to agree if that is acceptable or not, and as you may imagine that will take time. I would at least try to patch the build system to use an external luajit and lua, if that does not work then we can ask more people to reach a consensus on this. But in my opinion the path of least resistance is to at least try to unvendor dependencies.

toprinse:
> > @traversaro If I understood the conda-forge documentation correctly, is it acceptable to include the licenses of all third-party components used or vendored in MAQAO within the main LICENSE file, as done [here](https://gitlab.liparad.uvsq.fr/maqao_release/MAQAO/-/blob/master/LICENSE?ref_type=heads)?
> 
> It is acceptable, but it his not recommended. Introducing such a package with vendored dependencies may require more people from staged-recipes or core to agree if that is acceptable or not, and as you may imagine that will take time. I would at least try to patch the build system to use an external luajit and lua, if that does not work then we can ask more people to reach a consensus on this. But in my opinion the path of least resistance is to at least try to unvendor dependencies.

Okay, I will have a look on that and discuss the case with the maintainers of the project to have their opinion !

traversaro:
> Okay, I will have a look on that and discuss the case with the maintainers of the project to have their opinion !

To clarify, I see how this may be annoying for someone not used with package managers, but avoiding as much as possible vendoring dependencies (or at least provide a way to do not vendor them) is a common requirement for package managers, both at the Linux distribution level (like Debian, Fedora or nix) or multi-platform ones (like vcpkg or spack). For a nice introduction to the topic, see https://bssw.io/items/don-t-package-your-libraries-write-packageable-libraries , in particular https://youtu.be/sBP17HQAQjk?si=JcGlWjcejopAzwaq&t=859 .

toprinse:
> > Okay, I will have a look on that and discuss the case with the maintainers of the project to have their opinion !
> 
> To clarify, I see how this may be annoying for someone not used with package managers, but avoiding as much as possible vendoring dependencies (or at least provide a way to do not vendor them) is a common requirement for package managers, both at the Linux distribution level (like Debian, Fedora or nix) or multi-platform ones (like vcpkg or spack). For a nice introduction to the topic, see https://bssw.io/items/don-t-package-your-libraries-write-packageable-libraries , in particular https://youtu.be/sBP17HQAQjk?si=JcGlWjcejopAzwaq&t=859 .

Thanks for the links, I will have a read about it ! I fully understand why it is a problem for package managers. The only reason why I was not sure about the way to do is because this is not a project I'm familiar with, and so I did not had a really close look about the way it is built. The main issue here is that this MAQAO projects claimed to be fully static and have no dependencies, so here is the main thing ! But I'm working on a patch to keep the static build for the debian package distributed, and the patch will normally remove the local build for LuaJit. 

Thanks a lot for the help and recommendations !

toprinse:
Hello @traversaro. So I made a patch that works to remove inside build of LuaJit and libdwarf which already have conda-forge feedstocks. In the patch, I also clearly remove the sources of LuaJit and dwarf to be fully transparent. The thing is that the patch is now over 200000 lines. Is it okay if I put it here like that, or should I keep the sources of vendored packages inside the source code of Maqao but just don(t use them for the build ? My opinion is that it is more transparent to fully remove them from the sources so if you can just tell me what is the better solution ? Thanks by advance !

traversaro:
I think fully removing the source make sense, thanks! But it indeed it is not a good idea to include that in the patch, especially as it would make the patch really prone to conflicts. Probably the patch can just change the CMake, and to avoid that the vendored source code is used, you can delete the related folder as part of the build script, see for example https://github.com/conda-forge/collada-dom-feedstock/blob/857fa2bc8c81ebffb5f804a7d211f2adaef58a11/recipe/build.sh#L3-L4 .

toprinse:
> I think fully removing the source make sense, thanks! But it indeed it is not a good idea to include that in the patch, especially as it would make the patch really prone to conflicts. Probably the patch can just change the CMake, and to avoid that the vendored source code is used, you can delete the related folder as part of the build script, see for example https://github.com/conda-forge/collada-dom-feedstock/blob/857fa2bc8c81ebffb5f804a7d211f2adaef58a11/recipe/build.sh#L3-L4 .

Perfect, I was also supposing that adding the whole patch will be problematic ! So I made the changes and checked that LuaJIT and libdwarf are now coming from conda-forge, could you review it ? Thanks !

traversaro:
Great work! I am looking at the installed files, and I see that there is a `lib/libcounting.so` file, is that expected? Is that something maqao related?

toprinse:
> Great work! I am looking at the installed files, and I see that there is a `lib/libcounting.so` file, is that expected? Is that something maqao related?

yes, libcounting is a performance measurement library created by MAQAO. It provides hardware performance counter instrumentation capabilities for runtime analysis.

traversaro:
> > Great work! I am looking at the installed files, and I see that there is a `lib/libcounting.so` file, is that expected? Is that something maqao related?
> 
> yes, libcounting is a performance measurement library created by MAQAO. It provides hardware performance counter instrumentation capabilities for runtime analysis.

I see. I guess we can't change its name, but consider that such generic library names are prone to naming collision. However I checked in https://conda-metadata-app.streamlit.app/Search_by_file_path and there is no current naming conflict, so I think we are good to go, thanks!

recipes/maqao/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_compiler_version:
+  - 13
+cxx_compiler_version:
+  - 13
```
traversaro, line 4: This will use both clang 13 on macos and gcc 13 on Linux, it is intentional?

recipes/maqao/use-system-deps.patch
```diff
@@ -0,0 +1,447 @@
...
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
+@@ -19,6 +19,10 @@
+ ##
+ 
+ CMAKE_MINIMUM_REQUIRED(VERSION 3.3)
++if(DEFINED ENV{PREFIX})
++   include_directories($ENV{PREFIX}/include)
++   message(STATUS "Adding conda include path: $ENV{PREFIX}/include")
++endif()
```
traversaro: Not blocking, but I think this should not be necessary.

recipes/maqao/conda_build_config.yaml
toprinse: Yes, because for now MAQAO does not support gcc15. It will be fixed in the next release in january

recipes/maqao/use-system-deps.patch
toprinse: I will try without it and see !

recipes/maqao/conda_build_config.yaml
traversaro: Ah my bad, this is not building on macOS, so this is correct.

traversaro:
Other then the minor comment, this is ready to merge for me. Can you squash the commits to avoid polluting the staged-recipes repo history? Otherwise I Can squash on merge if that is ok for you, thanks!

toprinse:
> Other then the minor comment, this is ready to merge for me. Can you squash the commits to avoid polluting the staged-recipes repo history? Otherwise I Can squash on merge if that is ok for you, thanks!

I added the minor changes and rebuilt the package, everything seems to be okay ! You can squash the commits for the merge. Thanks for the review and all the advice !

toprinse:
Hello @traversaro ! So is it possible to merge this PR ? We will need the package for some demonstration in a near future so if it can be merged before it coule be great !
Thanks a lot !

vincent-pollet:
Hello! May I kindly ask what is the status of this PR? I would love to have maqao on conda-forge. Thank you!

traversaro:
> Hello @traversaro ! So is it possible to merge this PR ? We will need the package for some demonstration in a near future so if it can be merged before it coule be great ! Thanks a lot !

Sorry, I missed this notification. In general, reviewers tend to ignore PRs with the label `Awaiting author contribution`. To ensure that the label is removed, please always ping the relevant team, in this case @conda-forge/help-c-cpp

recipes/maqao/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+  name: maqao
+  version: "2025.1.0"
+  sha256: "26d398e74ee0a3edfc3137183c87c4b13b6b8e0b99bdcdcc907e9204acbb0163"
+
+package:
+  name: ${{ name|lower }}
+  version: ${{ version }}
+
+source:
+  url: https://gitlab.liparad.uvsq.fr/maqao_release/MAQAO/-/archive/5463caa1666a6d86c068420aff58ac1ef0666978/MAQAO-5463caa1666a6d86c068420aff58ac1ef0666978.tar.gz
```
traversaro: It does not seem that the upstream projects have tags. Can you explain how this checkout is used for the `2025.1.0` version? Would it be possible for the upstream project to actually use tags?

recipes/maqao/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+about:
+  homepage: https://www.maqao.org/
+  summary: 'MAQAO - Modular Assembly Quality Analyzer and Optimizer'
+  description: |
+    MAQAO (Modular Assembly Quality Analyzer and Optimizer) is a performance analysis and optimization 
+    framework operating at binary level with a focus on core performance.
+    
+    This package uses system libraries for LuaJIT, libdwarf, readline, and ncurses.
+    It includes the following vendored components:
+    - libfcgi (Open Market license)
```
traversaro: If this library is vendored the `license_file` and `license` of the vendored library should be included.

toprinse: Thank you for the feedback regarding the vendored libraries and their licenses.

You're absolutely right about including the license information for vendored components. I've updated the recipe to properly document all vendored libraries in the description section.

The LICENSE file of MAQAO is already composed of the licenses of every vendored package aggregated in a single consolidated file. Is this consolidated approach acceptable regarding conda-forge best practices ?

traversaro: > The LICENSE file of MAQAO is already composed of the licenses of every vendored package aggregated in a single consolidated file. Is this consolidated approach acceptable regarding conda-forge best practices ?

You are perfectly right, I actually remember check that out back in time. For future users like me, perhaps we can mention this in the recipe itself to avoid confusion.

traversaro: However, the `  license: LGPL-3.0-or-later` should reflect this, to include the spdx license of all vendored deps.

toprinse: So I added `MIT` and `LGPL-2.1-or-later`. Concerning the `Open-Market`  license, it does not have a SPDX ID, so is it okay if it is documented in the global LICENSE file of the package and in the description of the package ?

recipes/maqao/recipe.yaml
toprinse: Hello ! So the maqao maintainers made a 2026.1.0 release that is now pinned in the package context section!

toprinse:
@traversaro @conda-forge/help-c-cpp I fixed the version problem, we are now pointing on the 2026.1.0 version pinned in the main gitlab of MAQAO. Is it okay for merge ? Thanks !

vincent-pollet:
Hello @traversaro, sorry to ping you directly. I think all the changes you requested have been implemented, could you have another look? Thanks! :pray:

PR #31051, opened 2025-09-18
fix: update linter workflow to include issue_comment event for PRs

beckermr:
This PR ensures the staged-recipes linter runs when folks comment on issues to check for maintainers approving.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -12,7 +12,7 @@ on:
       - reopened
       - labeled
       - unlabeled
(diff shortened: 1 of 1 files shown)


.github/workflows/linter.yml
```diff
@@ -12,6 +12,9 @@ on:
       - reopened
       - labeled
       - unlabeled
+  issue_comment: nll
```
beckermr:
```suggestion
  issue_comment: null
```

PR #31055, opened 2025-09-18
Add SnowMapPy recipe

Hbechri:
This PR adds the Conda recipe for the SnowMapPy package.


haytamelyo:
I confirm

1minds3t (changes requested):
Hi @haytamelyo! Thanks for contributing **SnowMapPy** to conda-forge — it looks like a really useful package for snow cover analysis!

### Key Issue: Over-Pinning Dependencies

The CI failures are caused by exact version pins for many dependencies. This creates impossible constraints for the conda solver, e.g., your `python_min` is 3.8, but some packages like `zarr==3.0.6` require Python ≥3.11.

### Suggested Solution: Flexible Version Constraints

Use `>=` for minimum versions rather than `==`. Only pin exact versions when absolutely necessary. This allows conda to find compatible combinations across platforms.

**Example update:**

```yaml
run:
  - python >={{ python_min }}
  - numpy >=2.0        # minimum required by your code
  - scipy >=1.10
  - pandas >=2.0
  - setuptools >=60    # only pin if needed
  - zarr >=2.10
  - rasterio >=1.3
  - geopandas >=0.14
  - xarray >=2023.1
  # ... other dependencies as needed
```

### Why This Helps

* **Solver Flexibility:** Finds newest compatible versions automatically
* **Cross-Platform:** Micro-versions differ by platform but remain compatible
* **Future-Proof:** Package survives dependency updates
* **Lower Maintenance:** No need to constantly update pins

### Extra Recommendations

1. Only set minimum versions your code truly requires
2. Use upper bounds sparingly (`<2.0`) if newer versions break your code
3. Remove unnecessary pins (e.g., `setuptools`, `python-dateutil`)
4. Update `python_min` if required by dependencies (≥3.11)

**Quick Test:** After updating the recipe, you can verify compatibility with:

```bash
conda create -n test-env -c conda-forge snowmap-py
python -c "import snowmap; print('Import successful')"
```

This approach follows conda-forge best practices and will make your package much easier to maintain. Let me know if you want a hand testing these changes!

Hbechri:

recipes/snowmappy/meta.yaml
```diff
@@ -0,0 +1,75 @@
...
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools >=60
+  run:
+    - python >={{ python_min }}
+    - numpy >=2.0
+    - scipy >=1.10
+    - pandas >=2.0
+    - setuptools >=60
```
travishathaway, line 28:
```suggestion
```

This isn't listed in the project dependencies, and it should only be listed in the host section because it's not needed at runtime (it's used to build the wheel).

travishathaway (changes requested):
Hi @Hbechri,

I just wanted to ping you on this review again. I think we can get it merged it we address the issue I pointed out below.

1. Remove `setuptools` from `run` dependencies. This should only be needed in `host` during the package build.
2. Convert to `v1` recipe format. You can do this by installing `conda-recipe-manager` with `pixi global install conda-recipe-manager` and using the `crm convert` command (`crm convert meta.yaml > recipe.yaml`).

Let me know if you have any more questions.

PR #31124, opened 2025-09-27
Adding backports.zstd

Rogdham:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/backports.zstd/meta.yaml
+++ b/recipes/backports.zstd/meta.yaml
@@ -1,0 +1,48 @@
+{% set name = "backports.zstd" %}
+{% set version = "0.5.0" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/backports_zstd-{{ version }}.tar.gz
+  sha256: 7130d6bded652a3139fafba875283a74380a4f308f24d2361ea8761e49f837e9
+
+build:
+  skip: true  # [py>=314]
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+
+requirements:
+  build:
+    - {{ compiler('c') }}
+  host:
+    - pip
+    - python >=3.9,<3.14
+    - setuptools
+  run:
+    - python >=3.9,<3.14
+
+test:
+  imports:
+    - backports
+    - backports.zstd
+  requires:
+    - pip
+  commands:
+    - pip check
+
+about:
+  summary: Backport of compression.zstd
+  home: https://github.com/rogdham/backports.zstd
+  license: BSD-3-Clause AND MIT AND EPL-2.0
+  license_file:
+    - LICENSE.txt
+    - src/c/pythoncapi-compat/COPYING
+    - src/c/zstd/LICENSE
+
+extra:
+  recipe-maintainers:
+    - rogdham


Rogdham:
Hello @conda-forge/help-python-c, I think this PR is ready for review.

To be transparent, I am unsure about the vendoring of `libzstd`: for example the C extension builds a binary file including code from `libzstd`:
```
$ ldd lib/python3.12/site-packages/backports/zstd/_zstd.cpython-312-x86_64-linux-gnu.so
	linux-vdso.so.1 (0x00007f505d921000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f505d63c000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f505d923000)
```

But I think a dynamic link to use `libzstd` at runtime would require changes upstream to support it. I am wondering if this is necessary and if the extra work is worth it.

Thank you for your inputs!

Rogdham:
Version 1.0.0 added support for dynamic linking. This is now added to the PR! Added `zstd` as a dependency as a result.

bollwyvl:
Thanks for starting this, starting to see it required downstream.

I know it's a bit of trouble, and somewhat uncommon for other package management tools, but it would be very nice indeed if this installed a no-op package on python 3.14+ so that downstream `noarch: python` feedstocks could depend on this without multiplying their builds.

If you'd like, I can make a sub-pr to propose a path forward... as it will use a bunch of selectors, I'd likely move to the newer `recipe.yaml` which eschews magic comments for a more declarative jinja DSL with `if/then/else` blocks.

Rogdham:
I was thinking about recipes having `backports.zstd` as a dependency to include it with a marker on the python version needed.

But I am unfamiliar with much of the conda ecosystem, so if you think of a better path forward, I don't have any issue with that. Feel free to suggest any changes!

bollwyvl:
> suggest any changes

I had some issues actually creating the PR against your branch, so here's another PR:
- https://github.com/conda-forge/staged-recipes/pull/31252

The linter is angry, but backport packages like this are (thankfully) pretty rare, and could be tightened up after the feedstock is created.

Of note, I am seeing:

```
 │ │ ⚠ warning Warnings:
 │ │ ⚠ warning Overdepending against python_abi
```

Could it be this could be an `abi3` build? I see PyPI upstream has builds for the full matrix, but we're always on the lookout on conda-forge. It takes a couple switches, but slashes the number of builds by a large margin. I'll try another PR with that delta, see what comes up.

If that _were_ the case, it's possible it would just need two `python_min` variants: `3.14`, built Once, and `3.10` built Once Per OS Per Architecture, which would also appease the linter.

>  marker on the python version needed.

Yeah, we don't have these formalized yet in a cross-platform- and -ecosystem way, but there is something in the works. As things stand today, it would mean adding about as much complexity as the above PR to _every_ downstream, which we try to avoid.

Rogdham:
> Could it be this could be an abi3 build?

Unfortunately, no. When creating the backport, we discussed this with @emmatyping and the conclusion was that it was too much work for uncertain results. Indeed, each version of Python needed something else to be backported with different code paths.

Moreover, I see `backports.zstd` as a package to ease the transition to `compression.zstd`, and as such to be _temporary_ (by that I mean around 5 years). It was quite a lot of work to do already, so I am trying not to extend the scope if it is not a hard requirement.

bollwyvl:
Alternatively, adding the `no-op` could be a follow-on PR for the feedstock after review/merge by the appropriate team.

Rogdham:
If I'm being honest for a minute, I would say that I don't understand much of what you are doing in #31252 but trust you to know better.

So I would suggest to move forward with your PR, I have no issue with closing this one.

@bollwyvl do you have any insights on how to proceed from here to get your PR merged?

bollwyvl:
> move forward with your PR

Sure, I'll try some more things to see if it can be simplified and better annotated based on your review over there.

> get your PR merged

Welp, i'm not a staged-recipes maintainer, but I think if we make the case here, wait a tick, and then chime in on [zulip](https://conda-forge.zulipchat.com), we could hopefully get it out soon....

bollwyvl:
I've marked that ready for review: with more tinkering, I was able to get rid of the variants file, and do everything off `python_min` and `build_platform` and a `noarch: generic` no-op package. 

This leaves the no-op package with the cosmetic `py310` in the build string, but this has no impact on installability.

Once it is a feedstock and has been migrated to 3.14, it can even tighten up a bit so it gets a more useful build string, so:

```
Build variant: backports.zstd-1.0.0-py310h6d54471_0
╭─────────────────┬──────────────────────────────────────────────────────────╮
│ Variant         ┆ Version                                                  │
╞═════════════════╪══════════════════════════════════════════════════════════╡
│ build_platform  ┆ "linux-64"                                               │
│ channel_sources ┆ "/home/conda/staged-recipes/build_artifacts,conda-forge" │
│ channel_targets ┆ "conda-forge main"                                       │
│ python          ┆ "3.10.* *_cpython"                                       │
│ python_min      ┆ "3.10"                                                   │
│ target_platform ┆ "noarch"                                                 │
╰─────────────────┴──────────────────────────────────────────────────────────╯
```

... would be more like:

```
Build variant: backports.zstd-1.0.0-py314abcd1234_1
╭─────────────────┬──────────────────────────────────────────────────────────╮
│ Variant         ┆ Version                                                  │
╞═════════════════╪══════════════════════════════════════════════════════════╡
│ build_platform  ┆ "linux-64"                                               │
│ channel_sources ┆ "/home/conda/staged-recipes/build_artifacts,conda-forge" │
│ channel_targets ┆ "conda-forge main"                                       │
│ python          ┆ "3.14.* *_cpython"                                       │
│ target_platform ┆ "noarch"                                                 │
╰─────────────────┴──────────────────────────────────────────────────────────╯
```

Thanks again, for getting this started! We wouldn't have run into until some downstream needed it, and it would have been even more of an adventure without you doing the heavy lifting upstream!

bollwyvl:
Huzzah, merged. Once that's up at [conda-forge/backports.zstd-feedstock](https://github.com/conda-forge/backports.zstd-feedstock) (could be an hour or two), I'll see about doing the additional migrations:

- adding python 3.14, and fixing the no-op package
- adding `osx-arm64`, `linux-aarch64`, `linux-ppc64le`

Thanks again!

PR #31172, opened 2025-10-03
Adding opencarp, carputils

minrk:
Checklist
- [x] Build number is 0.


minrk:

recipes/carputils/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+  build: 0
+
+package:
+  name: ${{ name | lower }}
+  version: ${{ version }}
+
+source:
+  # carputils doesn't appear to make releases on its own repo
+  # Could get version from big -inclSubmodules.zip release bundle
+  # but rattler cannot unzip the zip file
```
danielnachun, line 15: Is this failing because the `.zip` file is too big?  Or because `rattler` doesn't support `.zip` archives? Alternatively, if that doesn't work, is it possible to get the archive with something like `https://git.opencarp.org/openCARP/openCARP/-/archive/0b56f65dd32649bfd8033c5407273aa8ff006608.tar.gz`.  That works for GitHub but I don't know if the same is true on this site.

minrk: Yeah, git archive should work fine. Updated.

The tagged zip release fails with:

```
Error:   × Failed to read from zip: invalid Zip archive: Could not find EOCD
```

I'm not sure why, it unpacks fine with other tools. It's also a 300MB download to get a 4MB subdirectory, so getting it from the carputils archive is probably better, even though it means it won't get bot bumps

minrk: works fine with the archive url, thanks. Should be all set now.

recipes/carputils/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+context:
+  name: carputils
+  version: 18.1.0
+  version_xy: ${{ (version | split('.'))[:2] | join('.') }}
+  build: 0
+
+package:
+  name: ${{ name | lower }}
+  version: ${{ version }}
```
danielnachun, line 10:
```suggestion
context:
  version: 18.1.0
  version_xy: ${{ (version | split('.'))[:2] | join('.') }}
  build: 0

package:
  name: carputils
  version: ${{ version }}
```
There is recent RFC to discourage/drop usage of the `name` context since it cannot be changed without an admin request: https://github.com/conda-forge/conda-forge.github.io/issues/2712.  Our examples and recipes generators are almost all updated to remove this, but we're still trying to filter it out with recipe submissions too.

recipes/opencarp/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+context:
+  name: opencarp
+  version: 18.1.0
+  version_xy: ${{ (version | split('.'))[:2] | join('.') }}
+  build: 0
+
+package:
+  name: ${{ name | lower }}
+  version: ${{ version }}
```
danielnachun, line 10:
```suggestion
context:
  version: 18.1.0
  version_xy: ${{ (version | split('.'))[:2] | join('.') }}
  build: 0

package:
  name: opencarp
  version: ${{ version }}
```
See above comment about this.

recipes/opencarp/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+build:
+  number: ${{ build }}
+  skip: win
+
+requirements:
+  build:
+    - git
+    - if: linux
+      then:
+      - binutils
```
danielnachun: What was the reason for needing to add this?  I believe the relevant executables for `binutils` are already supposed to be pulled in automatically by `${{ compiler('c') }}` and `${{ compiler('cxx') }}`.  If it's failing without `binutils` being added explicitly it would be good to know the underlying issue in case there is a problem with our current compiler toolchain.

recipes/carputils/recipe.yaml
danielnachun: Thanks for sharing the error, at some point we should report that to the prefix-dev team in case they want to dig into it.  Since the archive URL works, that is a good compromise here because it still gets us a checksum.  Hopefully updates aren't too much of a pain in the future!

danielnachun (changes requested):

recipes/opencarp/recipe.yaml
minrk: I removed it and it appears to be fine.

This is an old recipe I've been building myself outside conda-forge for a long time, so I guess it's a leftover. My guess is it was some hardcoded executables, because the base names of binutils are _not_ available by default, only the $name-$platform variants, while `binutils` has the regular names. But it seems no longer necessary, so removed!

recipes/opencarp/build.sh
```diff
@@ -0,0 +1,17 @@
...
+cmake \
+  ${CMAKE_ARGS} \
+  -DDLOPEN=ON \
+  ..
+
+make VERBOSE=1 -j"${CPU_COUNT:-1}"
+# cmake --install by default default installs duplicate petsc that doesn't work (?!)
+for component in core tool; do
+  cmake --install . --component $component
+done
```
danielnachun:
```suggestion
export OPENCARP_DIR="$PREFIX"
cmake -S ${SRC_DIR} -B _build \
  ${CMAKE_ARGS} \
  -DDLOPEN=ON

cmake --build _build -j"${CPU_COUNT:-1}"
# cmake --install by default default installs duplicate petsc that doesn't work (?!)
for component in core tool; do
  cmake --install _build --component $component
done
```
We're trying to standardize our CMake calls to be generator agnostic and to be more concise - this should achieve that without changing the build logic.

recipes/opencarp/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+  build:
+    - git
+    - ${{ compiler('c') }}
+    - ${{ compiler('cxx') }}
+    - ${{ stdlib('c') }}
+    - cmake
+    - make
+    - gengetopt
+    - pkg-config
+    - python 3.12.*
```
danielnachun:
```suggestion
    - python ${{ python_min }}
```
If you are setting the python version here to avoid exporting a run dependency on each Python version, it might be better to just use `${{ python_min }}` here so that whenever Python 3.12 hits EOL this doesn't have to be manually updated.  If Python 3.10 is too old, then you can set `python_min` in the context and the our bots should hopefully handle that in the future.

danielnachun (changes requested):

minrk:
@danielnachun thanks again! Should be all set now.

minrk:
@danielnachun anything left to do here?

danielnachun (approved):

PR #31188, opened 2025-10-08
Add dcm4che-tools

Dhananjhay:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/dcm4che-tools/build.sh
+++ b/recipes/dcm4che-tools/build.sh
@@ -25,4 +25,3 @@ ls ${PREFIX}/libexec/dcm4che-tools/bin | grep -v ".bat" | xargs -I % bash -c "ex
 
 cd dcm4che-assembly
 mvn license:download-licenses -Dgoal=download-licenses
-exit 1


Dhananjhay:
@conda-forge/staged-recipes

Dhananjhay:
@conda-forge/help-java

danielnachun:
I've pushed a version of this that builds the recipe from source as per our requirements.  Two things need to be resolved before we can merge this:
- Can you explain the motivation behind the patch you added to uncomment those class declarations in the `.properties` file?  Is this a bug fix or is there some other reason.  While patching is sometimes needed in conda-forge, we like to have a clear explanation of the reason for the patch.
- This package should be `noarch: generic` because it does not itself contain any compiled code.  However it does vendor `libopencv_java`, the JNI interface for Java applications to access OpenCV.  Unfortunately, our current OpenCV package does not build the JNI interface.  This can and probably should be changed.  The other detail I'm unclear about is whether the JDK will know to search the right paths for the native library.  There is an option `-Djava.library.path` that could be added to the wrapper script which would specify this, but I don't know if this is actually needed or if the search paths are already in the library.  I need to confer with other maintainers about this as it's a somewhat complex issue and I'll follow up with more details later.

Dhananjhay:
Hi @danielnachun, thank you for your help with this! Yes, the reason we uncommented those lines was because otherwise we weren't able to retrieve enhanced MR (a DICOM flavor) from a DICOM server.

danielnachun:
Was this issue reported upstream?  I think it's okay for us to include the patch here for now but we don't want this issue to go unreported to the original developers.  We'd like to make sure it is ultimately fixed in a new release.

Dhananjhay:
That's a great suggestion and I realize it was an oversight on my end. Thank you for that! I've reported this as an issue on the upstream repository. Hopefully it will be patched in the next release.

Dhananjhay:
@conda-forge/help-java

danielnachun:
Thanks for reporting the issue.  I'm still looking into the OpenCV Java question. I may try to make a PR and see if we can build it and report back here how it goes!

PR #31205, opened 2025-10-09
Initial recipe for openshift-installer

phreed:
Checklist
- [x] Build number is 0.


recipes/okd-install/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+  number: 0
+  skip:
+    - win
+  script:
+    - go-licenses save ./cmd/openshift-install --save_path ./library_licenses/
+    - export SOURCE_GIT_COMMIT=${{ version }}
+    - export BUILD_VERSION=${{ version }}
+    - export TAGS="include_gcs include_oss containers_image_openpgp"
+    - export OUTPUT="$PREFIX/bin/openshift-install"
+    - unset LDFLAGS
```
xhochy: Why are you unsetting this?

recipes/okd-install/conda_build_config.yaml
```diff
@@ -0,0 +1,10 @@
+# This file contains build configuration for the okd-install recipe
```
xhochy: Please delete this files, the compiler version is meaningless and the target platform should be set via skips

xhochy (changes requested):

recipes/okd-install/recipe.yaml
phreed: The LDFLAGS for the C build process were contaminating the linking for the GO compiler.
`go build -gcflags "" -ldflags "-s -w" -o ../../bin/linux_amd64/cluster-api-provider-aws "$path";`
See the `-ldflags "-s -w"` that is coming from the LDFLAGS.

recipes/okd-install/conda_build_config.yaml
phreed: I want to be able to trigger packages for  linux_aarch64, and linux linux_ppc64le and thought this might be the place to do that. I see how to do it once there is a proper feedstock but I do not see how to do it from staged-recipes.

recipes/okd-install/recipe.yaml
phreed: Odd, it builds on my linux machine but not with github actions.
Here is the error:
```
building openshift-install
 │ │ + echo 'building openshift-install'
 │ │ + go build -modcacherw -buildmode=pie -trimpath -gcflags '' -ldflags '-Wl,-O2 -Wl,--sort-common -Wl,--as-needed -Wl,-z,relro -Wl,-z,now -Wl,--disable-new-dtags -Wl,--gc-sections -Wl,--allow-shlib-undefined -Wl,-rpath,$PREFIX/lib -Wl,-rpath-link,$PREFIX/lib -L$PREFIX/lib -X github.com/openshift/installer/pkg/version.Raw=4.19.15 -X github.com/openshift/installer/pkg/version.Commit=4.19.15 -X github.com/openshift/installer/pkg/version.defaultArch=amd64 -s -w' -tags 'include_gcs include_oss containers_image_openpgp release' -o $PREFIX/bin/openshift-install ./cmd/openshift-install
 │ │ # github.com/openshift/installer/cmd/openshift-install
 │ │ flag provided but not defined: -Wl,-O2
```
As you can see the `-Wl,-O2` LDFLAGS are causing a problem.
I will not restore the fix (via `unset LDFLAGS`) until you have a chance to take a look and provide your feedback.

danielnachun: This is why for Go recipes we shouldn't use custom build scripts from upstream but instead directly invoke the `go build` command whenever possible.  The problem is that the upstream build script just blindly grabs whatever is defined by `LDFLAGS`, which in the case of our compiler packages are flags intended for C/C++/Fortran compilers, not Go.  Instead, just replace your call to `./hack/build.sh` with the standard call we use for building Go packages:
```
go build -o=${PREFIX}/bin/${PKG_NAME} -ldflags="-s -w -X github.com/openshift/installer/pkg/version.Raw={{ version }} -X github.com/openshift/installer/pkg/version.Commit={{ version }}" --tags "include_gcs include_oss containers_image_openpgp" ./cmd/openshift-install
```
That makes it much more clear to maintainers what exactly is being called.

phreed: The package hides a bunch of 'go build' commands in the  'hack/build.sh' script which includes the use of a Makefile. I have added the 'go build' commands from the 'Makefile' which is a fancy script so the reasoning to move into the build.sh should apply.

phreed:
@conda-forge/staged-recipes I have not heard from you in a while.

recipes/okd-install/build.sh
```diff
@@ -0,0 +1,103 @@
...
+go build \
+    -C cluster-api/cluster-api \
+    -gcflags "$GCFLAGS" -ldflags "-s -w" \
+    -o "$CLUSTER_API_BIN_DIR/cluster-api" \
+    ./vendor/sigs.k8s.io/cluster-api
+
+# I presume we skip the ENVTEST at least until we have a proper kub-apiserver.
+
+# Zip every binary in the folder into a single zip file (if any exist)
+zip -j1 "cluster-api.zip" $CLUSTER_API_BIN_DIR/*
```
danielnachun:
```suggestion
# Build cluster API binaries first
. ${SRC_DIR}/build-cluster-api.sh"
make -C cluster-api all
copy_cluster_api_to_mirror
```

After reviewing the upstream code base much more carefully, I think we can safely run these lines from the upstream `build.sh` script.  Unlike the main `go build` command which blindly pulls in `LDFLAGS`, the `Makefile` in `cluster-api` does **not** do this, making the whole recipe much simpler.

danielnachun (changes requested):
The only other change here is remove the `readme.md` - it will be ignored when the feedstock is generated and we try to keep the git commit history for staged-recipes free of redundant files.

recipes/okd-install/build.sh
```diff
@@ -0,0 +1,23 @@
...
+set -euo pipefail
+
+# OpenShift Installer Build Script
+# https://github.com/openshift/installer/blob/main/hack/build.sh
+#
+go-licenses save ./cmd/openshift-install --save_path ./library_licenses/
+
+# Build cluster API binaries first
+# Unset C-linker LDFLAGS so they don't interfere with Go's cgo/make build
+unset LDFLAGS
```
danielnachun, line 11:
```suggestion
# Tweak Makefile so it does not try to use `LDFLAGS` for the `-ldflags` argument
sed -i 's/LDFLAGS/GO_LDFLAGS/' cluster-api/Makefile
export SKIP_ENVTEST=y
```
I came up with a better fix that still allows us to use the C `LDFLAGS` and set the proper Go `-ldflags` argument.  The environment variable I added should also fix your current failure.

danielnachun (changes requested):

phreed:
I no longer care about this package.

PR #31252, opened 2025-10-20
backports.zstd (with noarch: generic no-op package for python 3.14+)

bollwyvl:
Checklist
- [x] Build number is 0.

References:
- alternative to (and contains #31124)

### DIFF
--- a/recipes/backports.zstd/conda_build_config.yaml
+++ b/recipes/backports.zstd/conda_build_config.yaml
@@ -1,8 +1,0 @@
-# This would ideally use CFEP-25 `python_min`, but `staged-recipes` doesn't
-# yet build python 3.14 by default.
-is_noop:
-  - "false"
-  - "true"
-# python_min:
-#  - "3.10"  # this would need to be updated yearly
-#  - "3.14"
--- a/recipes/backports.zstd/recipe.yaml
+++ b/recipes/backports.zstd/recipe.yaml
@@ -19,20 +19,16 @@ build:
 # on `backports.zstd` without having their _own_ variants, this builds multiple
 # `outputs` with the same name, but different build strings, per upstream release:
 #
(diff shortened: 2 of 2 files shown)


Rogdham:
I must say that I don't understand quite a few things, but it's probably on me!

- Why not using `match(python, ">=3.14")` only on the first package and `match(python, "<3.14")` on the second? In which case the `>= 3.14` could be removed on `host` & `run`?
- What is the use of `is_noop`? And why the definition based on OS and architecture?
- To make the intent clearer, would it be possible to remove the `host` and `tests` keys of the second package?

bollwyvl:
> Why not using match(python, ">=3.14") only on the first package and match(python, "<3.14") on the second?> What is the use of is_noop? And why the definition based on OS and architecture?

We need _something_ to delineate a "special" package, which we only want to build Once, Ever, Per Upstream Release. By convention, `conda-forge` does these kinds of builds on `linux-64`.

It's possible these could be condensed to the linter-appeasing:

```yaml
python_min:
 - 3.10
 - 3.14
```

With more of the logic done off that. I'll try that as a separate PR.

> In which case the >= 3.14 could be removed on host & run?
>
> To make the intent clearer, would it be possible to remove the host and tests keys of the second package?

This is a `noarch: python` package which would work on e.g. `python 3.15.0a0`.

bollwyvl:
> separate PR

yeah, that didn't work out. as noted on https://github.com/conda-forge/staged-recipes/pull/31124#issuecomment-3422542789 , using `python_min` might have to wait for an actual feedstock.

bollwyvl:
@conda-forge/help-python @conda-forge/help-python-c @Rogdham ready for review :clamp:

bollwyvl:
Thank you!

PR #31265, opened 2025-10-22
Add Khiops Python library package recipe

popescu-v:
Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/khiops/meta.yaml
+++ b/recipes/khiops/meta.yaml
@@ -1,4 +1,3 @@
-{% set python_min = "3.8" %}
 {% set name = "khiops" %}
 {% set version = "10.3.2.1" %}
 


tramora:
I agree to be listed as a maintainer

popescu-v:
@conda-forge/help-python, ready for review

recipes/khiops/meta.yaml
```diff
@@ -0,0 +1,62 @@
+{% set python_min = "3.8" %}
```
ocefpaf: Only set `python_min` if the minimum required version is higher than the conda-forge default.

```suggestion
```

popescu-v: Ok. Changed accordingly.

recipes/khiops/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+  script: {{ PYTHON }} -m pip install . --no-deps --ignore-installed --no-cache-dir --no-build-isolation -vvv
+  number: 0
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools
+  run:
+    - python >=3.8
```
ocefpaf:
```suggestion
    - python >={{ python_min }}
```

popescu-v: Actually, the `khiops` Python library supports Python >= 3.8, as documented in the `setup.py` file: https://github.com/KhiopsML/khiops-python/blob/888280dc01c90128e02761363cd53d41386a47a2/setup.py#L33.
Hence, we need to make sure the package can be installed on Python >= 3.8.
Can this constraint be specified otherwise that explicitly setting `python >=3.8`?

ocefpaf: Yep. But conda-forge doesn't support it. That is why we have a global min.

popescu-v: Thank you for the explanation. Changed accordingly.

PR #31337, opened 2025-10-30
Add recipe for Briefcase

freakboy3742:
Adds a recipe for BeeWare's Briefcase app packaging tool (v0.3.25, the current release).

Checklist
- [x] Build number is 0.

### DIFF
--- a/recipes/briefcase/meta.yaml
+++ b/recipes/briefcase/meta.yaml
@@ -10,7 +10,7 @@ source:
   sha256: 0d3ca50714da8bed30e80ea33f322e6b8be7e80a1fb651814d98561aa269de72
 
 build:
-  skip: true  # [py<310]
+  noarch: python
   entry_points:
     - briefcase = briefcase.__main__:main
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
@@ -18,12 +18,12 @@ build:
 
 requirements:
   host:
-    - python
+    - python {{ python_min }}
     - setuptools ==80.9.0
     - setuptools-scm ==9.2.0
     - pip
   run:
-    - python
+    - python >={{ python_min }}
     - packaging >=24.2
     - pip >=24.3
     - setuptools >=60
@@ -38,8 +38,11 @@ requirements:
     - python-dateutil >=2.9.0.post0
     - httpx >=0.20,<1.0
     - rich >=12.6,<15.0
-    - tomli >=2.0,<3.0  # [py<=310]
+    - tomli >=2.0,<3.0
     - tomli-w >=1.0,<2.0
+    - __unix  # [unix]
+    - __win   # [win]
+    - __osx   # [osx]
 
 test:
   imports:
--- a/recipes/briefcase/conda-forge.yml
+++ b/recipes/briefcase/conda-forge.yml
@@ -1,0 +1,4 @@
+noarch_platforms:
+  - linux_64
+  - win_64
+  - osx_64


freakboy3742:

I believe this recipe is ready for review.

freakboy3742:
@conda-forge/help-python I believe this recipe is ready for review.

recipes/briefcase/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - truststore >=0.10.1
+    - cookiecutter >=2.6.0,<3.0
+    - dmgbuild >=1.6.4,<2.0  # [osx]
+    - gitpython >=3.1,<4.0
+    - platformdirs >=2.6,<5.0
+    - psutil >=5.9,<8.0
+    - python-dateutil >=2.9.0.post0
+    - httpx >=0.20,<1.0
+    - rich >=12.6,<15.0
+    - tomli >=2.0,<3.0  # [py<=310]
```
ocefpaf: If we allow this to be installed in py>=310 we can make this noarch. It is preferred b/c we save on CI time. We can drop it when droppping py<=310.

recipes/briefcase/meta.yaml
```diff
@@ -0,0 +1,63 @@
...
+  run:
+    - python
+    - packaging >=24.2
+    - pip >=24.3
+    - setuptools >=60
+    - wheel >=0.37
+    - python-build >=0.10
+    - truststore >=0.10.1
+    - cookiecutter >=2.6.0,<3.0
+    - dmgbuild >=1.6.4,<2.0  # [osx]
```
ocefpaf, line 34: Let's make it OS-noarch. See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

freakboy3742: > Let's make it OS-noarch. See https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies

Ok, but... how? If you check the commit history, I tried making the package noarch; but if the `[osx]` selector is in place, it fails linting because of having a selector; but if I add the `conda-forge.yml` addition to set `noarch_platform`, it fails linting because there are modified files outside the recipes directory.

The tomli dependency is also a complication; it's not a *problem* having tomli installed on Py3.11+ - but it's unnecessary code.

Is this a situation where linting is *expected* to fail? Or am I missing something here?

ocefpaf: I'll sent a commit to implement that. It should be in the docs, but I'm too "in the known" to know how to do it. Maybe it is not as clear as we expected it to be.

ocefpaf: > Ok, but... how? I

Please read step 1 and 2 from the docs I linked above and see the changes here. Let me know why it was difficult to implement those changes so we can improve the docs.

> The tomli dependency is also a complication; it's not a problem having tomli installed on Py3.11+ - but it's unnecessary code.

It is fine in conda-forge. It is likely that another dependency already does that. We are all hoping to drop py310 to get rid of this.

freakboy3742: > > Ok, but... how? I
> 
> Please read step 1 and 2 from the docs I linked above and see the changes here. Let me know why it was difficult to implement those changes so we can improve the docs.

I read those steps, and - as I said in my last reply, and the commit history on this PR indicates - I tried that approach.

The difficulty was caused because the linked instructions say to "Use conda-forge.yml's noarch_platforms" option, but, there's no documentation on what that *means*. There is a `conda-forge.yml` in the root of the repository, and if you modify that file, the change will pass local linting, but then fails linting on the PR. 

In the discussion of this feature, every file reference talks about `recipe/meta.yml`; but the reference to `conda-forge.yml` is just `conda-forge.yml`, so it's not at all clear that adding a `condo-forge.yml` is a valid option *in* the recipe folder.

ocefpaf: >  it's not at all clear that adding a condo-forge.yml is a valid option in the recipe folder.

Indeed I guess that is the most obscure part of the docs there. We are so used that everything goes into the recipe directory in staged recipes that we forgot to spell it out in the docs. I'll send a PR amending that soon. Thanks for the feedback.

PR #31407, opened 2025-11-06
Add janet package manager (jpm) 1.1.0

tdejager:
Checklist
- [x] Build number is 0.

This adds the `jpm` package manager for the janet language, which is also on conda-forge. The recipe contains a number of tests to test that the package manager works as expected.

### DIFF
--- a/recipes/jpm/build.sh
+++ b/recipes/jpm/build.sh
@@ -17,22 +17,36 @@ manpath="${PREFIX}/share/man/man1"
 
 target_platform="${target_platform:-$(uname | tr '[:upper:]' '[:lower:]')}"
 
-cc_bin="${CC:-cc}"
-cxx_bin="${CXX:-c++}"
-ar_bin="${AR:-ar}"
-
+# The recipe declares the conda-forge C and C++ compiler packages for host+run,
+# so CC/CXX/AR must point at those wrapped toolchains. Fail fast with clear
+# guidance if something is misconfigured.
+if [[ -z "${CC:-}" ]]; then
+  echo "error: CC is unset. The conda-forge compiler('c') package should set it." >&2
+  exit 1
+fi
+cc_bin="${CC}"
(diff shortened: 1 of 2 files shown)


tdejager:
Let me cross-post my zulip question here:

jpm is a bit strange in that you can use it to compile a janet program into a standalone executable and it needs compilers for it. Also the configuration is statically bootstrapped into jpm at build time. Thats why compilers are added both to the host dependencies as well as the runtime dependencies. I'd rather not have compilers in there at all however.

So what happens is:
```
          ┌────────────────────┐
          │   your_script.janet│
          └─────────┬──────────┘
                    │
                    │ jpm quickbin
                    ▼
          ┌────────────────────┐
          │  Janet compiler    │
          │  (janet  + jpm)   │
          └─────────┬──────────┘
                    │
                    │ compiles .janet source
                    │ to Janet bytecode (.jbc)
                    ▼
         ┌─────────────────────┐
         │  Bytecode (.jbc)    │
         └─────────┬───────────┘
                   │
                   │ embedded into
                   │ a C stub (quickbin.c)
                   ▼
        ┌─────────────────────────┐
        │  C compiler (cc/clang)  │
        └──────────┬──────────────┘
                   │
                   │ compiles & links:
                   │   • quickbin.c (stub)
                   │   • libjanet (dylib for conda-forge, statically normally)
                   │   • embedded bytecode
                   ▼
         ┌────────────────────────┐
         │   Standalone binary       │
         │   (your_script)        │
         │   • no external Janet  │
         │   • bundled VM + code  │
         └────────────────────────┘
```
So we need to compile it at some point. And because of the static configuration as you can see in the feedstock at: https://github.com/conda-forge/staged-recipes/pull/31407/files#diff-9c12f8a0ba264ba046534782bbf189358655cc78d9147c5e3116e6d5b1105332

I added compilers to the host-section. Now I was thinking of making the compilers run-exports in that case, so in a conda environment you always get the correct version. Anyone have an opinion?

PR #31424, opened 2025-11-07
add asciinema-agg

ytausch:
Checklist

- [x] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged). -> rust package
- [x] Build number is 0.

### DIFF
--- a/recipes/asciinema-agg/recipe.yaml
+++ b/recipes/asciinema-agg/recipe.yaml
@@ -20,7 +20,11 @@ build:
       CARGO_PROFILE_RELEASE_STRIP: symbols
       CARGO_PROFILE_RELEASE_LTO: fat
     content:
-      - cargo auditable install --locked --no-track --bins --root ${{ PREFIX }} --path .
+      - if: unix
+        then:
+          - cargo auditable install --locked --no-track --bins --root ${{ PREFIX }} --path .
+        else:
+          - cargo auditable install --locked --no-track --bins --root %LIBRARY_PREFIX% --path .
       - cargo-bundle-licenses --format yaml --output ./THIRDPARTY.yml
 
 


pavelzw:
i am okay being maintainer

ytausch:
@conda-forge/help-rust ready for review!

recipes/asciinema-agg/recipe.yaml
```diff
@@ -0,0 +1,56 @@
...
+build:
+  number: 0
+
+  script:
+    env:
+      CARGO_PROFILE_RELEASE_STRIP: symbols
+      CARGO_PROFILE_RELEASE_LTO: fat
+    content:
+      - cargo auditable install --locked --no-track --bins --root ${{ PREFIX }} --path .
+      - cargo-bundle-licenses --format yaml --output ./THIRDPARTY.yml
```
danielnachun, line 28:
```suggestion
    env:
      CARGO_PROFILE_RELEASE_STRIP: symbols
      CARGO_PROFILE_RELEASE_LTO: fat
    content:
      - if: unix
      - then:
        - cargo auditable install --locked --no-track --bins --root ${{ PREFIX }} --path .
      - else:
        - cargo auditable install --locked --no-track --bins --root %LIBRARY_PREFIX% --path .
      - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
```
@xhochy @ocefpaf @pavelzw This recipe follows our example but I noticed that as the example is currently written, we would install to `%PREFIX%` on Windows instead of `%LIBRARY_PREFIX%`.  My suggestion above would change that, but I'm not sure it's actually needed.

I know from experience that both `%PREFIX%` and `%LIBRARY_PREFIX%` are in `%PATH%` so the executables will still be found.  Historically `%LIBRARY_PREFIX%` is used for Unix-style software so that it can have the standard POSIX file hierarchy (`bin`, `lib`, etc.), but does that matter any more for conda-forge, especially in the era of pixi?  If not then we could simplify many cross-platform compiled recipes as we wouldn't have to condition on Windows if the only difference was `%PREFIX%` vs `%LIBRARY_PREFIX%`.  Let me know what you think.

danielnachun (commented):
@ytausch Thank you for following the example recipe for Rust.  This recipe looks great but there is a question I've asked some other reviewers/maintainers for feedback about regarding the installation paths in our example.  The suggestion I made will only be relevant if it turns out that `%PREFIX%` vs `%LIBRARY_PREFIX%` actually matters any more on Windows.  If it does not then we'll be ready to merge as-is.

recipes/asciinema-agg/recipe.yaml
xhochy: We should write these things into `%LIBRARY_PREFIX%`. This has been a past decision that everything MinGW-related goes into PREFIX and everything else (which is the majority) will be in `Library/`.

pavelzw: i'll create a pr in the conda-forge docs 👍🏻

pavelzw: https://github.com/conda-forge/conda-forge.github.io/pull/2651

ytausch: fixed!

ytausch:
@danielnachun

PR #31432, opened 2025-11-08
Add mgclient

MementoRC:
Checklist

- [x] Build number is 0.


MementoRC:
@traversaro This recipe is ready. The recipe make sure the static library is not installed, but it is needed in order to run the tests. From past experience, running the tests help catch broken packages in feedstocks

MementoRC:
@traversaro This recipe is ready

MementoRC:
@conda-forge/help-c-cpp Hello, ready for review

recipes/mgclient/build.bat
```diff
@@ -0,0 +1,30 @@
...
+set BUILD_PREFIX=%_BUILD_PREFIX:D:=/d/%
+set SRC_DIR=%_SRC_DIR:D:=/d/%
+set RECIPE_DIR=%_RECIPE_DIR:D:=/d/%
+
+set MSYSTEM=MINGW64
+set MSYS2_PATH_TYPE=inherit
+set MSYS2_ARG_CONV_EXCL="*"
+set CHERE_INVOKING=1
+bash -lce "./conda_build.sh"
+if errorlevel 1 exit 1
```
traversaro: The library seems to support MSVC (see https://github.com/memgraph/mgclient/blob/55cc7f5fa291d649e4965a08751a4e05ed75696a/README.md?plain=1#L124). Is there any reason why can't we just compiled with MSVC and keep the recipe simplr?

traversaro:
> @traversaro This recipe is ready

Deeply sorry for the review delay, the recipe seems nice, but I wonder if we can simplify it a bit.

recipes/mgclient/recipe.yaml
```diff
@@ -0,0 +1,63 @@
...
+    - cmake >3.10
+    - ninja
+    - if: unix
+      then: sed
+      else:
+        - m2-findutils
+        - m2-sed
+        - m2-which
+  host:
+    - openssl
```
traversaro, line 31: The package installs a shared library, perhaps we can add an appropriate run_exports ?

recipes/mgclient/build.bat
MementoRC: No disrespect, if I am to be the long term maintainer, t is much simpler to have just one file related to that OS, not to mention much more pleasant and mindful. Dealing with that OS always generates irritations

But of course, if there are technical downstream considerations or conda-forge users inconveniences that I may have missed, please let me know

recipes/mgclient/recipe.yaml
MementoRC: Thanks, I always forget :(

MementoRC:
@traversaro Added the run_exports and made a few simplifications, though holding back on the 'only maintain Unix-style build process' personal preference, if no user implications

traversaro:
> if no user implications

It may not be user implications, but it make the recipe much harder to review. I understand preferring not to use autotools clang or other similar things, but in this case using CMake and Visual Studio reduces the build.bat to like ~4 lines. Furthermore, building a project with CMake like this disregard the `%CMAKE_ARGS%`, that is used to easily pass coherent options to all packages built with CMake.

recipes/mgclient/build.bat
```diff
@@ -0,0 +1,30 @@
+@echo off
+setlocal enabledelayedexpansion
+
+echo source D:/Miniforge/etc/profile.d/conda.sh       > conda_build.sh
```
traversaro: This may not be due for local builds or build on self-hosted runners.

MementoRC: Why this should be a consideration?

traversaro:
Thanks!

PR #31530, opened 2025-11-21
add napari-omero-downloader-cci

leclercsimon74:
This PR adds the conda-forge recipe for the napari plugin
**napari-omero-downloader-cci**.

- Source: https://github.com/CCI-GU-Sweden/napari-omero-downloader-cci
- Version: 0.3.2
- Pure Python / `noarch: python`
- Depends on `omero-py` (from conda-forge) to provide Ice bindings
- Tested locally with napari-base + omero-py

Recipe maintainer:
- @leclercsimon74

### DIFF
--- a/.ci_support/build_all.py
+++ b/.ci_support/build_all.py
@@ -290,7 +290,7 @@ def build_folders_rattler_build(
     specs = OrderedDict()
     for f in config.exclusive_config_files:
         specs[f] = conda_build.variants.parse_config_file(
-            os.path.abspath(os.path.expanduser(os.path.expandvars(f))), config
+            os.path.abspath(os.path.expanduser(os.path.expandvars(f))), config, loader=yaml.SafeLoader
         )
 
     variants = list(Path(recipes_dir).glob(f"**/conda_build_config.yaml")) \
@@ -302,7 +302,7 @@ def build_folders_rattler_build(
         )
     if variants and os.path.isfile(variants[0]):
         specs[variants[0]] = conda_build.variants.parse_config_file(
-            os.path.abspath(variants[0]), config
+            os.path.abspath(variants[0]), config, loader=yaml.SafeLoader
         )
 
     # Combine all the variant config files together


leclercsimon74:
All checks are passing, and I’m happy to adjust anything you’d like to see changed. Thanks for reviewing!

recipes/napari-omero-downloader-cci/meta.yaml
```diff
@@ -0,0 +1,51 @@
+{% set name = "napari-omero-downloader-cci" %}
+{% set version = "0.3.2" %}  # or whatever you are releasing
+{% set python_min = "3.9" %}
```
synapticarbors: The minimum supported version of python on conda-forge is 3.10, so you should likely remove this line and just let it use the default (3.10). I did notice that you restrict the runtime version of python to `<3.10`. Is this necessary? I don't see that in the upstream `pyproject.toml`.

recipes/napari-omero-downloader-cci/meta.yaml
```diff
@@ -0,0 +1,51 @@
...
+
+about:
+  home: https://github.com/CCI-GU-Sweden/napari-omero-downloader-cci
+  summary: "A plugin that allows napari to connect to the OMERO CCI server to visualize and download image data."
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - leclercsimon74  # GitHub handle
```
synapticarbors: You can remove comments like this.

recipes/napari-omero-downloader-cci/meta.yaml
leclercsimon74: Thanks for the comment.
Omero-py and its dependency (infamous zero-ice) works only on python 3.9 (on a window machine):

```
LibMambaUnsatisfiableError: Encountered problems while solving:
  - package omero-py-5.6.0-py_0 requires zeroc-ice36-python, but none of the providers can be installed
```

zero-ice put out the version 3.7, which work with python 3.12 ([link](https://zeroc.com/ice/downloads/3.7/python) 2 years ago). But the conda-forge channel version of omero-py is still working with 3.6.

On pip, I am assuming that the users of the package already did 'conda install omero-py', which put them in the correct version of python.

I will investigate more how easy/difficult it is to just pip zeroice and omero-py on Window. Which will require to fully rewrite my yaml file (very likely).

leclercsimon74:
Let me know if something needs to be changed here!

recipes/napari-omero-downloader-cci/meta.yaml
leclercsimon74: Done

PR #31550, opened 2025-11-25
Adding fvdb-core

swahtz:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/fvdb-core/meta.yaml
+++ b/recipes/fvdb-core/meta.yaml
@@ -47,6 +47,7 @@ requirements:
     - cuda-nvrtc-dev
     - cuda-nvtx-dev
     - libcublas-dev
+    - libcudnn-dev
     - libcusolver-dev
     - libcusparse-dev
     - libpng


swahtz:

This pytorch plugin requires CUDA 12.9.  In the current 12.6 ci_support configuration, the conda environments for the build will fail to resolve and so I had to modify the ci_support config to be able to run properly which you can see running here:

https://github.com/conda-forge/staged-recipes/runs/56342507840

And I reverted that change for the current state of the PR.  Unfortunately the machine it's running on seems to run out of memory partway through the build, though I can confirm the recipe runs successfully on my local machine with `build-locally.py`.

Thank you

swahtz:

swahtz:

This pytorch plugin requires CUDA 12.9. In the current 12.6 ci_support configuration, the conda environments for the build will fail to resolve and so I had to modify the ci_support config to be able to run properly which you can see running here:

https://github.com/conda-forge/staged-recipes/runs/56342507840

And I reverted that change for the current state of the PR. Unfortunately the machine it's running on seems to run out of memory partway through the build, though I can confirm the recipe runs successfully on my local machine with build-locally.py.

Thank you

recipes/fvdb-core/conda_build_config.yaml
```diff
@@ -0,0 +1,7 @@
+# conda_build_config.yaml
+c_compiler_version:
+  - 13
+cxx_compiler_version:
+  - 13
+pytorch:
+  - 2.8.0
```
flferretti: The last conda-forge PyTorch that still has `cuda126_*` builds is 2.7.1. From 2.7.1 onwards, the GPU builds are all `cuda129_*`. Therefore you may want to decrease this to 2.7.1 (or remove the pin) and restore it to 2.8.0 in the feedstock with CUDA12.9 available

swahtz: Hi @flferretti thanks for having a look.

In general for our framework, the minimum version of CUDA we wanted to support was 12.8 (and PyTorch 2.8.0).  Trying it out, I don't believe our software can build in CUDA 12.6.  Is there a better way to specify that PyTorch 2.8.0 and CUDA 12.8 are the minimum versions this recipe should be built for?

flferretti: Could you try for now to unpin Pytorch and see if this builds correctly? I believe the environment will be solved with 2.7.1.

Once the feedstock gets published, you can trigger a migration to CUDA 12.9 and eventually pin the PyTorch version there

swahtz: Sure I've given that a try.  Though I do expect the build to fail because I have tried to build a PyTorch 2.7.1 (CUDA 12.6) environment and build it locally but the build fails because of CUDA features we need in 12.8 and above (hence our minimum requirement I'm trying to pin).

swahtz:
> I'm sorry I didn't notice this at first. Are you perhaps missing a build.sh file? Otherwise, since you're downloading the tarball from pypi maybe you wanted to install this with pip?

No the sdist tarball should have the build.sh script, if you see in this previous run (where I switched to be able to use CUDA 12.9 and pytorch 2.8.0), the build was running correctly, it just happened to run out of system RAM on the runner:

https://github.com/conda-forge/staged-recipes/runs/56342507840

Currently we seem to be unable to resolve the conda environment after I unpinned pytorch:

```
└─ pytorch ==2.7,(>=2.7.1,<2.8.0a0) *cuda* does not exist (perhaps a typo or a missing channel).

```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - make
+    - pkg-config
+    - git
+    - procps-ng  # provides 'free'
+    - coreutils  # provides 'nproc', 'awk', etc.
+  host:
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch {{ pytorch }} *{{ torch_proc_type }}*
```
flferretti: We should keep both entries: `pytorch` picks up the globally pinned version from conda-forge, and `pytorch =*={{ torch_proc_type }}*` constrains the build string to the desired CPU/CUDA variant
```suggestion
    - pytorch
    - pytorch =*={{ torch_proc_type }}*
```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+  skip: true  # [py<310 or cuda_compiler_version == "11.8" or cuda_compiler_version == "None" or win or osx]
+  ignore_run_exports:
+    - python
+    - numpy
+    - libcusolver
+    - libpng
+    - libcublas
+    - libcusparse
+    - cuda-nvtx
+  script: ./build.sh install verbose --cuda-arch-list="7.5;8.0;9.0;10.0;12.0+PTX"
```
flferretti: In general, it would be better to vendor a build script inside the recipe, so it is clear what is happening during the build and we can achieve better reproducibility. It can be replaced by a small `build.sh` or an inline script in which you just set TORCH_CUDA_ARCH_LIST and run `$PYTHON -m pip install . --no-deps --no-build-isolation -vv`

Looking at the current [CMakeLists.txt](https://github.com/openvdb/fvdb-core/blob/main/CMakeLists.txt), CPM is used to download NanoVDB at configure time. For conda-forge we try to avoid fetching sources during the build and instead rely on packaged dependencies. Could you patch [get_nanovdb.cmake](https://github.com/openvdb/fvdb-core/blob/main/src/cmake/get_nanovdb.cmake) so that it finds and uses the nanovdb library from conda-forge rather than downloading NanoVDB via CPM? Then you can add `nanovdb` to `host` dependencies

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/fvdb_core-{{ version }}.tar.gz
+  sha256: f6ec0d5e6a2b601632720ed0be8aa7c8ba6f2d9bd63fac477cef2cfd088b5f16
+
+build:
+  skip: true  # [py<310 or cuda_compiler_version == "11.8" or cuda_compiler_version == "None" or win or osx]
+  ignore_run_exports:
+    - python
```
flferretti: We normally want the standard Python ABI pinning. Is there any specific reason why this is needed?
```suggestion
```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - numpy
+    - libcusolver
+    - libpng
+    - libcublas
+    - libcusparse
```
flferretti: If you use the CUDA `-dev` packages, your package will automatically depend on the matching runtime libs so environments don’t end up missing the `.so` files.

Moreover, if fvdb actually links to those libraries, you do want those run exports or you need to add `libcublas`, `libcusolver`, `libcusparse`, `cuda-nvtx` explicitly to `run`

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/fvdb_core-{{ version }}.tar.gz
+  sha256: f6ec0d5e6a2b601632720ed0be8aa7c8ba6f2d9bd63fac477cef2cfd088b5f16
+
+build:
+  skip: true  # [py<310 or cuda_compiler_version == "11.8" or cuda_compiler_version == "None" or win or osx]
+  ignore_run_exports:
+    - python
+    - numpy
```
flferretti: This is most likely not needed since you already have `{{ pin_compatible('numpy') }}`
```suggestion
```

recipes/fvdb-core/conda_build_config.yaml
flferretti, line 1: Is this file needed? The pinned versions match with https://github.com/conda-forge/staged-recipes/blob/fd73b847d7148e30c0ad875ade6f80c6e0c0d81d/.ci_support/linux64_cuda126.yaml#L11-L14

recipes/fvdb-core/meta.yaml
swahtz: Regarding the build script, I'm happy to rework the recipe so we're just calling `$PYTHON -m pip install . --no-deps --no-build-isolation -vv`

The `nanovdb` package that is on conda-forge seems to be the built python bindings.  NanoVDB is a header-only library and we don't have any runtime dependencies on NanoVDB.  Is it appropriate to use the `nanovdb` package at build time for obtaining those headers?

recipes/fvdb-core/meta.yaml
swahtz: I tried to do this but got an error when tryign to resolve a pytorch 2.7.0 enviroment:

```
└─ pytorch ==2.7,(>=2.7.1,<2.8.0a0) *cuda* does not exist (perhaps a typo or a missing channel).
```

Is it appropriate to contrain the build with:

    - pytorch >=2.8.0
    - pytorch =*={{ torch_proc_type }}*

swahtz: Oh I cannot do that either because I can't resolve a PyTorch 2.8.0 version with a version of CUDA as low as 12.6:

```
Getting pinned dependencies: ...working... failed
WARNING: failed to get package records, retrying.  exception was: Unsatisfiable dependencies for platform linux-64: {MatchSpec("pytorch==2.8.0=cuda129_generic_py310_h1fab1ae_200"), MatchSpec("cuda-version[version='>=12.9,<13']")}
  - package pytorch-2.8.0-cuda129_generic_py310_h1fab1ae_200 requires cuda-version >=12.9,<13, but none of the providers can be installed

├─ cuda-version =12.6 * is requested and can be installed;
└─ pytorch =*,>=2.8.0 cuda* is not installable because it requires
   └─ cuda-version >=12.9,<13 *, which conflicts with any installable versions previously reported.

   ```

**Comment by jakirkham** [comment]
@jameslamb could you please take a look at this one?

Thank you in advance for your help! 🙏

File: recipes/fvdb-core/meta.yaml [hunk 7]
```diff
@@ -0,0 +1,95 @@
...
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/fvdb_core-{{ version }}.tar.gz
+  sha256: f6ec0d5e6a2b601632720ed0be8aa7c8ba6f2d9bd63fac477cef2cfd088b5f16
+
+build:
+  skip: true  # [py<310 or cuda_compiler_version == "11.8" or cuda_compiler_version == "None" or win or osx]
```
jameslamb:
```suggestion
  skip: true  # [cuda_compiler_version == "None" or win or osx]
```

I think this could be simplified.

Python 3.10 is already the minimum on conda-forge:

https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/c2ba328d2f46df28a7b2749dba7a3ac677e3e27e/recipe/conda_build_config.yaml#L888-L899

And CUDA 11 packages are no longer being built by default: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/7431

recipes/fvdb-core/meta.yaml
jameslamb: > _Is it appropriate to use the nanovdb package at build time for obtaining those headers?_

My opinion: yes, using another conda-forge package would be preferable to using CPM for the reasons @flferretti mentions, even if the Python bindings are more than you need.

It would be nice for this case to eventually have the headers available as their own package, like `libboost-headers` offers:

https://github.com/conda-forge/boost-feedstock/blob/9913203923952479838c21b32f7c9cc28d106037/recipe/meta.yaml#L46

recipes/fvdb-core/meta.yaml
jameslamb: Also is it intentional that you're only building CUDA packages (skipping all builds where `cuda_compiler_version == "None"`)?

If so, other indirection in the recipe could be removed. For example, the host environment could list `libcusolver` etc. unconditionally and not need to guard them with a `{% if cuda_compiler_version != "None" %}`.

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - git
+    - procps-ng  # provides 'free'
+    - coreutils  # provides 'nproc', 'awk', etc.
+  host:
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch {{ pytorch }} *{{ torch_proc_type }}*
+    - numpy
+    - tqdm
```
jameslamb:
```suggestion
```

Looking in the repo, I only see `tqdm` used in examples and docs: https://github.com/search?q=repo%3Aopenvdb%2Ffvdb-core%20tqdm&type=code

Could this be removed?

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,95 @@
...
+    - pybind11 2.13.6
+    - psutil
```
jameslamb:
```suggestion
```

In the repo, I see `psutil` in 2 conda env YAML files but never directly imported in Python code.

https://github.com/search?q=repo%3Aopenvdb%2Ffvdb-core+psutil&type=code

Could this be removed?

jameslamb (commented):
At @jakirkham's request, I left a couple of review comments here.

The build logs have all expired because too much time has passed. I'd be happy to provide a more thorough review once there's a new build and logs I could look at. Hope it helps.

recipes/fvdb-core/meta.yaml
swahtz: I have put together a `nanovdb-dev` package which is a package of the header-only nanovdb library:
https://github.com/conda-forge/openvdb-feedstock/pull/37

recipes/fvdb-core/meta.yaml
swahtz: Yes it can

recipes/fvdb-core/meta.yaml
swahtz: Yes it can, it had been used in our previous build setup

recipes/fvdb-core/meta.yaml
swahtz: I got an error when the package was imported after being built that suggested doing this so I tried it, but I can try removing it.

recipes/fvdb-core/meta.yaml
swahtz: OK, I can remove these.  I had added these due to an error I saw when the package had been built and tried to be imported by the `build-locally` script which advised me to add them to this list.

recipes/fvdb-core/conda_build_config.yaml
swahtz: No, I can remove this, it was previously here to try and control the minimum PyTorch version we wanted to build with.

recipes/fvdb-core/meta.yaml
swahtz: OK, I have removed these unnecessary `cuda_compiler_version` guards

swahtz:
Hi @jameslamb @flferretti 

Thanks very much for having a look at this.

I've gone through and adopted most of the changes you both suggested.  I am happy to have a different build entry point, which I've implemented as a small `build.sh` and eventually I can patch our build to pickup a `nanovdb-dev` package's headers (I have an [open PR ](https://github.com/conda-forge/openvdb-feedstock/pull/37)in that feedstock to provide one).  The reason we use CPM to fetch nanovdb is because we are frequently developing on NanoVDB features to support fvdb and so there often isn't a published/packaged version number to correspond to those changes before we release in our development cycles.

I needed to change the `run` section specification of `pytorch` to match the `host` one and I can resolve an environment.

However, the thing I am stuck on is that even if I have resolved a CUDA 12.6 environment, the source code of our project will fail to build.  We require at least CUDA 12.8 to build the plugin.  What's an appropriate way to specify this so that combinations which involve any CUDA <12.8 are ignored and we only get build variants of CUDA versions that are valid for our project?

jameslamb:

jameslamb:
> _We require at least CUDA 12.8 to build the plugin. What's an appropriate way to specify this so that combinations which involve any CUDA <12.8 are ignored and we only get build variants of CUDA versions that are valid for our project?_

Ok so first, the good news... the default `{{ cuda_compiler_version }}` here on conda-forge is now 12.9, as of about a week ago: https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8152

I tried re-triggering CI (following https://conda-forge.org/docs/maintainer/infrastructure/#conda-forge-admin-please-restart-ci), let's see if we see success on a new build.

I'll leave some comments inline on how to reflect this in the recipe.

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/fvdb_core-{{ version }}.tar.gz
+  sha256: f6ec0d5e6a2b601632720ed0be8aa7c8ba6f2d9bd63fac477cef2cfd088b5f16
+
+build:
+  skip: true  # [cuda_compiler_version == "None" or win or osx]
+  number: 0
+  script_env:
+    - TORCH_CUDA_ARCH_LIST=7.5;8.0;9.0+PTX
```
jameslamb: In earlier iterations in this PR you had:

```shell
--cuda-arch-list="7.5;8.0;9.0;10.0;12.0+PTX"
```

Is it intentional to build for fewer architectures now?

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
+{% set name = "fvdb-core" %}
+{% set version = "0.3.0" %}
+{% set torch_proc_type = "cuda" if cuda_compiler_version != "None" else "cpu" %}
+
+package:
+  name: {{ name|lower }}
```
jameslamb:
```suggestion
  name: {{ name }}
```

The name is already all-lowercase, we can skip `|lower`.

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+  host:
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - numpy
+    - pybind11 2.13.6
+    - gitpython
+    - cuda-version {{ cuda_compiler_version }}
```
jameslamb, line 45: Since you said the project won't build without at least CUDA 12.8, I think here you want something like this:

```suggestion
    - cuda-version {{ cuda_compiler_version }}
    - cuda-nvcc-impl >=12.8
```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - libpng
+    - zlib
+  run:
+    - python
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - {{ pin_compatible('numpy') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-cudart
+    - libcudnn >=9.10.0
```
jameslamb:
```suggestion
    - libcudnn >=9.10.0
  run_constrained:
    - __cuda >= 12.8
```

If at least CUDA 12.8 is also needed at runtime, then you'll want something like this.

In case you haven't seen it, there are docs here:

* `run_constrained`: https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#run-constrained
* `__cuda` virtual package: https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-virtual.html

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+test:
+  imports:
+    - fvdb
```
jameslamb:
```suggestion
```

Does `fvdb` load shared objects (including Cython extensions) on an `import fvdb`? And will those fail if there's not a GPU and drivers present?

If so I *think* this `import` test won't work by default here on conda-forge. I believe these builds are all running on CPU-only runners. There are GPU runners available (https://conda-forge.org/docs/maintainer/infrastructure/#cirun) but I'm unsure of the process for getting access to them.

(cc @jakirkham for your advice)

I'd remove this for now, more correctness tests could be figured out in the feedstock once the initial packages are built.

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - procps-ng  # provides 'free'
+    - coreutils  # provides 'nproc', 'awk', etc.
+  host:
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - numpy
+    - pybind11 2.13.6
```
jameslamb:
```suggestion
    - pybind11 >=2.13
```

This is an extremely tight pin, and tight pins can lead to pain in upgrades (new Python versions, new versions of dependencies, etc.)

Is it necessary? If not, could we loosen this?

recipes/fvdb-core/meta.yaml
swahtz: Well CUDA 12.6 doesn't support compute arch's 10.0 and 12.0 (it's too early) so these weren't recognized arguments and so builds failed immediately with 12.6.  If we can successfully use >=12.8, I will revert to the originally intended arch's.

recipes/fvdb-core/meta.yaml
swahtz: Sure, that shouldn't be a problem.  Let's do that.

recipes/fvdb-core/meta.yaml
swahtz: Done

jameslamb: Ah right of course, I really should have remembered that 😅

recipes/fvdb-core/build.sh
```diff
@@ -0,0 +1,70 @@
...
+    echo "  Constraint: Estimated RAM ($((CMAKE_BUILD_PARALLEL_LEVEL * NVCC_THREADS * JOB_RAM_GB))) GB <= Available RAM ($RAM_GB GB)"
+
+    export CMAKE_BUILD_PARALLEL_LEVEL
+    export NVCC_THREADS
+  fi
+}
+
+
+setup_parallel_build_jobs
+$PYTHON -m pip install . --no-deps --no-build-isolation -vv
```
jameslamb:
```suggestion
export CMAKE_GENERATOR=Ninja
$PYTHON -m pip install \
    --no-deps \
    --no-build-isolation \
    -vv \
    -C 'skbuild.ninja.make-fallback=false' \
    .
```

The latest build is failing like this:

```text
      -- Configuring done (68.4s)
      CMake Error in .cpm_cache/vulkanheaders/6fc921fffac75636a2e70c97404ed7ccbe1cfabf/CMakeLists.txt:
        The target named "Vulkan-Module" has C++ sources that may use modules, but
        modules are not supported by this generator:

          Unix Makefiles

        Modules are supported only by Ninja, Ninja Multi-Config, and Visual Studio
        generators for VS 17.4 and newer.  See the cmake-cxxmodules(7) manual for
````

Let's try forcing the use of `ninja`.

Comment (jameslamb) [reply recipes/fvdb-core/build.sh]: Ha looks like you beat me to it: https://github.com/conda-forge/staged-recipes/pull/31550/changes/5bb4a7e87922f950a97830b83a1af30f5fb7639b

I think you still might want to also set `skbuild.ninja.make-fallback=false` to be sure we get a loud error if `ninja` can't be found (instead of silently falling back to `make`). But not too important either way.

Comment (swahtz) [reply recipes/fvdb-core/build.sh]: Will do!

File: recipes/fvdb-core/meta.yaml [hunk 17]
```diff
@@ -0,0 +1,89 @@
...
+    - pip
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - numpy
+    - pybind11 >=2.13
+    - gitpython
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-nvcc-impl >=12.8
+    - cuda-nvrtc-dev
+    - cuda-nvtx-dev
```
jakirkham, line 48: It looks like this isn't getting detected by CMake

```
  -- CPM: Adding package nvtx3@3.1.0 (v3.1.0-c-cpp)
  -- get_nvtx.cmake: Forced nvtx3_dir to CACHE: /home/conda/staged-recipes/build_artifacts/fvdb-core_1771020205707/work/build/cp312-cp312-linux_x86_64-Release/_deps/nvtx3-src/include (from CPM source)
```

So we may need to adjust the CMake options so it is found

Alternatively we can use the `nvtx-c` package [where we have confirmed CMake detection works]( https://github.com/conda-forge/nvtx-c-feedstock/blob/2cc59332c12788b6b2ab851567545591e6631ec2/recipe/recipe.yaml#L50-L54 )

If we need to push more changes here, we can include one of these fixes. Otherwise it is ok to handle in the feedstock

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - numpy
+    - pybind11 >=2.13
+    - gitpython
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-nvcc-impl >=12.8
```
jakirkham: We can drop this package and bound, now that conda-forge has moved to CUDA 12.9 as the minimum (refs below)

Refs:

* https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/8152
* https://github.com/conda-forge/staged-recipes/pull/32069

```suggestion
```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - {{ stdlib('c') }}
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-command-line-tools
+    - cmake >=3.25
+    - ninja
+    - make
```
jakirkham: Given we are using `ninja` to build do we still use `make`?

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+requirements:
+  build:
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - {{ stdlib('c') }}
+    - {{ compiler('c') }}
+    - {{ compiler('cxx') }}
+    - {{ compiler('cuda') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-command-line-tools
```
jakirkham, line 29: What do we use from here?

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+  run:
+    - python
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - {{ pin_compatible('numpy') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-cudart
+    - libcudnn >=9.10.0
+  run_constrained:
+    - __cuda >=12.8
```
jakirkham: The CUDA compiler [adds a lower version bound]( https://github.com/conda-forge/cuda-nvcc-feedstock/blob/bb5b0a95fa80eeb09a357a512beaa8312658d146/recipe/meta.yaml#L56-L59 ) automatically. So the package will only be installed on CUDA 12.9 or when CUDA 13.0 is added in the feedstock 13.x

```suggestion
```

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - libcusparse-dev
+    - libpng
+    - zlib
+  run:
+    - python
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - {{ pin_compatible('numpy') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - cuda-cudart
```
jakirkham: Given the compiler adds `cuda-version` with an appropriate constraint, we should be good without adding more here

By default NVCC [statically links `libcudart`]( https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#cudart-none-shared-static-cudart ), so we shouldn't need `cudart` unless we are dynamically linking to it

If we are dynamically linking to `cudart`, would still drop this line and instead and `cuda-cudart-dev` to `requirements/host`. It will [automatically add the right runtime dependency via `run_exports`]( https://github.com/conda-forge/cuda-cudart-feedstock/blob/d3cc16506bcdc5c9c2cb32333e42117c645653f1/recipe/meta.yaml#L117-L120 )

If we are not sure whether we are statically or dynamically linking `libcudart`, we should get some clues via the linking check run in CI

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,89 @@
...
+    - coreutils  # provides 'nproc', 'awk', etc.
+  host:
+    - python
+    - scikit-build-core
+    - pip
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - numpy
+    - pybind11 >=2.13
+    - gitpython
```
jakirkham, line 44: Am curious how `gitpython` is being used both during the build and in the test phase

jakirkham:
> I think my most recent round of comments should help answer the "how do I ensure CUDA >= 12.8" question, and hopefully that'll mean we can see some working builds.

If we wanted to do this, we could adjust the `skip` to exclude CUDA versions we don't support

The reason being we [treat the CUDA version as a matrix variable]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/350c3e596268505c418338c26b2a68099ba04b20/recipe/conda_build_config.yaml#L81-L83 ) (like [Python]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/350c3e596268505c418338c26b2a68099ba04b20/recipe/conda_build_config.yaml#L890-L897 )). So we either build or don't build for a specific CUDA version. So within any specific build it is pinned to a version. It isn't being solved for like any other dependency

Hope that makes more sense. Happy to discuss more

recipes/fvdb-core/meta.yaml
jameslamb: I agree with this suggestion, thanks for correcting this. @swahtz sorry for taking you down a bad path here.

I'd recommended this because `cuda-version` only coonstrains within a major version like `__cuda >=12`, but I wasn't considering that we also have a pin like `cuda-version {{ cuda_compiler_version }}` here (which would pin to the specific version from the matrix.

recipes/fvdb-core/meta.yaml
jameslamb: Thanks John, this was also my mistake: https://github.com/conda-forge/staged-recipes/pull/31550#discussion_r2801459971

recipes/fvdb-core/meta.yaml
swahtz: `gitpython` is used in our tests utils module (`fvdb.utils.tests`) which is used to clone a repo that houses some data we use during tests (both for some input data and output reference data we compare against).  It is not used during build, only during runtime.

jakirkham: Should we drop it from `requirements/host` then and stick to using it in `test/requires` then?

jakirkham:
It looks like we got through the tests! 🥳

The tests that fail reference the absence of a GPU. Is there an easy way to skip those tests?

recipes/fvdb-core/meta.yaml
swahtz: Yes or have it in the `run` requirements, I'll change that.  (at one point it was needed in our fvdb-core `build.sh` but we're not using that currently)

recipes/fvdb-core/meta.yaml
swahtz: I believe just nvcc, I thought this was the appropriate way to pull in the CUDA compiler.  I had found at some point that `{{ compiler('cuda') }}` wasn't enough

recipes/fvdb-core/meta.yaml
swahtz: When I `readelf -d libfvdb.so` I see libcudart as a needed library, so I believe we are dynamically linking `cudart`.  I'll try to add `cuda-cudart-dev` to the `requirements/host`

recipes/fvdb-core/meta.yaml
swahtz: I think I'd prefer to handle it in the feedstock just because we're gearing up for a new release in early March where I can make these build changes (as well as some of the other changes to support the conda-forge build better) and not have to patch the old 0.3.0 release.

swahtz:
> It looks like we got through the tests! 🥳
> 
> The tests that fail reference the absence of a GPU. Is there an easy way to skip those tests?

Yes this is great, thanks @jakirkham!  I pushed up a change to the test script to check for the presence of GPU

However, with some of the other recipe changes, I'm now getting these overlinking errors on packaging.  I believe some of the lines I removed/changed in my last commit were changes I discovered earlier I needed to resolve these overlinking errors.

https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1463111&view=logs&j=317836d8-2a86-5766-e372-44e4459f9db0&t=3a0bcf2e-03de-567a-9c0c-03d6e9873a3d&l=2019

recipes/fvdb-core/meta.yaml
swahtz: I adopted this suggestion.  Thanks

recipes/fvdb-core/meta.yaml
swahtz: Indeed, I have removed this.

recipes/fvdb-core/meta.yaml
swahtz: All done

jakirkham:
Thanks Jonathan! Credit goes to you and James for getting this into shape 🙏

Based on [the CI overlinking error]( https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1463375&view=logs&j=317836d8-2a86-5766-e372-44e4459f9db0&t=3a0bcf2e-03de-567a-9c0c-03d6e9873a3d&l=1916 ), it looks like we are linking to `cudnn`

```
  ERROR (fvdb-core,lib/python3.12/site-packages/fvdb/libfvdb.so): Needed DSO lib/libcudnn.so.9 found in ['conda-forge/linux-64::libcudnn==9.10.2.21=hf7e9902_0']
  ERROR (fvdb-core,lib/python3.12/site-packages/fvdb/libfvdb.so): .. but ['conda-forge/linux-64::libcudnn==9.10.2.21=hf7e9902_0'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
```

Does that sound correct? Should we be depending on `cudnn`? Or is this coming in accidentally?

swahtz:
> Thanks Jonathan! Credit goes to you and James for getting this into shape 🙏
> 
> Based on [the CI overlinking error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1463375&view=logs&j=317836d8-2a86-5766-e372-44e4459f9db0&t=3a0bcf2e-03de-567a-9c0c-03d6e9873a3d&l=1916), it looks like we are linking to `cudnn`
> 
> ```
>   ERROR (fvdb-core,lib/python3.12/site-packages/fvdb/libfvdb.so): Needed DSO lib/libcudnn.so.9 found in ['conda-forge/linux-64::libcudnn==9.10.2.21=hf7e9902_0']
>   ERROR (fvdb-core,lib/python3.12/site-packages/fvdb/libfvdb.so): .. but ['conda-forge/linux-64::libcudnn==9.10.2.21=hf7e9902_0'] not in reqs/run, (i.e. it is overlinking) (likely) or a missing dependency (less likely)
> ```
> 
> Does that sound correct? Should we be depending on `cudnn`? Or is this coming in accidentally?

Well @jakirkham we do use `cudnn_frontend` and `cudnn` that comes from PyTorch (which it has dynamically linked I believe), not sure if that's the issue?

recipes/fvdb-core/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+  run:
+    - python
+    - pytorch
+    - pytorch =*=*{{ torch_proc_type }}*
+    - {{ pin_compatible('numpy') }}
+    - cuda-version {{ cuda_compiler_version }}
+    - gitpython
+    - git
+    - parameterized
+
```
jameslamb, line 64: Anticipating more comments for the cudnn question, could we use a thread?

I see the following in the CMake configure logs:

```text
-- USE_CUDNN is set to 0. Compiling without cuDNN support
```

But then later see `cudnn_frontend` pulled in via CPM:

```text
-- CPM: Adding package cudnn_frontend@1.3.0 (v1.3.0)
-- Populating cudnn_frontend
-- Configuring done (0.0s)
-- Generating done (0.0s)
-- Build files have been written to: /home/conda/staged-recipes/build_artifacts/fvdb-core_1771202868458/work/build/cp312-cp312-linux_x86_64-Release/_deps/cudnn_frontend-subbuild
[1/9] Creating directories for 'cudnn_frontend-populate'
[1/9] Performing download step (git clone) for 'cudnn_frontend-populate'
Cloning into 'cudnn_frontend-src'...
HEAD is now at 1b0b5ea cudnn frontend v1.3 release notes. (#72)
[2/9] Performing update step for 'cudnn_frontend-populate'
-- Already at requested tag: v1.3.0
[3/9] No patch step for 'cudnn_frontend-populate'
[5/9] No configure step for 'cudnn_frontend-populate'
[6/9] No build step for 'cudnn_frontend-populate'
[7/9] No install step for 'cudnn_frontend-populate'
[8/9] No test step for 'cudnn_frontend-populate'
[9/9] Completed 'cudnn_frontend-populate'
```

Either way, could the `libcudnn.so` dependency be coming through transitively from `libtorch.so`?

```shell
conda create \
  --name torch-test \
  --yes \
    'libtorch 2.9.1 cuda129_mkl_hf3b6726_305'

source activate torch-test

find ${CONDA_PREFIX} -name 'libtorch.so*' -exec ldd {} \+
```

```text
...
libcudnn.so.9 => /home/jlamb/miniforge3/envs/torch-test/lib/././libcudnn.so.9 (0x000078d7fe200000)
...
```

And if it is, shouldn't the `pytorch =*=*{{ torch_proc_type }}*` dependency in `run:` be enough to pull in `libtorch` and therefore `libcudnn`? 🤔 

I will try to look into this soon.

swahtz: I would have thought the same @jameslamb 

`cudnn_frontend` is header-only afaik.  But I do see that we have dynamically linked `libcudnn` directly:

```
❯ readelf -d libfvdb.so | grep NEEDED
0x0000000000000001 (NEEDED)             Shared library: [libtorch.so]
 0x0000000000000001 (NEEDED)             Shared library: [libc10.so]
 0x0000000000000001 (NEEDED)             Shared library: [libc10_cuda.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudnn.so.9]
 0x0000000000000001 (NEEDED)             Shared library: [libtorch_cpu.so]
 0x0000000000000001 (NEEDED)             Shared library: [libtorch_cuda.so]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.12]
 0x0000000000000001 (NEEDED)             Shared library: [libstdc++.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libm.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [libgcc_s.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libc.so.6]
 0x0000000000000001 (NEEDED)             Shared library: [ld-linux-x86-64.so.2]

```

Should I add `libcudnn` to reqs/run as the error message implies or should I add `libcudnn-dev` to reqs/host like the other CUDA libraries?

swahtz: I added the requirement to the run requirements and now we're back to passing checks!

jameslamb: Ok great! I think it's fine for now to just take on the dependency and to investigate removing it later in the feedstock.

By the way, if you want (no pressure!) I'd be happy to join as a maintainer for this recipe if you'd like another set of hands to keep things moving.

swahtz: That would be amazing, thank you!  Someone who can lend a hand with knowledge in this domain would be fantastic

jameslamb: Ok sure, thanks! You could add me in the `recipe-maintainers:` section in the recipe here, or we can do it later in the feedstock, following https://conda-forge.org/docs/maintainer/updating_pkgs/#updating-the-maintainer-list

Either is fine.

recipes/fvdb-core/meta.yaml
jakirkham: Yeah `{{ compiler('cuda') }}` should be enough for `nvcc`

jakirkham:
Thanks Jonathan and James! 🙏

Sounds like a plan

recipes/fvdb-core/meta.yaml
jakirkham: Have filed this as issue: https://github.com/conda-forge/fvdb-core-feedstock/issues/3

recipes/fvdb-core/meta.yaml
jakirkham: Have filed this as issue: https://github.com/conda-forge/fvdb-core-feedstock/issues/2

recipes/fvdb-core/meta.yaml
jakirkham: Filed an issue to follow up on cuDNN: https://github.com/conda-forge/fvdb-core-feedstock/issues/4

PR #31574, opened 2025-11-27
Add smartg recipe

Mustapha-hygeos:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/smartg/meta.yaml
+++ b/recipes/smartg/meta.yaml
@@ -10,17 +10,17 @@ source:
   sha256: 6f6a81314da6a3fab92395d4c0bb9bb7b3c10420280f9e11439ecaf54dea2c08
 
 build:
+  noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
-  skip: True  # [osx] 
 
 requirements:
   host:
-    - python
+    - python {{ python_min }}
     - setuptools >=42
     - pip
   run:
-    - python
+    - python >={{ python_min }}
     - numpy
     - pandas
     - scipy
@@ -44,7 +44,7 @@ test:
   commands:
     - pip check
   requires:
-    - python
+    - python {{ python_min }}
     - pip
 
 about:


musmoulana:
@conda-forge/help-python Hi all checks passed, ready for review. Thanks.

recipes/smartg/meta.yaml
```diff
@@ -0,0 +1,62 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/smartg-{{ version }}.tar.gz
+  sha256: 6f6a81314da6a3fab92395d4c0bb9bb7b3c10420280f9e11439ecaf54dea2c08
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: True  # [osx]
```
ocefpaf: Looks like this one can be noarch. What is failing on osx?
If you must skip it please see https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies on how to keep it noarch while skip a platform.

musmoulana: Thank you for the comment !

The osx build fails because pycuda cannot be found during the build/test phase.

From your suggestion, I understand that the intended approach would be to come back to noarch but mark the problematic dependency as platform-specific by adding:
```yaml
requirements:
  run:
    - pycuda  # [not osx]
```
Before moving in that direction, I should clarify that pycuda is a core dependency of SMART-G. It is required for the GPU-based radiative transfer simulations, which are the main purpose of the package. Without it, only some intermediate modules (less central to SMART-G) would be usable.
An installation without pycuda would therefore require upstream changes to make pycuda optional.

With this clarification, would you still recommend keeping the package noarch and use a platform selector for pycuda, or is it more appropriate to keep skipping osx builds?

musmoulana: Thank you again for your time @ocefpaf

ocefpaf: In that case no need for a PS-noarch. You can make it "normal" noarch and let it fail on macos. as long as the error is an unavailable dependency the package won't be installable there and the recipe can be noarch.

PR #31581, opened 2025-11-28
add `flamegraph-pl` recipe

lucascolley:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/flamegraph-pl/build.sh
+++ b/recipes/flamegraph-pl/build.sh
@@ -2,5 +2,5 @@
 set -ex
 
 mkdir -p "${PREFIX}/bin"
-sed -i"" -e "1s|/usr/bin/perl|/usr/bin/env perl|" "flamegraph.pl"
+sed -i"" -E -e "1s|/usr/bin/perl( -w)?|/usr/bin/env perl|" "flamegraph.pl"
 install -m 0755 "flamegraph.pl" "${PREFIX}/bin/flamegraph"


lucascolley:
@conda-forge/help-perl ready for review :)

recipes/flamegraph-pl/build.bat
```diff
@@ -0,0 +1,12 @@
...
+
+if not exist "%PREFIX%\Library\bin" (
+    mkdir "%PREFIX%\Library\bin"
+)
+
+copy "flamegraph.pl" "%PREFIX%\Library\bin\flamegraph.pl" >nul
+
+(
+echo @echo off
+echo perl "%PREFIX%\Library\bin\flamegraph.pl" %%*
```
danielnachun:
```suggestion
echo call perl "%PREFIX%\Library\bin\flamegraph.pl" %%*
```
I think you need `call` here so that the terminal doesn't close when command ends.

recipes/flamegraph-pl/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -ex
+
+mkdir -p "${PREFIX}/bin"
+sed -i"" -e "1s|/usr/bin/perl|${PREFIX}/bin/perl|" "flamegraph.pl"
```
danielnachun:
```suggestion
sed -i"" -e "1s|/usr/bin/perl|/usr/bin/env perl|" "flamegraph.pl"
```
It is better to use `/usr/bin/env` here - there is a maximum character limit for the interpreter that can easily be exceeded if the prefix is long enough.

recipes/flamegraph-pl/recipe.yaml
```diff
@@ -0,0 +1,35 @@
...
+source:
+  url: https://github.com/brendangregg/FlameGraph/archive/v${{ version }}.tar.gz
+  sha256: c5ba824228a4f7781336477015cb3b2d8178ffd86bccd5f51864ed52a5ad6675
+
+build:
+  number: 0
+
+requirements:
+  run:
+    - perl
```
danielnachun, line 18:
```suggestion
build:
  noarch: generic
  number: 0

requirements:
  run:
    - perl
    - if: unix
      then:  __unix
      else: __win
```
This should be `noarch: generic`, but built separately on Unix and Windows because have not yet completed the unification of the Unix and Windows `perl` distributions yet (but this is imminent!).  You'll also need to add a `conda-forge.yml` file with these contents:
```
noarch_platforms:
  - linux_64
  - win_64
```

danielnachun (changes requested):
This looks good, the main change is to make this `noarch: generic` but I think the suggestion I made should achieve that with our current Perl setup.

lucascolley:
thanks for taking a look @danielnachun !

lucascolley:
linux failure is:

```
 │ Testing commands:
 │ env: ‘perl -w’: No such file or directory
 │ env: use -[v]S to pass options in shebang lines
```

recipes/flamegraph-pl/build.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+set -ex
+
+mkdir -p "${PREFIX}/bin"
+sed -i"" -e "1s|/usr/bin/perl|/usr/bin/env perl|" "flamegraph.pl"
```
chrisburr: The issue here is that on linux `env` doesn't support arguments, i.e. `#!/usr/bin/env perl -w` doesn't work.

https://unix.stackexchange.com/questions/63979/shebang-line-with-usr-bin-env-command-argument-fails-on-linux

It's probably best to just drop the `-w`:

```suggestion
sed -i"" -E -e "1s|/usr/bin/perl( -w)?|/usr/bin/env perl|" "flamegraph.pl"
```

lucascolley:
green!

lucascolley:
thanks both!

PR #31600, opened 2025-12-02
Adding `smmregrid`

oloapinivad:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/oceanval/meta.yaml
+++ b/recipes/oceanval/meta.yaml
@@ -1,0 +1,62 @@
+{% set version = "0.1.3" %}
+
+package:
+  name: oceanval
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/o/oceanval/oceanval-{{ version }}.tar.gz
+  sha256: 67384dc8b9fa9ceae6bdc0705e989bcc0c5554abf4a0c8aaf068d6936eea3bde 
+
+
+build:
+  noarch: python
+  number: 0
+  script: {{ PYTHON }} -m pip install . -vv
+
+requirements:
+  host:
+    - python {{ python_min }}
+    - pip
+    - setuptools_scm
+    - setuptools-git
+    - setuptools
+  run:
+    - python >={{ python_min }}
+    - dill
+    - multiprocess
+    - fiona
+    - jupyter-book  <=1.0.4
+    - jupytext
+    - plotnine
+    - nctoolkit
+    - notebook
+    - ipykernel
+    - rpy2
+    - r-base
+    - r-rcpp
+    - r-cowplot
+    - r-ggthemes
+    - r-maps
+    - r-plotrix
+    - r-tidyverse
+    - r-ggtext
+
+test:
+  imports:
+    - oceanval
+  requires:
+    - python {{ python_min }}
+    - pip
+  commands:
+    - pip check
+
+about:
+  home: https://oceanval.readthedocs.io/en/ 
+  license: GPL-3.0-only
+  license_file: LICENSE
+  summary: Automated ocean model validation in Python. 
+
+extra:
+  recipe-maintainers:
+    - robertjwilson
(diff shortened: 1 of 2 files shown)


oloapinivad:
@conda-forge-admin, please ping team @conda-forge/help-python

This should be ready to review, despite I am not sure how to deal with the CDO dependency and python minimum version. This is a pure python package which depends on a binary (CDO) which is not available on windows. I am not sure I configured the `meta.yaml` in the right way.

recipes/smmregrid/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/smmregrid-{{ version }}.tar.gz
+  sha256: 44837c85481f146f0a1f01ea8ebf633e908e40a8230d3d0c22d937d0cde1b718
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  skip: true  # [win]
```
ocefpaf: Looks like this one can be noarch.

oloapinivad: Thanks for the feedback! I think it is should now in place, despite the failure for windows, but please let me know if this is correct.

recipes/smmregrid/LICENSE
```diff
@@ -0,0 +1,201 @@
+                                 Apache License
```
ocefpaf: This file is in the tarball. No need to add it here again.

recipes/smmregrid/meta.yaml
```diff
@@ -0,0 +1,54 @@
+{% set name = "smmregrid" %}
+{% set version = "0.1.3" %}
+{% set python_min = "3.8" %}
```
ocefpaf: Only set `python_min` if the minimum required version is higher than the conda-forge default.

```suggestion
```

oloapinivad: thanks! so I should remove it everywhere? where can I find the info on the minimum version?

ocefpaf: Conda-forge tries to follow Python's EoL as close as possible, so right now it is py310.

recipes/smmregrid/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+  sha256: 44837c85481f146f0a1f01ea8ebf633e908e40a8230d3d0c22d937d0cde1b718
+
+build:
+  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
+  number: 0
+  noarch: python
+
+requirements:
+  host:
+    - python
```
ocefpaf:
```suggestion
    - python {{ python_min }}
```

recipes/smmregrid/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - setuptools >=61.0
+    - wheel
+    - pip
+  run:
+    - python
```
ocefpaf:
```suggestion
    - python >={{ python_min }}
```

recipes/smmregrid/meta.yaml
```diff
@@ -0,0 +1,53 @@
...
+    - sparse
+    - python-cdo
+    - numba
+    - cfunits
+    - cdo >=2.5.0
+
+test:
+  requires:
+    - pip
+    - python
```
ocefpaf:
```suggestion
    - python {{ python_min }}
```

recipes/smmregrid/LICENSE
oloapinivad: removed!

recipes/smmregrid/meta.yaml
oloapinivad: I hope I fix it now

PR #31627, opened 2025-12-04
Add cuda_tileiras

billysuh7:
Checklist

- [X] Build number is 0.

I have turned off applying patchelf to change RPATH , because doing so causes segfault when the `tileiras --version` is run during testing.

`patchelf --force-rpath --set-rpath $ORIGIN/../lib:$ORIGIN/../targets/x86_64-linux/lib bin/tileiras ...`

`/home/conda/staged-recipes/build_artifacts/cuda-tileiras-split_1764881330886/test_tmp/run_test.sh: line 9:   384 Segmentation fault      (core dumped) tileiras --version`

Looking at the ldd output it seems setting RPATH isn't needed anyhow, but I'd appreciate someone to weigh in on this

```
$ ldd tileiras
	linux-vdso.so.1 (0x00007ffc580ed000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f9d2a430000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f9d2a42b000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f9d2a426000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f9d2a33f000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f9d2a116000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f9d2a444000)
```

```
$ ./tileiras --version
tileiras: NVIDIA (R) Cuda Tile IR optimizing assembler
Copyright (c) 2005-2025 NVIDIA Corporation
Built on Fri_Nov__7_19:49:56_PST_2025
Cuda compilation tools, release 13.1, V13.1.80
Build local.local.36836380_
```


recipes/cuda_tileiras/meta.yaml
```diff
@@ -0,0 +1,81 @@
+{% set version = "13.1.80" %}
+{% set cuda_version = "13.1" %}
+{% set platform = "linux-x86_64" %}  # [linux64]
+{% set platform = "linux-sbsa" %}  # [aarch64 and arm_variant_type == "sbsa"]
+{% set platform = "linux-aarch64" %}  # [aarch64 and arm_variant_type == "tegra"]
```
carterbox:
```suggestion
```
There is no tegra variant for CTK 13.x

recipes/cuda_tileiras/build.sh
```diff
@@ -0,0 +1,37 @@
...
+#
+#    if [[ $i == "bin" ]]; then
+#        for j in `ls "${i}"`; do
+#	          [[ -f "bin/${j}" ]] || continue
+#
+#            echo patchelf --force-rpath --set-rpath "\$ORIGIN/../lib:\$ORIGIN/../${targetsDir}/lib" "${i}/${j}" ...
+#            patchelf --force-rpath --set-rpath "\$ORIGIN/../lib:\$ORIGIN/../${targetsDir}/lib" "${i}/${j}"
+#        done
+#    fi
+#done
```
carterbox: > Looking at the ldd output it seems setting RPATH isn't needed anyhow, but I'd appreciate someone to weigh in on this

You are correct. We only need to use patchelf when installing binaries to `$PREFIX/$targetsDir/` because the targets directory is a non-standard conda install path, so if we don't modify the rpath in the binaries, they won't be able to find their dependencies. You know this already 😄.

Since this binary is installed directly into $PREFIX/bin, we also don't need most of this script including creating the targets directory, and we don't need patchelf as a dependency.

recipes/cuda_tileiras/conda_build_config.yaml
```diff
@@ -0,0 +1,2 @@
+arm_variant_type: # [aarch64]
+  - sbsa          # [aarch64]
```
carterbox: We don't need arm_variant_type because starting with CUDA 13.0 there is no tegra variant, so you can delete all the things in the recipe related to arm-variant. You know this already.

I assume tileira will not be backported to CUDA 12.x

recipes/cuda_tileiras/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+  name: cuda-tileiras-split
+  version: {{ version }}
+
+source:
+  url: https://developer.download.nvidia.com/compute/cuda/redist/cuda_tileiras/{{ platform }}/cuda_tileiras-{{ platform }}-{{ version }}-archive.{{ extension }}
+  sha256: {{ sha256 }}
+
+build:
+  number: 0
+  binary_relocation: false
```
carterbox:
```suggestion
```
Why disable this? I believe it's needed to ensure that the binary can find it's dependencies.

recipes/cuda_tileiras/test-rpath.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+[[ ${target_platform} == "linux-64" ]] && targetsDir="targets/x86_64-linux"
+# https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html?highlight=tegra#cross-compilation
+[[ ${target_platform} == "linux-aarch64" && ${arm_variant_type} == "sbsa" ]] && targetsDir="targets/sbsa-linux"
```
carterbox: Probably don't need this script. Running `tileiras --version` is sufficient.

billysuh7: Done, thanks!

recipes/cuda_tileiras/meta.yaml
billysuh7: Done, thanks!

recipes/cuda_tileiras/conda_build_config.yaml
billysuh7: I removed the file itself, thanks!

recipes/cuda_tileiras/build.sh
billysuh7: I greatly simplified the file. Thanks!

PR #31663, opened 2025-12-09
Adding zeem

eunos-1128:
This pull request introduces a new recipe for the `zeem` C++ XML library, providing build, test, and packaging instructions for multiple platforms. It also adds a variant configuration for macOS SDK versioning. The main changes are grouped below:

**Addition of `zeem` recipe:**

* Added a new `recipe.yaml` for the `zeem` package, specifying version `2.0.5`, source URL and checksum, build scripts for Unix and Windows, build and host requirements, test scripts for both platforms, and metadata (homepage, license, summary, and maintainers).

**Platform-specific configuration:**

* Added a `variants.yaml` file to specify `MACOSX_SDK_VERSION` as `10.15` for macOS x86_64 builds.

---

Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/zeem/recipe.yaml
+++ b/recipes/zeem/recipe.yaml
@@ -57,7 +57,7 @@ requirements:
     # Patch version update may be breakable
     - ${{ pin_subpackage("zeem", upper_bound="x.x.x") }}
   ignore_run_exports:
-    from_package:
+    by_name:
       - catch2
 
 tests:


eunos-1128:

recipes/zeem/recipe.yaml
```diff
@@ -0,0 +1,85 @@
...
+        ctest -V -C Release --test-dir build
+        cmake --install build
+      else: |
+        echo on
+        cmake -S . -B build -DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=ON %CMAKE_ARGS% -DBUILD_SHARED_LIBS=ON
+        if errorlevel 1 exit 1
+        cmake --build build --config Release --parallel %CPU_COUNT%
+        if errorlevel 1 exit 1
+        ':': ctest -V -C Release --test-dir build
+        ':': if errorlevel 1 exit 1
```
danielnachun: I'm assuming you've commented this out because the tests fail on Windows?  If so it would be good to add a comment above these lines briefly explaining what failed.

recipes/zeem/recipe.yaml
```diff
@@ -0,0 +1,85 @@
+context:
+  name: zeem
+  version: "2.0.5"
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  url: https://github.com/mhekkel/${{ name }}/archive/refs/tags/v${{ version }}.tar.gz
```
danielnachun:
```suggestion
context:
  version: "2.0.5"

package:
  name: zeem
  version: ${{ version }}

source:
  url: https://github.com/mhekkel/zeem/archive/refs/tags/v${{ version }}.tar.gz
```
This Jinja key is not needed anywhere else so it can be removed.

danielnachun (changes requested):

recipes/zeem/recipe.yaml
eunos-1128: I added the comment.

recipes/zeem/recipe.yaml
```diff
@@ -0,0 +1,85 @@
...
+    - cmake
+    - if: unix
+      then: make
+    - pkg-config
+    - if: unix
+      then: sed
+      else: m2-sed
+  host:
+    - fast_float
+    - howardhinnant_date
```
danielnachun, line 55:
```suggestion
  host:
    - fast_float
    - howardhinnant_date
    - if: win
      then: catch2
```
Can you check if this fixes the test failure in Windows?  The error you posted in the comment makes me think that it might work if `catch2` is available.

eunos-1128: It failed in spite of catch2 included in windows.

eunos-1128: @danielnachun 

I've opened a PR in catch2-feedstock as the Windows build seem to lack shared libraries.

https://github.com/conda-forge/catch2-feedstock/pull/74

After it is merged, I will uncomment the lines.

eunos-1128: @danielnachun 

Catch2 now uses shared libraries, but for some reason, Catch2 tests seem to fail on Windows.

Therefore, I disabled post-build testing via `ctest` on Windows.

I believe the tests in the test section should be fine since they find the necessary files.

Could you please merge this?

danielnachun:
@eunos-1128 I think we will end up merging with testing disabled, but before doing that, can you re-enable the test once so I can see what the actual error is?  Due to an annoying quirk of Azure Pipelines when a run is interrupted, the logs are not saved as far as I can see.  So for this case we have to wait until all the jobs finish, then I can check the log, then we can disabled the tests and merge.

The only reason I'm pushing on this is because on occasion the tests can reveal some other underlying issue.  If it turns out that the cause is obscure or unrelated to this package then we will merge with the tests disabled.

eunos-1128:
@danielnachun 

I was testing without installing after building, so Windows couldn't find the shared libraries located in `build/`.

By adding paths to `PATH`, the tests have passed on Windows as well!

PR #31668, opened 2025-12-09
Adding files_collections

robin-cls:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/files_collections/recipe.yaml
+++ b/recipes/files_collections/recipe.yaml
@@ -1,15 +1,14 @@
 schema_version: 1
 
 context:
-  conda_name: files_collections
   version: 0.1.2
 
 package:
-  name: ${{ conda_name }}
+  name: files_collections
   version: ${{ version }}
 
 source:
-  url: https://files.pythonhosted.org/packages/b4/f6/c099ecbe3c94d640d799e5a9aa5c2ba3ed7f8f60293f7485231b36da4454/files_collections-${{ version }}.tar.gz
+  url: https://pypi.org/packages/source/f/files-collections/files_collections-${{ version }}.tar.gz
   sha256: cd171d9d3d0fb3ee81101d55ae80d5cb7a12bce79f8bb28cb10dad0147a57687
 
 build:


robin-cls:

robin-cls:
Hi @conda-forge/staged-recipes, @conda-forge/help-python, my PR is ready for review

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+  conda_name: files_collections
+  pypi_name: files-collections
+  version: 0.1.2
+
+package:
+  name: ${{ conda_name }}
+  version: ${{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/b4/f6/c099ecbe3c94d640d799e5a9aa5c2ba3ed7f8f60293f7485231b36da4454/files_collections-0.1.2.tar.gz
```
flferretti:
```suggestion
  url: https://files.pythonhosted.org/packages/b4/f6/c099ecbe3c94d640d799e5a9aa5c2ba3ed7f8f60293f7485231b36da4454/files_collections-${{ version }}.tar.gz
```
You may want to use `${{ version }}`, so it will be easier to update versions in the future

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - fcollections.missions
+        - fcollections.core
+        - fcollections.implementations
+        - fcollections.sad
+      python_version:
+        - "*"
+        - ${{ python_min }}.*
+  # --- Import tests with optional packages installed
+  - script:
+      - mamba install -y numba dask pyinterp shapely geopandas
```
flferretti: Please avoid installing packages with `mamba install`. If you need tests with optional dependencies, you can add them in the `tests:requires` section as e.g.:

```yml
tests:
  requires:
    - pip
    # Optional dependencies required for specific tests
    - numba
    - ...
```

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+source:
+  url: https://files.pythonhosted.org/packages/b4/f6/c099ecbe3c94d640d799e5a9aa5c2ba3ed7f8f60293f7485231b36da4454/files_collections-0.1.2.tar.gz
+  sha256: cd171d9d3d0fb3ee81101d55ae80d5cb7a12bce79f8bb28cb10dad0147a57687
+
+build:
+  noarch: python
+  number: 0
+  script:
+  # Install the generated package in the conda environment
+  - pip install . -vv
```
flferretti:
```suggestion
  - python -m pip install . -vv
```

Using `python -m pip` ensures that you are installing the package into exactly the same Python environment that is running the command

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+    - setuptools-scm
+  run:
+    - docstring_parser
+    - fsspec
+    - h5netcdf
+    - jinja2
+    - netCDF4
+    - numpy
+    # Numba not available for Python 3.14 as of 2025-12-09
+    - python >=${{ python_min }},<3.14
```
flferretti:
```suggestion
    - python >=${{ python_min }}
```
You can also let the solver do the work

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - fcollections.core
+        - fcollections.implementations
+        - fcollections.sad
+      python_version:
+        - "*"
+        - ${{ python_min }}.*
+  # --- Import tests with optional packages installed
+  - script:
+      - mamba install -y numba dask pyinterp shapely geopandas
+      - python -c "import fcollections.geometry; import fcollections.implementations"
```
flferretti, line 55: You could move these to the `imports` section

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+    # Numba not available for Python 3.14 as of 2025-12-09
+    - python >=${{ python_min }},<3.14
+    - requests
+    - typer
+    - xarray
+
+tests:
+  # --- Import tests without optional packages installed
+  - python:
+      imports:
```
flferretti, line 43:
```suggestion
  - python:
      pip_check: true
      imports:
```
Please add `pip_check` to verify integrity

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - fcollections
+        - fcollections.utilities
+        - fcollections.time
+        - fcollections.missions
+        - fcollections.core
+        - fcollections.implementations
+        - fcollections.sad
+      python_version:
+        - "*"
```
flferretti:
```suggestion
```
This is not needed

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,70 @@
+schema_version: 1
+
+context:
+  conda_name: files_collections
+  pypi_name: files-collections
```
flferretti: Are you using this?

flferretti (commented):
Thanks @robin-cls, I added a few comments

recipes/files_collections/recipe.yaml
robin-cls: I don't. I removed the pypi_name from the context

recipes/files_collections/recipe.yaml
robin-cls: My package is peculiar and has optional features that can be activated if the optional packages are installed. These optional packages are not declared in the dependencies and must instead be installed separately. This is to keep the core of the functionalities as lightweight as possible.

My goal in the tests is to check that everything goes well in case the optional packages are missing. The current setup is to test the import without declaring anything in ``test:requires``, meaning the 'core' functionalities behave properly. 

The ``mamba install`` and ``python -c`` commands check that if we add the 'optional' packages, the optional features can be imported.

From my understanding, ``test:requires`` applies to all tests, so if I declare the needed packages, I won't test the 'core' setup. Do you have any further suggestion to refine the testing section, or is my goal strange and I must reconsider my testing strategy ?

recipes/files_collections/recipe.yaml
robin-cls: Related to the previous comment on test:requires

robin-cls:
@flferretti Thanks a lot for your suggestions and your guidance ! I applied as much as I could. I still have some questions about the testing section though so we can continue to discuss it in the dedicated threads.

recipes/files_collections/recipe.yaml
flferretti: Thanks for the clarification. The dependencies in `tests:requires` are not installed when installing the package, they are only installed in the runner for testing. Are you saying that the following imports will fail if the optional dependencies are installed? 
```yaml
        - fcollections
        - fcollections.utilities
        - fcollections.time
        - fcollections.missions
        - fcollections.core
        - fcollections.implementations
        - fcollections.sad
```

robin-cls: The imports will work properly if the optional dependencies are installed. However, this means that they have not been tested with only the minimal dependencies installed.

Should I:
- Keep trying to test the import without the optional dependencies ?
- Or keep things simple, test the imports with everything (both mandatory and optional dependencies) installed and accept that the test I want to do won't be performed in the recipe ?

flferretti: If you need to try to import all the modules, you can keep the optional dependencies. I expect the core imports not to be affected from the optional dependencies.

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,68 @@
...
+  - script:
+      - mamba install -y numba dask pyinterp shapely geopandas
+      - python -c "import fcollections.geometry; import fcollections.implementations"
+
+about:
+  summary: "Files Collections — Select and open files in a collection"
+  license: Apache-2.0
+  homepage: https://github.com/robin-cls/fcollections
+  repository: https://github.com/robin-cls/fcollections
+  license_file: LICENCE
```
flferretti, line 69: This matches correctly with the filename inside your repo, but you may want to fix this typo anyway

recipes/files_collections/recipe.yaml
robin-cls: OK for that. I was able to remove the mamba install by adding the ``requirements.run`` section to the ``script`` testing. However, such equivalent is not available for ``tests[].python`` as far as I know.

I could follow your suggestion and use the ``tests.requires`` and ``import`` sections, but that would mean reverting to the old recipe format. Do you think it's worth it to drop the rattler format or can we keep it like this ?

recipes/files_collections/recipe.yaml
robin-cls: mamba installed replaced by ``requirements.run`` section

recipes/files_collections/recipe.yaml
robin-cls: I think we will need a release in the package to fix that, but we are not ready yet to make a new release (we have breaking changes pending on the ``main`` branch, with impact on downstream software). Would it be OK if I fix this in the future feedstock ?

flferretti: Sure, this is totally up to you

recipes/files_collections/recipe.yaml
flferretti: The solution you've found look pretty good to me. Thank you!

robin-cls:
Hi @conda-forge/staged-recipes, @conda-forge/help-python, would it be possible to merge this PR ?

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+context:
+  conda_name: files_collections
+  version: 0.1.2
+
+package:
+  name: ${{ conda_name }}
+  version: ${{ version }}
+
+source:
+  url: https://files.pythonhosted.org/packages/b4/f6/c099ecbe3c94d640d799e5a9aa5c2ba3ed7f8f60293f7485231b36da4454/files_collections-${{ version }}.tar.gz
```
h-vetinari: Don't use the URL to a single artefact, but one that works across versions
```
url: https://pypi.org/packages/source/f/files-collections/files_collections-${{ version }}.tar.gz
```

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,74 @@
+schema_version: 1
+
+context:
+  conda_name: files_collections
```
h-vetinari: remove this variable

recipes/files_collections/recipe.yaml
```diff
@@ -0,0 +1,74 @@
...
+  version: 0.1.2
+
+package:
+  name: ${{ conda_name }}
```
h-vetinari: replace it with the actual value
```suggestion
  name: files_collections
```

h-vetinari (commented):
Some small issues, but otherwise LGTM, thanks!

robin-cls:
Thanks @h-vetinari I appreciate !

PR #31710, opened 2025-12-14
SPEC0 package

hmaarrfk:
I'm advocating for a SPEC0 package to be added to help package maintainers opt-in to being more agressive than conda-forge's much longer support window.

https://scientific-python.org/specs/spec-0000/

I think that it will just help shed alot of the "support" for new packages, and historical packages.

It should:
1. Automatically export itself as a dependency
2. This should help "lift" the oldest version of packages as time moves forward.

### DIFF
--- a/recipes/spec0/recipe.yaml
+++ b/recipes/spec0/recipe.yaml
@@ -1,0 +1,66 @@
+context:
+  version: "2025.12"
+  # Update as defined in
+  # https://scientific-python.org/specs/spec-0000/#support-window
+  python_min: "3.12"
+
+package:
+  name: spec0
+  version: ${{ version }}
+
+build:
+  noarch: python
+  number: 0
+
+requirements:
+  host:
+  - python ${{ python_min }}.*
+  - pandas
+  - requests
+  - packaging
+  - ruamel.yaml
+  run:
+  - python >=${{ python_min }}
+  run_constraints:
+  - numpy >=2.0.0
+  - scipy >=1.12.0
+  - matplotlib >=3.9.0
+  - pandas >=2.2.0
+  - scikit-image >=0.23.0
+  - networkx >=3.3
+  - scikit-learn >=1.4.0
+  - xarray >=2023.12.0
+  - ipython >=8.19.0
+  - zarr >=2.17.0
+  run_exports:
+  - spec0 >=${{ version }}
+
+tests:
+- script: |
+    echo "No meaningful tests"
+
+about:
+  homepage: https://scientific-python.org/specs/spec-0000/
+  summary: 'Minimum Supported Dependencies'
+  description: |
+    This package helps packagers implement SPEC-0000 as a means to drop minimum
+    supported dependencies.
+
+    It is designed in a way that is Opt-in for packagers and not globally defined
+    accross all of conda-forge.
+
+    Add it to your run requirements, and it will export minimum versions for all
+    packages that are listed as core dependencies.
+
+    Once a new version of this package is released, make a PR to update the global
+    pin in https://github.com/conda-forge/conda-forge-pinning-feedstock/
+
+    One may override the "time bound" by simply pinning the spec0 package to the desired
+    version for their pckage. However, we strongly encourage packagers to simply use
+    the latest version of this package.
+  license: MIT
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+  - hmaarrfk
--- a/recipes/spec0/LICENSE
+++ b/recipes/spec0/LICENSE
@@ -1,0 +1,19 @@
+Copyright 2025, Mark Harfouche
+
+Permission is hereby granted, free of charge, to any person obtaining a copy of
+this software and associated documentation files (the “Software”), to deal in
+the Software without restriction, including without limitation the rights to
+use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+of the Software, and to permit persons to whom the Software is furnished to do
+so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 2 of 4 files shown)


hmaarrfk:
cc cc: @conda-forge/python @conda-forge/numpy  @conda-forge/scipy  @conda-forge/matplotlib  @conda-forge/pandas  @conda-forge/scikit-image  @conda-forge/networkx  @conda-forge/scikit-learn  @conda-forge/xarray  @conda-forge/ipython  @conda-forge/zarr 

thoughts appreciated

Carreau:
I want to be really careful to not "impose" this. A project can endorse spec0, and drop version accordingly, but it's more a request that a "you should". Maybe we should start with a pyproject.toml linter (repo-review?), then a bot opt-in github bot to update pyproject.toml than pushing this directly downstream in conda-forge ? Otherwise i'm a bit afraid we would antagonise a bunch of people. 

Maybe a field on condaforge.yaml please_spec0:bool,  and if true, on built of dependant that require <spec_min add a lint message : please consider dropping support for X, Y, Z?

Carreau:
Re-reading, I see that you must add it as a dependency; so ignore the `please_spec0:bool` part of my comment. I still think that having it first as an external package would be great.

I'm even thinking this could be extended for other packages / data-source where other project that are not really spec0 compliant but want to "publish" a different support windows could with:

'package_name|version|until-date'

recipes/spec0/recipe.yaml
```diff
@@ -0,0 +1,66 @@
...
+  summary: 'Minimum Supported Dependencies'
+  description: |
+    This package helps packagers implement SPEC-0000 as a means to drop minimum
+    supported dependencies.
+
+    It is designed in a way that is Opt-in for packagers and not globally defined
+    accross all of conda-forge.
+
+    Add it to your run requirements, and it will export minimum versions for all
+    packages that are listed as core dependencies.
```
h-vetinari, line 53:
```suggestion
    Add it to your run requirements, and it will impose minimum versions for all
    packages that are listed as core dependencies.
```

recipes/spec0/recipe.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+    Once a new version of this package is released, make a PR to update the global
+    pin in https://github.com/conda-forge/conda-forge-pinning-feedstock/
+
+    One may override the "time bound" by simply pinning the spec0 package to the desired
+    version for their pckage. However, we strongly encourage packagers to simply use
+    the latest version of this package.
```
h-vetinari, line 60: "latest version" is misleading, because `- spec0 *` wouldn't actually work. That's because if any of the bumped constraints were unsatisfiable with a newer spec0, the solver will just fall back to an older spec0.

The only way this works as intended is if feedstocks consume this as bare `- spec0`, with the pinning enforcing the (latest) version.

h-vetinari (commented):
Speaking with my numpy/scipy feedstock maintainer hat on, I wouldn't use this package, because I follow what upstream numpy/scipy are doing, and am not going to override that.

That said, I'm not against a spec0 package in principle. But there's a mismatch IMO; the `spec0` package would only have the intended effect in combination with the pinning. That's fine for feedstocks in conda-forge, but as implied by the initial statement, I think that's the wrong place to impose this. IOW, it should come from the upstream packages themselves. But those cannot (easily) use `spec0` as intended, because the solver will just fall back to whatever older version if a newer `spec0` version has bumped some dependency that's not yet reflected in whatever dev/CI environment this is being run outside conda-forge.

rgommers:
Agreed with the feedback that packages should just follow upstream.

> to help package maintainers opt-in to being more agressive than conda-forge's much longer support window.

I also don't understand this. Conda-forge does have issues with overly long support windows, but AFAIK not for dependencies covered by SPEC 0 - and not with any Python packages for that matter, but rather the things for which upstream cannot express a minimum version at all.

hmaarrfk:
from what i'm getting from here is that:

1. existing conda-forge maintainers of packages prefer to follow upstream
2. A solution that allows upstream to declare this more affirmatively would help.

I'll have to think of how to do this in the pip context, and circle back here.

hmaarrfk:
thank you all for your time and feedback

rgommers:
> A solution that allows upstream to declare this more affirmatively would help.

I don't think so. At least, I won't be using any dependency for something this simple that is already handled just fine by the existing metadata in `pyproject.toml`.

bollwyvl:
Speaking of upstreams and `pip`: this looks like it limits the versions of python pacakges and python itself. Therefore, perhaps this could be something that had an upstream `.tar.gz` on PyPI. While `pip` has no concept of `run_constraints` or `run_exports`, and `[extras]` are... not great, and not well supported by `conda-forge` tooling, this would at least represent something that upstreams could talk about concretely in a `pyproject.toml`, even if specifying it was long-winded: 

```toml
[project.optional-dependencies]
spec-0000 = ["spec-0000[numpy,pandas] >=2025.12"]
```

This would likely work with existing, not-conda-forge tooling like dependabot and `pip check` (sorta): I've long wanted a "better `pip check`" which could walk `[extras]` more effectively during the (especially built-in) `recipe.yaml#tests/python/pip_check`, but that's likely another matter.

hmaarrfk:
> I don't think so

I understand.

i feel the same way. 

it more maybe about tooling that helps you identify where you can shed maintenance burden.

* Is this a bot?
* A simple cli interface?

I'm not sure. I have previously created [nep29](https://pypi.org/project/nep29/) in the past.

However, my experience as a user of grayskull is that it is quite fragile. So i'm hesitant to make it into a bot that automatically changes your files.

I'm personally glad to see things have changed on the numpy+scipy front with the more aggressive releases. Perhaps my mental model is simply outdated.

hmaarrfk:
@bollwyvl

```
spec-0000[numpy,pandas]
```

this is interesting, i'll think about it and share ideas.

Carreau:
I would gladely add a 'spec0' tag to IPython repo and have a bot that open and issue/PR when the times come to bump the mininal version of the dependencies in pyproject.toml, or any other way that is similar to dependabot/pre-commit. Even if it just open an issue that would be quite useful.

PR #31567, opened 2025-11-26
 Add MedPython to Conda

alonyado:
##  Adding MedPython python Package to conda

Pepy Total Downloads
PyPI - License
GitHub contributors
GitHub commit activity

[GitHub Repo](https://github.com/Medial-EarlySign/MR_LIBS)
GitHub Repo stars

Our platform is designed to transform complex, semi-structured Electronic Medical Records (EMR) into **machine-learning-ready** data and reproducible model pipelines. The framework is optimized for the unique challenges of sparse, time-series EMR data, delivering **low memory usage** and **high-speed processing** at scale.

It was conceived as a **TensorFlow** for machine learning on medical data.

All software is now open-sourced under the MIT license. Some of the models developed by Medial EarlySign that are currently in production are available exclusively through our partners.

The framework was battle-tested in production across multiple healthcare sites and was a key component of an **award-winning** submission to the [CMS AI Health Outcomes Challenge](https://www.cms.gov/priorities/innovation/innovation-models/artificial-intelligence-health-outcomes-challenge).

## Why Use This Platform?

*   **High-Performance Processing:** Engineered for large-scale, sparse EMR time-series data where general-purpose libraries like pandas fall short.
*   **Reusable Pipelines:** Save valuable engineering time by providing shareable, tested pipelines and methods.
*   **Built-in Safeguards:** Mitigate common pitfalls like data leakage and time-series-specific overfitting.
*   **Production-Ready:** Designed for easy deployment using Docker or minimal distroless Linux images.

## Core Components

The platform is built on three key pillars:

*   **MedRepository:** A compact, efficient data repository and API for storing and accessing EMR signals. Querying categorical signals like perscriptions and diagnosis in an easy and efficient API. 
*   **MedModel:** An end-to-end machine learning pipeline that takes data from MedRepository or JSON EMR inputs to produce predictions and explainability outputs. It supports both training and inference.
*   **Medial Tools:** A suite of utilities for training, evaluation, and workflow management, including bootstrap analysis, fairness checks, and explainability.

More information on usage:

* [ETL_Infra](https://medial-earlysign.github.io/MR_Wiki/Tutorials/01.ETL%20Tutorial) - To load data into medpython data format
    - [API Documentation](https://medial-earlysign.github.io/MR_Tools)
* [AlgoMarker](https://medial-earlysign.github.io/MR_Wiki/Infrastructure%20Library/Medial%20Tools/Python/Python%20AlgoMarker%20API%20Server.md#simple-usage-example)
[...]

PR #31722, opened 2025-12-16
Add medpython-etl

alonyado:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/medpython-etl/meta.yaml
+++ b/recipes/medpython-etl/meta.yaml
@@ -1,5 +1,6 @@
 {% set name = "medpython-etl" %}
 {% set version = "1.0.1" %}
+{% set python_min = "3.8" %}
 
 package:
   name: {{ name|lower }}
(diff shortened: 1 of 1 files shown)


alonyado:

recipes/medpython-etl/meta.yaml
```diff
@@ -0,0 +1,49 @@
+{% set name = "medpython-etl" %}
+{% set version = "1.0.2" %}
+{% set python_min = "3.10" %}
```
travishathaway:
```suggestion
```

conda-forge sets this automatically for recipes, so if it's not import to override (which I'm guessing isn't the case here), it's better to rely on the value conda-forge sets so that this recipe remains compatible with conda-forge's minimum supported Python version.

travishathaway (commented):
Hi @alonyado,

I'd like to help you get your conda-forge recipes added. I was also looking at https://github.com/conda-forge/staged-recipes/pull/31567. We should get this added first so that recipe can build with a proper `pip check` call.

For both of these recipes, it's not very obvious where the source code lives. Is there a way that you could link to that in the recipe metadata? I think we could use `dev_url`. This isn't strictly required, but it really helps make the review process easier so that we can identify potential issues.

Thanks!

alonyado:
Thanks @travishathaway 
I have added the `dev_url`, removed the python_min and updated the version to the latest one.
After this is merged, I can update the medpython recipe and the pip check won't fail.

PR #31749, opened 2025-12-19
Add osm2pgsql

travishathaway:
Checklist

- [x] Build number is 0.


travishathaway:
I tried removing the usage of vendored libraries to the best of my abilities. I ended up following this package an example:

- https://github.com/conda-forge/osmium-tool-feedstock

I'm still not sure whether including the source of `libosmium` and `protozero` like this is advisable. I could open up separate pull requests for those dependencies if that's not okay, but that's going to be even more work :sweat_smile:.

travishathaway:
@conda-forge/staged-recipes and @conda-forge/help-c-cpp, ready for review!

recipes/osm2pgsql/variant_config.yaml
```diff
@@ -0,0 +1,4 @@
+c_stdlib:
+  - sysroot
+c_stdlib_version:
+  - 2.17
```
danielnachun:
```suggestion
```
This file is not needed because the default is still `glibc` 2.17 on Linux.

recipes/osm2pgsql/recipe.yaml
```diff
@@ -0,0 +1,66 @@
...
+requirements:
+  build:
+    - ${{ compiler('cxx') }}
+    - ${{ stdlib('c') }}
+    - cmake
+    - ninja
+    - pkg-config
+
+  host:
+    - boost
```
danielnachun:
```suggestion
    - libboost-devel
```
This is the package we now use for `boost`.

recipes/osm2pgsql/recipe.yaml
```diff
@@ -0,0 +1,66 @@
...
+    sha256: 567dad078f8a66d6d706ac1876b5251b688109d16974909d89ce2056d6e9f258
+    target_directory: osm2pgsql
+
+  - url: https://github.com/osmcode/libosmium/archive/refs/tags/v${{ libosmium_version }}.tar.gz
+    sha256: 8f74e3f6ba295baa7325ae5606e8f74ad9056f1d6ab4555c50bff6aa8246f366
+    target_directory: libosmium
+
+  - url: https://github.com/mapbox/protozero/archive/refs/tags/v${{ protozero_version }}.tar.gz
+    sha256: 6c7a896f1dc08435e8cd4f3780ff688cd0bfce6890599b755f6f3cb36398dc25
+    target_directory: protozero
```
danielnachun: These need to be made as separate recipes.  If the versions need to be constrained for the purposes of ABI compatibility, we can use run exports for this.

danielnachun (changes requested):

travishathaway:
@danielnachun,

Thanks for the review. I'll look into making those changes and adding the new recipes.

Do you know why including those dependencies was acceptable for the [osmium-tool-feedstock](https://github.com/conda-forge/osmium-tool-feedstock/blob/39d3493b77a265413d07c14c6c8eaea636f4fb9a/recipe/meta.yaml#L10-L20) and not this feedstock?

travishathaway:
Added this PR for `protozero` today:

- https://github.com/conda-forge/staged-recipes/pull/31949

danielnachun:
> @danielnachun,
> 
> Thanks for the review. I'll look into making those changes and adding the new recipes.
> 
> Do you know why including those dependencies was acceptable for the [osmium-tool-feedstock](https://github.com/conda-forge/osmium-tool-feedstock/blob/39d3493b77a265413d07c14c6c8eaea636f4fb9a/recipe/meta.yaml#L10-L20) and not this feedstock?

I was not involved in that review and that recipe was added several years ago, but I think that was not the right call and those dependencies should not have been bundled there in the first place.  Once you've added those dependencies I will make a PR to remove those dependencies in `osmium-tool`.  Maintaining shared libraries as separate packages is something which I think we have gotten much stricter about in recent years in conda-forge because it allows for much better coordination for things like ABI compatibility, security fixes and other maintenance tasks.

travishathaway:
@danielnachun,

That makes sense to me. Just FYI, I've open the following two new pull requests to add those missing packages:

- https://github.com/conda-forge/staged-recipes/pull/31949
- https://github.com/conda-forge/staged-recipes/pull/31952 (this one is broken until the previous one is added)

Let me know if there's anything else I'm missing. Looking forward to having all the osmium related packages in better shape.

travishathaway:
@danielnachun,

I was looking a little closer at this program and noticed one oddity. At the very end of the [CMakeLists.txt](https://github.com/osm2pgsql-dev/osm2pgsql/blob/58395bc4de5bfce9547a1de1e8929c0aacf9d1e1/CMakeLists.txt#L327-L335), it installs a Python script called `osm2pgsql-replication`. The script itself isn't very well isolated. It basically looks for any available Python on the user's computer and also depends on `psycopg` and `osmium`.

I'm wondering if I should just remove this script and package it separately as `osm2pgsql-replication`, or if I should just add the dependencies it needs to to this recipe. Do you have a preference?

danielnachun:
> @danielnachun,
> 
> I was looking a little closer at this program and noticed one oddity. At the very end of the [CMakeLists.txt](https://github.com/osm2pgsql-dev/osm2pgsql/blob/58395bc4de5bfce9547a1de1e8929c0aacf9d1e1/CMakeLists.txt#L327-L335), it installs a Python script called `osm2pgsql-replication`. The script itself isn't very well isolated. It basically looks for any available Python on the user's computer and also depends on `psycopg` and `osmium`.
> 
> I'm wondering if I should just remove this script and package it separately as `osm2pgsql-replication`, or if I should just add the dependencies it needs to to this recipe. Do you have a preference?

Since the script uses `#!/usr/bin/env python3`, it will correctly pick up the conda Python executable when the environment is activated.  So in this case you can just add the extra runtime dependencies it needs.

travishathaway:
Hi @danielnachun,

Sorry to bother you, but I'm currently stuck on this getting this pull request approved because I need the following recipe to be added:

- https://github.com/conda-forge/staged-recipes/pull/31949

Once that gets added, I'll be able to get this passing:

- https://github.com/conda-forge/staged-recipes/pull/31952

Once that's added, I'll be able to the `pyosmium` library too 😅:

- https://github.com/conda-forge/staged-recipes/pull/32575

I know this is a lot of work, but because you've already been helping here, I would really appreciate some more help to get those other pull requests approved and merged. Is there a chance you could also help out in those PR reviews?

Thanks in advance for any additional help you can provide!

travishathaway:
Hi @danielnachun,

I'm excited to finally announce that this pull request is ready for review! It was quite a bit of work getting the other dependencies prepared, but I learned a lot in the process, so it was definitely worth going this route.

The test script I added is also pretty extensive to make sure this package works correctly with other packages in conda-forge.

Let me know if you have any questions.

cc: @conda-forge/help-c-cpp for more visibility too.

recipes/osm2pgsql/recipe.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - libosmium
+
+  run:
+    - expat
+    - proj
+    - bzip2
+    - lua
+    - pyosmium
+    - psycopg
+    - libosmium
```
danielnachun, line 45: Most if not all of these (other than `lua`) should be handled by run exports.   Can you try to remove them and see if it complains about missing run time dependencies?  It should be pretty obvious which ones need to be added back.

travishathaway: `pyosmium` and `psycopg` are required by the Python script, so I'm going to keep those in `run`.

travishathaway:
@danielnachun,

Thanks for the review! Let me know if this looks good.

cc: @conda-forge/help-c-cpp for more visibility too.

recipes/osm2pgsql/build.bat
```diff
@@ -0,0 +1,30 @@
...
+
+cmake -S . ^
+ -B %BUILD_DIR% ^
+ -G "Ninja" ^
+ -D CMAKE_BUILD_TYPE=Release ^
+ -D EXTERNAL_LIBOSMIUM=ON ^
+ -D EXTERNAL_FMT=ON ^
+ -D EXTERNAL_CLI11=ON ^
+ -D EXTERNAL_PROTOZERO=ON ^
+ -D CMAKE_INSTALL_PREFIX="%LIBRARY_PREFIX%"
```
traversaro: Can you use %CMAKE_ARGS% here?

recipes/osm2pgsql/recipe.yaml
```diff
@@ -0,0 +1,78 @@
...
+    - ${{ compiler('cxx') }}
+    - ${{ stdlib('c') }}
+    - cmake
+    - ninja
+    - pkg-config
+    - catch2
+    - cli11
+    - fmt <12
+    - nlohmann_json
+    - libboost-devel
```
traversaro: Why these deps are in build and not in host?

recipes/osm2pgsql/recipe.yaml
traversaro: I think @danielnachun was referring to their own run_exports. The run_exports of a recipe are typically only used if a recipe package a shared lib, that I guess it is not the case here.

recipes/osm2pgsql/recipe.yaml
travishathaway: What's the difference? and how will this affect other packages that depend on it?

recipes/osm2pgsql/recipe.yaml
travishathaway: Okay, I'll take a closer look at this. I thought I knew what `run_exports` were, but I'm still a little unclear. Rattler Build has the best documentation on this, so I'll go back and read it:

- https://rattler-build.prefix.dev/latest/reference/recipe_file/#run-exports

recipes/osm2pgsql/build.bat
travishathaway: Is that a special variable set by conda-forge's Windows builders?

If everything is currently building just fine, I'm wondering what's the benefit of adding it 🤔.

recipes/osm2pgsql/recipe.yaml
traversaro: I am on the mobile now so I can go to in depth on this, but can you check https://conda-forge.org/docs/how-to/advanced/cross-compilation/#placing-requirements-in-build-or-host and see if it helps?

recipes/osm2pgsql/build.bat
traversaro: Yes, it is a variable set by the compilers activation package, see:
* https://conda-forge.org/docs/how-to/basics/cmake/
* https://conda-forge.org/docs/user/faq/#why-dont-the-cc-compilers-automatically-know-how-to-find-libraries-installed-by-conda
* https://github.com/conda-forge/vc-feedstock/blob/df2b81d31959ed4f706993fa0ab5b2130f716446/recipe/activate.bat#L109-L113

Unfortunately we do not have example CMake recipes in https://conda-forge.org/docs/maintainer/example_recipes/, but the main rationale for using CMAKE_ARGS is for consistency with the rest of the recipes mantained in conda-forge. For example, once you add `win-arm64` builds, `CMAKE_ARGS`  is required to make sure that the correct cross-compilation options are already set. Using `CMAKE_ARGS` ensures that the recipe behave as expected in the future, without any future adjustement. 

To frame it differently, what is the problem/downside of using `%CMAKE_ARGS%` ?

recipes/osm2pgsql/recipe.yaml
traversaro: In a nutshell, anything different from builds tools and compilers should go in host. This is important to ensure that:
* the libraries weak run_exports are correctly propagated to the final package, as otherwise for packages in build environments weak run_exports are ignored

recipes/osm2pgsql/build.sh
```diff
@@ -0,0 +1,39 @@
...
+# Ensure cstdlib is included to fix 'free' visibility issues with fmt library
+export CXXFLAGS="${CXXFLAGS} -include cstdlib"
+
+cd ${SRC_DIR}
+
+mkdir -p ${BUILD_DIR}
+
+rm -rf "contrib/"
+
+cmake ${CMAKE_ARGS} -S . \
```
matthewfeickert: This is more meant to be a drive by comment on modern CMake rather than any sort of critique or revisions suggestion, but with CMake there is no need to move directories as you can just do an out of source build and CMake generates your build directory for you automatically if it doesn't exist.

```suggestion
rm -rf "${SRC_DIR}/contrib/"

cmake ${CMAKE_ARGS} -S ${SRC_DIR} \
```

recipes/osm2pgsql/build.sh
```diff
@@ -0,0 +1,39 @@
...
+ -D CMAKE_FIND_FRAMEWORK=NEVER \
+ -D CMAKE_FIND_APPBUNDLE=NEVER \
+ -D CMAKE_FIND_USE_SYSTEM_ENVIRONMENT_PATH=OFF \
+ -D CMAKE_PREFIX_PATH="${PREFIX};${BUILD_PREFIX}" \
+ -D Boost_INCLUDE_DIR="${BUILD_PREFIX}/include" \
+ -D NLOHMANN_INCLUDE_DIR="${BUILD_PREFIX}/include" \
+ -D LUA_INCLUDE_DIR="${PREFIX}/include" \
+ -D LUA_LIBRARY="${PREFIX}/lib/liblua${SHLIB_EXT}"
+
+cmake --build ${BUILD_DIR} --target all
```
matthewfeickert: This might intentionally be non-parallel, but if not, then can speed up the build.

```suggestion
cmake --build ${BUILD_DIR} --target all --parallel ${CPU_COUNT}
```

recipes/osm2pgsql/build.sh
```diff
@@ -0,0 +1,39 @@
...
+
+cmake --build ${BUILD_DIR} --target install
```
matthewfeickert: This is actually a faster operation.

```suggestion
cmake --install ${BUILD_DIR}
```

recipes/osm2pgsql/build.bat
```diff
@@ -0,0 +1,30 @@
...
+set BUILD_DIR=build
+
+cd %SRC_DIR%
+
+if exist %BUILD_DIR% rmdir /s /q %BUILD_DIR%
+mkdir %BUILD_DIR%
+
+if exist osm2pgsql\contrib rmdir /s /q osm2pgsql\contrib
+
+cmake -S . ^
```
matthewfeickert:
```suggestion
if exist %SRC_DIR%\osm2pgsql\contrib rmdir /s /q %SRC_DIR%\osm2pgsql\contrib

cmake -S %SRC_DIR% ^
```

recipes/osm2pgsql/build.bat
```diff
@@ -0,0 +1,30 @@
...
+ -D CMAKE_BUILD_TYPE=Release ^
+ -D EXTERNAL_LIBOSMIUM=ON ^
+ -D EXTERNAL_FMT=ON ^
+ -D EXTERNAL_CLI11=ON ^
+ -D EXTERNAL_PROTOZERO=ON ^
+ -D CMAKE_INSTALL_PREFIX="%LIBRARY_PREFIX%"
+
+if errorlevel 1 exit 1
+
+cmake --build %BUILD_DIR% --target all
```
matthewfeickert:
```suggestion
cmake --build %BUILD_DIR% --target all --parallel %CPU_COUNT%
```

matthewfeickert (commented):
These are just some drive-by comments about CMake. Feel free to ignore these.

recipes/osm2pgsql/build.bat
```diff
@@ -0,0 +1,30 @@
...
+
+if errorlevel 1 exit 1
+
+cmake --build %BUILD_DIR% --target install
```
matthewfeickert:
```suggestion
cmake --install %BUILD_DIR%
```

recipes/osm2pgsql/recipe.yaml
traversaro: In this case, I think you need to move:

    - catch2
    - cli11
    - fmt <12
    - nlohmann_json
    - libboost-devel

in the host section.

travishathaway: Thanks for the advice. Will give it go!

recipes/osm2pgsql/build.bat
travishathaway: Thanks for the explanation. I only thought it was something optional, but if it helps with compatibility, then it sounds like something that should definitely be included.

travishathaway:
> These are just some drive-by comments about CMake. Feel free to ignore these.

Appreciate the feedback! 🚗

recipes/osm2pgsql/recipe.yaml
travishathaway: Turns out I didn't need `catch2` (didn't realize this was a unit testing framework). Now, I'm just getting some over depending warnings for `libboost-devel` and `fmt`, but I think that's okay, right?

traversaro: >  Now, I'm just getting some over depending warnings for libboost-devel and fmt, but I think that's okay, right?

They are definitely not super-important, but just to comment on each one:

> 2026-05-15T21:08:18.8610473Z  │ │ ⚠ warning Overdepending against libboost

I checked this one. The executable only depend on boost geometry, that is an header-only library. So probably you can just depend on `libboost-headers`, this will work as well, but you will not have the unnecessary run_exports

> 2026-05-15T21:08:18.8611016Z  │ │ ⚠ warning Overdepending against fmt

This is less clear to me. From the code, I guess that fmt is configured to be used in header-only way, see https://github.com/osm2pgsql-dev/osm2pgsql/blob/8b7d01a2ba89884ea151f5b53843a291e929de03/src/format.hpp#L13 . To be honest, I do not know if this is problematic or not, but this being a executable, I guess they can't be symbol conflicts anyhow (unless osm2pqsl loads plugins of some kind), so probably to remove the warnings, you can just ignore the fmt run_exports.

travishathaway: @traversaro,

I tried switching out `libboost-devel` for `libboost-headers`, but `cmake` complained about not being able to find libboost 😭 (https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1527900&view=logs&j=4cabe686-70ae-553a-7fd0-310379f2cbac&t=6a4fc7c9-c31a-5115-eff9-6479d72b69ff)

Is it okay to proceed here regardless? I'm not sure how important those warnings are.

traversaro: > Is it okay to proceed here regardless? I'm not sure how important those warnings are.

Yes, I think it is totally fine to proceed ignoring the warnings.

traversaro (approved):

traversaro (approved):

PR #31778, opened 2025-12-26
adding scala2

alberto-scolari:
Checklist

- [ ] Package does not vendor other packages. (If a package uses the source of another package, they should be separate packages or the licenses of all packages need to be packaged) - all licenses are packaged here
- [x] Build number is 0.

Starting here from Scala 2.12.20, because there are still users of Scala 2.12. Once accepted, also adding Scala 2.13.

### DIFF
--- a/recipes/scala2/recipe.yaml
+++ b/recipes/scala2/recipe.yaml
@@ -18,7 +18,7 @@ requirements:
     - maven
     - sbt
     - sed
-    - openjdk >=11,<17
+    - openjdk =11
   run:
     - openjdk >=11,<=21
 


alberto-scolari:
I confirm that I am willing to be listed here.

alberto-scolari:
@conda-forge/help-java this is ready for review. Since the package is `noarch` the Windows failures can be ignored.

recipes/scala2/recipe.yaml
```diff
@@ -0,0 +1,42 @@
...
+  noarch: generic
+
+requirements:
+  build:
+    - maven
+    - sbt
+    - sed
+    - openjdk >=11,<=23
+  run:
+    - openjdk =11
```
danielnachun:
```suggestion
    - openjdk =21
  run:
    - openjdk >=11
```
For the build requirement, we should just pin to the newest LTS version that can be used to build, which I think for this case would be OpenJDK 21.  

Can you clarify if `scala2` strictly requires OpenJDK 11 at run time?  The reason I ask is that in general JDK distributions put considerable effort into backwards compatibility.  Scala 2 obviously makes more complex use of the JDK than many other packages and very recent JDKs do seem to be making compatibility breaks for security reasons, but I want to make sure this restriction is actually needed as OpenJDK 11 is obviously quite old now.  It's also okay if we have to set an upper bound but allow a range of versions.

recipes/scala2/build.sh
```diff
@@ -0,0 +1,46 @@
...
+mkdir -p ${PREFIX}/bin
+mkdir -p ${PREFIX}/libexec/scala2
+
+cp -r ./build/pack/* ${PREFIX}/libexec/scala2/
+
+for FILE in fsc scala scalac scaladoc scalap; do
+    tee ${PREFIX}/bin/${FILE} << EOF
+#!/bin/sh
+LAUNCHER_PATH="\$(cd -- "\$(dirname "\$0")/../libexec/scala2/bin" >/dev/null 2>&1 ; pwd -P)"
+exec \${LAUNCHER_PATH}/${FILE} "\$@"
```
danielnachun:
```suggestion
exec \${CONDA_PREFIX}/libexec/scala2/bin/${FILE} "\$@"
```
The above suggestion would be simpler - can you explain your motivation for wanting to print the directory in the wrapper?

recipes/scala2/build.sh
```diff
@@ -0,0 +1,46 @@
...
+EOF
+    chmod +x ${PREFIX}/bin/${FILE}
+
+    tee ${PREFIX}/bin/${FILE}.bat << EOF
+@echo off
+"%~dp0..\libexec\scala2\bin\${FILE}.bat" %*
```
danielnachun:
```suggestion
call "%CONDA_PREFIX%\libexec\scala2\bin\${FILE}.bat" %*
```
We normally want to use `call` in the Windows wrappers so that the terminal does close immediately after the execution finishes.  While this behavior can be disabled, it is unfortunately enabled by default.  Also it is preferable to use `%CONDA_PREFIX%` to reference the path of the executable since that is an absolute path that will always be defined.

recipes/scala2/build.sh
```diff
@@ -0,0 +1,46 @@
...
+EOF
+done
+
+for CHANGE in "activate" "deactivate"
+do
+    mkdir -p "${PREFIX}/etc/conda/${CHANGE}.d"
+    for ext in ".sh" ".bat" ".ps1" "-win.sh"; do
+        cp "${RECIPE_DIR}/scripts/${CHANGE}${ext}" "${PREFIX}/etc/conda/${CHANGE}.d/${PKG_NAME}_${CHANGE}${ext}"
+    done
+done
```
danielnachun, line 43:
```suggestion
for CHANGE in "activate" "deactivate"
do
    mkdir -p "${PREFIX}/etc/conda/${CHANGE}.d"
    cp "${RECIPE_DIR}/scripts/${CHANGE}.*" "${PREFIX}/etc/conda/${CHANGE}.d"
done
```
I think you can just use wildcards to save yourself the inner `for` loop here.  Also, it is my understanding that only `.bat` activation/deactivation scripts are used on Windows, regardless of which shell the user is in, so I think you can remove the `.ps1` and `-win.sh` files.

recipes/scala2/recipe.yaml
alberto-scolari: `openjdk =11` as runtime dependency is a typo of mine, it makes indeed no sense to be so restrictive: thanks for catching that and sorry for not realising it myself! I am referring to the [JDK COMPATIBILITY page](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#running-versus-compiling) for the choice of build/run versions.

I probably flipped run version and build version. In the following I am explaining the choices I actually meant.

### Run versions: `openjdk >=11,<=21`
I chose OpenJDK 11 as minimum run version because it's the minimum version [reported in the docs](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-11-compatibility-notes) for which all tests pass, although it's not clear to which Scala version it refers to.

For the maximum run version, OpenJDK 21 is the safest choice for Scala 2.12.20, since OpenJDK 25 is [supported only from Scala 2.12.21](https://docs.scala-lang.org/overviews/jdk-compatibility/overview.html#jdk-25-compatibility-notes). We could of course "stretch the boundary", but with concerns about actual compatibility (was it ever tested?). Furthermore, users using OpenJDK 25 may use Scala 2.12.21 (released while I was writing this recipe and to be added right after it is accepted).

### Build versions: `openjdk >=11,<17`
I chose OpenJDK 11 as exact build version to be build a Scala distribution that is compatible with the run constraint `openjdk >=11,<=21`, thus using the minimum run version for build ("use a newer JDK for your daily work but do release builds on JDK 8", with 8 -> 11 in this case).

---

What do you think?

recipes/scala2/build.sh
alberto-scolari: The first line prints the path into a variable, so not in stdout. It is a way to depend only on the path of the current script, without relying on other env variables.

But `CONDA_PREFIX` can be assumed to be there and your solution is more concise, so I adopted it. Thanks for it!

recipes/scala2/build.sh
alberto-scolari: In my understanding (https://stackoverflow.com/a/2608151 https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/call) `call` creates a subprocess, which I wanted to avoid for this simple case.

I guess it doesn't matter much anyway, especially if it causes the terminal to close immediately automatically.

recipes/scala2/build.sh
alberto-scolari: I agree with using the wildcard, it is just shorter.

I am not sure that `.bat` activation files are used for any Windows shell: the OpenJDK feedstock [produces different files](https://github.com/conda-forge/openjdk-feedstock/tree/main/recipe/scripts), although their content is the same. I took the example from them indeed.

recipes/scala2/build.sh
danielnachun: Yeah it's an annoying quirk of the default settings of Windows terminals!

recipes/scala2/build.sh
danielnachun: I see, I had not seen that before, but we can stick to what you have for now in this case.  I will attempt understand more about how activation scripts are handled on Windows, because I've never seen activation scripts other than `.bat` being used on Windows.

recipes/scala2/recipe.yaml
danielnachun: The run version constraint makes sense now and is good to me.  Just to make our build environment "stable", let's pin the build version to `openjdk=11`.  Strict pinning for build dependencies is much less of an issue because it only ever needs to be installed in the runner while the package is being built and has no effect on the dependency requirements at run time.

recipes/scala2/build.sh
alberto-scolari: I think it just depends on the shell you are activating the environment from: the (de)activation script must be written in the language of the actual shell. So if you're activating the environment from a PowerShell, it must be a `.ps1` script. There's also a script for Bash on Windows (`(de)activate-win.sh`), which assumes Bash syntax, while the UNIX version uses a more generic shell syntax that is compatible with most UNIX shells.

danielnachun: I think what you said makes sense.  There is actually another platform independent way of specifying environment variables in a JSON file to activate/deactivate that we don't need to look into for now, but I'm not actually sure if that really works with Powershell, for example.  I'll have to look into this more and perhaps we'll eventually have a tutorial that shows how to do it that way which may greatly reduce the need to make activation scripts like this in the first place!  But for now we can stick with what you have here since we know it works.

danielnachun (approved):
This looks great now, thanks again for putting in the work to get Scala 2 added.  Please feel free to ping me if you have issues in the feedstock. We've made huge progress in the last year or so in improving support for JDK packages so hopefully maintenance of the feedstock goes smoothly.

alberto-scolari:
Thanks a lot for the in-depth review (which indeed spotted out some mistakes of mine) and for the support!

PR #31795, opened 2025-12-30
Add recipe for pybammsolvers

Saransh-cpp:
Successor of #28748

Adds [pybammsolvers](https://github.com/pybamm-team/pybammsolvers), the C/C++ solvers used by [PyBaMM](https://github.com/pybamm-team/pybamm), to conda-forge.

Checklist

- [X] Build number is 0.


recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+  home: https://github.com/pybamm-team/pybammsolvers
+  summary: Python interface for the IDAKLU solver
+  license: BSD-3-Clause
+  license_file: LICENSE
+
+extra:
+  recipe-maintainers:
+    - prady0t
+    - Saransh-cpp
+    - agriyakhetarpal
```
Saransh-cpp, line 70: @prady0t @agriyakhetarpal could you please comment here to make the linter happy?

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,70 @@
...
+{% set suitesparse_version = "7.8.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/pybamm-team/{{ name }}/archive/refs/tags/v{{ version }}.tar.gz
+    sha256: f5b1841b43394cd3173d93100f85f67ab783f50fc6a7c5951758e61df5ef71bf
+  - url: https://github.com/LLNL/sundials/archive/v{{ sundials_version }}.tar.gz
```
Saransh-cpp: Should we include SUNDIALS' License file in this PR?

recipes/pybammsolvers/bld.bat
Saransh-cpp, line 1: Vcpkg installs things correctly in the CI, and cmake finds everything correctly too; however, the build fails with:

```
2025-12-30T11:26:12.3977092Z   -- Running vcpkg install - done
2025-12-30T11:26:12.4350128Z   -- Selecting Windows SDK version 10.0.26100.0 to target Windows 10.0.20348.
2025-12-30T11:26:15.5443228Z   -- The C compiler identification is MSVC 19.44.35222.0
2025-12-30T11:26:16.6760357Z   -- The CXX compiler identification is MSVC 19.44.35222.0
2025-12-30T11:26:16.7311673Z   -- Detecting C compiler ABI info
2025-12-30T11:26:17.4982101Z   -- Detecting C compiler ABI info - done
2025-12-30T11:26:17.5158839Z   -- Check for working C compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe - skipped
2025-12-30T11:26:17.5168982Z   -- Detecting C compile features
2025-12-30T11:26:17.5176488Z   -- Detecting C compile features - done
2025-12-30T11:26:17.5270092Z   -- Detecting CXX compiler ABI info
2025-12-30T11:26:18.4433045Z   -- Detecting CXX compiler ABI info - done
2025-12-30T11:26:18.4686684Z   -- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.44.35207/bin/Hostx64/x64/cl.exe - skipped
2025-12-30T11:26:18.4689293Z   -- Detecting CXX compile features
2025-12-30T11:26:18.4699265Z   -- Detecting CXX compile features - done
2025-12-30T11:26:18.4854547Z   CMake Warning (dev) at D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/FindPythonLibsNew.cmake:101 (message):
2025-12-30T11:26:18.4857608Z     Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
2025-12-30T11:26:18.4859376Z     are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
2025-12-30T11:26:18.4861546Z     the cmake_policy command to set the policy and suppress this warning, or
2025-12-30T11:26:18.4862610Z     preferably upgrade to using FindPython, either by calling it explicitly
2025-12-30T11:26:18.4863412Z     before pybind11, or by setting PYBIND11_FINDPYTHON ON before pybind11.
2025-12-30T11:26:18.4865324Z   Call Stack (most recent call first):
2025-12-30T11:26:18.4867032Z     vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
2025-12-30T11:26:18.4869635Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Tools.cmake:44 (find_package)
2025-12-30T11:26:18.4871076Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Common.cmake:226 (include)
2025-12-30T11:26:18.4873554Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Config.cmake:257 (include)
2025-12-30T11:26:18.4875677Z     vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
2025-12-30T11:26:18.4876949Z     CMakeLists.txt:35 (find_package)
2025-12-30T11:26:18.4878311Z   This warning is for project developers.  Use -Wno-dev to suppress it.
2025-12-30T11:26:18.4879580Z 
2025-12-30T11:26:18.4944502Z   CMake Warning (dev) at vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package):
2025-12-30T11:26:18.4946428Z     Policy CMP0148 is not set: The FindPythonInterp and FindPythonLibs modules
2025-12-30T11:26:18.4948269Z     are removed.  Run "cmake --help-policy CMP0148" for policy details.  Use
2025-12-30T11:26:18.4989096Z     the cmake_policy command to set the policy and suppress this warning.
2025-12-30T11:26:18.5000645Z 
2025-12-30T11:26:18.5015837Z   Call Stack (most recent call first):
2025-12-30T11:26:18.5021369Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/FindPythonLibsNew.cmake:114 (find_package)
2025-12-30T11:26:18.5026581Z     vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
2025-12-30T11:26:18.5028684Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Tools.cmake:44 (find_package)
2025-12-30T11:26:18.5030556Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Common.cmake:226 (include)
2025-12-30T11:26:18.5044801Z     D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/share/cmake/pybind11/pybind11Config.cmake:257 (include)
2025-12-30T11:26:18.5065407Z     vcpkg/scripts/buildsystems/vcpkg.cmake:859 (_find_package)
2025-12-30T11:26:18.5067770Z     CMakeLists.txt:35 (find_package)
2025-12-30T11:26:18.5076206Z   This warning is for project developers.  Use -Wno-dev to suppress it.
2025-12-30T11:26:18.5077031Z 
2025-12-30T11:26:18.5893637Z   -- Found PythonInterp: D:/bld/pybammsolvers_1767091197636/_h_env/python.exe (found suitable version "3.11.14", minimum required is "3.8")
2025-12-30T11:26:18.7071784Z   -- Found PythonLibs: D:/bld/pybammsolvers_1767091197636/_h_env/libs/python311.lib
2025-12-30T11:26:18.7075394Z   -- Performing Test HAS_MSVC_GL_LTCG
2025-12-30T11:26:19.9847757Z   -- Performing Test HAS_MSVC_GL_LTCG - Success
2025-12-30T11:26:19.9850503Z   -- Found pybind11: D:/bld/pybammsolvers_1767091197636/_h_env/Lib/site-packages/pybind11/include (found version "3.0.1")
2025-12-30T11:26:19.9874202Z   PYBAMM_IDAKLU_EXPR_CASADI: ON
2025-12-30T11:26:20.0444937Z   Trying to link against any casadi package apart from the Python one
2025-12-30T11:26:22.7509193Z   -- Found OpenMP_C: -openmp (found version "2.0")
2025-12-30T11:26:23.5389841Z   -- Found OpenMP_CXX: -openmp (found version "2.0")
2025-12-30T11:26:23.5435096Z   -- Found OpenMP: TRUE (found version "2.0")
2025-12-30T11:26:24.4505910Z   -- Found SUNDIALS: D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/include
2025-12-30T11:26:24.4621292Z   SUNDIALS found in D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/include: D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_idas.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_sunlinsolklu.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_sunlinsoldense.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_sunlinsolspbcgs.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_sunmatrixsparse.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_nvecserial.lib;D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/sundials_nvecopenmp.lib
2025-12-30T11:26:24.4635009Z   -- Looking for sgemm_
2025-12-30T11:26:25.4782990Z   -- Looking for sgemm_ - not found
2025-12-30T11:26:25.5146641Z   -- Performing Test CMAKE_HAVE_LIBC_PTHREAD
2025-12-30T11:26:26.6522916Z   -- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
2025-12-30T11:26:26.6726542Z   -- Looking for pthread_create in pthreads
2025-12-30T11:26:27.6697552Z   -- Looking for pthread_create in pthreads - not found
2025-12-30T11:26:27.6698527Z   -- Looking for pthread_create in pthread
2025-12-30T11:26:27.8760156Z   -- Looking for pthread_create in pthread - not found
2025-12-30T11:26:27.8779177Z   -- Found Threads: TRUE
2025-12-30T11:26:30.4150161Z   -- Looking for sgemm_
2025-12-30T11:26:31.5578370Z   -- Looking for sgemm_ - found
2025-12-30T11:26:31.5939244Z   -- Found BLAS: D:/bld/pybammsolvers_1767091197636/work/build/temp.win-amd64-cpython-311/vcpkg_installed/x64-windows-static-md/lib/blas.lib
2025-12-30T11:26:31.5996439Z   -- Using VCPKG FindLAPACK from package 'lapack-reference'
2025-12-30T11:26:31.7028371Z   -- Looking for cheev_
2025-12-30T11:26:32.4978565Z   -- Looking for cheev_ - found
2025-12-30T11:26:32.4979864Z   -- A library with LAPACK API found.
2025-12-30T11:26:32.5098201Z   -- Configuring incomplete, errors occurred!
```

which is weird as everything above looks alright? Some help here would be greatly appreciated!

recipes/pybammsolvers/build.sh
Saransh-cpp, line 1: This linux build fails with:

```
import: 'pybammsolvers'
  File "/home/conda/staged-recipes/build_artifacts/pybammsolvers_1767091358793/test_tmp/run_test.py", line 2, in <module>
    import pybammsolvers
  File "/home/conda/staged-recipes/build_artifacts/pybammsolvers_1767091358793/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/lib/python3.11/site-packages/pybammsolvers/__init__.py", line 6, in <module>
    idaklu = il.module_from_spec(idaklu_spec)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: /home/conda/staged-recipes/build_artifacts/pybammsolvers_1767091358793/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pl/lib/python3.11/site-packages/pybammsolvers/idaklu.cpython-311-x86_64-linux-gnu.so: undefined symbol: _ZN6casadi8Function11deserializeERKSs
WARNING: Tests failed for pybammsolvers-0.3.3-py311h64d7783_0.conda - moving package to /home/conda/staged-recipes/build_artifacts/broken
```

which I am guessing is because `rpath` is not correctly set (for casadi's lib file, which is in `lib/python3.XX/site-packages/casadi`), but, the osx build passes without any issues, which is very weird. I have tried setting different values for `rpath` but none of them work. Some help here would be greatly appreciated!

Unfortunately, the script stalls locally while building one of the C++ files in pybammsolvers, which is why I have been unable to debug this error locally (but I think that has more to do with my local build configuration).

recipes/pybammsolvers/meta.yaml
prady0t: Linter rules are crazy 🙃

Saransh-cpp (commented):
@conda-forge/help-python-c could you please help us with the CI failure? See my comments below. Thanks!

If required, please also see #28748 for previous discussions.

agriyakhetarpal:
Hi @Saransh-cpp, it's great to see our progress in this direction again! Thanks for your work I have no clue about the configuration errors in the Windows build and will wait for the conda-forge packaging experts to respond further. Still, I can comment about the `_ZN6casadi8Function11deserializeERKSs` undefined symbol for the Linux build: this is essentially a problem with the Python CasADi wheels, which I've experienced on a few occasions, the first one being over two years ago when trying to cross-compile PyBaMM wheels from Linux x86_64 to Linux AArch64. You may read through the thread here: https://github.com/pybamm-team/pybammsolvers/issues/19#issuecomment-3238685537.

This undefined symbol comes from the standard library, due to CasADi's use of an ancient compiler version (GCC<5). This ships with the Dockcross images they use to cross-compile the wheels, meaning that they set `_GLIBCXX_USE_CXX11_ABI=0`. I tried to fix this in my PR at https://github.com/casadi/casadi/pull/4221, but it's been a while since I started that, and my expectation there (or so I think) is that the CasADi developers will pick my PR up (casadi/casadi#4222) with the 3.8 release, which may take many months. It is highly likely that an issue of a similar kind exists in CasADi's interactions with conda-forge's machinery. One way is to avoid depending on CasADi's Python package and instead compile it ourselves from source (i.e., not compile the Python wheel but only the C library). I can't remember whether we tried that in our previous iteration in #28748, but I recall trying it in the https://github.com/pybamm-team/pybammsolvers repo at some point.

Saransh-cpp:
Thanks a lot for your comment, @agriyakhetarpal!! I scratched my head a lot over the missing symbol and I'm glad that it is not an issue on our side 😅 

I'll try building CasADi from source on linux and check if it works.

Saransh-cpp:
Hmm, compiling CasADi from source actually did not work. @agriyakhetarpal, should I be compiling it from your fix branch?

agriyakhetarpal:
Yes, please try it from my branch, @Saransh-cpp! I think it's worth trying. https://github.com/casadi/casadi/pull/4221/ removes the `OLD_GCC` behaviour (you may see the change I made to `casadi-config.cmake.in`).

Saransh-cpp:
@agriyakhetarpal, the branch unfortunately gives the same error :( I think we should just wait for @conda-forge/help-python-c's review on this.

recipes/pybammsolvers/build.sh
danielnachun: This error isn't due to incorrect RPATHs but instead due to a missing LDFLAG.  I'm wondering if:
```
export LDFLAGS="-L${SP_DIR}/casadi -lcasadi"
```
or something similar will work.  Basically I think what's happening here is that at link time the `casadi` library is not being linked.

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,41 @@
...
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+      export USE_PYTHON_CASADI=FALSE
+      cd $SRC_DIR
+
+      git clone -b build-linux-rooms-natively --single-branch https://github.com/agriyakhetarpal/casadi
+
+      mkdir -p build_casadi
+      cd build_casadi
+
```
Saransh-cpp:
```suggestion

        export LDFLAGS="-L${PREFIX}/lib/casadi -lcasadi"

```

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,43 @@
...
+        export LDFLAGS="-L${PREFIX}/lib/casadi -lcasadi"
```
Saransh-cpp:
```suggestion
      cd build_casadi

      export LDFLAGS="-L${PREFIX}/lib/casadi -lcasadi"
```

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,43 @@
...
+      cmake -DWITH_PYTHON=OFF\
+            -DWITH_PYTHON3=OFF\
+            -DCMAKE_INSTALL_PREFIX=$PREFIX\
+            $SRC_DIR/casadi
+      make install -j10
+fi
+
+cd $SRC_DIR
+
+python -m pip install -vv --no-deps --no-build-isolation .
```
Saransh-cpp:
```suggestion
      cmake -DWITH_PYTHON=OFF\
            -DWITH_PYTHON3=OFF\
            -DCMAKE_INSTALL_PREFIX=$PREFIX\
            $SRC_DIR/casadi
      make install -j10
fi

cd $SRC_DIR

export LDFLAGS="-L${PREFIX}/lib/casadi -lcasadi"

python -m pip install -vv --no-deps --no-build-isolation .
```

recipes/pybammsolvers/build.sh
```diff
@@ -0,0 +1,43 @@
...
+      cmake -DWITH_PYTHON=OFF\
+            -DWITH_PYTHON3=OFF\
+            -DCMAKE_INSTALL_PREFIX=$PREFIX\
+            $SRC_DIR/casadi
+      make install -j10
+fi
+
+cd $SRC_DIR
+
+export LDFLAGS="-L${PREFIX}/lib/casadi -lcasadi"
```
Saransh-cpp:
```suggestion
export LDFLAGS="-L${PREFIX}/lib -lcasadi"
```

danielnachun:
I didn't notice this earlier, but why does the recipe clone `casadi` and build it locally instead of using the conda-forge version?  If we could use that the linker might behave better.

agriyakhetarpal:
Hmm, trying to build CasADi from source was an experiment we tried in order to avoid mismatched ABIs between the PyPI CasADi wheel and the conda-forge PyBaMM build here, but it didn't work, as it turns out.

I went through our previous PR again now, and I don't recall using CasADi from conda-forge, so it might be good to try as Daniel points out 🤔

Saransh-cpp:
Hi both, thanks a lot for the help! Let me clarify the timeline here:

This PR built pybammsolvers on conda-forge's CasADi before switching to a manual build as @agriyakhetarpal pointed out a flaw in their ABIs. It never built pybammsolvers with CasADi's PyPI wheels. However, we get the same error when we use conda-forge's CasADi and when we build CasADi from scratch, so, as pointed out be @danielnachun, it should be a linking issue. Let me try fixing this today.

Saransh-cpp:
Hi @agriyakhetarpal @danielnachun, sorry for being slow on this, but setting the flag (or `LD_LIBRARY_PATH` does not work. Moreover, I get a library not found error when linking against `casadi` using the flags, as casadi does not ship a `libcasadi.so` (?) which is why we link against a directory while building for PyPI (?)

I think @agriyakhetarpal knows more about our PyPI build so he can comment better.

agriyakhetarpal:
The glibc-too-old issue I had mentioned in the past was resolved via https://github.com/dockcross/dockcross/pull/931 recently. So if we want to fix it via our original means, we would need a new Dockcross image and a new CasADi wheel built with it, which would happen with a new CasADi release. However, I hope CasADi 3.8 will come before that and fix the problem at the root, i.e., by not using a cross-compiling image at all. IMO, there are too many parts to this...

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+{% set suitesparse_version = "7.8.3" %}
+
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  - url: https://github.com/pybamm-team/{{ name }}/archive/v{{ version }}.tar.gz
+    sha256: f5b1841b43394cd3173d93100f85f67ab783f50fc6a7c5951758e61df5ef71bf
+  - url: https://github.com/LLNL/sundials/archive/v{{ sundials_version }}.tar.gz
```
Saransh-cpp: Should we include the license of SUNDIALS in this PR?

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,73 @@
+{% set name = "pybammsolvers" %}
+{% set version = "0.3.3" %}
```
Saransh-cpp: Still at `0.3.3` as `0.4.0`+ requires CasADi `3.7.2`, which is still not up on conda-forge - https://github.com/conda-forge/casadi-feedstock/pull/140

Saransh-cpp (commented):
Hi @danielnachun @agriyakhetarpal @prady0t, I got the builds to work! :)

This should be ready to review!

Saransh-cpp:
A gentle ping to @conda-forge/help-python-c, just in case this was missed.

recipes/pybammsolvers/meta.yaml
danielnachun: Sorry if I missed this before, but why can't the conda-forge version of `sundials` be used here?  The version you are using is definitely available.  It's okay to symlink from the folder if CMake is hardcoded to a specific directory, though it's better to patch it to accept a path if that isn't too hard.

Saransh-cpp: Hmm, IIRC, we are building it locally as the SUNDIALS feedstock does not build it with `-DENABLE_OPENMP=ON` for OSX -

https://github.com/conda-forge/sundials-feedstock/blob/df6bd5f66bc007bafbab6d66433362047e43b267/recipe/build-sundials.sh#L7-L10

I can try updating their recipe and build it with OpenMP to see how it goes. Thanks!

Saransh-cpp: Thanks, @danielnachun! https://github.com/conda-forge/sundials-feedstock/pull/129 should fix this. However, I think this will not propagate to older versions of SUNDIALS; hence, we should still build it in our CI until we start using the latest version?

agriyakhetarpal: While the conda-forge team reviews this, maybe we can try bumping SUNDIALS and SuiteSparse to newer versions for `pybammsolvers`. The last few updates from the v6.x series to the v7.x series were fairly trivial for us!

Otherwise, I am also alright with deviating from our PyPI wheels for our recipe here, and choosing the latest version of SUNDIALS whose builds now ship with OpenMP included – if @danielnachun and other conda-forge reviewers allow us to make this choice (considering that it might take us some time until we are able to get a new `pybammsolvers` release out).

Saransh-cpp: It looks like updating the version in the repository works - https://github.com/pybamm-team/pybammsolvers/pull/96

However, yes, I agree, it will take us some time to get a new PyPI release up, so we should not wait for it. Too many upstream moving pieces for this PR 😬 

It looks good to me at this stage. We can revert to conda-forge SUNDIALS once the feedstock is up and a new release for pybammsolvers is up on PyPI.

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,73 @@
...
+    sha256: 97d0ae56391812013928143d5226a2ea789ff0bd657a8a6eeebf9e26d60186c0
+    folder: sundials
+  - patches:
+    - 0001-Fix-CasADi-include-dir.patch
+    - 0002-Fix-CASADI_DIR.patch
+    - 0003-Remove-CXX11-ABI-Definition.patch
+ 
+build:
+  number: 0
+  skip: true   # [py>313]
```
Saransh-cpp:
```suggestion
  skip: true   # [py>314]
```

recipes/pybammsolvers/meta.yaml
danielnachun: I think a better solution is to backport your fix to 7.3.0 in the feedstock.  Let me see how to do this and then you'll just be able to use that version on conda-forge.

Saransh-cpp: Thanks for looking into this! The good news is - we won't need backporting as we are planning to release a new version of pybammsolvers (with the latest SUNDIALS) sometime next week.

I'll ping you again once that is done and once I've updated this recipe to use conda-forge SUNDIALS. Thanks for being patient with this :)

recipes/pybammsolvers/bld.bat
```diff
@@ -0,0 +1,18 @@
+@echo off
+setlocal enabledelayedexpansion
+
+git clone https://github.com/microsoft/vcpkg --branch 2025.01.13
```
traversaro: Why are you using vcpkg instead of relying on the conda-forge libraries?

traversaro:
> A gentle ping to @conda-forge/help-python-c, just in case this was missed.
> 
> @conda-forge/help-python-c, ready for review!

Can you provide a bit of introduction to the library and the recipe, and the specific choices behind the recipe, like the use of vcpkg (?) on Windows, and the choice of vendoring sundials on *nix ?

Saransh-cpp:
Hi @traversaro, thanks for the review!

Tldr; the recipe now uses everything from conda-forge.

You can look at the discussions above to understand why we were building SUNDIALS from scratch + using VCPKG for windows. The fix included multiple upstream PRs, but now pybammsolvers uses the latest SUNDIALS and SuiteSparse, so we can finally use the conda-forge distributions.

Hi @danielnachun, thanks a lot for guiding in the right direction!

This should now be ready to review again. @conda-forge/help-python-c, ready for review! (to update the PR label; I hope it is okay to ping this handle whenever needed)

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+requirements:
+  build:
+    - {{ compiler('c') }}
+    - {{ stdlib("c") }}
+    - {{ compiler('cxx') }}
+  host:
+    - python
+    - setuptools >=71
+    - cmake  # [not win]
+    - make  # [not win]
```
traversaro: Why are these in `host` and not in `build`, and why are not there on Windows?

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,69 @@
...
+    - casadi =={{ casadi_version }}
+    - pybind11
+    - pip
+    - suitesparse =={{ suitesparse_version }}
+    - sundials =={{ sundials_version }}
```
traversaro: Why do you have this strict pinnings? I would understand having those as minimal version, but what prevent this libraries to be built with a future casadi 3.7.3 ?

recipes/pybammsolvers/meta.yaml
Saransh-cpp: Oops, just moved them to build. I did not install them on Windows because the CI passed without them (was being conservative with dependencies), but I can definitely remove the selector. Thanks!

recipes/pybammsolvers/meta.yaml
Saransh-cpp: We have strict pins in pybammsolvers as it rarely works with new CasADi/SUNDIALS/SuiteSparse versions (the wheels almost never build out-of-the-box, and our C/C++ API uses these three libraries). We update these versions once in a while, which I was planning to do here too.

Do you think it is better to not have these pins and deviate from the PyPI distribution?

traversaro: In a nutshell, yes. 

The PyPI packages probably does not work as new `CasADi`/`SUNDIALS`/`SuiteSparse` have a different ABI, and PyPI does not track that in any way. On the other hand, `conda-forge` has mechanism to explicitly handle that (`run_exports`, ABI migration bot) but to make sure that those mechanism work fine, you need to avoid to have these kind of strict pins. I write a bit about this pypi vs conda-forge differences in the context of lerobot packaging in https://github.com/conda-forge/lerobot-feedstock/tree/main/recipe#dependencies-metadata-changes, you can read more about this in:
* https://conda-forge.org/docs/maintainer/pinning_deps/#specifying-run_exports
* https://conda-forge.org/blog/2019/12/06/cfep09/
* https://pypackaging-native.github.io/

Saransh-cpp: Thanks a lot for the resources, @traversaro; amazing reads!

I have removed the pins and SuiteSparse + SUNDIALS from `run`. Should I add CasADi and SUNDIALS to the global pin repo (they're missing) or is it not required (we are the only feedstock using their C/C++ interface (?) + they have `run_exports` in their recipe)?

traversaro: > Should I add CasADi and SUNDIALS to the global pin repo (they're missing)

Good catch! Yes, they should definitely be added, as there are a few feedstock they depend on them, I opened https://github.com/conda-forge/conda-forge-pinning-feedstock/issues/8392 to avoid forgetting about it, but feel free to work on this if you like.

Anyhow, that should not block this PR.

recipes/pybammsolvers/meta.yaml
```diff
@@ -0,0 +1,64 @@
...
+    - suitesparse
+    - sundials
+  run:
+    - python
+    - casadi
+    - numpy
+    - llvm-openmp  # [osx]
+    - _openmp_mutex  # [linux]
+    - vcomp14  # [win]
+
```
traversaro: I am a bit confused, why you depend on this libraries, and you do not depend at host time with openmp? See https://conda-forge.org/docs/maintainer/knowledge_base/#openmp .

Saransh-cpp: My bad, should be fixed now. The only problem I see is the Windows CI erroring with overlinking error when `vcomp14` is not present as a runtime dependency. Everything else is in `host` now (and now it follows the recommendations in knowledge base).

Saransh-cpp:
A gentle review ping on this @traversaro @danielnachun

recipes/pybammsolvers/0003-Fix-Windows-cmake-prefix-path.patch
```diff
@@ -0,0 +1,19 @@
...
++    set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
++    list(APPEND CMAKE_IGNORE_PATH "${_conda_prefix}/lib/cmake/sundials")
++    file(WRITE "${CMAKE_BINARY_DIR}/conda_cmake/FindBLAS.cmake"
++        "set(BLAS_LIBRARIES \"${_conda_prefix}/lib/blas.lib\")\n"
++        "set(BLAS_LINKER_FLAGS \"\")\n"
++        "set(BLAS_FOUND TRUE)\n")
++    list(PREPEND CMAKE_MODULE_PATH "${CMAKE_BINARY_DIR}/conda_cmake")
++endif ()
++
+ find_package(pybind11 CONFIG REQUIRED)
```
traversaro, line 19: Can you explain why this patch is necessary?

recipes/pybammsolvers/0001-Fix-CasADi-include-dir.patch
```diff
@@ -0,0 +1,25 @@
...
+ 
+@@ -41,10 +41,7 @@ def set_vcpkg_environment_variables():
+ 
+ 
+ def get_casadi_python_include_dir():
+-    import casadi
+-
+-    casadi_path = Path(casadi.__file__).parent
+-    include_dir = casadi_path / "include"
++    include_dir = Path(os.environ["PREFIX"]) / "include" / "casadi"
```
traversaro: I guess this is wrong on Windows?

recipes/pybammsolvers/0002-Fix-CASADI_DIR.patch
```diff
@@ -0,0 +1,25 @@
...
+ 
+ if (${USE_PYTHON_CASADI})
+-    execute_process(
+-            COMMAND "${PYTHON_EXECUTABLE}" -c
+-            "import os; import sysconfig; print(os.path.join(sysconfig.get_path('purelib'), 'casadi', 'cmake'))"
+-            OUTPUT_VARIABLE CASADI_DIR
+-            OUTPUT_STRIP_TRAILING_WHITESPACE)
+-
+-    if (CASADI_DIR)
++    set(CASADI_DIR "$ENV{PREFIX}/lib/cmake/casadi" CACHE PATH "Path to casadi CMake package from Python casadi installation")
```
traversaro: This seems wrong on Windows, where casadi is installed in $ENV{PREFIX}/Library`.

recipes/pybammsolvers/meta.yaml
traversaro: I think the knoledge base is wrong, `vcomp14` should be in the host section, see https://github.com/conda-forge/conda-forge.github.io/pull/2804 .

matthewfeickert: Yeah, that PR still needs to be merged, but @Saransh-cpp ( :wave: Good to "see" you!) c.f. this Zulip discussion for more details: [#general > ✔ Help with Windows, CMake, Ninja, and overlinking @ 💬](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.E2.9C.94.20Help.20with.20Windows.2C.20CMake.2C.20Ninja.2C.20and.20overlinking/near/584590316)

matthewfeickert: Hm. Maybe `vcomp14` isn't the way to go given https://github.com/conda-forge/conda-forge.github.io/pull/2804#issuecomment-4245924382.

recipes/pybammsolvers/0001-Fix-CasADi-include-dir.patch
Saransh-cpp: Thanks, @traversaro! This patch is a past relic; not needed anymore as we now use conda casadi everywhere.

recipes/pybammsolvers/0002-Fix-CASADI_DIR.patch
Saransh-cpp: Missed it, thanks for spotting! Ditto for this; past relic, not needed anymore.

recipes/pybammsolvers/meta.yaml
Saransh-cpp: Hi both 👋🏽 

Glad bumping into you, @matthewfeickert 😄 thanks for the suggestion! What do you suggest I do in the meantime? Should I put `vcomp14` as a host dependency for now?

recipes/pybammsolvers/meta.yaml
```diff
@@ -17,7 +17,11 @@ build:
   number: 0
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   skip: true   # [py>314]
-
+  # need sundials 7.6 for openmp support on osx builds
+  # remove once there is a migrator for sundials 7.6
+  ignore_run_exports:
+    - sundials
```
traversaro: If you want to ignore the pin, just put `sundials 7.6.*` or `sundials 7.7.*` with a comment explaining why.

Saransh-cpp: Thanks!

recipes/pybammsolvers/0003-Fix-Windows-cmake-prefix-path.patch
Saransh-cpp: IIRC, cmake was unable to find BLAS in the CI. The bundled `FindSuiteSparse.cmake` internally calls `FindBLAS` (which fails), so I switch cmake to try CONFIG mode before MODULE mode. However, then I want cmake to not find `SUNDIALSConfig.cmake` as it does not set the old-style `SUNDIALS_LIBRARIES` variable that pybammsolvers links against. Finally, the shim for finding BLAS at the end was needed because cmake's built-in `FindBLAS.cmake` uses `NO_DEFAULT_PATH` in its `find_library()` calls (?), which was skipping `CMAKE_PREFIX_PATH` entirely.

recipes/pybammsolvers/meta.yaml
matthewfeickert: @Saransh-cpp I _think_ that you want to use `llvm-openmp` but with the `/openmp:llvm` option, which for a `v0` / `conda-build` recipe like this one would (I think) be something like:

```yaml
build:
   script:
    - set "CXXFLAGS=%CXXFLAGS% /openmp:llvm"  # [win]

...

requirements:
  host:
    - libgomp      # [linux]
    - llvm-openmp  # [not linux]
```

recipes/pybammsolvers/meta.yaml
```diff
@@ -15,8 +15,10 @@ source:
  
 build:
   number: 0
-  script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   skip: true   # [py>314]
+  script_env:
+    - "CXXFLAGS=%CXXFLAGS% /openmp:llvm"  # [win]
```
matthewfeickert: I don't know enough about how `conda-build` handles things compared to how `rattler-build` does to know if [`script_env`](https://docs.conda.io/projects/conda-build/en/stable/user-guide/environment-variables.html#inherited-environment-variables) gets set before or after the other environmental variables. If it gets set **after** this is fine, but if it gets set **before** then this could get overwritten by the `CXXFLAGS` value set for `{{ compiler('cxx') }}`. It this get set in the script 

```yaml
  script:
   - set "CXXFLAGS=%CXXFLAGS% /openmp:llvm"  # [win]
   - {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
```

then you know that things are correct.

Saransh-cpp: Thanks for all the help on this, @matthewfeickert! Using `script_env` worked, so I think this should work too.

matthewfeickert: > Using `script_env` worked, so I think this should work too.

Ah okay, if `script_env` worked then I'd go with whatever is easiest to maintain for you, though what you have with everything in `script` works now. :+1:

Saransh-cpp:
A gentle ping on this, @traversaro 😅 

There is just one unresolved comment from my end.

recipes/pybammsolvers/0003-Fix-Windows-cmake-prefix-path.patch
traversaro: Ok, can you add a comment in the recipe or the patch on this? This info may not be trivial to derive for future maintainers, and it is important for maintaining the patch.

Saransh-cpp: Thanks, @traversaro! I have added comments, so should be good to review again!

Saransh-cpp: Hi, @traversaro, a quick bump on this!

traversaro (approved):

PR #31799, opened 2025-12-30
Add recipe for cpcmx

awvwgk:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/cpcmx/meta.yaml
+++ b/recipes/cpcmx/meta.yaml
@@ -47,7 +47,6 @@ test:
 about:
   home: https://github.com/grimme-lab/cpcm-x
   license: LGPL-3.0-or-later
-  license_family: LGPL
   license_file:
     - LICENSE
   summary: Extended conductor-like polarizable continuum solvation model


awvwgk:
@conda-forge/help-c-cpp This recipe is ready for review

recipes/cpcmx/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+source:
+  url: https://github.com/grimme-lab/CPCM-X/releases/download/v{{ version }}/cpx-{{ version }}.tar.xz
+  sha256: 255dfa9b70c41580163fa3636125a0103c3b553c52fc8558d3a63f4654cc41c1
+
+build:
+  number: 0
+  skip: true  # [win]
+
+  run_exports:
+    - {{ pin_subpackage(name, max_pin='x.x') }}
```
danielnachun:
```suggestion
{% set version = "1.1.0" %}

package:
  name: cpcm-x
  version: {{ version }}

source:
  url: https://github.com/grimme-lab/CPCM-X/releases/download/v{{ version }}/cpx-{{ version }}.tar.xz
  sha256: 255dfa9b70c41580163fa3636125a0103c3b553c52fc8558d3a63f4654cc41c1

build:
  number: 0
  skip: true  # [win]

  run_exports:
    - {{ pin_subpackage("cpcm-x", max_pin='x.x') }}
```
The package name is not going to change, so we don't need this Jinja key.

recipes/cpcmx/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - numsa
+    - toml-f
+
+test:
+  requires:
+    - pkg-config
+  commands:
+    - cpx --help
+    - test -f $PREFIX/lib/libcpx$SHLIB_EXT  # [unix]
+    - pkg-config cpx --exact-version {{ version }}
```
danielnachun, line 45:
```suggestion
    - cpx --help
    - test -f $PREFIX/lib/libcpx$SHLIB_EXT  # [unix]
    - if not exist "%LIBRARY_PREFIX%\lib\cpx.lib" exit /b 1  # [win]
    - if not exist "%LIBRARY_PREFIX%\bin\cpx.dll" exit /b 1  # [win]
    - pkg-config cpx --exact-version {{ version }}
```
Since we are also building a shared library on Windows we should test for the presence of the necessary files.

recipes/cpcmx/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+
+about:
+  home: https://github.com/grimme-lab/cpcm-x
+  license: LGPL-3.0-or-later
+  license_family: LGPL
```
danielnachun:
```suggestion
```
This key is deprecated and is no longer needed.

recipes/cpcmx/meta.yaml
```diff
@@ -0,0 +1,56 @@
...
+    - {{ stdlib('m2w64_c') }}    # [win]
+    - {{ compiler('m2w64_c') }}  # [win]
+    - {{ compiler('m2w64_fortran') }}  # [win]
+    - meson
+    - ninja
+    - pkg-config
+    - libgomp  # [not osx]
+  host:
+    - libblas
+    - liblapack
```
danielnachun:
```suggestion
  host:
    - libblas
    - liblapack
    - libgomp  # [not osx]
```
This is should be in `host`, not `build`.

recipes/cpcmx/build.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -ex
+for pkg in lapack blas; do
+if [ -f $PREFIX/lib/pkgconfig/$pkg.pc ]; then
+   mv $PREFIX/lib/pkgconfig/$pkg.pc $SRC_DIR
+fi
+done
```
danielnachun, line 8: Why are these files being copied to `SRC_DIR`?  Are they not already found by the default `PKG_CONFIG_PATH`?

awvwgk: They unfortunately break the pkgconf export in meson, resulting in an unusable pkgconf file. The best option I found is to remove them from the environment for the build since meson can reliably find blas and lapack without them.

awvwgk:
Thanks for the comments, I updated the recipe accordingly.

recipes/cpcmx/build.sh
danielnachun: Thanks for explaining this.  This workaround is okay for now, but we should report this.  Can you open an issue in https://github.com/conda-forge/blas-feedstock explaining how to reproduce the failure you see when you don't move those files?  This may not be specific to just this package but we don't have a too many `meson` packages that use `libblas` and `liblapack`.

PR #31820, opened 2026-01-01
Added fbgemm recipe building fbgemm, fbgemm-gpu, and fbgemm-gpu-genai

das-intensity:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/fbgemm/meta.yaml
+++ b/recipes/fbgemm/meta.yaml
@@ -70,7 +70,9 @@ outputs:
         # Set CUDA architectures: starts at 6.0 because code uses __hfma2 (half-precision FMA)
         # which requires Compute Capability 6.0+ (Pascal). CC 5.0 (Maxwell) only supports FP16 storage, not arithmetic.
         # Compiling with 10.0 failed with error: nvcc fatal: Unsupported gpu architecture 'compute_100' - cuda 12.6 issue ?
-        - TORCH_CUDA_ARCH_LIST=6.0;7.0;7.5;8.0;8.6;9.0+PTX  # [cuda_compiler_version != "None"]
+        # TODO switch back to full arch list before merging
+        # - TORCH_CUDA_ARCH_LIST=6.0;7.0;7.5;8.0;8.6;9.0+PTX  # [cuda_compiler_version != "None"]
+        - TORCH_CUDA_ARCH_LIST=9.0  # [cuda_compiler_version != "None"]
       string: cuda{{ cuda_compiler_version | replace('.', '') }}_py{{ CONDA_PY }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
       string: cpu_py{{ CONDA_PY }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version == "None"]
     requirements:
@@ -129,7 +131,9 @@ outputs:
         # Set CUDA architectures: starts at 6.0 because genai code uses __hfma2 (half-precision FMA)
         # which requires Compute Capability 6.0+ (Pascal). CC 5.0 (Maxwell) only supports FP16 storage, not arithmetic.
         # Compiling with 10.0 failed with error: nvcc fatal: Unsupported gpu architecture 'compute_100' - cuda 12.6 issue ?
-        - TORCH_CUDA_ARCH_LIST=6.0;7.0;7.5;8.0;8.6;9.0+PTX  # [cuda_compiler_version != "None"]
+        # TODO switch back to full arch list before merging
+        # - TORCH_CUDA_ARCH_LIST=6.0;7.0;7.5;8.0;8.6;9.0+PTX  # [cuda_compiler_version != "None"]
+        - TORCH_CUDA_ARCH_LIST=9.0  # [cuda_compiler_version != "None"]
       string: cuda{{ cuda_compiler_version | replace('.', '') }}_py{{ CONDA_PY }}_h{{ PKG_HASH }}_{{ PKG_BUILDNUM }}  # [cuda_compiler_version != "None"]
     requirements:
       build:


recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+  version: {{ version }}
+
+source:
+  # Need to use git_url instead of tarball due to extensive git submodules
+  git_url: https://github.com/pytorch/FBGEMM.git
+  git_rev: v{{ version }}
+
+build:
+  number: 0
+  skip: true  # [py<38]
```
h-vetinari: obsolete

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+  skip: true  # [win]
+  skip: true  # [aarch64]  # git_url source requires git on build system, problematic for cross-compilation
```
h-vetinari, line 16: Even if that were the case (which I doubt), we would be able to use `$BUILD_PREFIX/bin/git`

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+
+outputs:
+  - name: fbgemm
+    build:
+      script: |
+        git submodule update --init --recursive
```
h-vetinari: conda should normally default to recursive checkouts of submodules? Did you test that this is required?

As an aside, why not start out with a v1 recipe right away?

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+        {% if cuda_compiler_version == "None" %}
+        python setup.py --package_variant=cpu --package_channel=release install --prefix=$PREFIX --single-version-externally-managed --record=record.txt
+        {% else %}
+        python setup.py --package_variant=cuda --package_channel=release install --prefix=$PREFIX --single-version-externally-managed --record=record.txt
+        {% endif %}
+      script_env:
+        # Set CUDA architectures: starts at 6.0 because code uses __hfma2 (half-precision FMA)
+        # which requires Compute Capability 6.0+ (Pascal). CC 5.0 (Maxwell) only supports FP16 storage, not arithmetic.
+        # Compiling with 10.0 failed with error: nvcc fatal: Unsupported gpu architecture 'compute_100' - cuda 12.6 issue ?
+        - TORCH_CUDA_ARCH_LIST=6.0;7.0;7.5;8.0;8.6;9.0+PTX  # [cuda_compiler_version != "None"]
```
h-vetinari: yeah, `compute_100` needs CUDA 12.8+; this is easily fixable on the feedstock by applying the CUDA migrator

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+        - tabulate
+        - jinja2
+        - pyyaml
+        - cuda-cudart-dev  # [cuda_compiler_version != "None"]
+        - cuda-nvrtc-dev  # [cuda_compiler_version != "None"]
+        - cuda-nvtx-dev  # [cuda_compiler_version != "None"]
+        - libcublas-dev  # [cuda_compiler_version != "None"]
+        - libcusolver-dev  # [cuda_compiler_version != "None"]
+        - libcusparse-dev  # [cuda_compiler_version != "None"]
+        - libcurand-dev  # [cuda_compiler_version != "None"]
```
h-vetinari, line 103: all of this should be in `host:`, not `build:`

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+  git_rev: v{{ version }}
+
+build:
+  number: 0
+  skip: true  # [py<38]
+  skip: true  # [win]
+  skip: true  # [aarch64]  # git_url source requires git on build system, problematic for cross-compilation
+
+outputs:
+  - name: fbgemm
```
h-vetinari: IMO we should call this output 
```suggestion
  - name: libfbgemm
```

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+        - libcusolver-dev  # [cuda_compiler_version != "None"]
+        - libcusparse-dev  # [cuda_compiler_version != "None"]
+        - libcurand-dev  # [cuda_compiler_version != "None"]
+      host:
+        - python
+        - pip
+        - setuptools
+        - setuptools-git-versioning
+        - wheel
+        - pytorch
```
h-vetinari, line 110: Is there any reasonable/feasible way to use fbgemm without pytorch? Presumably yes, otherwise pytorch wouldn't be able to vendor it without fbgemm becoming entangled with core pytorch. I'm asking because it'd obviously be interesting to depend on `fbgemm` for the pytorch feedstock, and then we need to break this cycle.

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
+      run:
+        - python
+        - pytorch
+        - numpy
+        - cuda-version >={{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
+        - cuda-cudart  # [cuda_compiler_version != "None"]
+    test:
+      imports:
+        - fbgemm_gpu
```
h-vetinari, line 122: this is the correct import even for the cpu variant?

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+      commands:
+        - python -c "import fbgemm_gpu; print(fbgemm_gpu.__version__)"
```
h-vetinari: What is the purpose of this test?

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+      host:
+        - llvm-openmp  # [osx]
+        - libgomp  # [linux]
+      run:
+    test:
+      commands:
+        - test -f $PREFIX/lib/libfbgemm${SHLIB_EXT}  # [unix]
+        - test -f $PREFIX/include/fbgemm/FbgemmBuild.h  # [unix]
+
+  - name: fbgemm-gpu
```
h-vetinari: I suspect it might be better to use the name `fbgemm` for this output (though we can keep `fbgemm-gpu` as an alias wrapper if necessary)

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+        - fbgemm_gpu
+      commands:
+        - python -c "import fbgemm_gpu; print(fbgemm_gpu.__version__)"
+
+  - name: fbgemm-gpu-genai
+    build:
+      skip: true  # [cuda_compiler_version == "None"]
+      script: |
+        cd fbgemm_gpu
+        python setup.py --package_variant=genai --package_channel=release install --prefix=$PREFIX --single-version-externally-managed --record=record.txt
```
h-vetinari: This looks problematic to me; how does the `genai` variant interact with the `cpu/cuda` variants? I would have expected this output to depend on the the previous one. Otherwise, we would have to implement pretty complex mutex rules, which is definitely not the right approach here.

recipes/fbgemm/meta.yaml
```diff
@@ -0,0 +1,211 @@
...
+    build:
+      script: |
+        git submodule update --init --recursive
+        mkdir build && cd build
+        cmake \
+          -DCMAKE_INSTALL_PREFIX=$PREFIX \
+          -DCMAKE_PREFIX_PATH=$PREFIX \
+          -DCMAKE_BUILD_TYPE=Release \
+          -DFBGEMM_LIBRARY_TYPE=shared \
+          -DASMJIT_STATIC=OFF \
```
h-vetinari, line 28: Can we build this on top of https://github.com/conda-forge/staged-recipes/pull/31498#issuecomment-3704256250? You can have several recipes per PR; if `fbgemm` depends on `asmjit`, CI here will determine the DAG and build the recipes in the correct order (so that `fbgemm` can depend on `asmjit`)

recipes/fbgemm/meta.yaml
das-intensity: Will drop, but can you point me to the source of truth for this?

recipes/fbgemm/meta.yaml
das-intensity: > conda should normally default to recursive checkouts

Right you are, I think I just missed removing this output. Will drop.

> why not start out with a v1 recipe right away?

I think I read this: https://github.com/conda-forge/staged-recipes/blob/main/recipes/example-v1/README.md?plain=1#L3

> but is not yet fully supported by conda-forge's automation.

and figured I didn't know enough conda-forge to know whether what's "not yet fully supported" would come back to bite me (plus I was more familiar with legacy style).

recipes/fbgemm/meta.yaml
das-intensity: That does sound like a logical answer, but `fbgemm_gpu` is the actual module name.

If we changed `fbgemm` -> `libfbgemm` like you suggested, it would free up the package name, **but** from a user standpoint, the [docs](https://github.com/pytorch/FBGEMM?tab=readme-ov-file#project-overview) clearly show 3 related but distinct pacakges:

1. FBGEMM
2. FBGEMM_GPU   <--- this output package
3. FBGEMM_GPU_GENAI

(don't shoot me, I'm just the ~messenger~ maintainer)

So given this, calling this `fbgemm` would be confusing to users.

recipes/fbgemm/meta.yaml
das-intensity: Yep :( see my comment below here for full info: https://github.com/conda-forge/staged-recipes/pull/31820#discussion_r2656974409

I guess you can see the gpu-cpu insanity coming, but that's how it is on pypi also: https://pypi.org/project/fbgemm-gpu-cpu/ (vs the GPU variant https://pypi.org/project/fbgemm-gpu/)

recipes/fbgemm/meta.yaml
das-intensity: See my above comment https://github.com/conda-forge/staged-recipes/pull/31820#discussion_r2656974409, while it uses the same `setup.py` file it's really a whole different codebase.

I will say there's some overlap, but `fbgemm-gpu` and `fbgemm-gpu-genai` both have lots that the other doesn't.

recipes/fbgemm/meta.yaml
das-intensity: I think this is the cycle I was mentioning in https://github.com/conda-forge/staged-recipes/pull/31498#issuecomment-3568478700, but if `fbgemm` is getting booted (or already did) out of `pytorch` package, then that'd break the

> fbgemm <- pytorch <- fbgemm-gpu

cycle issue.

As to whether fbgemm can be used independently, I too believe so, but am unsure how many people actually would use it like that.

recipes/fbgemm/meta.yaml
h-vetinari: Lowest version we build for is [here](https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/b9400c9f43005c44e535c843f139ef4b3a7c42e7/recipe/conda_build_config.yaml#L885); the de facto policy is described [here](https://github.com/conda-forge/python-feedstock/issues/794)

recipes/fbgemm/meta.yaml
das-intensity: No real purpose, I think this is just me being new and wanting to be overly-covered in case test -> imports wasn't actually run anymore.

Assuming we can trust test -> imports I'll drop this

recipes/fbgemm/meta.yaml
h-vetinari: `fbgemm_gpu_cpu` is braindead naming for something CPU-only, and the rest is not much better. There's a fundamental difference between `fbgemm` (what I call `libfbgemm`), the C++ library, and `fbgemm_gpu` (what I want to call `fbgemm`), the python bindings. The upstream naming does not reflect that at all. 

> So given this, calling this `fbgemm` would be confusing to users.

I don't think so, or rather, it doesn't matter. If a user installs `fbgemm` (in my proposed naming), they'd get both the library that the want, as well as the python bindings (that they perhaps don't need). If that user cares about slimming down their environment, they can learn about the `libfbgemm` naming.

Summing up, this is what I mean. We can't stop upstream from loading a shotgun and blowing their feet off, but we don't have to follow suit on this particular aspect.

| thing | name upstream | name conda-forge<br/>(proposed) |
|--------|--------|--------|
| library | `fbgemm`<br/>(not installable via PyPI) | `libfbgemm` |
| python bindings<br/>(CUDA) | `fbgemm-gpu` | `fbgemm` (build string `cuda*`)<br/>(possible to keep upstream name as compat. wrapper) |
| python bindings<br/>(CPU) | `fbgemm-gpu-cpu` | `fbgemm` (build string `cpu*`)<br/>(possible to keep upstream name as compat. wrapper) |
| genAI extension | `fbgemm-gpu-genai` | `fbgemm-genai`<br/>(depending on `fbgemm`) |

CC @conda-forge/pytorch-cpu for viz.

das-intensity: I guess the lack of any `fbgemm-gpu` will ensure that users wanting that "perhaps look harder", and then perhaps we can start the description field with:

> Upstream name FBGEMM_GPU

so it shows up on anaconda search pages e.g. https://anaconda.org/search?q=pytorch

Will make the change!

das-intensity:
The cuda build timed out after 6hrs. Unfortunately I'm not surprised. It didn't take quite that long on my local machine IIRC, but my local is pretty powerful.

How can I go about debugging the pipeline machines? E.g. how can I know if it's fully pegging the CPU, such that perhaps a cmake/etc flag might help.

h-vetinari:
> How can I go about debugging the pipeline machines? E.g. how can I know if it's fully pegging the CPU, such that perhaps a cmake/etc flag might help.

Most effective is reducing the GPU arches for now (to a single one). We can switch to cirun once the feedstock is created.

recipes/fbgemm/meta.yaml
h-vetinari: > but is not yet fully supported by conda-forge's automation.

That comment is 2 years old. By now things work pretty much without a hitch.

recipes/fbgemm/meta.yaml
das-intensity: So interestingly it uses git to clone, but then says git isn't available. See here: https://gist.github.com/das-intensity/7f5a5bc9d238bbcd63940863a6ab3404

h-vetinari: Try adding
```
- git
- git-lfs
```
to the build environment. For whatever reason, the checkout procedure tries to look in there
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/conda/staged-recipes/build_artifacts/fbgemm_1767418241898/_build_env/bin/git'
```

`git-lfs` may not be strictly necessary, but addresses
```
git: 'lfs' is not a git command. See 'git --help'.
```

recipes/fbgemm/meta.yaml
h-vetinari: See #31836

recipes/fbgemm/meta.yaml
das-intensity: I suspect you're probably right about SOME of these, but I switched from:
```
      build:
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        - {{ compiler('cuda') }}  # [cuda_compiler_version != "None"]
        - {{ stdlib('c') }}
        - cmake
        - make
        - ninja
        - git 
        - python
        - pip 
        - setuptools-git-versioning
        - pytorch
        - pytorch * *cuda*  # [cuda_compiler_version != "None"]
        - scikit-build
        - tabulate
        - jinja2
        - pyyaml
        - cuda-cudart-dev  # [cuda_compiler_version != "None"]
        - cuda-nvrtc-dev  # [cuda_compiler_version != "None"]
        - cuda-nvtx-dev  # [cuda_compiler_version != "None"]
        - libcublas-dev  # [cuda_compiler_version != "None"]
        - libcusolver-dev  # [cuda_compiler_version != "None"]
        - libcusparse-dev  # [cuda_compiler_version != "None"]
        - libcurand-dev  # [cuda_compiler_version != "None"]
      host:
        - python
        - pip 
        - setuptools
        - setuptools-git-versioning
        - wheel
        - pytorch
        - scikit-build
        - numpy
        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
```
to
```
      build:
        - {{ compiler('c') }}
        - {{ compiler('cxx') }}
        - {{ compiler('cuda') }}  # [cuda_compiler_version != "None"]
        - {{ stdlib('c') }}
        - cmake
        - make
        - ninja
        - git 
      host:
        - python
        - pip 
        - setuptools
        - setuptools-git-versioning
        - wheel
        - pytorch
        - pytorch * *cuda*  # [cuda_compiler_version != "None"]
        - scikit-build
        - numpy
        - tabulate
        - jinja2
        - pyyaml
        - cuda-version {{ cuda_compiler_version }}  # [cuda_compiler_version != "None"]
        - cuda-cudart-dev  # [cuda_compiler_version != "None"]
        - cuda-nvrtc-dev  # [cuda_compiler_version != "None"]
        - cuda-nvtx-dev  # [cuda_compiler_version != "None"]
        - libcublas-dev  # [cuda_compiler_version != "None"]
        - libcusolver-dev  # [cuda_compiler_version != "None"]
        - libcusparse-dev  # [cuda_compiler_version != "None"]
        - libcurand-dev  # [cuda_compiler_version != "None"]
```
but the cuda build failed with:
```
[  7%] Building CXX object CMakeFiles/asmjit.dir/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/external/asmjit/src/asmjit/core/builder.cpp.o
/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_build_env/bin/x86_64-conda-linux-gnu-c++ -DPROTOBUF_USE_DLLS -DUSE_C10D_GLOO -DUSE_C10D_NCCL -DUSE_DISTRIBUTED -DUSE_RPC -DUSE_TENSORPIPE -Dasmjit_EXPORTS -isystem /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/python3.11/site-packages/torch/include -isystem /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/python3.11/site-packages/torch/include/torch/csrc/api/include -isystem /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/include -isystem /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_build_env/targets/x86_64-linux/include -DNO_AVX512=1 -O3 -DNDEBUG -std=c++20 -fPIC -Wno-deprecated-anon-enum-enum-conversion -Wno-deprecated-declarations -D_GLIBCXX_USE_CXX11_ABI=1 -MD -MT CMakeFiles/asmjit.dir/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/external/asmjit/src/asmjit/core/builder.cpp.o -MF CMakeFiles/asmjit.dir/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/external/asmjit/src/asmjit/core/builder.cpp.o.d -o CMakeFiles/asmjit.dir/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/external/asmjit/src/asmjit/core/builder.cpp.o -c /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/external/asmjit/src/asmjit/core/builder.cpp
In file included from /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/include/ATen/cuda/CUDAContext.h:3,
                 from /home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/work/fbgemm_gpu/src/embedding_inplace_ops/embedding_inplace_update_gpu.cpp:11:
/home/conda/staged-recipes/build_artifacts/fbgemm_1767562352076/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/include/ATen/cuda/CUDAContextLight.h:7:10: fatal error: cusparse.h: No such file or directory
    7 | #include <cusparse.h>
      |          ^~~~~~~~~~~~
compilation terminated.
```

PR #31824, opened 2026-01-01
Add cellucid

inecik:
Cellucid is a browser-first viewer for exploring large single-cell datasets in real time: fly through 2D/3D embeddings (UMAP/t-SNE/PCA), color by genes or metadata, filter and compare populations, and share reproducible views with collaborators.


inecik:
this is ready for review

PR #31834, opened 2026-01-03
Add python package `optmagic` to staged-recipes.

xapple:
Adding `optmagic` from PyPI for distribution on conda-forge.

PR #31835, opened 2026-01-03
Add python package `crest4` to staged-recipes.

xapple:
Adding `crest4` from PyPI for distribution on conda-forge.

### DIFF
--- a/recipes/crest4/meta.yaml
+++ b/recipes/crest4/meta.yaml
@@ -1,4 +1,4 @@
-{% set version = "4.4.5" %}
+{% set version = "4.4.6" %}
 
 package:
   name: crest4
@@ -6,7 +6,7 @@ package:
 
 source:
   url: https://pypi.org/packages/source/c/crest4/crest4-{{ version }}.tar.gz
-  sha256: 4b2282244b2184af91555561154723c024a3b597fbf2cc72df665874ce3f60e5
+  sha256: 9bac26cb2f00aa1d6c4759fa801d85829702142fba4a9e7bdb0ccb1ca0deb625
 
 build:
   noarch: python
@@ -20,17 +20,17 @@ requirements:
     - setuptools
   run:
     - python >={{ python_min }}
-    - autopaths =>1.6.3
-    - plumbing =>2.11.4
-    - fasta =>2.3.6
-    - seqsearch =>2.2.4
-    - optmagic =>1.1.1
-    - biopython =>1.86
-    - rich =>14.2.0
-    - ete4 =>4.3.0
-    - pandas =>2.3.3
-    - pytest =>9.0.2
-    - pytest-asyncio =>1.3.0
+    - autopaths >=1.6.3
+    - plumbing >=2.11.4
+    - fasta >=2.3.6
+    - seqsearch >=2.2.4
+    - optmagic >=1.1.1
+    - biopython >=1.86
+    - rich >=14.2.0
+    - ete4 >=4.3.0
+    - pandas >=2.3.3
+    - pytest >=9.0.2
+    - pytest-asyncio >=1.3.0
 
 test:
   imports:


xapple:
This recipe depends on two another packages that I also submitted for review earlier (so those ones should be accepted first before this one builds):
- [`seqsearch`](https://github.com/conda-forge/staged-recipes/pull/31833)
- [`optmagic`](https://github.com/conda-forge/staged-recipes/pull/31834)

xapple:
So the osx_64 build failed with this error:

```
2026-03-18T07:33:46.6229140Z conda.CondaMultiError: Encountered an SSL error. Most likely a certificate verification issue.
2026-03-18T07:33:46.6230710Z Exception: [SSL] record layer failure (_ssl.c:2580)
```

I'm going to assume that's just transient.

xapple:

recipes/crest4/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+    - plumbing ==2.11.4
+    - fasta ==2.3.6
+    - seqsearch ==2.2.4
+    - optmagic ==1.1.1
+    - biopython ==1.86
+    - rich ==14.2.0
+    - ete4 ==4.3.0
+    - pandas ==2.3.3
+    - pytest ==9.0.2
+    - pytest-asyncio ==1.3.0
```
ocefpaf: Are all these pins upstream? That will make it very hard to install this package along side anything else.

Does this package really requires pytest as a runtime dependency?

recipes/crest4/meta.yaml
```diff
@@ -0,0 +1,52 @@
...
+package:
+  name: crest4
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/c/crest4/crest4-{{ version }}.tar.gz
+  sha256: 4b2282244b2184af91555561154723c024a3b597fbf2cc72df665874ce3f60e5
+
+build:
+  skip: true  # [win or arm64 or aarch64]
```
ocefpaf: Looks like this package should be noarch.

recipes/crest4/meta.yaml
xapple: The strict pins of the package are to guarantee absolute reproducibility of the results. You see, this is a scientific package producing results that can get published, and thus need to be recreated by any reviewers or other labs exactly. Besides, the first 5 packages in the list are other conda-forge recipes that I maintain so that also reduces the issue.

About the `pytest` dependency, I debated a while to include it or not. But once again the users are expected to thoroughly check their installation to guarantee the computations are consistent. In the command line tool that is user-facing "--pytest" is an argument advertised in the help text and can be inputed to run the test suite.

What do you think?

recipes/crest4/meta.yaml
xapple: Yes actually you are right. It can't run on windows though, because one of its dependencies is missing there (ete4). But, I have updated the meta.yaml to mark it as `noarch`.

xapple:
This is what I feared. Now the checks don't pass, because it can't build on windows. The error is:


```
conda_libmamba_solver.conda_build_exceptions.ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: {MatchSpec("ete4=4.3.0")}
  - unsupported request

\u2514\u2500 ete4 =4.3.0 * does not exist (perhaps a typo or a missing channel).
##[error]Cmd.exe exited with code '1'.
```

@ocefpaf What would you recommend in such a case? Thanks!

recipes/crest4/meta.yaml
ocefpaf: Pins should be in apps or scientific workflows. In packages they only bring pain and suffering. I understand that people confuse the differences between applications and packages. but reproducibility is not broken or ensured with some pin relaxation. It is achieved via proper testing upstream.

ocefpaf:
> This is what I feared. Now the checks don't pass, because it can't build on windows. The error is:
> 
> ```
> conda_libmamba_solver.conda_build_exceptions.ExplainedDependencyNeedsBuildingError: Unsatisfiable dependencies for platform win-64: {MatchSpec("ete4=4.3.0")}
> Encountered problems while solving:
>   - unsupported request
> 
> Could not solve for environment specs
> The following package could not be installed
> \u2514\u2500 ete4 =4.3.0 * does not exist (perhaps a typo or a missing channel).
> ##[error]Cmd.exe exited with code '1'.
> ```
> 
> @ocefpaf What would you recommend in such a case? Thanks!

Nothing. It's okay if a noarch package fails on a specific platform due to a missing dependency. We prefer using noarch because it's faster to build, and the package will work as soon as the missing dependency becomes available on that platform.

xapple:
Ah I had no idea is was okay to merge even if all the checks don't pass. Well that's good news.
I have now unpinned the versions of the dependencies.
Anything else @ocefpaf ?

PR #31843, opened 2026-01-04
Add phast 1.9.3

asiepel:
Initial conda-forge recipe for PHAST (CMake-based phylogenetic analysis toolkit).


asiepel:
Thanks for the review.

I’ve since confirmed that PHAST is already packaged in Bioconda under the same name (for this same project), though it has not been updated recently. To avoid duplication and naming conflicts, I will update and maintain PHAST in Bioconda instead.

Closing this PR accordingly.

PR #31844, opened 2026-01-05
Add vine

asiepel:
Initial conda-forge recipe for VINE (variational inference toolkit; depends on PHAST).


asiepel:
This recipe depends on PHAST, which is already provided by Bioconda and will be updated there. I will submit VINE to Bioconda instead to avoid cross-channel dependency issues.

Closing this PR.

PR #31860, opened 2026-01-07
Add recipe for tea (Gitea CLI)

pavelzw:
This PR adds a recipe for [tea](https://gitea.com/gitea/tea), the official CLI for Gitea.

**Summary:**
- tea is a command line tool to interact with Gitea servers
- Version: 0.11.1
- License: MIT

**Build tested locally with:**
```
rattler-build build -r recipes/tea -m .ci_support/osx_arm64.yaml
```

All tests passed.

### DIFF
--- a/recipes/tea/recipe.yaml
+++ b/recipes/tea/recipe.yaml
@@ -1,13 +1,12 @@
 context:
-  name: tea
   version: "0.11.1"
 
 package:
-  name: ${{ name }}
+  name: tea
   version: ${{ version }}
 
 source:
-  - url: https://gitea.com/gitea/${{ name }}/archive/v${{ version }}.tar.gz
+  - url: https://gitea.com/gitea/tea/archive/v${{ version }}.tar.gz
     sha256: 1da6b6d2534bd6ffb0931400014bbdef26242cf4d35d4ba44c24928811825805
     target_directory: src
   - if: win


pavelzw:
@conda-forge/help-go this is ready for review

recipes/tea/recipe.yaml
```diff
@@ -0,0 +1,61 @@
+context:
+  name: tea
+  version: "0.11.1"
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
```
danielnachun, line 6:
```suggestion
package:
  version: ${{ version }}
```
This Jinja key is unneeded.

recipes/tea/recipe.yaml
```diff
@@ -0,0 +1,61 @@
...
+source:
+  - url: https://gitea.com/gitea/tea/archive/v${{ version }}.tar.gz
+    sha256: 1da6b6d2534bd6ffb0931400014bbdef26242cf4d35d4ba44c24928811825805
+    target_directory: src
+  - if: win
+    then:
+      url: https://github.com/mattn/go-localereader/raw/master/LICENSE
+      sha256: dceede03550621e65ec7b79fcba87cf6dbb9d31132a8ae71a8fe8ceec0af7da3
+      target_directory: go-localereader-license
+      file_name: LICENSE
```
danielnachun, line 17: I'm not firmly opposed to downloading licenses like this, but I'll note here that this can get very messy if the recipe adds more dependencies with missing licenses.  I've so far been advising contributors to instead follow the approach I used in https://github.com/conda-forge/go-gum-feedstock where the license files are added to the recipe itself, and then can be copied en masse.  Otherwise having to hard-code individual licenses in the recipe could become quite tedious.

recipes/tea/recipe.yaml
pavelzw: this is incorrect, `.package` requires both `name` and `version` to be set.

recipes/tea/recipe.yaml
pavelzw: > this can get very messy if the recipe adds more dependencies with missing licenses

i agree but as long as we only have a hand full, it's fine. I find direct downloads a bit easier to handle, otherwise you might end up with files that you don't need anymore. also, the pinned sha256 is a nice bonus that you actually get notified once the license changed. when checking them into the repo, you might end up with stale licenses

recipes/tea/recipe.yaml
```diff
@@ -0,0 +1,61 @@
...
+  version: "0.11.1"
+
+package:
+  name: ${{ name }}
+  version: ${{ version }}
+
+source:
+  - url: https://gitea.com/gitea/${{ name }}/archive/v${{ version }}.tar.gz
+    sha256: 1da6b6d2534bd6ffb0931400014bbdef26242cf4d35d4ba44c24928811825805
+    target_directory: src
```
danielnachun, line 11:
```suggestion
context:
  version: "0.11.1"

package:
  name: tea
  version: ${{ version }}

source:
  - url: https://gitea.com/gitea/tea/archive/v${{ version }}.tar.gz
    sha256: 1da6b6d2534bd6ffb0931400014bbdef26242cf4d35d4ba44c24928811825805
    target_directory: src
```
Sorry this is the change I wanted to suggest, I made a mistake before.

pavelzw: while correct, i still try to keep as close to the official recommendations as possible: https://conda-forge.org/docs/maintainer/example_recipes/go/

here we can see 

```yaml
context:
  name: example-package
  version: "0.1.0"

package:
  name: ${{ name|lower }}
  version: ${{ version }}

source:
  url: https://github.com/example-package/${{ name }}/archive/refs/tags/v${{ version }}.tar.gz
  sha256: 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
  target_directory: src
```

i updated the recipe accordingly to use `${{ name }}` in the url as well in [`f1d870a` (#31860)](https://github.com/conda-forge/staged-recipes/pull/31860/commits/f1d870ade0bd92b5cf89df03e4c66a1f67299f50)

danielnachun: I see, I need to update the example because we want to move away using `name` as a Jinja key altogether.  See https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/.E2.9C.94.20Motivation.20for.20.60name.60.20field.20in.20.60context.60.20for.20recipes.3F/with/567005291 for the explanation, but essentially this is a relic from a different time that should be removed for at least two reasons:
- It negatively impacts readability of the URL
- It incorrectly implies that the package name can be changed trivially, which is not true - changing the name requires filing an admin request to change the package outputs, and the new name must be checked for any collisions with existing conda-forge and bioconda packages.

Does that seem like an acceptable reason to go against the example?  I don't want to block merging of this package over a minor issue, but I guess this a bit of weird situation where I've only recently become aware that some of our examples are still using this.  Let me know what you think!

pavelzw:
thanks for the review @danielnachun! PTAL again

recipes/tea/recipe.yaml
pavelzw: example updated in https://github.com/conda-forge/conda-forge.github.io/pull/2713

PR #31888, opened 2026-01-09
Add `cascode` package

daniellovell:
## Adding cascode

Cascode is a synthesized analog description language (ADL) for rapid, portable analog/mixed-signal design. It provides a CLI for compiling `.cas` files to ACIR (Analog Circuit Intermediate Representation) and generating SPICE netlists.

Github Repo: https://github.com/daniellovell/cascode

Checklist

- [x] Build number is 0.


daniellovell:
@conda-forge-admin please ping staged-recipes

daniellovell:
It is a C#/dotnet package, so maybe conda-forge/help-c-cpp is the closest?

daniellovell:
@conda-forge-admin please ping conda-forge/help-c-cpp (I believe that would be the closest/most relevant to this dotnet package)

traversaro:
> This package ships prebuilt self-contained .NET binaries from GitHub Releases (no .NET runtime dependency)

Binary repackaging is in general discouraged in conda-forge, see https://conda-forge.org/docs/maintainer/guidelines/#compiled-objects . Can you explain the reason behind this? Once you motivated this, we need to get an explicit approval from core members.

daniellovell:
Hi @traversaro, thanks for taking a look!

I think the sentence you've highlighted could have been worded better. This is not binary repackaging - the full source is available, its .NET/C# app that is open source here: https://github.com/daniellovell/cascode 

We build from the source in CI using `dotnet publish` which is self-contained (bundles the .NET runtime, like Java does nowadays so that it does not require the JRE )

The conda recipe downloads our own CI-built artifacts

So to clarify,
1. The full source is publicly available
2. All binaries are built from source in public CI
3. The app is fully self-contained with no runtime dependencies
4. This is a distribution of our own software, not repackaging of third party binaries

It's the same pattern used by other .NET-based tools distributed via package managers (like the `dotnet` tool itself)

I hope this clarifies, sorry about that!

daniellovell:
@traversaro I've reimplemented the build process here such that the recipe does not vendor binaries. We pull source and build from source in the recipe :) 

Thanks again!

traversaro:
> I think the sentence you've highlighted could have been worded better.

I tried to clarified the requirements in https://github.com/conda-forge/conda-forge.github.io/pull/2717, let me know if it is more clear.

daniellovell:
@traversaro Just to clarify, I meant my own sentence could have been worded better, but yes I agree that the clarification you proposed in https://github.com/conda-forge/conda-forge.github.io/pull/2717 would have caught my misunderstanding here :)

daniellovell:
I believe I've revised the recipe in this PR to meet that requirement - the recipe pulls the source and builds in conda-forge CI.  The linter/build checks all pass so it should be ready for re-review.

@conda-forge-admin please ping conda-forge/help-c-cpp

traversaro:
I do not have a lot of experience in dotnet/C#, perhaps @conda-forge/staged-recipes or @conda-forge/dotnet may have better inputs. @daniellovell just to understand, there is any specific reason why you are tagging @conda-forge/help-c-cpp in this PR?

recipes/cascode/build.sh
```diff
@@ -0,0 +1,33 @@
...
+  *)
+    echo "Unsupported target_platform: ${target_platform}" >&2
+    exit 1
+    ;;
+esac
+
+dotnet publish tools/cli/Cascode.Cli.csproj \
+  -c Release \
+  -r "${rid}" \
+  -p:SelfContained=true \
```
dhirschfeld: I don't think this should be self-contained as that vendors the dotnet runtime.

Instead, you should have a run dependency on the `dotnet-runtime` package.

daniellovell: Sure, I can make that change!

daniellovell:
@traversaro 

> I do not have a lot of experience in dotnet/C#, perhaps @conda-forge/staged-recipes or @conda-forge/dotnet may have better inputs. @daniellovell just to understand, there is any specific reason why you are tagging @conda-forge/help-c-cpp in this PR?

I don't see that there is an @conda-forge/help-dotnet tag in the bot's directory, so I chose `help-c-cpp` as the closest family of languages to C# (though `help-java`might be "closer" given its a managed runtime language..?)

recipes/cascode/build.sh
daniellovell: Done! 😄

recipes/cascode/meta.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - /usr/lib/swift/libswift*.dylib  # [osx]
+
+requirements:
+  build:
+    - dotnet-sdk
+    - {{ stdlib('c') }}
+  run:
+    - dotnet-runtime >=10.0
+    - libgcc  # [linux]
+    - libstdcxx  # [linux]
```
dhirschfeld: I'm not sure if the application built against .NET 10 would run with the .NET 11 or 12 runtime. Maybe the runtimes are forward compatible? If not, you could pin them:

```suggestion
requirements:
  build:
    - dotnet-sdk ~=10.0
    - {{ stdlib('c') }}
  run:
    - dotnet-runtime ~=10.0
    - libgcc  # [linux]
    - libstdcxx  # [linux]
```

daniellovell: They should be forward compatible, but I agree being conservative & pinning has limited downside & seems prudent.

dhirschfeld (approved):
Other than one take-it-or-leave-it comment, it seems ok to me now it's not vendoring the .NET runtime.

recipes/cascode/meta.yaml
dhirschfeld: It can always be relaxed later.

dhirschfeld:
LGTM, but you'll have to wait for someone with merge rights to actually do the honours.

:shipit:

recipes/cascode/bld.bat
```diff
@@ -0,0 +1,4 @@
+@echo off
+mkdir %PREFIX%\Scripts
+copy cascode.exe %PREFIX%\Scripts\
+if errorlevel 1 exit 1
```
traversaro: Can you delete this file if we are not building on Windows?

traversaro (changes requested):

recipes/cascode/bld.bat
daniellovell: Shoot, good catch. I did not mean to disable Windows - there is no reason to skip it.

daniellovell: Resolved by re-enabling Windows

daniellovell:
Bumping this for re-review since addressing the requested changes from @traversaro 

@conda-forge-admin please ping conda-forge/help-c-cpp

daniellovell:
Bumping this maybe redirecting to conda-forge/help-java since this is a C#/dotnet package which is similarly a managed runtime language?

@conda-forge-admin please ping conda-forge/help-java

dhirschfeld:
You might get more visibility messaging on the Zulip chat.

danielnachun:
Does this package actually have any compiled code?  We have many .NET packages now which are `noarch: generic` - see https://github.com/conda-forge/t4-feedstock for an example of how to do this.  I can also push a commit here if you are unsure how to implement this.

dhirschfeld:
> *We have many .NET packages now which are `noarch: generic`*

Good to know! I'll keep that in mind for future .NET recipes. If you're otherwise happy with the PR I'd probably push the change and merge to avoid the inevitable async comms delay.

daniellovell:
Thanks @danielnachun (nice name btw 😉 )
My only concern with this is that the package has two NuGet dependencies with native components - SQLitePCLRaw and Google.OrTools. 

noarch package bundles native libraries for all platforms instead of one, so that means we would have:

- OrTools linux-x64: 18.7 MB, linux-arm64: 17.3 MB, osx-x64: 15.3 MB, osx-arm64: 13.6 MB, win-x64: 10.7 MB - **75.6 MB total**
- SQLite (SourceGear.sqlite3, all platforms): **33.3 MB**
- Grand total native deps: ~109 MB compressed, ~200-300 MB installed

By  comparison, the per-platform build includes only one platform's native builds @ 20-35MB each.

I think I'm okay with this 10x package size increase for the benefit of recipe simplicity and single CI build (and unblocking the PR 😛 )

I'll implement noarch and update this PR momentarily

daniellovell:
I've made the requested change to adopt `noarch: generic` 

To reduce some of this bloat I'm adding a pruning step to keep only the runtime assets for the conda-forge target platforms. I feel like this is not really idiomatic with the intent of `noarch` but shipping `ios` and `android` runtime assets for a desktop CLI app also feels worse.

daniellovell:
In retrospect, I’m rolling back the `noarch: generic` attempt for this recipe.

I originally was onboard that `noarch` would simplify things (single build, simpler matrix), and technically it does run. But in retrospect, because Cascode vendors NuGet packages with native binaries (SQLitePCLRaw + OR-Tools), CI turned into binary-repackaging territory very quickly.

  - Linux CI started failing with overlinking errors on vendored `runtimes/*/native/*.so` assets.
  - Windows CI failed to find `cascode` in test env without extra launcher/script handling.
  - I also had to add runtime pruning logic to keep package bloat under control.

I do think `noarch` in conceptually possible here, but I think it adds more fragility for this particular package 😓

danielnachun:
I didn't realize this had native dependencies, that is rare for .NET packages, but it does happen.  Can you give some idea of how hard it would be to build the native dependencies from source?  This would be particularly important for OR-tools, which ideally should be using the conda-forge libraries like CBC etc. instead of vendoring its own copies.  

I realize this is an unusual request as one of the selling points of the .NET and JDK ecosystems is not having to deal with compiling native code.  But there are a lot of issues with shipping pre-compiled binaries so our general policy on conda-forge is to build native code from source except for closed source binaries where this is not possible.

PR #31903, opened 2026-01-11
add model-hosting-containers-standards

h-vetinari:
New dependency required by https://github.com/conda-forge/vllm-feedstock/pull/20

### DIFF
--- a/recipes/model-hosting-containers-standards/recipe.yaml
+++ b/recipes/model-hosting-containers-standards/recipe.yaml
@@ -46,7 +46,6 @@ tests:
       pip_check: true
   - script:
       - generate-supervisor-config -h
-      - standard-supervisor -h
 
 about:
   homepage: https://github.com/aws/model-hosting-container-standards


h-vetinari:
This is ready @conda-forge/help-python. windows cannot work because https://github.com/conda-forge/supervisor-feedstock is unix-only.

recipes/model-hosting-containers-standards/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+source:
+  url: https://pypi.org/packages/source/m/model-hosting-container-standards/model_hosting_container_standards-${{ version }}.tar.gz
+  sha256: 27a1333410dde2719286a300a2803e24fdde407baa91894eb845c0f268aa194d
+
+build:
+  number: 0
+  noarch: python
+  script:
+    - poetry build
+    - pip install -f dist model-hosting-container-standards
```
danielnachun, line 17: What's the reason we would want to use this over the standard `${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation` we usually use?  There doesn't seem to be anything wrong with installing this way, I just haven't seen a Python package installation done this way before.

recipes/model-hosting-containers-standards/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+    - jmespath
+    - pydantic
+    - starlette >=0.49.1
+    - supervisor >=4.2.0
+
+tests:
+  - python:
+      python_version:
+        - ${{ python_min }}.*
+        - 3.*
```
danielnachun, line 43: Is this second line `3.*` to test on multiple versions?  I haven't seen this before but maybe it's a new feature.

recipes/model-hosting-containers-standards/recipe.yaml
h-vetinari: The project seems to require poetry to build (no other documented install instructions in its README), which is relatively rare. The primary installation variant suggested upstream -- `poetry install` -- does stuff we don't want to do, but `poetry build` just creates a wheel, which we then install from the dist folder. Before pushing here, I looked around in conda-forge for packages built by poetry and took this pattern from [here](https://github.com/conda-forge/uhd-feedstock/blob/d6920ad4565dcd4d6a4438c9a78a658a1949ecf3/recipe/bld.bat#L79-L81). I haven't investigated further if it'd work with `pip install`. Feel free to push into this PR if you'd like to try that.

recipes/model-hosting-containers-standards/recipe.yaml
h-vetinari: Yup, it's v1-only, see https://github.com/conda-forge/conda-forge.github.io/issues/2617.

h-vetinari:
> we don't need the windows build, right?

Yes, as I noted above:

> windows cannot work because https://github.com/conda-forge/supervisor-feedstock is unix-only.

---

> this is a blocker for getting new VLLM versions out.

The much heavier lifting will be flashinfer (also new dependency in 0.11). I briefly looked upstream and IMO this will not be straightforward. That shouldn't stop anyone from trying though; I'll try to help where I can.

PR #31912, opened 2026-01-12
Adding the Python package `luminet`

bgmeulem:
Checklist

- [x] Build number is 0.

### DIFF
--- a/.azure-pipelines/azure-pipelines-linux.yml
+++ b/.azure-pipelines/azure-pipelines-linux.yml
@@ -22,10 +22,17 @@ jobs:
     displayName: Manage disk space
 
   - script: |
-      sudo fallocate -l 10GiB /swapfile || true
-      sudo chmod 600 /swapfile || true
-      sudo mkswap /swapfile || true
-      sudo swapon /swapfile || true
+      SWAPFILE=/swapfile
+      # If there is already a swapfile, disable it and remove it
+      if swapon --show | grep -q $SWAPFILE; then
+          echo "Disabling existing swapfile..."
+          sudo swapoff $SWAPFILE || true
+      fi
+      [ -f "$SWAPFILE" ] && sudo rm -f $SWAPFILE
+      sudo fallocate -l 10GiB $SWAPFILE
+      sudo chmod 600 $SWAPFILE
+      sudo mkswap $SWAPFILE
+      sudo swapon $SWAPFILE
     displayName: Create swap file
 
   - script: |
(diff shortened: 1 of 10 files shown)


bgmeulem:

recipes/luminet/meta.yaml
```diff
@@ -0,0 +1,47 @@
+{% set name = "luminet" %}
+{% set version = "0.2.3" %}
+{% set python_min = "3.8" %}
```
MridulS: The global conda-forge pin of python_min is 3.10 right now so this is not needed. It is only used if the needed python min is higher than the global pin.

recipes/luminet/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+    - numpy >=1.14,<3
+    - scipy >=0.17.0,<2
+
+test:
+  imports:
+    - luminet
+  commands:
+    - pip check
+  requires:
+    - pip
```
MridulS, line 34: Hmm, the pip here may not be needed.

recipes/luminet/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+  name: {{ name|lower }}
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/{{ name[0] }}/{{ name }}/luminet-{{ version }}.tar.gz
+  sha256: 42fa9b8e18b20f3e92c66532cb8f234dcc6ce70de208193b13fd1d71a362dfe0
+
+build:
+  noarch: python
+  script: {{ PYTHON }} -m pip install . -vv
```
MridulS: The usual recommendation is to use `${{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation`, any specific reason to change that here?

recipes/luminet/meta.yaml
bgmeulem: I need the `pip` dependency for the `pip check` test. I read on [PyOpenSci](https://www.pyopensci.org/python-package-guide/tutorials/publish-conda-forge.html#step-4-tests-for-conda-forge) this was a recommended check to see if installation dependencies check out. If this is already done by `conda-forge` or otherwise superfluous, I can remove it?

recipes/luminet/meta.yaml
bgmeulem: Nope. Don't know why I have a different one. Fixed in most recent commit. Thanks for letting me know!

recipes/luminet/meta.yaml
travishathaway: When conda environments are created with Python, pip is automatically added unless you tell it not to via configuration.

That being said, it doesn't hurt anything to leave this in there either. I personally think it's okay.

ocefpaf: @MridulS for v1 it is not necessary b/c pip check is a built-in test. We do need it here though b/c this recipe is the old format.

PR #31933, opened 2026-01-15
Add new noarch python packages

claudiushaag:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/pymicro/recipe.yaml
+++ b/recipes/pymicro/recipe.yaml
@@ -1,0 +1,62 @@
+context:
+  version: "0.6.1"
+
+package:
+  name: pymicro
+  version: ${{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pymicro/pymicro-${{ version }}.tar.gz
+  sha256: 2cac40fd5f422cf2e8708abfc54c4995ed6948b3f58701e0117e370a7d97f145
+
+build:
+  noarch: python
+  script:
+    env:
+        # as set in the upstream setup.py
+        BASICTOOLS_DISABLE_MKL: 1
+        BASICTOOLS_DISABLE_OPENMP: 1
+    content: python -m pip install . -vv
+  number: 0
+
+
+requirements:
+  host:
+    - python ${{ python_min }}.*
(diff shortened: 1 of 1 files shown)


claudiushaag:
@conda-forge/help-python This is ready for review! These are just some noarch python packages I wanted to bring to cf, they do not depend on each other.
Only anomaly is the disabled `pip check` on `pymicro`, which is because of the dependency `basictools`, which does not pass it itself, and is archived by now.
Thanks for having a look!

recipes/pymicro/recipe.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - requests
+    - pytest
+    - basictools
+    - pyqt
+
+tests:
+  - python:
+      imports:
+        - pymicro
+      # pip check disabled because the dependency basictools does not pass it, but feedstock is archived by now
```
ocefpaf, line 50: We should unarchive  and fix it if other pacakges will be using it. Do you mind sending the unarchive request for it?

claudiushaag: I agree with your intention, let me add a little context:
The basictools package is archived because there was a namechange with the update to major version 2: [archiving-PR](https://github.com/conda-forge/admin-requests/pull/1396), [archiving-issue](https://github.com/conda-forge/basictools-feedstock/issues/55).
pymicro itself will move to the new package called `muscat`, as discussed [here](https://github.com/heprom/pymicro/issues/27), and it would be the first package to depend on basictools in the channel as of right now. My initial idea was to turn pip check on as soon as the new version with the `muscat`-dependency was released. And even in this package there are potential problems with `pip check`, as I pointed out in this [issue](https://github.com/conda-forge/eigency-feedstock/issues/26).

But I can also hold back on adding the package until this is done to ensure our standards, which I'm also perfectly fine with. What do you think?

ocefpaf: OK. In that case it is not worth the effort.

claudiushaag: Thank you!

PR #31940, opened 2026-01-16
Add libcuobjclient

adibbley:
Checklist

- [x] Build number is 0.


recipes/libcuobjclient/build.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Install to conda style directories
+[[ -d lib64 ]] && mv lib64 lib
+mkdir -p ${PREFIX}/lib
+mkdir -p ${PREFIX}/lib/pkgconfig
+[[ -d pkg-config ]] && mv pkg-config/* ${PREFIX}/lib/pkgconfig/
+[[ -d "$PREFIX/lib/pkgconfig" ]] && sed -E -i "s|cudaroot=.+|cudaroot=$PREFIX|g" $PREFIX/lib/pkgconfig/cuobjclient*.pc
```
billysuh7, line 8: I wonder if we need to package PC files. E.g. NVPL team provided PC files and said they need not be packaged for conda format because conda works in their own env where $PREFIX is set and thus can find all libs relative to this, without the need for PC files.

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_version }}
+        - libcufile
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - {{ pin_compatible("libcufile", max_pin="x.x") }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
billysuh7: Since `arm-variant` is specified in the `build:` section, the `arm-variant` here should no longer be needed because it has `strong` `run_exports` from https://github.com/conda-forge/arm-variant-feedstock/blob/main/recipe/meta.yaml#L13

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
+        - {{ stdlib("c") }}
+      host:
+        - cuda-version {{ cuda_version }}
+      run:
+        - {{ pin_compatible("cuda-version", max_pin="x.x") }}
+        - {{ pin_subpackage("libcuobjclient", exact=True) }}
+      run_constrained:
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
billysuh7: Same here

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+  sha256: bd88f9e3ce8499d94faf444c176fd6d9f461337952a7e9b0ccceba0b77fd4a3e  # [aarch64]
+
+build:
+  number: 0
+  skip: true  # [not (linux64 or aarch64)]
+
+requirements:
+  build:
+    - cf-nvidia-tools 1.*  # [linux]
+    - patchelf <0.18.0     # [linux]
```
billysuh7, line 28: I'm not sure if still need to constrain the `patchelf` version here?

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,133 @@
+{% set name = "libcuobjclient" %}
+{% set version = "1.0.0.26" %}
+{% set lib_version = version.split(".")[:-1] | join(".") %}
+{% set cuda_version = "13.1" %}
```
billysuh7, line 4: We are doing this for upcoming 13.2, correct ?

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,133 @@
...
+    - patchelf <0.18.0     # [linux]
+
+outputs:
+  - name: libcuobjclient
+    build:
+      binary_relocation: false
+      missing_dso_whitelist:
+        - "*libmlx5*"
+        - "*librdmacm*"
+        - "*libnuma*"
```
billysuh7: `libnuma` package is found in `conda-forge` and `defaults` channel now so we should be able to use the conda version here instead of white-listing and relegating to system-installed one?

recipes/libcuobjclient/build.sh
adibbley: CTK components still include them so I think we should have them here for consistency. For non-CTK components it is much less likely for us to add pc files in general.

recipes/libcuobjclient/meta.yaml
adibbley: The client libraries can be added for 13.1.1. cuObject server shipped with a reduced support matrix, but we should be able to add those conda packages in a future release.

recipes/libcuobjclient/meta.yaml
adibbley: Thanks, dropped these.

recipes/libcuobjclient/meta.yaml
adibbley: It does seem to be pinned in general now, getting `0.17.2-h58526e2_0` without the constraint so went ahead and dropped it here.

recipes/libcuobjclient/meta.yaml
adibbley: Added, thanks!

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+  description: |
+    cuObject (GPUDirect Storage for Objects) is a high performance suite of libraries 
+    designed to enable direct data transfers between GPU memory or system memory and S3 
+    compatible object storage using RDMA
+  doc_url: https://docs.nvidia.com/gpudirect-storage/cuobject/index.html
+
+extra:
+  recipe-maintainers:
+    - conda-forge/cuda
+  feedstock-name: libcuobjclient
```
jakirkham: Would swap the order here

```suggestion
extra:
  feedstock-name: libcuobjclient
  recipe-maintainers:
    - conda-forge/cuda
```

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+  url: https://developer.download.nvidia.com/compute/cuda/redist/{{ name }}/{{ platform }}/{{ name }}-{{ platform }}-{{ version }}-archive.{{ extension }}
+  sha256: aba19ffc2425b9981020d182f27093a52d150274482645621eefa029f276ef7c  # [linux64]
+  sha256: bd88f9e3ce8499d94faf444c176fd6d9f461337952a7e9b0ccceba0b77fd4a3e  # [aarch64]
+
+build:
+  number: 0
+  skip: true  # [not (linux64 or aarch64)]
+
+requirements:
+  build:
```
jakirkham:
```suggestion
  build:
    - {{ stdlib("c") }}
```

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    - cf-nvidia-tools 1.*  # [linux]
+    - patchelf             # [linux]
```
jakirkham: The GLIBC version is used by `cf-nvidia-tools`. So we need to make sure `{{ stdlib("c") }}` is added here so GLIBC is defined

Also we want to add a version guard to ensure `patchelf` does not run into this bug: https://github.com/NixOS/patchelf/issues/492

```suggestion
  build:
    - {{ stdlib("c") }}
    - cf-nvidia-tools 1.*  # [linux]
    - patchelf <0.18.0     # [linux]
```

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+        - "*librdmacm*"
+        - "*libibverbs*"
+    files:
+      - lib/libcuobjclient*.so.*
+      - targets/{{ target_name }}/lib/libcuobjclient*.so.*
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
carterbox: arm-variant is no longer relevant/needed after CTK 13.0.

```suggestion
```

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,130 @@
...
+    files:
+      - lib/libcuobjclient*.so
+      - lib/pkgconfig
+      - targets/{{ target_name }}/include
+      - targets/{{ target_name }}/lib/libcuobjclient*.so
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
+        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```
carterbox:
```suggestion
```

billysuh7 (approved):
LGTM but I'd leave @jakirkham to take a look at the CI errors

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+      license_url: https://docs.nvidia.com/cuda/eula/index.html
+      summary: Library for NVIDIA GPUDirect Storage for Objects
+      description: |
+        cuObject (GPUDirect Storage for Objects) is a high performance suite of libraries 
+        designed to enable direct data transfers between GPU memory or system memory and S3 
+        compatible object storage using RDMA
+      doc_url: https://docs.nvidia.com/gpudirect-storage/cuobject/index.html
+
+  - name: libcuobjclient-dev
+    build:
```
jakirkham, line 78:
```suggestion
    build:
      binary_relocation: false
```

recipes/libcuobjclient/meta.yaml
```diff
@@ -0,0 +1,128 @@
...
+        - "*libmlx5*"
+        - "*librdmacm*"
+        - "*libibverbs*"
+    files:
+      - lib/libcuobjclient*.so.*
+      - targets/{{ target_name }}/lib/libcuobjclient*.so.*
+    requirements:
+      build:
+        - {{ compiler("c") }}
+        - {{ compiler("cxx") }}
```
jakirkham, line 44:
```suggestion
        - {{ compiler("cxx") }}
        - arm-variant * {{ arm_variant_type }}  # [aarch64]
```

adibbley: This was dropped [here](https://github.com/conda-forge/staged-recipes/pull/31940#discussion_r2710090634), sounds like it shouldn't be necessary as this component is only for 13.1.1+

adibbley:
> Also would add a `conda-forge.yml` file in the recipe with the following content as we know the overlinking check is broken for CUDA `redist`s
> 
> ```yaml
> conda_build:
>   error_overlinking: false
> ```

Are we able to add that in staged-recipes? I thought we would just need to follow up with this in the resulting feedstock.

jakirkham:
> Are we able to add that in staged-recipes? I thought we would just need to follow up with this in the resulting feedstock.

We are able to add a `conda-forge.yml` (in `staged-recipes`), which can affect the build process here and will be used in the resulting feedstock

adibbley:
> We are able to add a `conda-forge.yml` (in `staged-recipes`), which can affect the build process here and will be used in the resulting feedstock

Good to know, thanks! Is [this](https://github.com/conda-forge/staged-recipes/pull/31940/changes/097df1fe3b44b29cd9d9f439cb90536f8e6e387d) the correct spot to add it?

recipes/libcuobjclient/meta.yaml
jakirkham: Ah ok. Yeah that is fair. We can drop it. Thanks for the context! 🙏

jakirkham:
Yep. Thanks Alex! 🙏

Looks like `staged-recipes` isn't using that flag in the build. Though let's not worry about that for now. This will at least ensure the feedstock has this flag configured correctly

jakirkham:
Thanks Alex for the PR! Also thanks Billy and Daniel for the reviews! 🙏

PR #31949, opened 2026-01-18
Adding protozero v1.8.1

travishathaway:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/protozero/build.bat
+++ b/recipes/protozero/build.bat
@@ -10,8 +10,6 @@ if errorlevel 1 exit /b 1
 cmake %CMAKE_ARGS% ^
     -GNinja ^
     -DWERROR=OFF ^
-    -DCMAKE_PREFIX_PATH=%PREFIX% ^
-    -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% ^
     %SRC_DIR%
 if errorlevel 1 exit /b 1
 
@@ -19,4 +17,7 @@ ninja
 if errorlevel 1 exit /b 1
 
 ninja install
+if errorlevel 1 exit /b 1
+
+ninja test
 if errorlevel 1 exit /b 1
--- a/recipes/protozero/build.sh
+++ b/recipes/protozero/build.sh
@@ -9,9 +9,8 @@ extra_cmake_args=(
 mkdir build && cd build
 
 cmake ${CMAKE_ARGS} "${extra_cmake_args[@]}" \
-    -DCMAKE_PREFIX_PATH=$PREFIX \
-    -DCMAKE_INSTALL_PREFIX=$PREFIX \
     $SRC_DIR
 
 ninja
 ninja install
+ninja test
--- a/recipes/protozero/recipe.yaml
+++ b/recipes/protozero/recipe.yaml
@@ -21,11 +21,10 @@ requirements:
 
 tests:
   - package_contents:
-      files:
-        exists:
-        - ${{ "Library/" if win }}include/protozero/pbf_builder.hpp
-        - ${{ "Library/" if win }}include/protozero/config.hpp
-        - ${{ "Library/" if win }}include/protozero/version.hpp
+      include:
+        - protozero/pbf_builder.hpp
+        - protozero/config.hpp
+        - protozero/version.hpp
 
 about:
   homepage: https://github.com/mapbox/protozero


travishathaway:
@OliverSchmitz and @kordejong,

I justed wanted to put this PR on your radar because you are the maintainers of https://github.com/conda-forge/osmium-tool-feedstock. Once I get this merged, I would be happy to open a PR in your feedstock so that it uses this package instead.

travishathaway:
@conda-forge/staged-recipes and @conda-forge/help-c-cpp, ready for review.

traversaro:
Thanks for the contribution! 

A few reference for this package:
* repology: https://repology.org/project/protozero/versions
* homebrew recipe: https://github.com/Homebrew/homebrew-core/blob/f1c3f5afd9f7295bb2b770fd41345286729cd7c3/Formula/p/protozero.rb
* vcpkg recipe: https://github.com/microsoft/vcpkg/blob/f14401ca0f2754347c3864da7488a9b955b4e47a/ports/protozero/portfile.cmake
* Alpine Linux recipe: https://gitlab.alpinelinux.org/alpine/aports/-/blob/8eb605743cff97343a23111244cd5feb22f2245b/community/protozero/APKBUILD

So it seems quite a classical CMake recipe with no deps.

recipes/protozero/build.bat
```diff
@@ -0,0 +1,22 @@
...
+if errorlevel 1 exit /b 1
+
+cd build
+if errorlevel 1 exit /b 1
+
+cmake %CMAKE_ARGS% ^
+    -GNinja ^
+    -DWERROR=OFF ^
+    -DCMAKE_PREFIX_PATH=%PREFIX% ^
+    -DCMAKE_INSTALL_PREFIX=%LIBRARY_PREFIX% ^
```
traversaro: This should be already part of `%CMAKE_ARGS%`, is there any reason for adding them explicitly?

traversaro: Also, `CMAKE_PREFIX_PATH` should point to `%LIBRARY_PREFIX%` (unless I am missing something).

recipes/protozero/build.sh
```diff
@@ -0,0 +1,17 @@
...
+extra_cmake_args=(
+    -GNinja
+    -DWERROR=OFF
+)
+
+mkdir build && cd build
+
+cmake ${CMAKE_ARGS} "${extra_cmake_args[@]}" \
+    -DCMAKE_PREFIX_PATH=$PREFIX \
+    -DCMAKE_INSTALL_PREFIX=$PREFIX \
```
traversaro: These should be already part of $CMAKE_ARGS, is there any reason for adding them explicitly?

recipes/protozero/recipe.yaml
```diff
@@ -0,0 +1,42 @@
...
+    - cmake
+    - ninja
+
+tests:
+  - package_contents:
+      files:
+        exists:
+        - ${{ "Library/" if win }}include/protozero/pbf_builder.hpp
+        - ${{ "Library/" if win }}include/protozero/config.hpp
+        - ${{ "Library/" if win }}include/protozero/version.hpp
```
traversaro: Nice to ahve, to avoid the `${{ "Library/" if win }}` you can probably use the `include` key (see https://rattler-build.prefix.dev/latest/reference/recipe_file/#check-for-package-contents):

```suggestion
  - package_contents:
      include:
        - protozero/pbf_builder.hpp
        - protozero/config.hpp
        - protozero/version.hpp
```

recipes/protozero/build.sh
```diff
@@ -0,0 +1,17 @@
...
+    $SRC_DIR
+
+ninja
+ninja install
```
traversaro, line 15: The library is header-only, the only thing that is compiled are tests. Given that we are compiling them, does it make sense to run them by adding:


```suggestion
ninja install
ninja test
```

travishathaway:
Hi @traversaro,

Thanks for the review and helpful comments, especially links to reference implementations. I'm still fairly new to the C++ packaging world, so if anything doesn't quite make sense, that's why.

I'll take some time to integrate your feedback and will ping you again when I am ready for another review.

travishathaway:
@traversaro,

This is ready for another review.

traversaro:
> @traversaro,
> 
> This is ready for another review.

Great work! Not blocking for me (other reviewers may have other opinions) but one option to cleanup further the recipe is to switch to just build two `noarch: generic`, one for unix and one for Windows, see https://conda-forge.org/docs/maintainer/knowledge_base/#noarch-packages-with-os-specific-dependencies .

travishathaway:
@conda-forge/help-c-cpp,

I just wanted to ping you all once more on this recipe. I believe everything should be in order 🙏. Thanks in advance!

traversaro:
Thanks a lot for the contribution! I think we can merge as there are many non-noarch header-only libraries in conda-forge, and any further improvement can be done post-PR.

danielnachun:
We should open a discussion on Zulip on how better to handle these header-only libraries.  There is a plan to merge `%LIBRARY_PREFIX%` with `%PREFIX%` on Windows which is one source of friction at the moment.  The other issue is that you often need a C/C++/Fortran compiler to run the tests, but adding those compilers then adds a bunch of native run exports that aren't actually needed.  I think I actually managed at one point to persuade a v1 recipe header-only recipe to be properly `noarch: generic` as a test, but it was a pain.  I feel like we may need some more standardized format in v1 recipes to handle this as it's a somewhat common type of package and it shouldn't be such a pain to make it `noarch: generic`.

travishathaway:
FYI: [#general > Support C++ Header only libraries as &#96;noarch: generic&#96;](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Support.20C.2B.2B.20Header.20only.20libraries.20as.20.60noarch.3A.20generic.60/with/574034349)

PR #31959, opened 2026-01-20
Add roboplan and toppra libraries

sea-bass:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/toppra/recipe.yaml
+++ b/recipes/toppra/recipe.yaml
@@ -1,15 +1,14 @@
 context:
   name: toppra
-  version: "0.6.2"
+  version: "0.6.4"
 
 recipe:
   name: toppra
   version: ${{ version }}
 
 source:
-  # TODO: Once this is versioned, replace with a versioned tarball.
-  url: https://github.com/hungpham2511/toppra/archive/6b11235b304455a944479f1cb250559f1eb44b7b.tar.gz
-  sha256: cc0318173769d9dd88c7315e6e0d5cf1b63d570fdc05ea1e0fe1339da53c10fd
+  url: https://github.com/hungpham2511/toppra/archive/refs/tags/v${{ version }}.tar.gz
+  sha256: 31213336fc36a69e8003808e7317773ba7c78e503c8cbd6d01fe2d50660e3294
 
 build:
   number: 0


sea-bass:
@eholum Do you want to be a maintainer on here? If yes, please officially agree on here; else, I can remove you.

sea-bass:

(and @traversaro)

sea-bass:
> ❌ This recipe is using a compiler, which now requires adding a build dependence on ${{ stdlib("c") }} as well. Note that this rule applies to each output of the recipe using a compiler.

Adding this makes my local builds fail, so not sure what the right approach is here.

traversaro:
> > ❌ This recipe is using a compiler, which now requires adding a build dependence on ${{ stdlib("c") }} as well. Note that this rule applies to each output of the recipe using a compiler.
> 
> Adding this makes my local builds fail, so not sure what the right approach is here.

Ah yes, this is a common pitfall. The problem is that the `${{ stdlib("c") }} ` requires the variable `c_stdlib` to be defined in a [variant configuration file](https://rattler-build.prefix.dev/v0.17.1/variants/). So to build conda-forge-compliant recipe, it is necessary to pass via the `-m` parameter a proper variant file. The variant file used by conda-forge infrastructure is https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/main/recipe/conda_build_config.yaml, so a simple solution is to download the file and pass it via `-m` to rattler-build.

However, once a recipe is in staged-recipes, you can you the pixi task of staged-recipes to build it locally, that handle this aspects automatically.

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+  name: roboplan
+  version: "0.1.0"
+
+recipe:
+  name: roboplan
+  version: ${{ version }}
+
+source:
+  git: https://github.com/open-planning/roboplan.git
+  tag: ${{ version }}
```
traversaro: conda-forge requires to use tarball with hashes for reproducibility, see https://conda-forge.org/docs/maintainer/adding_pkgs/#build-from-tarballs-not-repos .

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+      name: libroboplan_example_models
+      version: ${{ version }}
+    build:
+      skip:
+        - not linux
+      script: |
+        cmake -S "${SRC_DIR}/roboplan_example_models" \
+          -B build/roboplan_example_models \
+          -G Ninja \
+          -DCMAKE_INSTALL_PREFIX=$PREFIX
```
traversaro: Please use `${CMAKE_ARGS}` here. In this way, you can omit `-DCMAKE_INSTALL_PREFIX=$PREFIX` and any other conda-forge-specific boilerplate, and also cross-compilation (required to build on `linux-aarch64` and `osx-arm64` will work out of the box, see https://conda-forge.org/docs/maintainer/knowledge_base/#using-cmake .

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+    requirements:
+      run_exports:
+        - ${{ pin_subpackage("libroboplan_example_models", upper_bound="x.x.x") }}
+      build:
+        - ${{ stdlib('c') }}
+        - ${{ compiler('cxx') }}
+        - cmake
+        - ninja
+      host:
+        - gtest
```
traversaro, line 39: This will add a run-dependency on gtest, that is spurious as I guess you are not actually depending on gtest. You can add a block:

~~~
      ignore_run_exports:
        from_package:
          - gtest
~~~

to ignore this spurious dep, see

https://github.com/conda-forge/gz-sim-yarp-plugins-feedstock/blob/main/recipe/recipe.yaml#L75C1-L79C18

recipes/roboplan/recipe.yaml
sea-bass: I saw this, but it doesn't work out of the box because apparently the tarballs don't include submodules?! Is there any way you know to address this?

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+    tests:
+      - script:
+        - test -f "${PREFIX}/lib/libroboplan_example_models${SHLIB_EXT}"
```
traversaro: You can be more explicit (and platform indipendent) with:

```suggestion
    - package_contents:
        lib:
          - roboplan_example_models
```

recipes/roboplan/recipe.yaml
sea-bass: One alternative I could think of was to also submit a recipe for `toppra` (which is the submodule in question here), but...

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+        - ${{ stdlib('c') }}
+        - ${{ compiler('cxx') }}
+        - cmake
+        - ninja
+      host:
+        - gtest
+        - libcoal
+        - libpinocchio
+        - yaml-cpp
+        - ${{ pin_subpackage("libroboplan_example_models", upper_bound="x.x.x") }}
```
traversaro: Why you use `pin_subpackage` here? Isn't sufficient a normal `libroboplan_example_models` dependency, so that you get the correct run dependency via run_exports?

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+      script: |
+        # Build TOPP-RA dependency from source
+        cmake -S "${SRC_DIR}/external/toppra/cpp" \
+          -B build/toppra \
+          -G Ninja \
+          -DBUILD_TESTS=OFF \
+          -DPYTHON_BINDINGS=OFF \
+          -DCMAKE_INSTALL_PREFIX=$PREFIX
+        cmake --build build/toppra
+        cmake --install build/toppra
```
traversaro: This is a bit problematic, as it could interfere with other uses of toppra. Could it make sense to have a proper conda-forge recipe of toppra as well? You can just add it as part of this recipe as a new `recipes/toppra` folder, the CI will sort it out that it needs to be built.

traversaro: Ah, I see that the last tag is really old. Do you think we can convince the upstream mantainers to do a new tag?

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,232 @@
...
+      - script:
+        - test -f "${PREFIX}/lib/libroboplan_simple_ik${SHLIB_EXT}"
+
+  - package:
+      name: libroboplan_toppra
+      version: ${{ version }}
+    build:
+      skip:
+        - win
+        - osx
```
traversaro: I would either use `not linux` or this, not both that is confusing.

recipes/roboplan/recipe.yaml
sea-bass: Is the tag for the python package though? That one is on pypi, didn't think it was on conda-forge.

I'm specifically building the cmake/cpp project in there, so it's probably a totally different target.

The repo just happens to have two parallel implementations

sea-bass:
Thanks @traversaro -- I think I've addressed all your comments. That was super helpful!

@eholum Do you want to be a maintainer on the roboplan recipe? If yes, please officially agree on here; else, I can remove you.

@hungpham2511 Do you want to be a maintainer on the toppra recipe (for now it's just C++ lib)? If yes, please officially agree on here; else, I can remove you.

hungpham2511:
Hi @sea-bass,

I am not familiar with conda in general and have little bandwith to handle toppra at the moment, so please remove me from the maintainer list.

sea-bass:
> I am not familiar with conda in general and have little bandwith to handle toppra at the moment, so please remove me from the maintainer list.

No problem at all, thought I would check just in case. Thanks for confirming!

recipes/toppra/recipe.yaml
```diff
@@ -0,0 +1,68 @@
...
+  license_file: LICENSE
+  summary: Path parameterization for robots
+  repository: https://github.com/hungpham2511/toppra
+  documentation: https://hungpham2511.github.io/toppra/index.html
+
+extra:
+  feedstock-name: toppra
+  recipe-maintainers:
+    - sea-bass
+    - hungpham2511
```
sea-bass:
```suggestion
```

recipes/toppra/recipe.yaml
```diff
@@ -0,0 +1,67 @@
...
+        - eigen
+      ignore_run_exports:
+        by_name:
+          - gtest
+        from_package:
+          - gtest
+    tests:
+      - package_contents:
+          lib:
+            - libtoppra
```
traversaro:
```suggestion
            - toppra
```

The logic handles prefix like `lib`

traversaro:
On a more generic comment, did you consider using `-` instead of `_` for package names? conda-forge allow both, but most packages use `-` (see https://conda-forge.org/packages/), and that is also what robostack uses https://robostack.github.io/jazzy.html .

recipes/toppra/recipe.yaml
```diff
@@ -0,0 +1,67 @@
...
+  version: "0.6.2"
+
+recipe:
+  name: toppra
+  version: ${{ version }}
+
+source:
+  # TODO: Once this is versioned, replace with a versioned tarball.
+  url: https://github.com/hungpham2511/toppra/archive/6b11235b304455a944479f1cb250559f1eb44b7b.tar.gz
+  sha256: cc0318173769d9dd88c7315e6e0d5cf1b63d570fdc05ea1e0fe1339da53c10fd
```
traversaro: Let's try to get this sorted out https://github.com/hungpham2511/toppra/pull/268, as 0.6.2 version was already released ~2 years ago (see https://github.com/hungpham2511/toppra/releases/tag/v0.6.2) and packaging as 0.6.2 something that is not 0.6.2 is just really confusing.

sea-bass: Good point -- I'll look into this and update accordingly when we have a solution: https://github.com/hungpham2511/toppra/issues/269

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+        - gtest
+      ignore_run_exports:
+        by_name:
+          - gtest
+        from_package:
+          - gtest
+    tests:
+      - package_contents:
+          lib:
+            - libroboplan_example_models
```
traversaro: Minor, package_contents/lib will automatically add the `lib`:

```suggestion
            - roboplan_example_models
```

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+        - libroboplan-example-models
+      ignore_run_exports:
+        by_name:
+          - gtest
+        from_package:
+          - gtest
+    tests:
+      - package_contents:
+          lib:
+            - libroboplan
```
traversaro:
```suggestion
            - roboplan
```

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+        - cmake
+        - ninja
+      host:
+        - gtest
+        - libroboplan
+        - libroboplan-example-models
+    tests:
+      - package_contents:
+          lib:
+            - libroboplan_rrt
```
traversaro:
```suggestion
            - roboplan_rrt
```

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+        - libtoppra
+      ignore_run_exports:
+        by_name:
+          - gtest
+        from_package:
+          - gtest
+    tests:
+      - package_contents:
+          lib:
+            - libroboplan_toppra
```
traversaro:
```suggestion
            - roboplan_toppra
```

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+
+  - package:
+      name: roboplan-python
+      version: ${{ version }}
+    build:
+      skip:
+        - not linux
+      script: |
+        export CXX=${CXX:-c++}
+        export CMAKE_ARGS="-DCMAKE_CXX_COMPILER=${CXX}"
```
traversaro: `CMAKE_ARGS` is already set, if you want to set something more it you can append to it:


```suggestion
        export CMAKE_ARGS="${CMAKE_ARGS} -DCMAKE_CXX_COMPILER=${CXX}"
```

recipes/toppra/recipe.yaml
traversaro: Ok, in case upstream is not responding, I think we can go for another way to proceed. Checking the changes between 0.6.2 and the latest changes (see https://github.com/hungpham2511/toppra/compare/v0.6.2...develop). The only C++ change that is not in test is the change from https://github.com/hungpham2511/toppra/pull/247 in `cpp/src/toppra/solver/seidel-internal.hpp`, and it is not a change in the public interface, so it make sense to backport it. So probably an alternative route is to package the tag https://github.com/hungpham2511/toppra/releases/tag/v0.6.2 and just add that fix as a patch, that would be perfectly fine for conda-forge and would permit to unblock the PR.

sea-bass: i'm OK waiting for upstream!

recipes/roboplan/recipe.yaml
```diff
@@ -0,0 +1,247 @@
...
+        - libroboplan-example-models
+      ignore_run_exports:
+        by_name:
+          - gtest
+        from_package:
+          - gtest
+    tests:
+      - package_contents:
+          lib:
+            - libroboplan_simple_ik
```
sea-bass:
```suggestion
            - roboplan_simple_ik

recipes/toppra/recipe.yaml
sea-bass: @traversaro Upstream is good and I just updated the `toppra` recipe -- woooo

traversaro:
The only missing thing for the linter and for the review in general is @eholum accepting of being a maintainer. Otherwise you can remove him from maintainers and the add him in the recipe when he agrees.

eholum:
Accepted! 👍

PR #31993, opened 2026-01-23
Adding pptreport

hschult:
Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/pptreport/meta.yaml
+++ b/recipes/pptreport/meta.yaml
@@ -9,19 +9,18 @@ source:
   sha256: ced0cbcacb8d67cd875456382460531abb380e69e19e5e78997c270ce539ad46
 
 build:
-  # due to pymupdf
-  skip: true  # [not linux]
+  noarch: python
   script: {{ PYTHON }} -m pip install . -vv --no-deps --no-build-isolation
   number: 0
 
 requirements:
   host:
-    - python
+    - python {{ python_min }}
     - setuptools >=42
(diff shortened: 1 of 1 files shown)


hschult:

recipes/pptreport/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: pptreport
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/p/pptreport/pptreport-{{ version }}.tar.gz
+  sha256: ced0cbcacb8d67cd875456382460531abb380e69e19e5e78997c270ce539ad46
+
+build:
+  # due to pymupdf
+  skip: true  # [not linux]
```
ocefpaf: Looks like this one can be noarch. What is the issue with `pymupdf`?

hschult: They only provide a Linux build. Not sure why

ocefpaf: It's okay if a noarch package fails on a specific platform due to a missing dependency. We prefer using noarch because it's faster to build, and the package will work as soon as the missing dependency becomes available on that platform.

hschult: I see. I will change it to noarch. Thanks for the explanation :)

hschult: Should I address the failing non-Linux builds somehow?

PR #32004, opened 2026-01-24
Add `nvidia-cutlass-dsl`

iamthebot:
This is a dependency for `flashinfer` which is a dependency for VLLM 0.11+ (see conda-forge/vllm-feedstock#20). Unforunately, these are only published as binary wheels on pypi. The license appears, however, identical to what we use for CUDA packages which does allow redistribution.

Given that we only have binary wheels to work with, I have the upstream sources set to the binary wheels that are available. I.e we *must* use binary repackaging for this package. There's no other way. Even the cutlass repo pulls these in from pypi.

IMO it makes sense this be a separate package from `cutlass` which is just the C++ lib.

Checklist

- [x] Build number is 0.

### DIFF
--- a/recipes/nvidia-cutlass-dsl/recipe.yaml
+++ b/recipes/nvidia-cutlass-dsl/recipe.yaml
@@ -5,6 +5,7 @@
 context:
   name: nvidia-cutlass-dsl
   version: "4.3.5"
+  build: 0
 
 package:
   name: ${{ name }}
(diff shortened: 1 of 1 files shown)


bdice:
Hi! Thanks for this -- I can help find the right folks to help with this, it may take a few days.

I think we may be able to build this from source rather than repackage wheels. That's the typical best practice for conda-forge. I'm not super familiar with this package but I think the source is around here: https://github.com/NVIDIA/cutlass/tree/main/python

danielnachun:
The source code for this appears to be available at https://github.com/NVIDIA/cutlass.  Can you confirm if that is correct?  If so we will build from source there instead of using binary wheels.

bdice:
Hi @iamthebot, thanks for your help. You're correct that some components are closed source. We have precedent for handling this with NVIDIA packages on conda-forge. I am going to work with the CUTLASS team to create a recipe that meets the appropriate requirements and will follow up soon.

bdice:
A tracking issue has been filed here for this work: https://github.com/conda-forge/staged-recipes/issues/32172

danielnachun:
In a nutshell, the open source components should be built from source, and we should debundle the packages/libraries into separate recipes.  The starting point should be to list out the components that are in the repo and indicate which parts are closed source and which are open source.

PR #32031, opened 2026-01-28
add opencolorio

carlodri:
https://opencolorio.org/

Checklist

- [x] Build number is 0.


ndeybach:
Hi @carlodri ,

I see you are struggling on  the linux / macos builds. I am under linux and can debug directly.

I have done as such, and I see the build seems to fail because the recipe includes for python packages of openclorio :

```
build:
   [...]
  python:
    entry_points:
      [...]
      - pyocioamf = PyOpenColorIO.bin.pyocioamf.pyocioamf:main
      - pyociodisplay = PyOpenColorIO.bin.pyociodisplay.pyociodisplay:main
[...]
tests:
  - python:
      imports:
        - PyOpenColorIO
      pip_check: true
  - script:
      [...]
      - pyocioamf --help
      - pyociodisplay --help
```

Is it really expected to build the bindings / python packages in the "opencolorio" package ? Wouldn't the python packages / builds be for another recipe / build ? Or even better maybe a split forge here that would output both `pyopencolorio` and `opencolorio` packages ?

carlodri:
Hi @ndeybach! Yes I am indeed struggling, thanks for stepping in!

It would be great (and the right thing) to make a split package with two outputs, `opencolorio`  and `pyopencolorio`. I've always struggled with split-output recipes, so it would be great if you can help.

ndeybach:
Hi @carlodri ,

I recently worked on some PRs for https://github.com/conda-forge/libjxl-split-feedstock/ , which may be a good example on how to handle a split forge ?

I am working a bit offline up to something that I am satisfied about and works under linux before pushing it here (could you add me so I have the rights for that ? Though I am not familiar with how that would work on such a first PR, do I have to register for conda forge admin ?)

I am not that much experienced in conda forge builds, but I do develop quite a bit of private rattler build recipes for my company usage, so I'll try to help on this one. My goal is also to have openimageio updated to 3.0 on conda forge down the line.

I'll update here as soon as I have something working.

ndeybach:
Oh and for now I am working toward having 3 outputs: `opencolorio` (for the lib), `pyopencolorio` (for the python bindings) and `opencolorio-tools` (for the cli tools). Does that sound good to you ?

ndeybach:
Also, I see you have not added a windows build. Since openimageio will require one, what do you think about we add it ? (I don't have a windows machine on hand though so either we do ci/cd builds iterations or if you have one locally to tests builds quickly it would be best)

carlodri:
Thanks for the feedback! The three-split is perfect I think. On how to practically contribute, I'm not sure either. I think only conda-forge core maintainers are allowed to push here and I am not one of them. 

The only practical way I see would be for me to give you access to my staged-recipes fork for the time we work together on this PR. Because the alternative is that you open a PR against my branch and I merge it here, but that is quite a cumbersome workflow I guess. Let me know what you prefer.

I'm not sure if you started from scratch or from my commits (maybe better from scratch I think because I made quite a mess with my attempts :-)). 

Concerning the Windows build, I just recently moved practically all of my work on linux machines (ha enough of Windows, 11 in particular :grimacing: ), but I think I can spin up a Windows VM if necessary.

carlodri:
Oh and thanks for the libjxl reference, it is really helpful to understand the logic of the split recipes!

ndeybach:
Super nice if you could have a windows session at hand (I have to see one day if I can setup a VM or dual boot for just basic testing, but didn't find the time yet (at least in a reproducible / fully up and running way dev wise ). No need to do so yet though, only if we have road bumps on the windows build.

I started my work from you branch on staged-recipes. Not so terrible imo. I changed a lot already (I took a pystring error in the face too like your last fix and more xD). Maybe the best would be if we could work in the same staged-recipes fork, no ? (https://github.com/carlodri/staged-recipes/tree/opencolorio )

I just got a successful build of all three under linux. I'll have a second review pass tomorrow for it and will add a windows build script (will be a "dumb" auto generated port of the unix build script though as powershell is  not my strong point ^ ^).

carlodri:
Hi @ndeybach I invited you to my stages-recipes fork so you should be able to push here now!

ndeybach:
Hi @carlodri,

Thanks for the invite. Seems to have worked well.

Sorry for the delay, it took me more time than expected to get some last few features working correctly. Notably enabling opengl acceleration headless mode under linux for the tools.

I had to scrap a lot of your previous recipe with the introduction of the split, I kept the overall build script skeleton but added some separation because of the split.

Awaiting to see how the macos and windows build will develop.

Though some major choices are to be made regarding some points, as noted in the commit message :
- the usage of openimageio (oiio) is disabled in `opencolorio-tools` because otherwise it would be a looped dependency starting with the future 3.X versions of oiio package.
Possible fixes :
	- we regroup oiio and the current 3 outputs into one big split recipe : so that builds are coherent and use the same dependencies (one does not "need" the other to be published first .. they are published atomically.
	- or we keep it as currently done: but the tools are only compatible with exr (thanks to openexr dep) and can't use more common formats (png, jpg, jxl, ..)
- under linux I took the decision to enable headless openGL acceleration compatibility ... which means that additional dependencies and steps take size and time for it under linux. It is debatable imo because of size increase and niche use case. I did not had time to compare size before / after though.


What is your opinion on those and my overall commit / draft ?

carlodri:
Wow @ndeybach you did an incredible work! I don't have specific comments on the choices you have made, I think it's great to have the packages split like this.

Concerning the oiio thing, I'm not sure which is the best solution here, honestly I don't know the packages well enough.

I think the osx failure is due to the fact that, as the [docs](https://opencolorio.readthedocs.io/en/latest/quick_start/installation.html#building-from-source) say at a certain point, libminizip-ng has some issues when being installed as an external package. We could try to remove minizip-ng on osx and let ocio vendor it, so see if the problem goes away...

Oh and by the way , please add yourself as maintainer, if you want/can it would be of great help!

ndeybach:
Thanks for the positive feedback !

I added myself in the maintainer list. Though it is going to be my first maintainer role so brand new to me and to what it entails and needs ^ ^. It has not been very long for me in the conda forge / build world and I'll probably need guidance on some aspects.

For oiio I think the best move here is to request expertise from someone with better understanding and more experience of conda forge, I'll follow up with a tag and detailed post for that once we have a working build on all platforms.

For the osx failure I pin pointed it to cmake on ocio side that looks for "minizip-ng" in the dependencies but on macos it seems conda forge ships only under the old name of "minizip" .. hence it not finding it. So I do not think letting ocio vendor the lib is best (especially in conda forge context where tracking which package was used is best). Hence, I have added a code patch for cmake file to recognize the other name as fallback and if that works we may want to contribute to upstream ocio. 

For windows it seems it is a python executable not correctly selected (and those no correct deps installed), so I gave precise flags to cmake for its path, but I'll let the ci/cd pass before further investigation.

carlodri:
thanks @ndeybach! I think you'll make a great maintainer, much better than I could do especially in these c/cpp projects. :-)

For what I can see this setup is great. About oiio, yes I think we should ask someone in conda-forge/core for guidance. Honestly, I didn't have time to think it through, but I can't really wrap my head around this circular oiio-ocio dependency: even if we build in a single recipe we still need one to be built before the other... (?)

ndeybach:
Thanks for your kind words @carlodri , I'll try to do my best on this one.

For oiio recursive problem, I started a topic on zulip : [#general > OpenColorIO (tools) Recursive Dependency handling? PR #32031](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/OpenColorIO.20.28tools.29.20Recursive.20Dependency.20handling.3F.20PR.20.2332031/with/577916645) to have more feedback from the conda forge core team and experienced members on the matter.

But as it goes we may go down the route of doing a different recipe for the tools...

carlodri:
Thanks for pushing this forward @ndeybach! AFAIU it seems like this PR will be transformed into a PR to the oiio feedstock, once the v1 conversion there is done, is that right?

ndeybach:
Hi @carlodri ,

Sorry for the delay, last week was quite busy.

From the zulip and oiio feedstock discussion it is indeed feeling like we are going toward a unified repository on that. Maybe though it will await this [CEP](https://github.com/conda/ceps/pull/102) to have staging enabled. But nothing sure yet... the v1 conversion still has to be agreed upon and merged for v2.5.x oiio before any v3.x thing..

PR #32071, opened 2026-01-30
Add cutile-python

bdice:
Checklist

- [x] Build number is 0.

This PR adds `cutile-python` (https://github.com/NVIDIA/cutile-python/), a programming language for NVIDIA GPUs that generates kernels based on Tile IR.

**Note to reviewers:** ⚠️ CI is expected to fail in `staged-recipes` -- CUDA 12.6 is enabled here but this recipe requires CUDA 13.1 to build. I verified locally that this builds with CUDA 13.1. Once this PR is merged and a feedstock repository is created, we can set up the `conda_build_config.yaml` needed for this to build properly.

Depends on https://github.com/conda-forge/cuda-tileiras-feedstock/pull/6.


recipes/cutile-python/recipe.yaml
```diff
@@ -0,0 +1,78 @@
...
+    kernels based on Tile IR. It requires NVIDIA Driver r580+ and CUDA Toolkit
+    13.1+.
+  license: Apache-2.0
+  license_file: LICENSES/Apache-2.0.txt
+  documentation: https://docs.nvidia.com/cuda/cutile-python
+  repository: https://github.com/NVIDIA/cutile-python
+
+extra:
+  recipe-maintainers:
+    - bdice
```
bdice, line 82: @conda-forge/cuda: I will change this from `bdice` to `conda-forge/cuda` in a PR after the repo is created. I believe we've had issues in the past with _creating_ feedstock repos where a team is set as the only maintainer.

leofang: You can also add `conda-forge/cuda-python`.

recipes/cutile-python/recipe.yaml
```diff
@@ -0,0 +1,78 @@
...
+  run:
+    - cuda-tileiras
+    - python
+    - typing-extensions
+  run_constraints:
+    # Requires CUDA driver >=580
+    - __cuda >=13.0
+
+tests:
+  # Cannot import cuda.tile without a CUDA driver, so just verify package contents
```
leofang: This seems like a very sharp edge... Good catch.

bdice: I think this was fixed in 1.1.0. I tested it locally again after updating and no longer observe this issue. I added better import tests.

bdice:
@conda-forge/staged-recipes this is ready for review.

Copying this from above:
> **Note to reviewers:** ⚠️ CI is expected to fail in `staged-recipes` -- CUDA 12.6 is enabled in `staged-recipes` but this recipe requires CUDA 13.1 to build. I verified locally that this builds with CUDA 13.1. Once this PR is merged and a feedstock repository is created, we can set up the `conda_build_config.yaml` needed for this to build properly.

recipes/cutile-python/recipe.yaml
```diff
@@ -0,0 +1,81 @@
...
+  patches:
+    - patches/0001-Use-standard-CMake-FindCUDAToolkit-module.patch
+    - patches/0002-Remove-nostdlib-flags-to-allow-linking-glibc.patch
+
+build:
+  number: 0
+  skip:
+    - osx
+    - win
+    - cuda_compiler_version == "None"
```
carterbox, line 22:
```suggestion
  skip:
    - true
    - osx
    - win
    - cuda_compiler_version == "None"
```

carterbox: We cannot merge recipes with failing builds because of branch protection rules.

carterbox: Unskip in feedstock after applying CUDA 13.0 migration.

recipes/cutile-python/recipe.yaml
```diff
@@ -0,0 +1,82 @@
...
+  patches:
+    - patches/0001-Use-standard-CMake-FindCUDAToolkit-module.patch
+    - patches/0002-Remove-nostdlib-flags-to-allow-linking-glibc.patch
+
+build:
+  number: 0
+  skip:
+    - true
+    - osx
+    - win
```
leofang, line 21: This is fine in the staged recipes, but we should enable the Windows builds in the feedstock.

recipes/cutile-python/recipe.yaml
```diff
@@ -0,0 +1,82 @@
...
+    - cuda_compiler_version == "None"
+  script:
+    - export CUDA_TILE_CMAKE_DLPACK_PATH=$PREFIX
+    - python -m pip install . -vv
+
+requirements:
+  build:
+    - ${{ compiler('c') }}
+    - ${{ compiler('cxx') }}
+    - ${{ compiler('cuda') }}
```
leofang, line 31: 1. I am actually unaware of any GPU code existing in the cuTile Python repo. Do we really need NVCC? I would think GCC/MSVC are enough. 
2. Regardless, we probably miss a few `ignore_run_exports[_from]` to make it MVC-compatible?

leofang (approved):

recipes/cutile-python/recipe.yaml
bdice: Yes, that's the plan. I'll start a PR for that.

recipes/cutile-python/recipe.yaml
bdice: Good question. I'll look at this in the downstream feedstock.

carterbox:
It was brought to my attention that it looks like I inappropriately added a naked skip to circumvent branch protection rules.

I should clarify that the correct skip statement should have been something like:
```
skip if cuda_compiler_version <=13.1
```
However, since I don't know how to implement that statement in the new recipe format, I opted to just skip all builds instead. This is because there is no CUDA 13.1 variant file available for staged-recipes, so even if I had figured out the correct syntax, the result would have been the same.

I believe it would have also been inappropriate to add a new variant file for CUDA 13.1 builds to staged-recipes because at the time we were still building CUDA 12.8 here. We have only just added CUDA 12.9 and the next will probably be CUDA 13.0 since that migrator is still open.

My apologies if that was unclear.

recipes/cutile-python/recipe.yaml
carterbox: Please see [my note at the bottom of this MR](https://github.com/conda-forge/staged-recipes/pull/32071#issuecomment-3881985367) about skipping all builds.

PR #32093, opened 2026-02-03
Adding `syster-cli` a tool for working with SysML v2 Files

phreed:
Checklist

- [x] Build number is 0.


jade-codes:
Happy to be listed

phreed:
@condo-forge/help-rust ready for review

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,55 @@
...
+context:
+  version: "0.2.3_alpha"
+  git_tag: ${{ version | replace('_', '-') }}
+
+package:
+  name: syster-cli
+  version: ${{ version }}
+
+source:
+  - url: https://github.com/jade-codes/syster-cli/archive/refs/tags/v${{ git_tag }}.tar.gz
```
bollwyvl, line 13: Recommend bumping to `v0.3.1-alpha` which includes a proper license (LICENSE.md), and remove the vendored one.

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,55 @@
...
+  - path: LICENSE
+
+build:
+  number: 0
+  script:
+    - if: unix
+      then:
+        - cargo auditable install --no-track --bins --root ${{ PREFIX }} --path .
+      else:
+        - cargo auditable install --no-track --bins --root %LIBRARY_PREFIX% --path .
```
bollwyvl: these could be collapsed to an inline conditional, e.g. 

```yaml
script:
  - cargo auditable install --no-track --bins --root ${{ PREFIX if unix else '%LIBRARY_PREFIX%' }} --path .
```

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,55 @@
...
+requirements:
+  build:
+    - ${{ compiler('rust') }}
+    - ${{ stdlib("c") }}
+    - cargo-bundle-licenses
+    - cargo-auditable
+
+tests:
+  - script:
+      - syster --help
```
bollwyvl, line 35: Consider adding a check which ensures that `--version` is at least indicative of _something_ (ideally the expected version). Better would be to _some_ kind of actual smoke test of a file.

phreed: This is a placeholder (it is still alpha). 
I have added the `- syster --version` though.
@jade-codes Let me know what you would consider a good smoke test?

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - syster --version
```
bollwyvl, line 36: Right, so this demonstrates it can report a version without exiting with a non-zero exit code. 

Is this version something useful? e.g.

```yaml
- syster --version  # does it even work
- syster --version | grep -iE "^syster ${{ git_tag | replace('.', '\\.') }}$"  # does it do the expected
```

(of note: this would sadly require a `dependencies: {run: [git]}` (or `m2-grep`, but...) to work on `win-64`)

phreed: I thought conda-forge recipes.yaml were only supposed to do basic smoke tests.
In any case, I added in the version check via the cross-platform nushell. 
The test runs a test_version.nu script (a new file).

recipes/syster-cli/test_version.nu
phreed, line 1: As you can see it is a fairly simple script.
It checks that `syster` runs successfully.
It then verifies that the version matches what was expected.

bollwyvl:
No further notes I would consider blocking, but: bad news, and should have lead with this: I'm just a maintainer of a bunch of (semi-related) feedstocks. To get someone who can _actually_ merge this, you'd want to read that checklist up top, and mention the `@` then `help-rust` group (it should syntax highlight) with a nice, thoughtful comment asking for a review; this will add a bunch of PR labels, and make it bubble up to someone with the rights to push the button.

On testing: it's true, "`conda-forge` is not a replacement for upstream CI". I'd say "does the CLI does one thing without failing, and returns an expected, important piece of data" is still in the "simple" category. As SysML is _not_ simple, I'd like to see something that actually did something with some SysML file and the standard library (e.g. do a little roundtrip through the export formats, get the same SHA out), but more about that below.

As for the syntax of the test: the `nushell`, while supported, seems more complex than a  `grep -iE` one-liner. That being said, with `rattler-build`, it's also possible to put (short) scripts inline in `build` and `tests` without first starting a `bash` or `cmd.exe` native shell or a separate file. This can avoid any `argv` wrangling, which often saves lines. See [pixi-feedstock](https://github.com/conda-forge/pixi-feedstock/blob/afa092c1809952b57d4a07e4372a84b50ede6fd5/recipe/recipe.yaml#L23) for an (the only?) example of inlining with `nushell`.

More broadly on someone actually being able to use this thing: similar to other "not-exacly-code" ecosystems like [openmodelica](https://github.com/conda-forge/staged-recipes/pull/31074/files/504c6ba378e27c7bab95b3e2fdc83565bb0b1270#diff-39deb5ffa0bcd7aecc299693b4e3093e812233d35704992cfca2c7bd14e80f4a): the upstream docs mention the "standard library," and the python bindings will go [download](https://github.com/jade-codes/syster-tree/blob/main/src/systree/cli.py)... something from github. 

Even if the upstream community can't decide on a well-known path that accommodates multiple versions (e.g. `$PREFIX/share/SysML-v2/2026-01`), this tool suite _could_ just fiat something, and start packaging that data as a standalone package, e.g. `sysml-v2-standard-library`. This could then enjoy the conda-forge benefits, and be versioned together with tools like `syster-cli` instead of having a big `curl`venture. This would also make it much easier to test that this package actually does something, and detect whether it  _doesn't_ do something with some new version of the standard library.

phreed:
@conda-forge/help-rust ready for review, I would like this merged.
This is a new application, I expect it to evolve rapidly

> As for the syntax of the test: the `nushell`, while supported, seems more complex than a `grep -iE` one-liner. That being said, with `rattler-build`, it's also possible to put (short) scripts inline in `build` and `tests` without first starting a `bash` or `cmd.exe` native shell or a separate file. This can avoid any `argv` wrangling, which often saves lines. See [pixi-feedstock](https://github.com/conda-forge/pixi-feedstock/blob/afa092c1809952b57d4a07e4372a84b50ede6fd5/recipe/recipe.yaml#L23) for an (the only?) example of inlining with `nushell`.

I have heard this argument a few times, and if we were only interested in building on unix-like systems it would make sense. I think your point about pixi-feedstock being the only use of nushell is a "survivor bias" fallacy.  Each time this argument gets used I convert the relatively simple nushell script into a slightly simple bash script (the core justification for the argument) and a convoluted cmd script with a bunch dependencies on flaky ported utilities. In this case it is not too bad, it is only `grep` (if it were autotools then a mess).  After going through all that, the package gets ignored.
Sorry for the rant, but I do not know where else to discuss this.

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+  - script:
+      - nu ./test_version.nu ${{ git_tag }}
+    requirements:
+      run:
+        - nushell
+    files:
+      recipe:
+        - test_version.nu
```
danielnachun:
```suggestion
```
This extra test is redundant here - we are already running `syster --version` above, and we don't need to verify that the developers set the right version in their code.  So let's just keep things simple and consistent with how we normally test CLI tools.

recipes/syster-cli/recipe.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+source:
+  - url: https://github.com/jade-codes/syster-cli/archive/refs/tags/v${{ git_tag }}.tar.gz
+    sha256: eeff7e737a79390293b485f8088ab9f838989f5964bb8ec0b6bd05588937c71a
+
+build:
+  number: 0
+  script:
+    - cargo auditable install --no-track --bins --root ${{ PREFIX if unix else '%LIBRARY_PREFIX%' }} --path .
+    - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
```
danielnachun:
```suggestion
  script:
    env:
      CARGO_PROFILE_RELEASE_STRIP: symbols
      CARGO_PROFILE_RELEASE_LTO: fat
    content:
      - cargo auditable install --no-track --bins --root ${{ PREFIX if unix else '%LIBRARY_PREFIX%' }} --path .
      - cargo-bundle-licenses --format yaml --output THIRDPARTY.yml
```

recipes/syster-cli/recipe.yaml
bollwyvl: that's on me on a previous suggestion: was really looking for _any_ verification that it did something that could be relied upon, even if a pre-release.

danielnachun:
> I have heard this argument a few times, and if we were only interested in building on unix-like systems it would make sense. I think your point about pixi-feedstock being the only use of nushell is a "survivor bias" fallacy. Each time this argument gets used I convert the relatively simple nushell script into a slightly simple bash script (the core justification for the argument) and a convoluted cmd script with a bunch dependencies on flaky ported utilities. In this case it is not too bad, it is only grep (if it were autotools then a mess). After going through all that, the package gets ignored.
> Sorry for the rant, but I do not know where else to discuss this.

I'm sorry if some reviews are getting dropped, we have a few more people stepping up to try to reduce this.  I will say though that you will probably have a much more success getting recipes reviewed quickly if they conform more closely to how our other recipes are written, and that will usually mean using batch scripts for Windows.  The important thing to keep in mind here for recipe reviewers and feedstock maintainers is that we have to review hundreds of submissions and maintain thousands of feedstocks, and for us the more homogenous and consistent recipes are, the easier it is for us to quickly scan them and identify potential issues.

Nushell in particular seems to be a stumbling block for a few reasons:
- There is a lot less documentation available than for bash and batch (if only due to how old they are), which makes it more time-consuming for reviewers/maintainers to check if something is implemented correctly or to troubleshoot issues.
- The syntax is a lot more verbose than bash and even batch scripting (if it's done right), and that is just an extra cognitive burden especially when switching between Nushell and bash/batch
- Nushell does not seem to have an easy and intuitive equivalent to `set -o xtrace` in bash and `@echo on` to show the values of environment variables when they are actually evaluated.  These are absolutely essential for debugging because a fair amount of conda-forge configuration relies on environment variables and we often need to see their actual values to troubleshoot problems.

On a related note, it seems you've wanted to use Nushell to implement custom testing code.  In general, this is something we really want to avoid, because over time that code will inevitably break.  While testing our packages is important, we also need it to be maintainable.  For executables, our main concern is just that they execute without segfaulting or crashing with an error that would indicate they were not built correctly.  For libraries, the main concern is that they can be found and loaded, again without crashing.  In most cases there are simple ways to test this with only one or a few lines of code.  There are build systems that provide built-in testing - this is mostly relevant for Make, CMake and Meson, where a set of tests can be run with a single standardized command, and those tests are provided by the upstream developer who is responsible for maintaining them.  Only for very specific cases, especially if we are patching/customizing the package for conda-forge in a way that deviates quite a bit from the way it is normally distributed, do we end up needing extensive custom test code.

I realize some of these best practices are not explicitly written out in our documentation.  This is a continually evolving process that we try to improve with user feedback and as the software ecosystems change over time.  However if there is one phrase to summarize the philosophy for making good recipes, I would say it is "less is more".  Conciseness and uniformity are essential for an ecosystem of this size, and I would try to understand the suggestions/feedback we give from that perspective whenever possible.

phreed:
@danielnachun 
Thanks for your detailed response.

danielnachun:
The `test_version.nu` isn't being used, so it needs to be deleted.  Then we can merge!

PR #32155, opened 2026-02-09
adding`pymedm` recipe

jGaboardi:
Checklist

- [x] Build number is 0.


jGaboardi:
* seems to actually need a the `jaxlib` pin after all.

jGaboardi:
@conda-forge/help-python

This Windows build giving me trouble, and I think I am missing something obvious for adding in some logic that simply skips Windows. In *theory*, we [support Windows](https://github.com/likeness-pop/pymedm/actions/runs/21839226268/job/63018692375), but in *practice* we do not (and never have) due to the `jax` compat issues. I have gone through the knowledge base and tried `skip: True  # [win]`, etc. But I must be missing the correct combination of additions/removals to get this passing.

recipes/pymedm/meta.yaml
```diff
@@ -0,0 +1,55 @@
...
+    - pip
+
+about:
+  home: https://github.com/{{ org }}/
+  summary: Penalized Maximum-Entropy Dasymetric Modeling (P-MEDM) in Python
+  license: LicenseRef-UT-BATTELLE
+  license_file:
+    - LICENSE
+    - NOTICE
+  doc_url: https://github.com/{{ org }}/{{ name }}/
```
ocefpaf: This obscures the recipe and makes it harder top read. Note that the package platforms will use this to render the page! Using variables in recipes is a balance between convenience vs readability. I would not use any in the metadata.

ocefpaf:
> @conda-forge/help-python
> 
> This Windows build giving me trouble, and I think I am missing something obvious for adding in some logic that simply skips Windows. In _theory_, we [support Windows](https://github.com/likeness-pop/pymedm/actions/runs/21839226268/job/63018692375), but in _practice_ we do not (and never have) due to the `jax` compat issues. I have gone through the knowledge base and tried `skip: True # [win]`, etc. But I must be missing the correct combination of additions/removals to get this passing.

The problem is that we do not have jaxlib on Windows. Leave it noarch b/c then that dependency is available, this package will be ready without the need to do anything here.

recipes/pymedm/meta.yaml
jGaboardi: OK, I was wondering if I was overdoing it there. And it seems I was. Good call. 

I made that change in another recipe, so I will revert that as well.

jGaboardi:
Yeah, I was digging around the [the logs](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=1457965&view=logs&j=53c3d5f4-e7c7-5b53-4b94-211701b2ba17&t=8f042650-8479-5761-7a4a-8d021c0bb2de) and thought that was the case.

So is the solution here to simply allow the Windows build to fail? And if so, should I add any extra logic in to make the check green?

ocefpaf:
> So is the solution here to simply allow the Windows build to fail? 

Yep. It won't be installable on Windows until that dependency can be satisfied.

jGaboardi:
@ocefpaf, thanks! You are the best!

PR #32198, opened 2026-02-13
Add newspaper4k and pydevd

cml37:
Checklist

- [x] Build number is 0.


cml37:
@conda-forge-admin, please ping team conda-forge/help-python

zklaus:
Looks like at least tqdm is already part of conda-forge.

cml37:
@zklaus  oh absolutely, but I don't think t**dq**m is!!  It was made for the explicit purpose of handling that very typo.  I have no idea why the upstream is using this, but hey, what can I say, haha!

cml37:
I did request [a change](https://github.com/WolfgangFahl/nicegui_widgets/commit/cad79e9eb366cf59219b9746f6b1d24f2669e84d#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R20), but looks like this logic hasn't changed in 3 years.  We shall see!!

recipes/tdqm/meta.yaml
```diff
@@ -0,0 +1,42 @@
+{% set version = "0.0.1" %}
+
+package:
+  name: tdqm
+  version: {{ version }}
+
+source:
+  url: https://pypi.org/packages/source/t/tdqm/tdqm-{{ version }}.tar.gz
+  sha256: f050004a76b1d22f70b78209b48781353c82440215b6ba7d22b1f499b05a0101
```
carterbox: Not sure whether this package should exist.

However, we shouldn't need to build an entire package to create an alias for an existing package. The alias should just be a metapackage which depends only on the package it is aliasing, and it should probably just be added as an additional output of the package it is aliasing, so that the version numbers will match.

cml37: I understand, but I guess this is the package that got used, so I kindof feel we should go with it.  To be honest, I actually don't have use for any of these anyway... trying to be a good maintainer and unjam https://github.com/conda-forge/geograpy3-feedstock/pull/10 which, well, I don't even use anymore.  But yea, this is getting more involved than I'd like... I'd rather fix than abandon, but it's been time consuming to even get to this point...

cml37:
Can we move forward with this?

PR #32225, opened 2026-02-15
Smspp project recommendations

hmaarrfk:
Checklist

- [ ] Build number is 0.

### DIFF
--- a/recipes/smspp-project/meta.yaml
+++ b/recipes/smspp-project/meta.yaml
@@ -9,7 +9,6 @@ source:
   sha256: 02d30a318b39491edc92b98cbe78ddbdd68891695493667b53728a80e3696967
 
 build:
-  skip: True  # [win]
   number: 0
   run_exports:
     - {{ pin_subpackage('smspp-project', max_pin='x.x.x') }}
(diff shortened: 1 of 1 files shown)


davide-f:
Many thanks for the investigation!

With the other maintainers, we agreed that the windows version shall be built static.
The reason is related to the dependencies we build on [coin-or] that require mkl that is built static.
If you switch the build_shared OFF, you'll experience the cyclical dependencies discussed in the other thread

hmaarrfk:
I really don't believe that windows should get any special treatment.

The more "special" consideration you give to windows, the longer your support tail will be.

We can likely resolve the error. It seems that a library dependency isn't declared.

Static linking makes your project susceptible to more errors, or at least requires you to compile when any of your (many) dependencies or subdependencies gets updated.

davide-f:
Agree on the point on windows.
To add more context, maintainers of coin-or-X packages have been kind to allow fixing a (MT/MD incompatibility)[https://github.com/conda-forge/coin-or-utils-feedstock/pull/28] and we also discussed [here](https://github.com/conda-forge/coin-or-utils-feedstock/issues/26) on how to move beyond the static mkl_static, which is the package leading to the issues experienced here. The former was solved, while the second issue was linked to the building process of the sources and the effort was deemed significant

